PR #2, opened 2014-02-05
Base preconfigured choices on UUID instead of label

mistydemeo:
preconfiguredChoices were previously being matched via description strings, which was error-prone. This could lead to processing MCP files being created which would fail to match anything, or strings that would accidentally match multiple microservices. This replaces matching with precise UUID matches instead.


Hwesta:
Looks good.  #5 (AIC) will have to be updated to take this into account before it can be merged (or vice versa)

Hwesta:
:shipit:

mistydemeo:
Will merge once the extract code is tested.

PR #3, opened 2014-02-05
Event detail for manually normalized files can be manually specified

Hwesta:
Manually normalized files can have the event detail populated by a person.  Form becomes available for manually normalized files after their are discovered (after normalization in most cases, after 'approve normalization' if 'manually normalize' was selected).  New waiting point for this information to be entered added.

Will need to be updated with preconfigured choice changes from #2.


mistydemeo:
Careful committing empty `.gitignores`. ;)

src/MCPClient/lib/clientScripts/storeAIP.py
```diff
@@ -73,7 +73,7 @@ def store_aip(aip_destination_uri, aip_path, sip_uuid, sip_name):
         package_type="AIP",
         size=os.path.getsize(aip_path)
     )
-    if new_file is not None and new_file['status'] != "FAIL":
+    if new_file is not None and new_file.get('status', '') != "FAIL":
```
mistydemeo, line 76: Is this meant to be true if `new_file` doesn't have a `'status'` attribute?

src/MCPClient/lib/clientScripts/normalize.py
```diff
@@ -117,7 +117,8 @@ def check_manual_normalization(opts):
             else:
                 return None
             print('Looking for', filename, 'in database')
-            return File.objects.get(sip=opts.sip_uuid, originallocation__iendswith=filename).currentlocation #removedtime = 0
+            # FIXME: SQL uses removedtime=0. Convince Django to express this
+            return File.objects.get(sip=opts.sip_uuid, originallocation__iendswith=filename) #removedtime = 0
```
mistydemeo, line 121: Maybe tidy up the comments since you're touching that part of the code?

src/MCPClient/lib/clientScripts/normalize.py
```diff
@@ -203,17 +195,43 @@ def once_normalized(command, opts, replacement_dict):
...
             sql = "INSERT INTO FilesIDs (fileUUID, formatName, formatVersion, formatRegistryName, formatRegistryKey) VALUES ('%s', '%s', NULL, NULL, NULL);" % (output_file_uuid, command.fpcommand.output_format.description)
             databaseInterface.runSQL(sql)
 
 
+def insert_derivation_event(original_uuid, output_uuid, derivation_uuid,
+        event_detail_output, outcome_detail_note, today=None):
+    """ Add the derivation link for preservation files and the event. """
+    if today is None:
+        today = str(datetime.date.today())
+    #Add event information to current file
```
mistydemeo: Ultra nitpick, but space after hash?

src/MCPClient/lib/clientScripts/retryNormalizeRemoveNormalized.py
```diff
@@ -70,6 +71,26 @@ def removePreservationFiles(SIPDirectory, SIPUUID):
...
         print inst.args
 
+    # Remove manually normalized derivation links
+    # TODO? Update this to delete for all derivations where the event is deleted
+    try:
+        sql = """DELETE FROM Derivations USING Derivations JOIN Files ON Derivations.derivedFileUUID = Files.fileUUID WHERE fileGrpUse='manualNormalization' AND sipUUID = '%s';""" % SIPUUID
+        databaseInterface.runSQL(sql)
+    except Exception as inst:
+        traceback.print_exc(file=sys.stdout)
+        print type(inst)     # the exception instance
```
mistydemeo: Is this necessary if we already printed the traceback? `traceback.print_exc()` should already print the exception class and its message.

src/dashboard/src/components/ingest/views.py
```diff
@@ -167,6 +177,27 @@ def ingest_metadata_edit(request, uuid, id=None):
 
     return render(request, 'ingest/metadata_edit.html', locals())
 
+def ingest_metadata_event_detail(request, uuid):
+    EventDetailFormset = modelformset_factory(models.Event, form=forms.EventDetailForm, extra=0)
```
mistydemeo, line 181: Break up the line lengths, maybe? Not a big deal though.

src/dashboard/src/main/models.py
```diff
@@ -132,6 +132,44 @@ class MetadataAppliesToType(models.Model):
     class Meta:
         db_table = u'MetadataAppliesToTypes'
 
+
+class Event(models.Model):
+    id = models.AutoField(primary_key=True, db_column='pk', editable=False)
+    event_id = UUIDField(auto=False, null=True, db_column='eventIdentifierUUID')
+    file_uuid = models.ForeignKey('File', db_column='fileUUID', to_field='uuid', null=True, blank=True)
```
mistydemeo, line 139: Underscores or no underscores? IIRC we recently dumped the underscores from a different model.

mistydemeo:
Only minor nitpicks, really. Looks good!

src/dashboard/src/main/models.py
Hwesta, line 139: I'm having convention conflicts where PEP8 says to use underscores (which I prefer), but the existing code doesn't and I can't decide which one overrides in a specific case.

mistydemeo, line 139: Underscores always IMO :snake: Maybe we should plan to just gradually introduce more and more underscores and overwhelm the original convention.

src/MCPClient/lib/clientScripts/retryNormalizeRemoveNormalized.py
Hwesta: Probably code I copied/moved from somewhere else that I didn't clean up - I'll fix it.  (logging.exception is probably more appropriate anyway)

src/MCPClient/lib/clientScripts/normalize.py
Hwesta, line 121: I added the FIXME and #removedtime = 0 comments, since I couldn't find a way to filter DateTimeFields on '0', which is what they're set to.  It's intended to find files that have not yet been removed.  Suggestions for a fix welcome!

src/MCPClient/lib/clientScripts/storeAIP.py
Hwesta, line 76: Yes, for compatibility if the storage service hasn't been updated to return a 'status'.  Possibly not an issue now, but it was when I wrote it.

Hwesta:
Oops, that .gitignore watchedDirectory was in the wrong place, moved it to the correct one.

A bunch of the code github says is new is actually moved from elsewhere (not nearby) that I didn't clean up.

PR #6, opened 2014-02-05
Move characterization to the FPR

mistydemeo:
This moves characterization to the FPR, allowing characterization rules to be defined and customized. Two new categories of FPR rules are added "characterization" and "default characterization".

FITS no longer runs on every file, and the old archivematicaFITS.py no longer runs at all. "Default characterization" rules will be run on any tool that doesn't have a specific characterization tool, and FITS will be the default shipped by Archivematica. An arbitrary number of characterization rules can be defined for a given format.

FPR database dump updates aren't covered here - FPR rules will get taken care of post-merge.

fiwalk will be a new characterization rule included when this is merged, and some fiwalk support files are included.

This depends on artefactual/archivematica-fpr-admin#1 in order for the new rules to show up in the UI.


src/MCPClient/lib/clientScripts/characterizeFile.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if format:
+        rules = FPRule.active.filter(format=format.uuid, purpose='characterization')
+
+    # Characterization always occurs - if nothing is specified, get one or more defaults
+    # specified in the FPR.
+    if not rules:
+        rules = FPRule.active.filter(purpose='default_characterization')
+
+    for rule in rules:
```
Hwesta, line 43: Confirming - we want to run _all_ the characterization rules we find?

Hwesta:
We'll have to address 4f499d5b1a5c2da3a51e9a0b1c3fcc1a84da397a / https://projects.artefactual.com/issues/6160 / https://projects.artefactual.com/issues/6161 before we can drop archivematicaFITS entirely :(  I think that's blocking on this commit until we figure out what goes into the METS there.

Also, I don't know anything about the fiwalk plugins, so I can't comment on them.

Other than that, looks good!

src/MCPClient/lib/clientScripts/characterizeFile.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if failed:
+        return -1
+    else:
+        return 0
+
+if __name__ == '__main__':
+    file_path = sys.argv[1]
+    file_uuid = sys.argv[2]
+    exit(main(file_path, file_uuid))
```
Hwesta: sys.exit?

src/MCPClient/lib/clientScripts/characterizeFile.py
mistydemeo, line 43: Yes: the idea was that multiple characterization tools could be defined for a given format, and all should be run. Same is also true for default characterization.

src/MCPClient/lib/clientScripts/characterizeFile.py
mistydemeo: That keeps tripping me up!

mistydemeo:
> We'll have to address 4f499d5 / https://projects.artefactual.com/issues/6160 / https://projects.artefactual.com/issues/6161 before we can drop archivematicaFITS entirely :(

That needs to get moved in to here either way - archivematicaFITS isn't going to be run under this! I'll talk to Evelyn.

mistydemeo:
Caught a couple minor bugs and pushed fixes. Will squash when merging.

The lxml/unicode one was a nasty surprise.

Hwesta:
Looks good.  Can't comment on the fiwalk plugin changes, and the parts that are in other pull requests have comments there.

mistydemeo:
Rebased and squashed; I think this is ready.

Hwesta:
Still looks good

mistydemeo:
The commit history here is messy because it's been rebased on top of #22, but otherwise I think this is good to go.

mistydemeo:
Rebased on top of qa/1.x now that #22 is merged. I'd like to get this merged ASAP to make sure that client branches that need this are okay.

PR #7, opened 2014-02-05
Add FIDO extensions

mistydemeo:
PRONOM doesn't currently support all of the formats that we need. Luckily FIDO supports extension files to add new local formats. This set here should be considered preliminary until we're ready to cut the 1.0 release, because the next PRONOM signature update will add at least a few of these.


src/archivematicaCommon/lib/externals/fido/archivematica_format_extensions.xml
```diff
@@ -0,0 +1,62 @@
...
+
+  <format>
+    <puid>archivematica-fmt/5</puid>
+    <name>Advanced Forensic Format</name>
+    <extension>aff</extension>
+    <signature>
+      <name>Advanced Forensic Format</name>
+      <pattern>
+        <position>BOF</position>
+        <regex>AFF10\r\n\x00</regex>
```
mistydemeo, line 58: Windows line endings in magic bytes :(

mistydemeo:
Added the new FPR formats to the DB. I'll merge these when the other two PRs get merged.

PR #8, opened 2014-02-05
Migrate certain unit variables to passVars in new chains

mistydemeo:
Certain chains need to be able to inherit passVars set in previous chains. An example of this is identification, where a second pass of identification happens after extraction during the transfer; the previous choice made needs to be able to apply.

Unit variables with the variable name `replacementDict` will now have their values included into passVars at the creation of every subsequent chain in the unit.

### DIFF
--- a/src/MCPClient/lib/clientScripts/identifyFileFormat.py
+++ b/src/MCPClient/lib/clientScripts/identifyFileFormat.py
@@ -12,8 +12,34 @@ if path not in sys.path:
     sys.path.append(path)
 os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
 from fpr.models import IDCommand, IDRule, FormatVersion
-from main.models import FileFormatVersion, File
+from main.models import FileFormatVersion, File, UnitVariable
 
+
+def save_idtool(file_, value):
+    """
+    Saves the chosen ID tool's UUID in a unit variable, which allows it to be
+    refetched by a later chain.
+
+    This is necessary in order to allow post-extraction identification to work.
+    The replacement dict will be saved to the special 'replacementDict' unit
+    variable, which will be transformed back into a passVar when a new chain in
+    the same unit is begun.
+    """
+
+    # The unit_uuid foreign key can point to a transfer or SIP, and this tool
+    # runs in both.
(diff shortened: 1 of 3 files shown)


src/MCPServer/lib/jobChain.py
```diff
@@ -33,6 +34,21 @@
 #
 #potentialToHold/getFromDB
 #-previous chain links
+def fetchUnitVariableForUnit(unit_uuid):
+    "Returns a dict combining all of the replacementDict unit variables for the specified unit."
```
Hwesta: Triple quotes around the docstring?

mistydemeo: Is that the preferred style for single-line docstrings?

Hwesta:
Looks good, I could only find the most minor of nitpicks.  :shipit:

src/MCPServer/lib/jobChain.py
mistydemeo: > Triple quotes are used even though the string fits on one line. This makes it easy to later expand it.

http://www.python.org/dev/peps/pep-0257/

Looks like that is preferred, will fix.

src/MCPClient/lib/clientScripts/identifyFileFormat.py
```diff
@@ -12,9 +12,29 @@
     sys.path.append(path)
 os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
 from fpr.models import IDCommand, IDRule, FormatVersion
-from main.models import FileFormatVersion, File
+from main.models import FileFormatVersion, File, UnitVariable
 
 
+def save_idtool(file_, value):
+    """
```
mistydemeo, line 19: @Hwesta I like the style of starting the comment on the line after the triple quotes, does that work for you?

Hwesta, line 19: Sure.

While we're talking docstring minutia, there should be a newline after the summary. 

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

mistydemeo:
Pushed!

PR #9, opened 2014-02-05
Add new extraction options

mistydemeo:
This adds two new extraction options:
- Make package extraction optional, not mandatory as it currently is
- Make package deletion optional; currently the original package is always deleted after it is extracted

This depends on #8. It should also be updated after #2 is merged to add preconfigured choices for these new options. Both should be on by default, to match current behaviour.

### DIFF
--- a/src/MCPClient/lib/clientScripts/extractContents.py
+++ b/src/MCPClient/lib/clientScripts/extractContents.py
@@ -18,10 +18,6 @@ from fpr.models import FPCommand
 from main.models import FileFormatVersion, File
 
 file_path_cache = {}
-# Controls whether or not to delete the package file after extracting its contents.
-# Default is true.
-# TODO: Make this configurable in some fashion.
-remove_package_file = True
 
 def output_directory(file_path, date):
     if file_path_cache.get(file_path):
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/extractContents.py
```diff
@@ -134,4 +130,11 @@ def main(transfer_uuid, sip_directory, date, task_uuid):
...
     task_uuid = sys.argv[4]
-    exit(main(transfer_uuid, sip_directory, date, task_uuid))
+    # Whether or not to remove the package file post-extraction
+    # This is set by the user during the transfer, and defaults to false.
+    if sys.argv[5] == "True":
+        delete = True
+    else:
+        delete = False
+    print("Deleting: {}".format(delete), file=sys.stderr)
+    exit(main(transfer_uuid, sip_directory, date, task_uuid, delete=delete))
```
Hwesta: Your favorite sys.exit reminder :)

mistydemeo: My kingdom for a top-level `exit()` :horse:

Hwesta:
Looks good.  :shipit: whenever it's supposed to go into 1.x

(Reviewing MicroServiceChainLinks changes is the worst)

mistydemeo:
> (Reviewing MicroServiceChainLinks changes is the worst)

Agreed \* 1000

PR #10, opened 2014-02-05
Add new transfer metadata form

mistydemeo:
This adds a new metadata form which is available during the transfer. The goal is to allow metadata to be associated with a transfer containing disk images, and all of the fields contained in the initial set are transfer-specific.

This requires updates to the UI and to the database, in order to provide room to save the metadata. 9f82847ae67fecfc11053e23f00566dc40768dda also makes a major change so that each transfer component (directory chosen from the "browse" menu) now creates an individual transfer, instead of combining into one transfer.

This badly needs to be squashed on merge.


mistydemeo:
@hwesta Special note 45b3462b013c2735f6e77aa54c1f718326ab96f0, which contains a separate change I probably shouldn't have lumped in. This allows arbitrary external XML files to be referenced in the AIP's METS, originally intended for FTK Imager XML files. They won't be mdWrapped though, just mdReffed.

PR #11, opened 2014-02-05
Deal more gracefully with non-ASCII, non-Unicode paths

mistydemeo:
It's possible for users to attempt to begin transfers using files whose names contain non-Unicode, non-ASCII characters. Since communication to/from the storage service and between the dashboard and JS frontend occurs via JSON responses, which mandates Unicode, it's necessary to find an alternate way to represent paths.

This moves to communicate paths back and forth in base64 encoding, and vendors a JS base64 library to let the UI decode and encode paths as appropriate.

This depends on v2 of the storage service API, in artefactual/archivematica-storage-service#3.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
@@ -28,10 +28,9 @@ import sys
 import lxml.etree as etree
 import MySQLdb
 from xml.sax.saxutils import quoteattr
-#from archivematicaCreateMETS2 import escape
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 import databaseInterface
-#from archivematicaFunctions import escape
+from archivematicaFunctions import escape
 
 
 UUIDsDic={}
@@ -55,10 +54,6 @@ XMLFile = opts.xmlFile
 includeAmdSec = opts.amdSec
 basePathString = "%%%s%%" % (opts.basePathString)
 fileGroupIdentifier = opts.fileGroupIdentifier
-
-def escape(string):
-    string = string.decode('utf-8')
-    return string
 
 
 def newChild(parent, tag, text=None, tailText=None):
--- a/src/MCPClient/lib/clientScripts/restructureForCompliance.py
+++ b/src/MCPClient/lib/clientScripts/restructureForCompliance.py
@@ -40,7 +40,11 @@ def restructureTRIMForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unit
         reqDirPath = os.path.join(unitPath, dir)
         if not os.path.isdir(reqDirPath):
             os.mkdir(reqDirPath)
-    
+
+    # The types returned by os.listdir() depends on the type of the argument
+    # passed to it. In this case, we want all of the returned names to be
+    # bytestrings because they may contain arbitrary, non-Unicode characters.
+    unitPath = str(unitPath)
     for item in os.listdir(unitPath):
         if item in requiredDirectories:
             continue
@@ -83,7 +87,11 @@ def restructureTRIMForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unit
 def restructureBagForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unitIdentifierType, unitPathReplaceWith = "%transferDirectory%"):
     bagFileDefaultDest = os.path.join(unitPath, "logs", "BagIt")
     requiredDirectories.append(bagFileDefaultDest)
-    unitDataPath = os.path.join(unitPath, "data")
+    # This needs to be cast to a string since we're calling os.path.join(),
+    # and any of the other arguments could contain arbitrary, non-Unicode
+    # characters.
+    unitPath = str(unitPath)
+    unitDataPath = str(os.path.join(unitPath, "data"))
     for dir in requiredDirectories:
         dirPath = os.path.join(unitPath, dir)
         dirDataPath = os.path.join(unitPath, "data", dir)
@@ -126,6 +134,7 @@ def restructureForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unitIden
     print unitUUID, unitType
 
 def restructureDirectory(unitPath):
+    unitPath = str(unitPath)
     for dir in requiredDirectories:
         dirPath = os.path.join(unitPath, dir)
         if not os.path.isdir(dirPath):
--- a/src/dashboard/src/components/archival_storage/views.py
+++ b/src/dashboard/src/components/archival_storage/views.py
@@ -233,7 +233,7 @@ def aip_delete(request, uuid):
 
 def aip_download(request, uuid):
     server_url = helpers.get_setting('storage_service_url', None)
-    redirect_url = server_url + 'api/v1/file/' + uuid + '/download'
+    redirect_url = server_url + 'api/v2/file/' + uuid + '/download'
     return HttpResponseRedirect(redirect_url)
 
 def aip_file_download(request, uuid):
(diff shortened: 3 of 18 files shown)


src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -37,7 +37,7 @@
     date = sys.argv[3]
     taskUUID = sys.argv[4]
 
-    command = 'clamdscan  - <"' + escapeForCommand(target).replace("$", "\\$") + '"'
+    command = "clamdscan  - <'" + escapeForCommand(target).replace("$", "\\$") + "'"
```
mistydemeo: Possibly no longer necessary?

mistydemeo:
This also updates a number of database fields to blog/longblob types.

src/dashboard/src/media/js/base64.js
```diff
@@ -0,0 +1,216 @@
+/*
+Copyright (c) 2008 Fred Palmer fred.palmer_at_gmail.com
```
Hwesta: I think this file should be in src/dashboard/src/media/js/vendor/

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
```diff
@@ -56,9 +56,12 @@
 basePathString = "%%%s%%" % (opts.basePathString)
 fileGroupIdentifier = opts.fileGroupIdentifier
 
-def escape(string):
-    string = string.decode('utf-8')
-    return string
+def escape(s):
```
Hwesta: Why not use the one from archivematicaFunctions?  (Though, this will soon be irrelevant because of your other METS work)

mistydemeo: Haha, wow, I missed that I'd edited both at some point. I see there's a commented out import of that function at L34, I might as well just use the same one.

mistydemeo: If python has a concept of function aliases I _will_ rename this function to `mangle()`.

src/MCPClient/lib/clientScripts/restructureForCompliance.py
```diff
@@ -41,6 +41,7 @@ def restructureTRIMForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unit
         if not os.path.isdir(reqDirPath):
             os.mkdir(reqDirPath)
     
+    unitPath = str(unitPath)
```
Hwesta, line 47: Why are we explicitly casting these to strings?  What effect does that have?

mistydemeo, line 47: `os.listdir()`'s results depend on the type of the argument passed to it.

mistydemeo, line 47: Below, it's done because of `os.path.join`, which can raise an exception for strings of different types.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -233,7 +233,7 @@ def aip_delete(request, uuid):
 
 def aip_download(request, uuid):
     server_url = helpers.get_setting('storage_service_url', None)
-    redirect_url = server_url + 'api/v1/file/' + uuid + '/download'
+    redirect_url = server_url + 'api/v2/file/' + uuid + '/download'
```
Hwesta, line 236: If you could move anything that uses the storage service URL to storageService.py, then we'd only have to change the version in one place going forward.  Might be better suited for another merge?

src/dashboard/src/components/ingest/views.py
```diff
@@ -276,6 +276,8 @@ def ingest_normalization_report(request, uuid, current_page=None):
     sipname = utils.get_directory_name_from_job(job)
 
     objects = getNormalizationReportQuery(sipUUID=uuid)
+    for o in objects:
+        o['location'] = o['location'].decode('utf-8', errors='replace')
```
Hwesta: Do we want all of these going through archivematicaFunctions.escape?

mistydemeo: Possibly, but I was wary of doing an extra import just for a single occurrence. Still makes sense to use it though, now that I think about it.

src/dashboard/src/main/models.py
```diff
@@ -191,6 +191,7 @@ class File(models.Model):
     uuid = models.CharField(max_length=36, primary_key=True, db_column='fileUUID')
     sip = models.ForeignKey(SIP, db_column='sipUUID', to_field = 'uuid')
     transfer = models.ForeignKey(Transfer, db_column='transferUUID', to_field = 'uuid')
+    # both actually `longblob` in the database
```
Hwesta, line 194: Is this going to cause problems where Django thinks the contents can be converted to a string?  There is a BinaryField, but it only accepts a 'bytes' assignment - though that might be what we want.

src/dashboard/src/main/views.py
```diff
@@ -108,6 +108,19 @@ def tasks(request, uuid):
     if (len(objects) == 0):
         return tasks_subjobs(request, uuid)
 
+    # Filenames can be any encoding - we want to be able to display
+    # unicode, while just displaying unicode replacement characters
+    # for any other encoding present.
+    for item in objects:
+        if item.filename:
```
Hwesta: Use archivematicaFunctions.escape?  Then you won't have to check if the attribute exists first.

src/dashboard/src/main/models.py
mistydemeo, line 194: Django seems to have no problems with this. I looked it up, and this seems to be the accepted solution. Django doesn't have types that directly map to blob and co.

Hwesta, line 194: Great.  (I wonder how we get the create SQL to generate longblobs, whenever we get to that point?)

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
Hwesta: `from archivematicaFunctions import escape as mangle`?

src/MCPClient/lib/clientScripts/restructureForCompliance.py
Hwesta, line 47: Can you add comments to that effect?  Otherwise it's confusing why we're casting a string(-like thing) to a string explicitly.  It's the sort of thing I might remove on cleanup if I didn't know better.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
Hwesta: What was switching the quotes supposed to do?

mistydemeo: I had encountered a filename which contained backticks and shell expansion was taking place when it was double-quoted.

Hwesta: `archivematicaFunctions.escapeForCommand` escapes backticks for just that reason (see ae890e3cffc303ad25549fa6de2c61fca0298403).  So probably not necessary anymore.

mistydemeo: Will remove this commit then - I wrote it before that.

Ideally though I'd like to redo this part so we don't spawn a shell, I'd rather we not be susceptible to shell escaping issues in the first place (especially when calling a virus scanner).

Hwesta: Very true!

mistydemeo:
Changes made as individual commits, will squash on merging.

PR #13, opened 2014-02-05
Move transfer creation logic into the dashboard

mistydemeo:
Currently a surprising amount of logic that goes into creating a transfer happens in the clientside JavaScript in the dashboard. Unfortunately, this means that if the user interrupts the dashboard during a lengthy file copy—for instance, if the user's connection is lost, or if the user closes their browser—then a "zombie transfer" is possible, leaving an incomplete transfer that will never appear in the dashboard.

No further information or input is needed from the user at this point, so it's easy to move this logic into the dashboard instead.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -203,16 +203,17 @@ def start_transfer(request):
     temp_base_dir = helpers.get_client_config_value('temp_dir') or None
     temp_dir = tempfile.mkdtemp(dir=temp_base_dir)
 
-    for path in paths:
-        copy_transfer_component(transfer_name=transfer_name,
+    for i, path in enumerate(paths):
+        target = transfer_name + '_' + str(i + 1)
+        copy_transfer_component(transfer_name=target,
                                 path=path, destination=temp_dir)
 
-    if len(paths) == 1 and helpers.file_is_an_archive(paths[0]):
-        filepath = os.path.join(temp_dir, os.path.basename(paths[0]))
-    else:
-        filepath = os.path.join(temp_dir, transfer_name)
-    copy_to_start_transfer(filepath=filepath,
-                           type=transfer_type, accession=accession)
+        if helpers.file_is_an_archive(path):
+            filepath = os.path.join(temp_dir, os.path.basename(path))
+        else:
+            filepath = os.path.join(temp_dir, target)
+        copy_to_start_transfer(filepath=filepath,
+                               type=transfer_type, accession=accession)
 
     response = {'message': 'Copy successful.'}
     return helpers.json_response(response)


src/dashboard/src/components/filesystem_ajax/urls.py
```diff
@@ -27,8 +27,7 @@
     (r'^delete/$', 'delete'),
     (r'^copy_to_originals/$', 'copy_to_originals'),
     (r'^copy_to_arrange/$', 'copy_to_arrange'),
-    (r'^copy_transfer_component/$', 'copy_transfer_component'),
     (r'^get_temp_directory/$', 'get_temp_directory'),
-    (r'^ransfer/$', 'copy_to_start_transfer'),
+    (r'^ransfer/$', 'start_transfer'),
```
Hwesta, line 31: Do we really want to keep the 'ransfer' endpoint, instead of something less typo-ish? (start_transfer?)

mistydemeo, line 31: I would be very happy to change this, but I have no idea why it is this way in the first place. ;)

Hwesta, line 31: Probably because 'transfer' conflicted with the transfer/ tab and 'ransfer' is close enough? At least, that's what I always assumed.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -187,11 +187,32 @@ def get_temp_directory(request):
...
     transfer_name = archivematicaFunctions.unicodeToStr(request.POST.get('name', ''))
-    path = archivematicaFunctions.unicodeToStr(request.POST.get('path', ''))
-    destination = archivematicaFunctions.unicodeToStr(request.POST.get('destination', ''))
+    transfer_type = archivematicaFunctions.unicodeToStr(request.POST.get('type', ''))
+    accession = archivematicaFunctions.unicodeToStr(request.POST.get('accession', ''))
+    paths = archivematicaFunctions.unicodeToStr(request.POST.getlist('paths[]', ''))
+
+    # Create temp directory that everything will be copied into
+    temp_base_dir = helpers.get_client_config_value('temp_dir') or None
+    temp_dir = tempfile.mkdtemp(dir=temp_base_dir)
```
Hwesta, line 204: Delete the temp directory this once copy_to_start_transfer has succeeded.  (Much easier now that it's in one function instead of several)

mistydemeo:
Rebased, and added a new commit to create fresh transfers from each transfer component.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -189,13 +189,36 @@ def get_temp_directory(request):
...
-    path = base64.b64decode(request.POST.get('path', ''))
-    destination = archivematicaFunctions.unicodeToStr(request.POST.get('destination', ''))
+    transfer_type = archivematicaFunctions.unicodeToStr(request.POST.get('type', ''))
+    accession = archivematicaFunctions.unicodeToStr(request.POST.get('accession', ''))
+    paths = request.POST.getlist('paths[]', '')
+    paths = [base64.b64decode(path) for path in paths]
+
+    # Create temp directory that everything will be copied into
+    temp_base_dir = helpers.get_client_config_value('temp_dir') or None
+    temp_dir = tempfile.mkdtemp(dir=temp_base_dir)
```
Hwesta, line 204: This should be deleted before returning so it doesn't clutter things up.

Hwesta:
Looks good.  When I last talked to Courtney she had some opinions on each transfer component becoming its own transfer, so you may want to speak to her before committing this (if you haven't already)

PR #14, opened 2014-02-05
Support JSON metadata

mistydemeo:
This allows JSON metadata to be supported, in addition to CSV metadata.

JSON metadata follows a similar format. It should contain an array of objects, one object per file to represent. Its keys should contain data in the same format that the CSV columns would contain. Rather than have separate JSON metadata code in archivematicaCreateMETS2, this just adds a new microservice to translate JSON into CSV at the end of ingest.

Supported JSON resembles this:

``` json
[
    {
        "filename": "foo.txt",
        "dc.title": "Some text file",
        "dc.author": "Me"
    }
]
```

In addition, this branch also adds supported for qualified Dublin Core. A couple of cleanups in these areas of code snuck in too.

This needs to be squashed when it's merged.

### DIFF
--- a/src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
+++ b/src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
@@ -1,0 +1,61 @@
+#!/usr/bin/python -OO
+
+import csv
+import json
+import os
+import sys
+
+
+def fetch_keys(objects):
+    keys = set()
+    for object in objects:
+        keys.update(object.keys())
+
+    # Column order is otherwise unimportant, but
+    # "filename" and "parts" must be column 0.
+    # (They are mutually exclusive.)
+    keys = list(keys)
+    if 'filename' in keys:
+        keys.remove('filename')
+        keys.insert(0, 'filename')
+    elif 'parts' in keys:
+        keys.remove('parts')
+        keys.insert(0, 'parts')
+
+    return keys
+
+
+# DictWriter will fail if any Unicode characters are in the keys or
+# values in a dict passed to writerow(). This encodes them all to
+# UTF-8 bytestrings.
+def fix_encoding(row):
+    return {key.encode('utf-8'): value.encode('utf-8') for key, value in row.iteritems()}
+
+
+def main(sip_uuid, json_metadata):
+    # Many transfers won't have JSON metadata, so just exit without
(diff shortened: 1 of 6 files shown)


src/MCPServer/share/mysql_dev1.sql
```diff
@@ -34,3 +34,11 @@ UPDATE MicroServiceChainLinks SET defaultNextChainLink='e19f8eed-faf9-4e04-bf1f-
 UPDATE MicroServiceChoiceReplacementDic SET replacementDic='{\"%ContentdmServer%\":\"http://111.222.333.444:81\", \"%ContentdmUser%\":\"usernamebar\", \"%ContentdmGroup%\":\"456\"}' WHERE pk='c001db23-200c-4195-9c4a-65f206f817f2';
 UPDATE MicroServiceChoiceReplacementDic SET replacementDic='{\"%ContentdmServer%\":\"http://localhost\", \"%ContentdmUser%\":\"usernamefoo\", \"%ContentdmGroup%\":\"123\"}' WHERE pk='ce62eec6-0a49-489f-ac4b-c7b8c93086fd';
 -- /Issue 5232
+
+-- Issue XXXX
```
Hwesta: Issue number?

src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
```diff
@@ -0,0 +1,61 @@
...
+        writer.writeheader()
+        for row in parsed:
+            writer.writerow(fix_encoding(row))
+
+    return 0
+
+if __name__ == '__main__':
+    try:
+        sip_uuid, json_metadata = sys.argv[1:]
+    except KeyError:
```
Hwesta: `ValueError` also generated if `sys.argv[1:]` isn't length 2

src/MCPClient/tests/test_json_conversion.py
```diff
@@ -0,0 +1,17 @@
+import os
+import subprocess
+
+JSON = '[{"dc.title": "This is a test item", "filename": "objects/test.txt"}]'
+CSV = 'filename,dc.title\r\nobjects/test.txt,This is a test item\r\n'
+
+
+def test_json_csv_conversion(tmpdir):
```
Hwesta, line 8: :+1: Yay tests!

PR #15, opened 2014-02-07
identifyFileFormat: Write identification into Events

mistydemeo:
This replaces the old archivematicaFITS event-writing code, which will allow us to remove archivematicaFITS and merge #6.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaFITS.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaFITS.py
@@ -32,7 +32,6 @@ from archivematicaFunctions import getTagged
 from archivematicaFunctions import escapeForCommand
 from databaseFunctions import insertIntoFPCommandOutput
 from databaseFunctions import insertIntoEvents
-from databaseFunctions import insertIntoFilesIDs
 import databaseInterface
 
 databaseInterface.printSQL = False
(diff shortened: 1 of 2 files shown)


mistydemeo:
This produces output in the METS that matches the old format:

``` xml
<format>
  <formatDesignation>
    <formatName>MPEG 1/2 Audio Layer 3</formatName>
    <formatVersion></formatVersion>
  </formatDesignation>
  <formatRegistry>
    <formatRegistryName>PRONOM</formatRegistryName>
    <formatRegistryKey>fmt/134</formatRegistryKey>
  </formatRegistry>
</format>
```

``` xml
<eventIdentifier>
  <eventIdentifierType>UUID</eventIdentifierType>
  <eventIdentifierValue>01fbf8da-6acf-48da-85a9-eeb65e2b6ec2</eventIdentifierValue>
</eventIdentifier>
<eventType>format identification</eventType>
<eventDateTime>2014-02-07T22:36:22</eventDateTime>
<eventDetail>program="Fido"; version="1"</eventDetail>
<eventOutcomeInformation>
  <eventOutcome>Positive</eventOutcome>
  <eventOutcomeDetail>
    <eventOutcomeDetailNote>fmt/134</eventOutcomeDetailNote>
  </eventOutcomeDetail>
</eventOutcomeInformation>
```

Hwesta:
Code looks good. :)

What if the format it's identified as doesn't have a PRONOM ID?  Eg. identify was done by extension to a FormatVersion that doesn't have a PRONOM ID.

Also, how does the <format> METS get generated?  It used to be from FilesIDs, but I don't see that getting inserted, or another related change.

mistydemeo:
> What if the format it's identified as doesn't have a PRONOM ID? Eg. identify was done by extension to a FormatVersion that doesn't have a PRONOM ID.

We maintain an internal map for this in the FPR. At the time we identify formats, we associate the file with a FormatVersion, and FormatVersion has a pronom_id column. The tool output isn't what we're recording.

> Also, how does the METS get generated? It used to be from FilesIDs, but I don't see that getting inserted, or another related change.

¯_(ツ)_/¯

This is still kind of a mystery to me but I'll look into it.

mistydemeo:
Here's a good example of how the PUID is still written when file extension is used: https://gist.github.com/mistydemeo/a9f6b74a1856002be3dd (second section)

mistydemeo:
Looks like FilesIDs wasn't being populated, so updated that. File IDs look the same for PRONOM file types, and for non-PRONOM file types they look like this:

``` xml
<format>
  <formatDesignation>
    <formatName>Generic Bitmap</formatName>
    <formatVersion></formatVersion>
  </formatDesignation>
  <formatRegistry>
    <formatRegistryName>Archivematica Format Policy Registry</formatRegistryName>
    <formatRegistryKey>.bmp</formatRegistryKey>
  </formatRegistry>
</format>
```

mistydemeo:
Tested, and removing the archivematicaFITS code has not caused obvious issues.

Hwesta:
archivematicaFITS is run on submission documentation as well (after normalization, MSCLs 33d7ac55-291c-43ae-bb42-f599ef428325 and 76d87f57-9718-4f68-82e6-91174674c49c).  Is this change going to affect that? I don't know what information about the submission documentation is actually needed.

mistydemeo:
Hm, that's a good question. I'll ask Evelyn about this.

mistydemeo:
Evelyn suggests adding running `identifyFileFormat` on submission documentation, and allowing a selection of tool there the same as in the other two places it's run. I like that idea.

PR #16, opened 2014-02-14
executeOrRun: allow file objects as stdIn

mistydemeo:
Previously, executeOrRun only allowed strings to be passed as stdin to launched subprocesses. However, the underlying `subprocess` module allows arbitrary `file` objects to be passed. This allows for a secure way to pipe data into processes without needing to invoke a subshell.


Hwesta:
Looks good. :shipit:

PR #17, opened 2014-02-25
Ensure paths are treated internally as base64 everywhere

mistydemeo:
- filesystem_ajax now also uses base64 between the dashboard and itself,
  instead of doing a translation before paths are sent to the SS.
- directory_to_dict path helper also translates parent directories.
- Filesystem browser uses base64 paths everywhere outside of HTML
  representation.

@hwesta, are you okay with pushing base64 changes deeper into the filesystem browser JS code like this?

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -432,7 +432,8 @@ def pad_destination_filepath_if_it_already_exists(filepath, original=None, attem
 
 def download(request):
     shared_dir = os.path.realpath(helpers.get_client_config_value('sharedDirectoryMounted'))
-    requested_filepath = os.path.realpath('/' + request.GET.get('filepath', ''))
+    filepath = base64.b64decode(request.GET.get('filepath', ''))
+    requested_filepath = os.path.realpath('/' + filepath)
 
     # respond with 404 if a non-Archivematica file is requested
     try:
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -66,10 +66,10 @@ def directory_to_dict(path, directory={}, entry=False):
     if (entry == False):
         entry = directory
         # remove leading slash
-        entry['parent'] = os.path.dirname(path)[1:]
+        entry['parent'] = base64.b64encode(os.path.dirname(path)[1:])
 
     # set standard entry properties
-    entry['name'] = os.path.basename(path)
+    entry['name'] = base64.b64encode(os.path.basename(path))
     entry['children'] = []
 
     # define entries
@@ -78,7 +78,7 @@ def directory_to_dict(path, directory={}, entry=False):
         new_entry = None
         if file[0] != '.':
             new_entry = {}
-            new_entry['name'] = file
+            new_entry['name'] = base64.b64encode(file)
             entry['children'].append(new_entry)
 
         # if entry is a directory, recurse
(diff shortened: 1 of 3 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -94,9 +94,9 @@ def directory_to_dict(path, directory={}, entry=False):
 def directory_children_proxy_to_storage_server(request, location_uuid, basePath=False):
     path = ''
     if (basePath):
-        path = path + basePath
-    path = path + request.GET.get('base_path', '')
-    path = path + request.GET.get('path', '')
+        path = path + base64.b64decode(basePath)
```
Hwesta: Just `path =  base64.b64decode(basePath)` should be the same, and less confusing.

mistydemeo: Good point.

src/dashboard/src/media/js/file-explorer.js
```diff
@@ -36,8 +36,15 @@
       return this.path().replace(/\//g, '_').replace('.', '__');
     },
 
+    // Decode parent and child elements, concatenate them as plaintext,
+    // then base64-encode them again for transfer to a server.
+    // This is primarily used when querying the storage service.
     path: function() {
-      return this.get('parent') + '/' + this.get('name');
+      parent = this.get('parent');
```
Hwesta: Can this return undefined?  (Do we care?)

src/dashboard/src/media/js/file-explorer.js
```diff
@@ -36,8 +36,15 @@
...
+      if (parent)
+        parent = Base64.decode(parent);
+      path = parent + '/' + this.get('name');
```
Hwesta: `this.get('name')` doesn't return a base64 encoded path that needs to be decoded?

src/dashboard/src/media/js/file-explorer.js
mistydemeo: Yes - the `if` statement below was meant to be used in case it does.

Hwesta: Derp, ok

src/dashboard/src/media/js/file-explorer.js
mistydemeo: Ah, I think I see why - an earlier part of the code, in `render`, mutates the entity. Going to change it to create a new object instead. (L434)

Hwesta:
You've looked into the crazy encoding problems more than I have - if you think this is best, then sure.

We'll have to remember the encoding is happening when making future changes - add something with a crazy name to our "unit tests"?

src/dashboard/src/media/js/file-explorer.js
```diff
@@ -36,8 +36,21 @@
       return this.path().replace(/\//g, '_').replace('.', '__');
     },
 
+    // Provides a human-readable version of the path, suitable for display.
+    // This is used by path() below.
+    displaypath: function() {
```
mistydemeo, line 41: This is not currently used, but I included it for future use in case we _do_ need a human-readable path (and for clarity that `path()` is not one).

src/dashboard/src/media/js/file-explorer.js
```diff
@@ -99,8 +118,13 @@
...
     render: function() {
-      var context = this.context()
-        , html = this.template(context);
+      // The actual path is stored as base64, so it needs to be decoded
+      // before being displayed in HTML.
+      var context = this.context();
+      if (context.name) {
+        context.name = Base64.decode(context.name);
+      }
+      html = this.template(context);
```
mistydemeo, line 127: This is the only place the path strings are actually decoded from base64. All paths' `'parent'` and `'name'` fields should be base64-encoded at the time they enter the file explorer, and the file explorer code takes care of joining paths in a safe way.

Hwesta:
Discussed in person, conclusions:
- Transfer source may have to deal with non-unicode paths (eg. Japanese encoded in shift-jis, Thai encoded in tis-620 etc) so the storage service, Django and javascript frontend all have to support this
- Maintaining two sets of code (notably the Javascript file browser & file explorer) to support both unicode paths and base 64 encoded paths is too much work
- All paths sent between the Javascript, dashboard and storage service will be encoded in base 64, and will be decoded as late as possible: before display to use, before path manipulation, before storage in DB
  - This includes paths for SIP arrangement, review AIP, and anything else that uses file-explorer/file-browser
- Other functions in filesystem_ajax that deal with paths were deprecated, or updated in dev branches and shouldn't be touched

PR #19, opened 2014-03-11
Allow DC metadata to be entered in transfer, persist to ingest

mistydemeo:
These commits allow DC metadata to be entered in the transfer phase. This is implemented by adding a new DC form, which exists in the transfer phase, and saving the metadata to disk in between transfer and ingest.

To save and load DC metadata, the data is serialized to disk as JSON at the very end of a transfer using the new `saveDublinCore` microservice. This happens after the user selects to create a SIP, so it's unlikely that users would be able to try to manually edit metadata after it runs.

At the very beginning of ingest, immediately after the SIP is verified, the new `loadDublinCore` microservice looks for a `metadata/dc.json` file inside the SIP; if present, it populates the DC metadata for the SIP in the database using that file. This allows the DC metadata to be editable by hand during ingest, and makes sure it will be used in the final AIP METS.

~~To support this, this now persists the contents of the transfer's `metadata/` directory when creating a SIP. In the past, anything there would be discarded when creating a single SIP.~~

This PR's commit history is still p. messy, will get cleaned up on merge.

### DIFF
--- a/src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
+++ b/src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
@@ -81,6 +81,14 @@ if __name__ == '__main__':
         path = os.path.join(tmpSIPDir, directory)
         if not os.path.isdir(path):
             os.makedirs(path)
+
+    # Copy the JSON metadata file, if present;
+    # this contains a serialized copy of DC metadata entered in the dashboard UI
+    src = os.path.normpath(os.path.join(objectsDirectory, "..", "metadata", "dc.json"))
+    dst = os.path.join(tmpSIPDir, "metadata", "dc.json")
+    # This file only exists if any metadata was created during the transfer
+    if os.path.exists(src):
+        shutil.copy(src, dst)
     
     #copy processingMCP.xml file
     src = os.path.join(os.path.dirname(objectsDirectory[:-1]), "processingMCP.xml") 
--- a/src/MCPClient/lib/clientScripts/loadDublinCore.py
+++ b/src/MCPClient/lib/clientScripts/loadDublinCore.py
@@ -1,0 +1,40 @@
+#!/usr/bin/python2
+
+import json
+import os
+import sys
+
+path = '/usr/share/archivematica/dashboard'
+if path not in sys.path:
+    sys.path.append(path)
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
+from main import models
+
+# This is the UUID of SIP from the `MetadataAppliesToTypes` table
+INGEST_METADATA_TYPE = '3e48343d-e2d2-4956-aaa3-b54d26eb9761'
+
+
+def main(sip_uuid, dc_path):
+    # If there's no metadata, that's not an error, and just keep going
+    if not os.path.exists(dc_path):
+        print "DC metadata not found; exiting", "(at", dc_path + ")"
+        return 0
+
+    print "Loading DC metadata from", dc_path
+    with open(dc_path) as json_data:
+        data = json.load(json_data)
+    dc = models.DublinCore(metadataappliestoidentifier=sip_uuid,
+                           metadataappliestotype=INGEST_METADATA_TYPE)
+    for key, value in data.iteritems():
+        try:
+            setattr(dc, key, value)
+        except AttributeError:
+            print >> sys.stderr, "Invalid DC attribute:", key
+
+    dc.save()
+    return 0
+
+if __name__ == '__main__':
+    sip_uuid = sys.argv[1]
+    dc_path = sys.argv[2]
+    sys.exit(main(sip_uuid, dc_path))
--- a/src/MCPClient/lib/clientScripts/saveDublinCore.py
+++ b/src/MCPClient/lib/clientScripts/saveDublinCore.py
@@ -1,0 +1,43 @@
+#!/usr/bin/python2
+
+import json
+import os
+import sys
+
+path = '/usr/share/archivematica/dashboard'
+if path not in sys.path:
+    sys.path.append(path)
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
+from main import models
+
+FIELDS = [
+    'title', 'creator', 'subject', 'description', 'publisher', 'contributor',
+    'date', 'type', 'format', 'identifier', 'source', 'relation', 'language',
+    'coverage', 'rights'
+]
+
+
+def main(transfer_uuid, target_path):
+    jsonified = {}
+    try:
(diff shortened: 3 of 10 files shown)


src/MCPClient/lib/clientScripts/loadDublinCore.py
```diff
@@ -0,0 +1,36 @@
...
+        print "DC metadata not found; exiting", "(at", dc_path, ")"
+        return 0
+
+    print "Loading DC metadata from", dc_path
+    with open(dc_path) as json_data:
+        data = json.load(json_data)
+    dc = models.DublinCore(metadataappliestoidentifier=sip_uuid,
+                           metadataappliestotype=INGEST_METADATA_TYPE)
+    for key, value in data.iteritems():
+        setattr(dc, key, value)
```
mistydemeo: Should probably do error checking in case users manually place `dc.json` files with invalid fields.

src/MCPClient/lib/clientScripts/saveDublinCore.py
```diff
@@ -0,0 +1,43 @@
...
+    jsonified = {}
+    try:
+        dc = models.DublinCore.objects.get(metadataappliestoidentifier=transfer_uuid)
+    except:
+        # There may not be any DC metadata for this transfer, and that's fine
+        print >> sys.stderr, "No DC metadata found; skipping"
+        return 0
+    for field in FIELDS:
+        attr = getattr(dc, field)
+        if attr:
```
mistydemeo, line 30: Can `getattr()` fail to return a result here? Not sure how Django models would deal w/ that. This does at least prevent writing empty strings to the JSON.

src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
```diff
@@ -63,6 +63,15 @@
     #move the objects to the SIPDir
     for item in os.listdir(objectsDirectory):
         shutil.move(os.path.join(objectsDirectory, item), os.path.join(tmpSIPDir, "objects", item))
+    # Move the metadata/ directory, too
+    metadata_dir = os.path.join(objectsDirectory, '..', 'metadata')
```
Hwesta: `os.path.normpath` might be useful here.

src/MCPClient/lib/clientScripts/loadDublinCore.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import sys
+
+path = '/usr/share/archivematica/dashboard'
+if path not in sys.path:
+    sys.path.append(path)
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
+from main import models
+
+INGEST_METADATA_TYPE = '3e48343d-e2d2-4956-aaa3-b54d26eb9761'
```
Hwesta, line 14: Add a comment indicating this is the UUID of 'SIP' from MetadataAppliesToTypes table, please.

src/MCPClient/lib/clientScripts/loadDublinCore.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def main(sip_uuid, dc_path):
+    # If there's no metadata, that's not an error, and just keep going
+    if not os.path.exists(dc_path):
+        print "DC metadata not found; exiting", "(at", dc_path, ")"
```
Hwesta: Why the extra `", "` between 'exiting' and '(at'?

src/MCPClient/lib/clientScripts/saveDublinCore.py
Hwesta, line 30: getattr raises an exception if no default value is provided - you probably want `getattr(dc, field, None)` or to handle the exception.  Using `models.DublinCore._meta.get_all_field_names()` instead of `FIELDS` might be useful as well, since it'll never return a field that doesn't exist (though you may not want all the fields - i.e. is_part_of (related to AIC work) isn't listed, and I assume that's intentional?

src/dashboard/src/components/transfer/views.py
```diff
@@ -141,3 +143,56 @@ def delete(request, uuid):
...
+            dc = models.DublinCore.objects.get(metadataappliestotype__exact=1,
+                    metadataappliestoidentifier__exact=uuid)
+            return redirect('components.transfer.views.transfer_metadata_edit', uuid, dc.id)
+        except ObjectDoesNotExist:
+            dc = models.DublinCore(
+                metadataappliestotype=transfer_metadata_type_id(),
+                metadataappliestoidentifier=uuid
+            )
+
+    fields = ['title', 'creator', 'subject', 'description', 'publisher',
```
Hwesta, line 177: Since we've defined the fields list a few times (here, saveDublinCore.py, ingest.forms.AICDublinCoreMetadataForm/DublinCoreMetadataForm), maybe we should make it an attribute of DublinCore model?  Several of the fields lists vary on whether they include is_part_of or not, though.

Hwesta:
It seems a shame to make almost identical templates for metadata_{edit,list}.html, but there's already a precedent set with microservices.html.  Not sure of an easy way to fix it, though, other than passing in all the links that differ, and that gets awkward quickly.

src/MCPClient/lib/clientScripts/saveDublinCore.py
mistydemeo, line 30: Yes, I left out `is_part_of` here because I only want the standard DC fields, not our extensions. `models.DublinCore._meta.get_all_field_names()` includes a few other undesirable fields, too (`id`, `metadataappliestoidentifier`, `metadataappliestotype`).

src/dashboard/src/components/transfer/views.py
mistydemeo, line 177: I think a `DC_TERMS` that includes the standard DC terms set would make sense; `is_part_of` should only get added on a case-by-case basis.

Hwesta, line 177: Sounds good, and that can be used in places where `_meta.get_all_field_names()` returns the extras we don't want.

mistydemeo:
Squashed and rebased.

I have no memory of why 223b8118838df063c648c1bc4305dc239b95476f is in this branch, but the method is no longer called anywhere so I have my suspicions it's totally unnecessary.

mistydemeo:
The only place it was called was this entry in urls, which is commented out: https://github.com/artefactual/archivematica/blob/qa/1.x/src/dashboard/src/components/filesystem_ajax/urls.py#L25 And has been since August: https://github.com/artefactual/archivematica/commit/4186da91d241d41c8e0972e97478a4b0d8072962#diff-222af4b335ab77a072b7bd5d18ed3960L23

mistydemeo:
Repushed without that commit.

PR #20, opened 2014-03-13
Switch from openfits to FITS

mistydemeo:
The currently-packaged version of FITS is 0.8.0.

This also updates the XML file to use the FITS 0.8.0 version while retaining our custom settings, and disables checksum generation (which was not an option before).


Hwesta:
Don't know enough to comment on the fits.xml changes, but looks fine.

PR #22, opened 2014-03-14
ReplacementDict: add constructor from DB data

mistydemeo:
This new constructor allows a ReplacementDict to be constructed by passing in rows from the database. This allows places outside of the Unit classes to gain access to the full suite of standard unit variables without having to fully reimplement the logic.

Eventually the three Unit classes can probably be migrated to use this constructor, ensuring that there's only one canonical way to produce this style of ReplacementDict.

refs #6451


src/MCPServer/lib/passClasses.py
```diff
@@ -22,12 +22,109 @@
...
+                sipdir = sip.currentpath.replace('%sharedPath%', shared_path)
+
+            rd['%SIPUUID%'] = sip.uuid
+            sip_name = os.path.basename(sipdir[:-1]).replace('-' + sip.uuid, '')
+            rd['%SIPName%'] = sip_name
+            rd['%currentPath%'] = sipdir
+            rd['%SIPDirectory%'] = sipdir
+            rd['%SIPDirectoryBasename%'] = os.path.basename(os.path.abspath(sipdir))
+            rd['%SIPLogsDirectory%'] = sipdir + 'logs/'
+            rd['%SIPObjectsDirectory%'] = sipdir + 'objects/'
```
Hwesta: `os.path.join(sipdir, 'objects', '')` ?

mistydemeo: Didn't realize adding an empty string would result in a trailing `/`!

src/MCPServer/lib/passClasses.py
```diff
@@ -22,12 +22,109 @@
...
+            current_location = file_.currentlocation.replace('%transferDirectory%', base_location)
+            rd['%currentLocation%'] = current_location
+            rd['%fileGrpUse%'] = file_.filegrpuse
+            if type_ == 'file':
+                rd['%relativeLocation%'] = current_location
+
+        rd['%processingDirectory%'] = config.get('MCPServer', "processingDirectory")
+        rd['%checksumsNoExtension'] = config.get('MCPServer', "checksumsNoExtention")
+        rd['%watchDirectoryPath%'] = config.get('MCPServer', "watchDirectoryPath")
+        rd['%rejectedDirectory%'] = config.get('MCPServer', "rejectedDirectory")
```
Hwesta: How expensive is config.get()?  Is it worthwhile looking into memoizing it somewhere?  (Probably a new bug)

mistydemeo: Good question, I'm not sure. Memoizing it would probably also best happen at the same time we separate it away from `archivematicaMCP`, so we can parse it in a central place that can be imported by more of Archivematica.

src/MCPClient/lib/clientScripts/normalize.py
```diff
@@ -53,6 +56,11 @@ def get_replacement_dict(opts):
         print("Unsupported command purpose", opts.purpose, file=sys.stderr)
         return None
 
+    # Populates the standard set of unit variables, so,
+    # e.g., %fileUUID% is avialable
```
Hwesta: Typo avialable -> available

Hwesta:
I'm hesitant to put a direct dependency between the MCPServer and MCPClient.  We're looking at making them different packages (eventually) with common dependencies on archivematicaCommon.  Perhaps move passClasses to archivematicaCommon (and rename it??), and import from there?  Both MCPServer and MCPClient already add archivematicaCommon to their path lots of places.

src/MCPServer/lib/passClasses.py
```diff
@@ -22,12 +22,107 @@
...
+from main import models
+
+# TODO This probably shouldn't be directly read here, but this file can't
+#      safely import archivematicaMCP because it has too many side effects.
+config = ConfigParser.SafeConfigParser({'MCPArchivematicaServerInterface': ""})
+config.read("/etc/archivematica/MCPServer/serverConfig.conf")
+
+
+def interpolate_string_values(string, **kwargs):
+    rd = ReplacementDict.frommodel(**kwargs)
```
Hwesta: Docstring?  Especially for **kwargs values

Hwesta:
Also, tests, since we're trying to actually do that?  It uses Django's ORM so it shouldn't be completely impossible.

mistydemeo:
> I'm hesitant to put a direct dependency between the MCPServer and MCPClient. We're looking at making them different packages (eventually) with common dependencies on archivematicaCommon. Perhaps move passClasses to archivematicaCommon (and rename it??), and import from there?

That makes a lot of sense to me; it already has no coupling to the rest of MCPServer. I'll move it to archivematicaCommon, and rename it to something that makes more sense than `passClasses`.

mistydemeo:
Moved to archivematicaCommon under the module name `dicts`. `dicts.ReplacementDict`, etc. has a reasonable ring to me, especially when namespaced under `archivematicaCommon`.

mistydemeo:
Updated with tests. The string argument form is currently untested (because that would need to mock the DB and/or have a test database, which we don't do yet), but various forms of the model argument form are.

mistydemeo:
Cherry-picked in a minor improvement from the characterize file branch; when passing in only a file, it will now try to find the sip via join.

Added a few other RD-related improvements: added a docstring to `fromstring`, and made `replace`'s docstring doctest-compatible. fromstring doesn't get one due to Reasons.

src/archivematicaCommon/lib/dicts.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/python -OO
+
+# This file is part of Archivematica.
```
Hwesta, line 3: I don't think we need the extensive copyright text in each file anymore.

mistydemeo, line 3: I'm glad to remove it in another commit! Let's discuss with Justin whether we need it. I didn't add it here, it's just showing new due to the rename.

Hwesta:
Looks great!  I'm very happy about the tests too.

mistydemeo:
Great! I have a few other branches depending on this, so I'm going to merge it ASAP.

mistydemeo:
Added one new commit which moves several of the synonyms and file-related variables from normalize into ReplacementDict, so they're available to other microservices. Hopefully this makes things like "%fileFullPath%", which only worked in normalization, less mysterious.

mistydemeo:
One more improvement I want to make before this merges:

I realized that there's a slight difference in how Unit classes treat these, and how other microservices do. Unit classes process dicts with paths that still reference common variables like %sharedPath%; that only gets expanded by the time it hits the client. On the other hand, client scripts are run via the client and require fully-expanded paths.

I'm thinking of adding another kwarg which controls whether paths are expanded. Once the unit classes are converted to use this dict, they can elect not to have path expansion.

mistydemeo:
@Hwesta Thoughts on making `expand_paths` on or off by default?

Hwesta:
1. false = closest to old behavior
2. true = least surprising, since I'd expect to get back useful paths not ones that still need replacing
3. make a required parameter = caller always has to think about what they're getting.

Not a strong opinion, but defaulting to true seems best.

mistydemeo:
> make a required parameter = caller always has to think about what they're getting.

I have Opinions about making boolean arguments non-keyword arguments. ;)

I'll make it on by default, since that's most useful outside the Unit classes.

mistydemeo:
Merged.

PR #23, opened 2014-03-17
No longer overwrite FITS configuration XML

mistydemeo:
In the past, the default FITS configuration XML was installed via its own package, and Archivematica would then overwrite it in its own install.

This custom configuration has been moved to our FITS package, and Archivematica no longer stomps over it. This should also ensure that FITS still works after Archivematica is uninstalled.


Hwesta:
Can/does this remove externals/fits/archivematicaConfigs/ directories?  Other than that, looks good.

mistydemeo:
This does - directories in git only exist insofar as files exist in them. Git doesn't track directories separately, so that directory structure will no longer exist as of this commit.

PR #25, opened 2014-03-24
Add link to download pointer files

Hwesta:
Add link to download pointer files from Archival Storage tab.  Proxy to storage service API URL that actually returns the data.  Minor cleanup in templates to use flexible URLs.

refs #5716, relies on artefactual/archivematica-storage-service/#5


mistydemeo:
:thumbsup:

PR #26, opened 2014-03-26
Improve dev-helper "update/install package requirements" task

mistydemeo:


localDevSetup/installDependsFromDebianFile.py
```diff
@@ -39,14 +39,23 @@
 while not line.startswith("Depends:"):
     line = f.readline()
 
-
+# Depends: statements can span lines
+followup = f.readline()
+while followup.startswith((' ', '\t')):
```
Hwesta, line 44: `startswith(string.whitespace)` ?  Not that it matters as much since this is a helper script.

mistydemeo, line 44: This isn't a regex - it takes a string for an exact match or a tuple. `string.whitespace` is a string (`'\t\n\x0b\x0c\r '`), so it would have to be modified to work here.

mistydemeo, line 44: I guess it could be

``` python
startswith(tuple(string.whitespace))
```

Hwesta:
Looks fine.

PR #28, opened 2014-04-04
Update pointer file

Hwesta:
Updated the pointer file to work better with some LOCKSS requirements (though this is an independent change).
1. Add compression event and Archivematica agent
2. Move ADMID to structMap

Also did some cleanup in archivematicaXMLNamesSpace and createPointerFile, since I've learned some about lxml and namespaces since I originally wrote it. (Notably that () around an ElementMaker structure are unnecessary, hence a lot of whitespace changes)

refs #6500

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaXMLNamesSpace.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaXMLNamesSpace.py
@@ -1,42 +1,27 @@
 #!/usr/bin/python -OO
-#
-# This file is part of Archivematica.
-#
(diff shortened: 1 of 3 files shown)


mistydemeo:
Looks good. Is there anything that consumes this pointer file outside of Archivematica? Just making sure we don't accidentally break anything.

Hwesta:
Pointer file is only used by the storage service currently, and even then it doesn't pay attention to events. ;P

mistydemeo:
Ha, fair enough then!

PR #29, opened 2014-04-11
Fix approve transfer API

Hwesta:
MicroServiceChainChoice had been updated to have a foreign key to MicroServiceChain, but not all cases that used chainavailable were updated.  Fixed the last one.

Other cleanup while fixing this:
- Add instructions to API whitelist template specifying the format
- Whitelisted IPs for API are now whitespaces separated, instead of \r\n
- Clean up import list
- Add logging
- Add unicode function to MicroServiceChain & MicroServiceChainChoice

Supports artefactual/archivematica-storage-service#7

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -209,9 +209,8 @@ def approve_transfer_via_mcp(directory, type, user_id):
                 # attempt to find appropriate choice
                 chain_to_execute = None
                 for choice in choices:
-                    chain = models.MicroServiceChain.objects.get(pk=choice.chainavailable)
-                    if chain.description == 'Approve transfer':
-                        chain_to_execute=chain.pk
+                    if choice.chainavailable.description == 'Approve transfer':
+                        chain_to_execute=choice.chainavailable.pk
 
                 # execute choice if found
                 if chain_to_execute != None:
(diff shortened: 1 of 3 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -15,13 +15,26 @@
...
 from tastypie.authentication import ApiKeyAuthentication
+
+# This project, alphabetical
 from contrib.mcp.client import MCPClient
-from main import models
 from components import helpers
+from main import models
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(filename="/tmp/archivematicaDashboard.log",
```
mistydemeo: I wonder if we should wait on cleaning up the logging permissions before adding more logs that hardcode to `/tmp/archivematicaDashboard.log`? The more we do this, the less it becomes possible to run tests as anything other than the archivematica user, since simple imports can run into permissions issues.

mistydemeo:
Looks good aside from my logging comment.

src/dashboard/src/components/api/views.py
Hwesta: Since #34 exists now, whichever one gets merged first can be updated to fix the logging here.  Sound reasonable?

Hwesta:
Removed logging module and config, but left the comment out logger.foo lines.

PR #30, opened 2014-04-17
Add SIP creation panel

Hwesta:
Add the SIP creation panel, which allows arranging files from transfers into SIPs.  Partially described by https://www.archivematica.org/wiki/Transfer_and_SIP_creation#Mockups
- arrange files from Transfers to a new SIP by storing new paths in SIPArrange table.  On create SIP, POST to SS to move from old path to specified path
- SIP arrangement panel
  - left side shows files from backlogged transfers, populated by the backlog search from ES
  - right side shows arranged SIPs, created by creating directories, dragging transfer files over and rearranging them.  Any folder can be selected to create a new SIP
  - logs and metadata cannot be copied over, though they can be viewed
- delete transfers whose contents have been completely stored

Other Features:
- run syncdb by dev-helper
  - to create SIPArrange table
- failed SIP hook
  - sets files in failed SIP available to be arranged again
- backlog now uses SS (artefactual/archivematica-storage-service#9)
- added SIP restructure for files with UUIDs after failing verification, for newly created arranged SIPs that don't have the expected structure
- separation of download from SS vs download from local FS (ie SIP Arrangement)
- cleanup
  - restructure for compliance to files that actually called them
  - required directories and optional files for SIP verification and restructuring moved to common
  - SS url generation centralized to storageService.py helper module
  - remove now unneeded backlog search page - backlog search populates the SIP arrangement panel

Requires artefactual/archivematica-storage-service#9

Still needs more history cleanup.


mistydemeo:
You already know this, so just marking down so I remember: the failed SIP hook should probably take https://projects.artefactual.com/issues/6636 into account. I imagine weird things would happen if a rejected DIP ends up being slurped back as a SIP for arrangement?

mistydemeo:
> backlog now uses SS (artefactual/archivematica-storage-service#9)

Yay! More and more stuff that should be using the SS is using it :D

src/MCPClient/lib/clientScripts/failedSIPCleanup.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/python2 -OO
+
+import argparse
+import os
+import sys
+
+path = '/usr/share/archivematica/dashboard'
+if path not in sys.path:
```
mistydemeo, line 8: I've seen this pattern a few times: just out of curiosity, do we anticipate these client scripts being imported by something that is already using dashboard paths?

src/MCPClient/lib/clientScripts/indexAIP.py
```diff
@@ -10,13 +10,14 @@
 import databaseInterface
 import elasticSearchFunctions
 from executeOrRunSubProcess import executeOrRun
+import storageService as storage_service
```
mistydemeo, line 13: Outside the scope of this PR, but let's rename the module at some point!

src/MCPClient/lib/clientScripts/moveToBacklog.py
```diff
@@ -0,0 +1,66 @@
...
+    sys.path.append(path)
+import storageService as storage_service
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(filename="/tmp/archivematica.log",
+    level=logging.INFO)
+
+
+def main(transfer_uuid, transfer_path):
+    current_location = storage_service.get_location(purpose="CP")[0]
```
mistydemeo, line 19: Do we have a constant for this purpose?

src/MCPClient/lib/clientScripts/restructureForCompliance.py
```diff
@@ -23,130 +23,25 @@
...
-            if item == "manifest.txt":
-                destDir = "metadata/submissionDocumentation"
-            dst = os.path.join(unitPath, destDir, item)
-            updateFileLocation2(src, dst, unitPath, unitIdentifier, unitIdentifierType, unitPathReplaceWith)
-            files = getFileUUIDLike(dst, unitPath, unitIdentifier, unitIdentifierType, unitPathReplaceWith)
-            for key, value in files.iteritems():
-                fileUUID = value
-                updateFileGrpUse(fileUUID, "TRIM metadata")
 
-def restructureBagForComplianceFileUUIDsAssigned(unitPath, unitIdentifier, unitIdentifierType, unitPathReplaceWith = "%transferDirectory%"):
```
mistydemeo, line 87: :D :D :D

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -23,9 +23,20 @@
 
 import lxml.etree as etree
 import os
-import sys
 import re
 
+REQUIRED_DIRECTORIES = [
```
mistydemeo, line 28: :thumbsup:

src/dashboard/src/components/filesystem_ajax/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # if entry is a directory, recurse
+        child_path = os.path.join(path, file)
+        if new_entry != None and os.path.isdir(child_path) and os.access(child_path, os.R_OK):
+            directory_to_dict(child_path, directory, new_entry)
+
+    # return fully traversed data
+    return directory
+
+def directory_contents(path, contents=[]):
```
mistydemeo: IIRC I deleted this in another branch, possibly in qa/1.x - just FYI in case you get a merge conflict from it.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -342,96 +283,382 @@ def copy_to_start_transfer(request):
...
+    shared_dir = helpers.get_server_config_value('sharedDirectory')
+    staging_sip_path = staging_sip_path.lstrip('/')
+    staging_abs_path = os.path.join(shared_dir, staging_sip_path)
+
+    # Copy default processing config
+    default_config_path = helpers.default_processing_config_path()
+    sip_config_path = os.path.join(staging_abs_path, 'processingMCP.xml')
+    shutil.copyfile(default_config_path, sip_config_path)
+
+    # # Remove pre-normalize file format ID, may not have been ID'd yet
```
mistydemeo: Planning to keep or remove this?

src/dashboard/src/main/models.py
```diff
@@ -227,6 +227,25 @@ class Transfer(models.Model):
     class Meta:
         db_table = u'Transfers'
 
+
+class SIPArrange(models.Model):
+    # original_location = models.CharField(max_length=36)
```
mistydemeo: Remove the commented out line?

mistydemeo: So this model represents a file in an arranged SIP?

src/dashboard/tests/test_filesystem_ajax.py
```diff
@@ -0,0 +1,54 @@
+#! /usr/bin/python2
```
mistydemeo, line 1: Yay, tests! :sparkling_heart:

mistydemeo:
This looks good! I admit I skimmed the JS, but otherwise looks like it's in good shape

mistydemeo:
Can the same file(s) appear in multiple SIPs?

src/MCPClient/lib/clientScripts/failedSIPCleanup.py
Hwesta, line 8: I noticed that the sys.path is often cluttered with multiple instances of '/usr/share/archivematica/dashboard', '/usr/lib/archivematica/archivematicaCommon' etc having been added, and was hoping to prevent that (what happens if sys.path gets absurdly large?).  It seems less useful in client scripts as they're probably run once and don't have this problem. I go back and forth on whether it's a worthwhile pattern to cart around.

src/MCPClient/lib/clientScripts/indexAIP.py
Hwesta, line 13: I originally named it storageService to match with everything else in archivematicaCommon, but import it as storage_service to make myself happy. :P PEP8 is conflicting - recommends pothole, but also says go with surrounding code style.  Would like to rename though!

src/MCPClient/lib/clientScripts/moveToBacklog.py
Hwesta, line 19: Yes, but it's in the storage service and not import-able here. :( Should I redefine some SS constants here?  In archivematicaCommon/storageService? Is there a better solution?

src/dashboard/src/components/filesystem_ajax/helpers.py
Hwesta: Thanks, deleted directory_contents in my branch as well.

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta: Removed it, since I don't think the processingConfig changes are needed after all, as SIP
creation requires the transfers to be on the same pipeline as the new SIP currently. That will eventually be replaced by parsing a transfer METS file into the DB, so all the file  information and events will be there, and file ID won't have to be redone.

Hwesta:
Updated failedSIPCleanup to run before the (likely to fail) move to failed directory.  Since the post-failed SIP hook only touches the SIPArrange table, DIPs should be unaffected.

The same file should not appear in multiple SIPs. Any entry in SIPArrange with the same original_path greys out the corresponding entry in the originals panel.  Once an AIP is created, store_aip checks to see if all files in a backlogged transfer are in AIPs (as tracked by the SIPArrange table), and submits a delete request if that's the case.  It's possible to get around these restrictions, but casual error should be avoided like this.

Hwesta:
Rebased on current qa/1.x, with updates mentioned above.

PR #31, opened 2014-04-21
Add new transcribe files microservice

mistydemeo:
This adds a new "transcribe files" microservice. The primary initial usecase is OCR. Transcription commands are FPR commands, like normalization. OCR should be saved into a SIP in objects/metadata/OCRfiles.

OCR data will be included in a DIP if it exists, and will be copied into an `OCRfiles` directory at the root of the DIP.

Requirements are here: https://www.archivematica.org/wiki/OCR_text_in_DIP

This depends on #22.

~~Note that this isn't 100% ready to merge: I need to unify the post-normalization chains first. Currently transcribeFile is only called in one of them.~~

### DIFF
--- a/src/MCPServer/share/mysql_dev1.sql
+++ b/src/MCPServer/share/mysql_dev1.sql
@@ -492,4 +492,11 @@ INSERT INTO MicroServiceChainLinks(pk, microserviceGroup, defaultExitMessage, cu
 
 INSERT INTO MicroServiceChainLinksExitCodes (pk, microServiceChainLink, exitCode, nextMicroServiceChainLink, exitMessage) VALUES ('22ebafb1-3ec3-406a-939d-4eb9f3b8bbd1', @transcribeChoiceMSCL, 0, @transcribeFileMSCL, 'Completed successfully');
 INSERT INTO MicroServiceChainLinksExitCodes (pk, microServiceChainLink, exitCode, nextMicroServiceChainLink, exitMessage) VALUES ('804d4d23-e81b-4d81-8e67-1a3b5470c841', @transcribeFileMSCL, 0, @postTranscribeMSCL, 'Completed successfully');
+
+-- Copy OCR data into DIP
+INSERT INTO StandardTasksConfigs (pk, requiresOutputLock, execute, arguments) VALUES ('5c4f877f-b352-4977-b51b-53ebc437b08c', 0, 'copyRecursive_v0.0', '"%SIPObjectsDirectory%metadata/OCRfiles" "%SIPDirectory%DIP"');
+INSERT INTO TasksConfigs (pk, taskType, taskTypePKReference, description) VALUES ('102cd6b0-5d30-4e04-9b62-4e9f12d74549', '36b2e239-4a57-4aa5-8ebc-7a29139baca6', '5c4f877f-b352-4977-b51b-53ebc437b08c', 'Copy OCR data to DIP directory');
+INSERT INTO MicroServiceChainLinks(pk, microserviceGroup, defaultExitMessage, currentTask, defaultNextChainLink) values ('43c72f8b-3cea-4b4c-b99d-cfdefdfcc270', 'Prepare DIP', 'Failed', '102cd6b0-5d30-4e04-9b62-4e9f12d74549', '7d728c39-395f-4892-8193-92f086c0546f');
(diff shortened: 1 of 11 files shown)


mistydemeo:
> I need to unify the post-normalization chains first.

~~Done in #6; that should be merged before this.~~ Merged!

mistydemeo:
Made numerous updates here in order to facilitate the "transcribe derivatives" usecase, and to fix the AIP METS document.

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -98,6 +98,13 @@ def index(request):
             "yes_option":   "9efab23c-31dc-4cbd-a39d-bb1665460cbe", # Store AIP
             "action":       "Store AIP"
         },
+        {
+            "name":         "transcribe_file",
```
Hwesta, line 99: Are you going to set a default in processingMCP?

src/MCPServer/share/mysql_dev1.sql
```diff
@@ -387,3 +387,47 @@ INSERT INTO MicroServiceChainLinks(pk, microserviceGroup, defaultExitMessage, cu
...
+SET @postTranscribeMSCL = 'f574b2a0-6e0b-4c74-ac5b-a73ddb9593a0' COLLATE utf8_unicode_ci;
+
+INSERT INTO TasksConfigs (pk, taskType, taskTypePKReference, description) VALUES (@transcribeChoiceTC, '9c84b047-9a6d-463f-9836-eafa49743b84', NULL, 'Transcribe SIP contents');
+INSERT INTO MicroServiceChainLinks (pk, microserviceGroup, defaultExitMessage, currentTask, defaultNextChainLink) VALUES (@transcribeChoiceMSCL, 'Normalize', 'Failed', @transcribeChoiceTC, @transcribeFileMSCL);
+UPDATE MicroServiceChainLinks SET defaultNextChainLink=@transcribeChoiceMSCL WHERE pk=@preTranscribeMSCL;
+UPDATE MicroServiceChainLinksExitCodes SET nextMicroServiceChainLink=@transcribeChoiceMSCL WHERE microServiceChainLink=@preTranscribeMSCL;
+
+SET @transcribeYes = '5a9985d3-ce7e-4710-85c1-f74696770fa9' COLLATE utf8_unicode_ci;
+SET @transcribeNo = '1170e555-cd4e-4b2f-a3d6-bfb09e8fcc53' COLLATE utf8_unicode_ci;
+INSERT INTO MicroServiceChoiceReplacementDic (pk, choiceAvailableAtLink, description, replacementDic) VALUES (@transcribeYes, @transcribeChoiceMSCL, 'Yes', '{"%transcribe%": "True"}');
```
Hwesta, line 485: Any particular reason you used replacement dicts instead of a new chain here?  It seems a little odd that the transcribe microservice will still run (even if it does nothing) if No is selected.

Hwesta:
Do any OCR tools need to be added to the dependency list?  Check with Justin, but we seem to be adding FPR package dependencies to Archivematica.

mistydemeo:
Yes, good point - I'll add tesseract-ocr to the MCPClient deps.

src/MCPServer/share/mysql_dev1.sql
mistydemeo, line 485: IIRC, there were some unit variables I didn't want to drop, and/or didn't want to add a microservice to preserve. Alternately, I like to sidestep the watch directory / new chain dance where possible.

src/dashboard/src/components/administration/views_processing.py
mistydemeo, line 99: After discussion with Courtney, preconfigured to "no".

src/MCPServer/share/mysql_dev1.sql
Hwesta, line 485: Unit variables is understandable.  Any time we stop for a user choice we should probably have a WD/new chain.  If AM is restarted while waiting at the question, it will restart the chain based on where the SIP is currently on disk.  If it's in currently processing dir, it'll just fail, otherwise it'll start up again at the last WD it was in - which may mean re-running MSCLs that shouldn't be run multiple times.  Where the SIP is currently/how likely the choice is to be automated/can the MSCLs of the previous WD be rerun safely/how much we care all can factor in, though in practice we tend to have a new WD at every choice.

PR #32, opened 2014-04-24
JSON metadata conversion: support repeated columns

mistydemeo:
This allows arrays to be provided as arguments to an object key in the JSON metadata. If an array is present, then the column will be repeated multiple times in the output CSV to provide room for multiple results for compatibility with how we support repeated values in CSV.

Sample JSON input:

``` json
[
    {
        "filename": "objects/foo.txt",
        "dc.title": "This is an object",
        "dc.subject": ["Subject 1", "Subject 2"]
    }
]
```

Which produces:

``` csv
filename,dc.subject,dc.subject,dc.title
objects/foo.txt,Subject 1,Subject 2,This is an object
```

refs #6266

### DIFF
--- a/src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
+++ b/src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
@@ -7,9 +7,27 @@ import sys
 
 
 def fetch_keys(objects):
+    """
+    Returns a list of keys in a set of dicts suitable for use as a
+    CSV header row.
+    If values in the dict are lists, then the header row will contain
+    N occurrences of the key, where N is the largest list under that
+    key in the set of dicts.
+    """
+    # Track the occurrence of each set of keys; we'll need to produce
+    # one column for each occurrence of that key if its value in the
+    # JSON object is an array.
+    occurrence_count = {}
     keys = set()
-    for object in objects:
-        keys.update(object.keys())
+
+    for o in objects:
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
```diff
@@ -7,9 +7,27 @@
...
     keys = set()
-    for object in objects:
-        keys.update(object.keys())
+
+    for o in objects:
+        o_keys = o.keys()
+        for key in o_keys:
+            if isinstance(o[key], list):
+                occurrence = len(o[key])
+                if not key in occurrence_count or occurrence_count[key] < occurrence:
```
Hwesta: Nitpick:  `if key not in occurrence_count ...` is easier to parse than `if not key in occurrence_count ...` (and more idiomatic I think)

src/MCPClient/tests/test_json_conversion.py
```diff
@@ -15,3 +18,15 @@ def test_json_csv_conversion(tmpdir):
         csvdata = csvfile.read()
 
     assert csvdata == CSV
+
+
+def test_json_csv_conversion_with_repeated_columns(tmpdir):
```
Hwesta, line 23: :thumbsup:

Hwesta:
Looks good.

mistydemeo:
Updated and merged.

PR #37, opened 2014-05-10
Enable viewing disabled processing config options (and dead code cleanup)

Hwesta:
Enable viewing/modifying configuration dropdowns even when the option is not enabled.  

Also cleaning up some dead code:
- Remove very deprecated transfer source & AIP storage JS, now replaced by the storage service.
- Remove baffling dead API code.
- Remove client scripts that are never referenced in the DB or client modules list
- Match transfer & ingest URLs from the beginning of the line


mistydemeo:
:thumbsup:

PR #38, opened 2014-05-13
Fix METS namespaces

Hwesta:
refs #6676

Fix the AIP METS to only declare the namespaces it's actually using, notably XSI, xlink, and METS as the default. Replace all `from archivematicaXMLNamesSpace import *` with explicit imports and update
references.


mistydemeo:
:thumbsup:

PR #40, opened 2014-05-26
Uncompressed AIP support

mistydemeo:


src/MCPClient/lib/clientScripts/compressAIP.py
```diff
@@ -35,6 +43,15 @@ def compress_aip(compression, compression_level, sip_directory, sip_name, sip_uu
     archive_path = '{name}-{uuid}'.format(name=sip_name, uuid=sip_uuid)
     uncompressed_location = sip_directory+archive_path
 
+    # Even though no actual compression is taking place,
+    # the location still needs to be set in the unit to ensure that the
+    # %AIPFilename% variable is set appropriately.
+    # Setting it to an empty string ensures the common
```
Hwesta, line 49: Is this comment still correct?  It looks like the AIP filename is being set to the uncompressed location, which is the folder the bagged (but not yet compressed) AIP is in.

src/MCPClient/lib/clientScripts/verifyAIP.py
```diff
@@ -24,20 +39,19 @@ def verify_aip():
...
+    if is_uncompressed_aip:
+        bag = aip_path
+    else:
+        try:
+            extract_dir = os.path.join(temp_dir, sip_uuid)
+        except:
+            print >>sys.stderr, 'Error extracting AIP at "{}"'.format(aip_path)
+            return 1
+
+        bag = extract_aip(aip_path, extract_dir)
```
Hwesta: Should this line be inside the try/except?  Also, we should probably just catch Exception, not everything.

Hwesta:
Looks good.

src/MCPClient/lib/clientScripts/compressAIP.py
mistydemeo, line 49: This is still correct; the AIP bag is already created in-place, e.g. the SIP becomes the AIP bag. Rather than move the entire SIP contents down a directory, I'm treating the SIP-bag-as-AIP as the uncompressed AIP.

src/MCPClient/lib/clientScripts/verifyAIP.py
mistydemeo: Fixed!

mistydemeo:
Rebased on qa/1.x post-SIP arrange.

src/MCPClient/lib/clientScripts/storeAIP.py
```diff
@@ -67,13 +67,20 @@ def store_aip(aip_destination_uri, aip_path, sip_uuid, sip_name, sip_type):
     elif sip_type == 'AIC':
         package_type = sip_type
 
+    # Uncompressed directory AIPs must be terminated in a /,
```
Hwesta, line 70: By experimentation, this doesn't do what it ought to. :( Uncompressed AIPs end up inside a folder of the same name.

mistydemeo, line 70: Hm, it was working before :(

mistydemeo, line 70: Was a dumb mistake: I was appending the `/` to `aip_path` (which is never sent to the storage service) instead of `relative_aip_path`. Fixed now.

Hwesta:
:+1:

PR #41, opened 2014-05-28
Display Archivematica version in administration

mistydemeo:
refs #6640

### DIFF
--- a/src/dashboard/src/templates/administration/version.html
+++ b/src/dashboard/src/templates/administration/version.html
@@ -1,0 +1,33 @@
+{% extends "layout_fluid.html" %}
+{% load breadcrumb %}
+
+{% block title %}Version{% endblock %}
+{% block h1 %}Version{% endblock %}
+{% block page_id %}Version{% endblock %}
+
+{% block content %}
+
+  <div class="row">
+
+    {% include "administration/sidebar.html" %}
+
+    <div class="span12">
+
+      <h3>About Archivematica</h3>
+
+      <table>
+        <tr>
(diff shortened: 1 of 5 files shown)


src/dashboard/src/templates/administration/version.html
```diff
@@ -0,0 +1,35 @@
+{% extends "layout_fluid.html" %}
+{% load breadcrumb %}
+
+{% block title %}Version{% endblock %}
+{% block h1 %}Version{% endblock %}
+{% block page_id %}Version{% endblock %}
+
+{% block js %}
```
Hwesta: Defining empty blocks isn't necessary for the templating language, unless you're explicitly saying you don't want to inherit from the parent (in which case a note to that effect will keep me from deleting this when I see it later).

Hwesta:
The original bug talked about listing the Agents - is that going to happen here or elsewhere?  (One whole page just for the version seems lonely...)

jhsimpson:
For 1.2.0 release, we should add the 'preservation system' Agent to this page.

mistydemeo:
Updated with agent info. Fields are out of alphabetical order because I figured it's better for version to come first.

I bumped AM's version number. Do we typically update the agent right in the DB, or does that happen elsewhere?

Hwesta:
The Agent's version currently gets set in the [installer](https://github.com/artefactual/archivematica/blob/dev/issue-6640-version/src/dashboard/src/installer/views.py#L62-L65), so changing the SQL is probably not necessary.

Looks good to merge.

PR #42, opened 2014-05-30
Add a few missing dependencies to MCPClient

mistydemeo:
Also adds a watched directory.

### DIFF
--- a/src/MCPClient/debian/control
+++ b/src/MCPClient/debian/control
@@ -9,12 +9,12 @@ Homepage: http://archivematica.org
 Package: archivematica-mcp-client
 Architecture: i386 amd64 
 Depends: ${shlibs:Depends}, ${misc:Depends}, archivematica-common,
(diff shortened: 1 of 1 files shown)


Hwesta:
If we're adding watched directories, there's a couple more that are missing (or should have their .gitignore updated to exclude everything in that directory except itself, instead of an empty file):
- watchedDirectories/system/createAIC
- watchedDirectories/workFlowDecisions/compressionAIPDecisions
- watchedDirectories/workFlowDecisions/extractPackagesChoice
- watchedDirectories/workFlowDecisions/metadataReminder

Also, `watchedDirectories/watchedDirectories` can be deleted, since there's no entry in the WatchedDirectories table for them:

mistydemeo:
We can't merge the whole branch here because bulk_extractor and sleuthkit won't be packaged for the release. I'll merge only the gitignore.

mistydemeo:
Merged everything else; the only commit left is for the bulk_extractor and sleuthkit deps that we can't actually do right now.

PR #43, opened 2014-05-30
Remove unused dev-helper-committer

mistydemeo:
This helper script was never completed, and it doesn't look like there's much interest in using it.

refs #4757


Hwesta:
:+1:

PR #44, opened 2014-05-30
dev-helper: be less incredulous about clearing transfer backlog

mistydemeo:
Every dev-helper question (`Would you like to ${part}?`) already includes a question mark. Including the question mark in the `$part` string meant there were two question marks.


Hwesta:
:-1:  You should be incredulous about deleting AIPs and backlogged transfers!!!

:P

PR #45, opened 2014-05-30
Characterization: update FITS command

mistydemeo:
This deactivates the old placeholder FITS command, and adds a new one that calls FITS using the new characterization feature.

This same SQL should also get run on the FPR server.


mistydemeo:
This was generated by the new tool in #46.

Hwesta:
:+1: 

It would be nice if the update only listed changed fields, to make it easier to read, but that's a tool feature request.

mistydemeo:
I'd like that too, but I'm not writing the SQL - it's being generated by Django in this case. Don't think I can influence the SQL that Django generates!

Hwesta:
Ha, can't argue with that!

PR #46, opened 2014-05-31
FPR rule commandline tool

mistydemeo:
This pull request adds a new commandline tool which generates FPR rules. It immediately applies the change to the local database, and also emits SQL to stdout that can be captured to reapply the change at a later point in time. That new SQL can be added to `fpr_dev.sql` files, or applied to the FPR server.

This is still WIP, so please request features and etc.!

Planned features:
- Add new FPR formats.
- Name linked objects by their description instead of UUID.

### DIFF
--- a/src/archivematicaCommon/lib/utilities/add-format
+++ b/src/archivematicaCommon/lib/utilities/add-format
@@ -1,0 +1,1 @@
+add-fpr-rule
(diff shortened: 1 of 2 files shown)


mistydemeo:
The generated SQL isn't the prettiest in the world, but it's what Django generates when it creates the new rows in the DB and/or updates existing rows.

mistydemeo:
Pushed a commit which adds a new feature: specify source code to the requested script on the commandline, and it will read that file instead of prompting you for it. That way you can edit your scripts in a real text editor, and don't have to copy/paste them into the CLI.

mistydemeo:
Implemented adding new formats. This also creates the associated FormatGroup and Format entries, if they don't already exist.

mistydemeo:
Closing; this is moving to archivematica-devtools.

PR #48, opened 2014-06-02
CreateMETS2: fix conversion of unicode documents into str

mistydemeo:
If an XML document string includes an encoding declaration, etree will refuse to parse the document if it's a unicode string. This fixes up the way it was converting into a bytestring before.


Hwesta:
:+1:

PR #50, opened 2014-06-06
Remove commandline utilities

mistydemeo:
Now that the archivematica-devtools repo exists, I think we're ready to remove these from the primary Archivematica repo and transfer them over.


Hwesta:
:+1:

mistydemeo:
Pushed.

PR #52, opened 2014-06-30
normalize: fall back to default rule for unidentified files

mistydemeo:


src/MCPClient/lib/clientScripts/normalize.py
```diff
@@ -273,37 +277,30 @@ def main(opts):
...
+        rule = get_default_rule(opts.purpose)
+        print(os.path.basename(file_.currentlocation), "not identified - falling back to default", opts.purpose, "rule")
+    else:
+        print('File format:', format_id.format_version)
         try:
-            rule = FPRule.active.get(purpose='default_'+opts.purpose)
-            print("No rule for", os.path.basename(file_.currentlocation),
-                "falling back to default", opts.purpose, "rule")
+            rule = FPRule.active.get(format=format_id.format_version,
+            purpose=opts.purpose)
```
Hwesta: Tiny nitpick - indent this one more level?

src/MCPClient/lib/clientScripts/normalize.py
```diff
@@ -273,37 +277,30 @@ def main(opts):
...
         except FPRule.DoesNotExist:
```
Hwesta, line 293: Also catch `FPRule.MultipleObjectsReturned`?  What is returned if `format_id.format_version` is `None`?  I think it's multiple objects (and that shouldn't happen) but we should catch it.

mistydemeo, line 293: It actually returns `FPRule.DoesNotExist` - `format_id` is not nullable (and is a foreign key enforced at the database level), so it's impossible for a row to ever have a NULL `format_id`.

src/MCPClient/lib/clientScripts/normalize.py
mistydemeo: Updated!

Hwesta:
:+1:

PR #55, opened 2014-07-03
Manual Normalization normalization.csv matches subfolders

Hwesta:
refs #4922

Update manual normalization with normalization.csv to match original files based on the full path from the objects folder, not just the basename.  This allows original files to be stored in subfolders, and the normalization.csv to reference them there.


mistydemeo:
:thumbsup:

PR #56, opened 2014-07-04
Change extract option choices to yes/no

mistydemeo:
Per https://projects.artefactual.com/issues/6539 - the "Extract" and "Do not extract" chains should be renamed "Yes" and "No".


Hwesta:
:shipit:

PR #57, opened 2014-07-05
Minor fixes for 1.2

Hwesta:
Fixed three small bugs:
- #6816 Restore ES helper function that was refactored away
- #6611 Save the FPRule after updating the counts
- #6885 Only display zip files in the transfer component browser if the transfer type is zipped bag

### DIFF
--- a/src/dashboard/src/media/js/transfer/component_directory_select.js
+++ b/src/dashboard/src/media/js/transfer/component_directory_select.js
@@ -21,7 +21,7 @@ var components;
 var transferMetadataSetRowUUID;
 var transferDirectoryPickerPathCounter = 1;
 
-function createDirectoryPicker(locationUUID, baseDirectory, modalCssId, targetCssId, pathTemplateCssId) {
+function createDirectoryPicker(locationUUID, baseDirectory, modalCssId, targetCssId, pathTemplateCssId, foldersOnly) {
   var selector = new DirectoryPickerView({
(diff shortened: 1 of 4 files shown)


mistydemeo:
:thumbsup:

mistydemeo:
Looks like this needs rebasing.

src/dashboard/src/media/js/transfer/component_directory_select.js
```diff
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-function createDirectoryPicker(locationUUID, baseDirectory, modalCssId, targetCssId) {
+function createDirectoryPicker(locationUUID, baseDirectory, modalCssId, targetCssId, foldersOnly) {
```
mistydemeo: This function signature is getting out of hand, but I don't know that we have a clean way to deal with it!

Hwesta:
Rebased, and made a couple more changes
- Update transcoder.py to  use Django's [F()](https://docs.djangoproject.com/en/dev/ref/models/queries/#django.db.models.F) to prevent race conditions when updating counts
- Update JS to show `#transfer-name-container` (instead of `this`) if not zipped bag

jhsimpson:
I like the use of F(), I thought at first that was only available in django 1.7, but it works in 1.5.x also.

:thumbsup:

PR #58, opened 2014-07-07
MCPClient: add bulk_extractor, sleuthkit dependencies

mistydemeo:
These cannot be added as dependencies until we have packages available, so this shouldn't be pulled yet.

sleuthkit will be used for an FPR rule that will be pushed post-release. bulk_extractor is required for the "examine contents" microservice that already exists.

### DIFF
--- a/src/MCPClient/debian/control
+++ b/src/MCPClient/debian/control
@@ -9,12 +9,12 @@ Homepage: http://archivematica.org
 Package: archivematica-mcp-client
 Architecture: i386 amd64 
 Depends: ${shlibs:Depends}, ${misc:Depends}, archivematica-common,
- atool, bagit, clamav, clamav-daemon, ffmpeg (>= 7:2.1.0), fits (>= 0.8.0-3),
- gearman, imagemagick, inkscape, libxml2-utils, logapp,
+ atool, bagit, bulk_extractor, clamav, clamav-daemon, ffmpeg (>= 7:2.1.0),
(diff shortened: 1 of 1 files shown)


jhsimpson:
I just realized there is a small change required here - the bulk extractor package in ppa;archivematica/externals is called bulk-extractor , not bulk_extractor.  I think launchpad or debian does not allow package names with underscores in them?  

Perhaps this change can be done in the dev/issue-xxxx-1.2.0-release branch?

jhsimpson:
After testing this PR, it turns out that sleuthkit can't be installed with the current backported package.  On a 12.04 machine, it gives this error:

The following packages have unmet dependencies:
 sleuthkit : Depends: libewf2 (>= 20090606) but it is not installable
             Depends: libtsk10 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.

In order to get sleuthkit installed in 12.04, we are going to have to package libewf-dev , which requires installing gcc and tons of stuff, or libewf2, which appears to be just a meta-package, but I couldn't figure out how to backport it to precise.

Either way, it doesn't look like sleuthkit can be easily backported to 12.04, so I am not sure if this PR should really make it into the 1.2.0 release of Archivematica or not.

jhsimpson:
Upon further research, I determined that I was wrong, the externals-dev repo was not added on my test machine, so sleuthkit does in fact install when the correct ppa is used.

PR #59, opened 2014-07-07
Run virus scan on extracted files

Hwesta:
Add virus scan node after extracting files.  In clamscan script, check if a scan event already exists, and skip re-running clamscan if so.

Also, cleanup the clamscan script - remove unused imports & dead comments, fix whitespace, tweak output to be more useful, etc.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
@@ -26,12 +26,20 @@ import sys
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 from executeOrRunSubProcess import executeOrRun
 from databaseFunctions import insertIntoEvents
+import databaseInterface
 
 if __name__ == '__main__':
     fileUUID = sys.argv[1]
     target = sys.argv[2]
     date = sys.argv[3]
     taskUUID = sys.argv[4]
+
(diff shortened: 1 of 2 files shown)


Hwesta:
A couple of questions before this gets merged:
1. If virus scan fails now, the MSCL is listed as failed but processing continues.  Should it fail the transfer if virus scan fails?
2. Virus scanning is done on the whole SIP, not just the objects directory.  When running virus scan after extracting, should it be limited to just the objects directory?  Since events are not generated for processingMCP.xml and log files, they are scanned a second time currently.

mistydemeo:
> If virus scan fails now, the MSCL is listed as failed but processing continues. Should it fail the transfer if virus scan fails?

I think failing the transfer makes the most sense - safer than proceeding with a possibly bad transfer.

>  When running virus scan after extracting, should it be limited to just the objects directory? Since events are not generated for processingMCP.xml and log files, they are scanned a second time currently.

I think we discussed this in IRC, but since extracted contents always go in objects, we can safely limit it to just the objects directory.

mistydemeo:
:thumbsup:

Hwesta:
Updated all virus scan nodes to fail the SIP/Transfer if virus scan fails.  Added new StandardTasksConfig for clamscan that only runs on objects.

PR #60, opened 2014-07-08
Extract: skip prompts unless relevant

mistydemeo:
This adds a new microservice which checks to see if the transfer contains any compressed packages. If no extractable packages are present, then the prompts to extract and delete will be skipped.

Fixes #6350.


Hwesta:
:+1: 

Seems a little silly to add a MSCL to skip some MSCLs, but the code looks good.

PR #61, opened 2014-07-08
extractContents: specify filename in extraction event

mistydemeo:
Fixes #6576


Hwesta:
:+1:

mistydemeo:
Fixed event to specify the correct filename.

Hwesta:
Still :+1:

PR #62, opened 2014-07-10
FPRClient: fix paging API responses

mistydemeo:
Rather than try to fetch all items at once (which doesn't work for sets of results larger than 1000), rework the existing code into a simple fetch function and a generator that iterates over every item of the requested type. These are now fetched in batches of 100 and the "next" URL provided by the API response is used to page through all results.


src/archivematicaCommon/lib/utilities/FPRClient/getFromRestAPI.py
```diff
@@ -23,18 +23,37 @@
...
 import requests
 import sys
 
+
 class FPRConnectionError(Exception):
     pass
 
-def getFromRestAPI(url, resource, params, verbose=False, auth=None, verify=True):
+
+def getFromRestAPI(resource="", params=None, url="https://fpr.archivematica.org/fpr/api/v2/", limit=0, start_at=None, verbose=False, auth=None, verify=True):
```
Hwesta: Docstring for this? What the parameters are and how this will behave is non-obvious. (I'm thinking interactions between resource and url)

src/archivematicaCommon/lib/utilities/FPRClient/getFromRestAPI.py
```diff
@@ -45,12 +64,31 @@ def getFromRestAPI(url, resource, params, verbose=False, auth=None, verify=True)
...
+
+    while response["meta"]["next"]:
+        next = urlparse.urljoin(base_url, response["meta"]["next"])
+        response = getFromRestAPI(url=next, verify=verify)
+
+        for object in response["objects"]:
+            yield object
 
 if __name__ == '__main__':
     parser = OptionParser()
```
Hwesta, line 153: Update this section to use each_record instead of getFromRestAPI? Useful for testing.

Hwesta:
Looks good!  Nice use of generator functions and urlparse (I hadn't heard of that library before).

Tests?  We don't have an easy way to specify vcrpy is a dependency, but it would work well here.

Out of scope for this, but do we want to move this out of utilities?  I'm not sure what benefit we get from keeping it there.

mistydemeo:
Added verbose docstrings. Prefixed `getFromRestAPI` with an underscore to mark it as private now.

I'll look at adding some vcr.py tests.

mistydemeo:
> Out of scope for this, but do we want to move this out of utilities? I'm not sure what benefit we get from keeping it there.

I am not clear why it's there in the first place! Was it seen as a commandline utility at one point?

src/archivematicaCommon/lib/utilities/FPRClient/getFromRestAPI.py
```diff
@@ -23,18 +23,75 @@
...
 import requests
 import sys
 
+
 class FPRConnectionError(Exception):
     pass
 
-def getFromRestAPI(url, resource, params, verbose=False, auth=None, verify=True):
+
+def _getFromRestAPI(resource="", params=None, url="https://fpr.archivematica.org/fpr/api/v2/", limit=0, start_at=None, verbose=False, auth=None, verify=True):
```
Hwesta: Since you're renaming it anyway, you could make it potholes instead of camels. :)  Totally optional.

Hwesta:
I can only assume that it was intended for command line use, but I really don't know.  Historical reasons?

src/archivematicaCommon/lib/utilities/FPRClient/getFromRestAPI.py
mistydemeo, line 153: Updated.

mistydemeo:
Renamed `_get_from_rest_api` to pothole case. Adjusted tests: now uses VCR both for the old `_get_from_rest_api` test and for a new `each_record` test, and converted `fprserver` into a constant.

Hwesta:
:+1:

mistydemeo:
Added test requirements to requirements.txt in f1955fbbb53c86b9781e4f56cea248cb2769c07a.

PR #63, opened 2014-07-11
Add new FPR formats, commands and tools

mistydemeo:
- Batch import of all PRONOM formats, as of PRONOM v74
- Add ffprobe, mediainfo, exiftool FPTools, commands and
  characterization rules


Hwesta:
We need to escape double quotes in FPCommands?  Isn't that handled by being inside single quotes?

Hwesta:
Too much SQL to review properly... it's probably fine.

mistydemeo:
> We need to escape double quotes in FPCommands? Isn't that handled by being inside single quotes?

Probably it is! This SQL was generated by Django, and I didn't alter it in any way - not sure if this is just Django being overcareful, or if it's actually necessary.

Hwesta:
Fair - let's assume Django isn't inserting spurious "s.

PR #64, opened 2014-07-15
Update FIDO command to use new format extensions

mistydemeo:
This updates the existing `fido` command to use `-loadformats` to read the new format extension file, teaching it about the new formats.


jhsimpson:
:thumbsup:

PR #69, opened 2014-07-18
Remove popup metadata reminder

mistydemeo:
This is no longer needed now that there are ways to ensure that there's an opportunity to input metadata before it's finalized.

Fixes #7013.


Hwesta:
:+1:

PR #71, opened 2014-07-18
createMETSMetadataCSV: use universal newlines

mistydemeo:
This fixes reading CSVs with unusual newlines, like \r.

Fixes #7015.


mistydemeo:
Should we use universal newlines for all places that read CSVs?

Hwesta:
Probably.

mistydemeo:
Updated commit to touch every place that reads CSV.

mistydemeo:
Also added comments so we remember why it's there.

Hwesta:
Haha, I was about to ask for comments explaining the U, but you committed it first.

:+1:

PR #72, opened 2014-07-18
Remove disabled 'Create SIP manually' option

Hwesta:
Workflow for SIP arrangement requires sending the transfer to backlog, so disabled 'Create SIP(s) manually' option is inaccurate and misleading.

refs #6965


mistydemeo:
:thumbsup: Might need to wait for #70 to merge though?

Hwesta:
Will probably actually require a new DB dump on that branch, sadly.

PR #73, opened 2014-07-18
Allow DIPs to be stored in the storage service

mistydemeo:


mistydemeo:
The related Storage Service branch has been merged, but not the Archivematica branch.

Hwesta:
Should probably add `src/MCPServer/share/sharedDirectoryStructure/www/DIPsStore/`, since that's the default location in the SS, and we do that for AIPStore.  Probably easiest to copy the file in `src/MCPServer/share/sharedDirectoryStructure/www/AIPsStore/transferBacklog/arrange/.gitignore`

Other than that, :+1:

mistydemeo:
Will merge Monday once we decide what to do with the DB stuff; this touches the DB.

PR #74, opened 2014-07-21
Move transcription to its own microservice

mistydemeo:
This moves it out of "normalize".


Hwesta:
:+1: If the release happens first, this will have to be updated, though.

mistydemeo:
This should happen before we release 1.2, but I guess I don't know if it comes before or after #70.

PR #76, opened 2014-07-21
databaseFunctions: remove debug print

mistydemeo:
Looks like this got committed by accident. It's been causing SQL statements to be printed into the stdout of microservice jobs by accident, which makes debugging a little noisier.


Hwesta:
:+1:

PR #77, opened 2014-07-22
Remove KryoFlux FIDO extension

mistydemeo:
This was officially included in PRONOM as of the v77 signature files, so we don't need our own custom ones anymore.


Hwesta:
Don't know much about FIDO, but looks good.

mistydemeo:
The new version of PRONOM includes these similar, but more detailed, signature definitions: http://apps.nationalarchives.gov.uk/pronom/fmt/655.xml http://apps.nationalarchives.gov.uk/pronom/fmt/656.xml So no more need for us to define an unofficial one.

I'll be posting another pull request to add the PRONOM v75/v77 formats, which includes the two new KryoFlux FormatVersions.

mistydemeo:
I'll merge this once we decide what we're doing with the new database stuff.

PR #79, opened 2014-07-24
get_transfer_file_info: only return precise matches

mistydemeo:
Elasticsearch's top match may not actually be an exact match, even if an exact match is in the set.

Fixes #7001


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -689,12 +689,23 @@ def get_transfer_file_info(field, value):
...
+        # the actual file being queried for; in that case only consider results
+        # where the value is an actual precise match.
+        filtered_results = [results for results in documents['hits']['hits']
+                            if results['_source'][field] == value]
+        result_count = len(filtered_results)
+        if result_count == 1:
+            results = filtered_results[0]['_source']
+        elif result_count > 1:
+            logging.warning('get_transfer_file_info returned %s results for query %s: %s (using first result)',
+                            result_count, field, value)
```
Hwesta, line 718: Results not getting set here, despite logs saying using first result.

mistydemeo, line 718: Fixed by moving `results = ...` outside the if statement and keeping only the `elif result_count > 1` case.

Also pushed a new commit that raises if no exact results are found.

mistydemeo:
We currently use `get_transfer_file_info` only in cases where we expect an exact result, so I feel comfortable raising in this case. But do you see using it in a more search-oriented context?

Hwesta:
Not familiar enough with how the rest of the ES code works, but it seems reasonable here.

PR #80, opened 2014-07-24
Improve Elasticsearch error handling in try_to_index

mistydemeo:
- Iterate the correct number of times, instead of half the number of specified times
- If indexing fails at the end, reraise the exception instead of silently swallowing it

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -287,18 +287,15 @@ def connect_and_index_aip(uuid, name, filePath, pathToMETS, size=None, aips_in_a
     try_to_index(conn, aipData, 'aips', 'aip')
 
 def try_to_index(conn, data, index, doc_type, wait_between_tries=10, max_tries=10):
-    tries = 0
+    if max_tries < 1:
+        raise ValueError("max_tries must be 1 or greater")
 
-    while tries < max_tries:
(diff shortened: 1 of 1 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -494,18 +496,23 @@ def rename_dict_keys_with_child_dicts(data):
...
+            new[key + '_' + type_of_list + '_list'] = rename_list_elements_if_they_are_dicts(data[key])
         else:
             new[key] = data[key]
     return new
 
-def rename_list_elements_if_they_are_dicts(list):
-    for index, value in enumerate(list):
+def rename_list_elements_if_they_are_dicts(data):
+    for index, value in enumerate(data):
         if type(value) is list:
```
Hwesta: The `type() is foo` checks should be updated to use `isinstance`

Hwesta:
:+1:

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: Good point. That would have made this bug obvious in the first case, for that matter.

mistydemeo: Added a new commit to do this, since it touches more functions than just renaming types did.

Hwesta:
More :+1:

PR #82, opened 2014-07-26
Add post store AIP hook

mistydemeo:


Hwesta:
:+1:  I feel like you need to review it too since I wrote half this. :P

mistydemeo:
:b And I didn't reject it, so that's fine.

PR #84, opened 2014-07-29
elasticsearch: Clear tool output before indexing

mistydemeo:
This is both huge, and difficult to normalize for Elasticsearch.

Fixes #7038

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -50,6 +50,30 @@ MAX_QUERY_SIZE = 50000  # TODO Check that this is a reasonable number
 class ElasticsearchError(Exception):
     pass
 
+def remove_tool_output_from_mets(doc):
+    """
+    Given an ElementTree object, removes all objectsCharacteristicsExtensions elements.
+    This modifies the existing document in-place; it does not return a new document.
+
+    This helps index METS files, which might otherwise get too large to
+    be usable.
+    """
+    root = doc.getroot()
+    nsmap = { #TODO use XML namespaces from archivematicaXMLNameSpaces.py
+        'dc': 'http://purl.org/dc/terms/',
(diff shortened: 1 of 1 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -50,6 +50,35 @@
...
+    """
+    root = doc.getroot()
+    nsmap = { #TODO use XML namespaces from archivematicaXMLNameSpaces.py
+        'dc': 'http://purl.org/dc/terms/',
+        'm': 'http://www.loc.gov/METS/',
+        'p': 'info:lc/xmlns/premis-v2',
+        'f': 'http://hul.harvard.edu/ois/xml/ns/fits/fits_output',
+    }
+
+    #before_length = len(ElementTree.tostring(root))
```
Hwesta: Remove this and after_length below?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -50,6 +50,35 @@
 class ElasticsearchError(Exception):
     pass
 
+def remove_tool_output_from_mets(doc):
+    """
+    Given an ElementTree object, returns a copy of the document with
```
Hwesta: Does this actually copy it, or return a modified ElementTree?

mistydemeo: It actually modifies it in-place, now that you mention it, which is not what it implies right now. Not really worth doing a deep-copy for this, so I'll just document that it modifies it in-place (and remove the return).

Hwesta:
:+1: Minor comments only.

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: Yeah, I'd kept them from the old version but I guess it's better just to kill them.

PR #85, opened 2014-07-31
bagit: shorten argument list

mistydemeo:
This prevents issues like #6624, where a long SIP name would cause the argument list to easily overflow the 1000-character limit in the job command column.


Hwesta:
:+1:

PR #89, opened 2014-08-05
createMETS2: prefix PREMIS namespacing

mistydemeo:
Leaving PREMIS elements unprefixed resulted in any other unprefixed elements being interpreted as being FITS - for example, un-namespaced documents from external sources.

Fixes #7055.

Unfortunately this isn't (yet) a _real_ fix - just a workaround. The outer document still uses a None namespace (for METS), so non-namespaced elements from emedded documents are now interpreted as being in the METS namespace instead of the PREMIS namespace. (This is also why `div` elements are interpreted as being `mets:div` elements, for example.)

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -300,49 +300,49 @@ def createTechMD(fileUUID):
     sqlLock.release()
 
     #OBJECT
-    object = etree.SubElement(xmlData, "object", nsmap={None: ns.premisNS})
-    object.set(ns.xsiBNS+"type", "file")
+    object = etree.SubElement(xmlData, ns.premisBNS + "object", nsmap={'premis': ns.premisNS})
+    object.set(ns.xsiBNS+"type", "premis:file")
     object.set(ns.xsiBNS+"schemaLocation", ns.premisNS + " http://www.loc.gov/standards/premis/v2/premis-v2-2.xsd")
     object.set("version", "2.2")
 
-    objectIdentifier = etree.SubElement(object, "objectIdentifier")
-    etree.SubElement(objectIdentifier, "objectIdentifierType").text = "UUID"
-    etree.SubElement(objectIdentifier, "objectIdentifierValue").text = fileUUID
+    objectIdentifier = etree.SubElement(object, ns.premisBNS + "objectIdentifier")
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -300,49 +300,49 @@ def createTechMD(fileUUID):
     sqlLock.release()
 
     #OBJECT
-    object = etree.SubElement(xmlData, "object", nsmap={None: ns.premisNS})
-    object.set(ns.xsiBNS+"type", "file")
+    object = etree.SubElement(xmlData, ns.premisBNS + "object", nsmap={'premis': ns.premisNS})
+    object.set(ns.xsiBNS+"type", "premis:file")
```
Hwesta, line 304: Do attributes have to be namespaced as well? It seems a little odd.

Hwesta:
I wish we could fix this by having ffprobe namespace it's XML properly, but ['fully_qualified=1'](http://www.ffmpeg.org/ffprobe.html#xml) doesn't namespace the element correctly.

The PREMIS in METS validator accepts that <mets:format ...> is fine, but <premis:format ...> isn't? It seems odd.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
mistydemeo, line 304: This one does, yes - doesn't validate otherwise. It is unusual!

mistydemeo:
> The PREMIS in METS validator accepts that is fine, but isn't?

PREMIS in METS validator doesn't actually validate that every element is from the namespace it claims to be, just that recognized tags are used correctly. Since none of the tags we're currently shoving in the `<objectCharacteristicsExtension>` clash with METS tags, the validator has no problem with it even though it is incorrect.

mistydemeo:
> I wish we could fix this by having ffprobe namespace it's XML properly, but 'fully_qualified=1' doesn't namespace the element correctly.

This'd still just be a workaround. :( After all, other XML we're including still wouldn't namespace its XML and would be namespaced incorrectly. I think the real issue is that default namespacing high up in the document is incompatible with wrapping arbitrary XML.

Hwesta:
Hmm, point. I'd rather all tools were smart enough to correctly namespace their output, but I'll keep that in mind for the metsrw.

:+1:  to commit

PR #90, opened 2014-08-06
Metadata processing: Add file ID, characterization

Hwesta:
refs #6575

Metadata files should have identification and characterization information in the METS file. Add those microservices to metadata processing. Update processing config to reflect this. Fix a couple microservice group discontinuities.

### DIFF
--- a/src/dashboard/src/components/administration/views_processing.py
+++ b/src/dashboard/src/components/administration/views_processing.py
@@ -161,7 +161,7 @@ def index(request):
         },
         {
             "name": "select_format_id_tool_submissiondocs",
-            "label": "Select file format identification command (Submission documentation)",
+            "label": "Select file format identification command (Submission documentation & metadata)",
             "choice_uuid": '087d27be-c719-47d8-9bbb-9a7d8b609c44'
         },
         {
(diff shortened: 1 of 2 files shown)


src/dashboard/src/components/administration/views_processing.py
```diff
@@ -161,7 +161,7 @@ def index(request):
         },
         {
             "name": "select_format_id_tool_submissiondocs",
-            "label": "Select file format identification command (Submission documentation)",
+            "label": "Select file format identification command (Submission documentation & metadata)",
```
mistydemeo, line 164: Don't think this will actually work - `087d27be-c719-47d8-9bbb-9a7d8b609c44` is the UUID of the submission doc ID MSCL, while you added the new `b2444a6e-c626-4487-9abc-1556dd89a8ae` for metadata.

mistydemeo:
:thumbsup: aside from comment above!

src/dashboard/src/components/administration/views_processing.py
Hwesta, line 164: I didn't add a 'choose file ID' microservice - I'm using the same one for metadata as for submission docs, since %IDCommand% is still around. Just updating the text on the processing config to make this clear.

mistydemeo, line 164: Oh, right! Sorry.

PR #92, opened 2014-08-09
createMETS2: namespace all elements

mistydemeo:
This ensures that intentionally non-namespaced elements (such as embedded tool output) get interpreted correctly as having no namespace, rather than being within a default namespace defined earlier.


Hwesta:
:+1:

PR #93, opened 2014-08-12
submissionDocs: if a submissionDocs already exists in objects, merge instead of nesting

mistydemeo:
When submissionDocumentation was being moved into the AIP before characterization, Archivematica was using mv. This meant that if the SIP already had a directory by that name, submissionDocumentation would end up nested inside the existing directory instead of merged with it.

This adds a new moveOrMerge clientscript that acts like mv if the destination doesn't exist, and merges the contents if it does.

The first commit here was from I was mistaken about the source of this problem. It's not actually _wrong_, but probably unnecessary; I'll probably just remove it.


Hwesta:
Should we do this for other reserved-name directories? metadata and objects off the top of my head (I know Justin was compiling a list of reserved word directories).

mistydemeo:
Looks like there's a comparable "move metadata to objects directory" microservice, but not for the other special directories. Will adjust that microservice to use this too.

mistydemeo:
Updated.

Hwesta:
:+1: 

Though maybe not until after 1.2, as you mentioned.

PR #94, opened 2014-08-12
PREMIS Rights: month and day optional

Hwesta:
refs #7099

PREMIS Rights dates should accept year, year-month or year-month-day. Update matching to make month and day optional. Update year regex to allow any 4 digit year, instead of just ones starting with 0, 1 or 2.

Update vendored library from 2011 version to version at 3.0.55 tag in https://github.com/RobinHerbots/jquery.inputmask  This allows optional segments in the regex.


mistydemeo:
:thumbsup:

PR #95, opened 2014-08-16
Unify Unit ReplacementDict creation code

mistydemeo:
In the past, each Unit class had its own logic for creating replacementDicts. This meant that the available unit variables for a given job was dependent on the type of unit class that was spawning the job, and introduced the possibility of other inconsistencies.

For Archivematica 1.2, a new constructor for ReplacementDict had already been created which used the database to populate a common set of fields; this is used by every other place in Archivematica that creates and consumes ReplacementDicts for this kind of task, including every client script (normalization, identification, etc.)

The File unit is the only unit which retains its original constructor, and only as a backup. The new constructor depends on the database, but the File unit has to spawn several jobs before files are ever added to the database. The old minimalistic ReplacementDict is kept as a fallback for this case only; as soon as every file is assigned a UUID, the new ReplacementDict can be used.

refs #6451

### DIFF
--- a/src/MCPServer/lib/unitDIP.py
+++ b/src/MCPServer/lib/unitDIP.py
@@ -26,10 +26,15 @@ from unitFile import unitFile
 import archivematicaMCP
 import os
 import sys
+
 import lxml.etree as etree
 
 sys.path.append("/usr/share/archivematica/dashboard")
 from main.models import File
+
+sys.path.append("/usr/lib/archivematica/archivematicaCommon")
+from dicts import ReplacementDict
+
 
 class UnitDIPError(Exception):
     pass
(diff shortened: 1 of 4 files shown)


src/MCPServer/lib/unitFile.py
```diff
@@ -36,21 +41,25 @@ def __init__(self, currentPath, UUID="None", owningUnit=None):
...
+                "%fileUUID%": self.UUID,
+                "%fileGrpUse%": self.fileGrpUse
             }
-            return ret
-    
+
     def reload(self):
-        return 
-    
+        return
```
Hwesta: Not related to the change, but since you touched it, I think `pass` is more correct here than `return` if you want to change it. Also below.

Hwesta:
unitDIP.getReplacementDic needs to be updated as well.

Other than that, looks good! :+1:

mistydemeo:
Updated, and folded in a commit with some other cleanups too.

mistydemeo:
124 lines deleted, 45 lines added. Feels like a good ratio. :smile_cat:

src/MCPServer/lib/unitDIP.py
```diff
@@ -71,48 +69,17 @@ def reloadFileList(self):
...
-
-        #no-op for reload on DIP
-        return
+        pass
 
 
     def getReplacementDic(self, target):
-        # self.currentPath = currentPath.__str__()
-        # self.UUID = uuid.uuid4().__str__()
         #Pre do some variables, that other variables rely on, because dictionaries don't maintain order
```
Hwesta: This was removed from the other units - probably should be removed here.

Hwesta:
:+1:

mistydemeo:
I'll look at merging this as soon as the 1.2 release branch is cut.

mistydemeo:
Rebased on qa/1.x

Hwesta:
Still :+1:

mistydemeo:
Rebased on qa/1.x, now that the models branch is in.

Hwesta:
Still :+1:

mistydemeo:
Caught a bug in the DIP dict, fixed.

The DIP dict is a weird and awkward edge case, because every other dict is simply taking data modeled in the database. The DIP doesn't exist in the DB, on the other hand, and the directories in the dict `SIPDirectory` and friends) aren't based directly on data in the database. I considered pushing some of this into the ReplacementDict constructor but that's getting really hairy already; I just made sure that unitDIP alters the dict it receives to fix up the few different variables.

PR #97, opened 2014-08-20
Manual Normalization: normalization.csv, non-Unicode

Hwesta:
Fix normalization.csv parsing to match on the unsanitized name in all cases, instead of the sanitized name.

Fix createMETS2 trying to sort files and folders in the fileSec (and structMap?) that broke on non-Unicode filenames. Since the sort is unecessary, removed that and the outdated PyICU dependency.

Diff may be easier to understand if done with -w (ignore whitespace changes), since I increased the identation in `createFileSec` without really making other changes.

Quick testing shows that all the expected files/directories are the fileSec and structMap, and the METS passes validation.


Hwesta:
This should get looked at by an analyst for the METS changes (mostly regarding ordering in the fileSec).

mistydemeo:
Looks good. :+1: I'll try testing with non-English filenames.

mistydemeo:
Tested, works fine with non-ASCII, Unicode filenames.

Have not tested with non-Unicode filenames, but that's probably not critical.

PR #98, opened 2014-08-20
Replace databaseInterface with Django models

mistydemeo:
~~`databaseInterface` is still imported in a few places in MCPServer, primarily for the `getUTCTime` function that will be moved elsewhere, and for database locking that may be removeable.~~

See comments below.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -42,13 +42,15 @@ import sys
 import threading
 import traceback
 
+os.environ['DJANGO_SETTINGS_MODULE'] = "settings.common"
+
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-from executeOrRunSubProcess import executeOrRun
 import databaseInterface
-from databaseFunctions import logTaskAssignedSQL
+from executeOrRunSubProcess import executeOrRun
 printOutputLock = threading.Lock()
 
-databaseInterface.printSQL = True
+sys.path.append("/usr/share/archivematica/dashboard")
+from main.models import Task
 
 config = ConfigParser.SafeConfigParser({'MCPArchivematicaServerInterface': ""})
 config.read("/etc/archivematica/MCPClient/clientConfig.conf")
@@ -94,17 +96,17 @@ def executeCommand(gearman_worker, gearman_job):
         sInput = ""
         clientID = gearman_worker.worker_client_id
 
-
-        sql = """SELECT Tasks.taskUUID FROM Tasks WHERE taskUUID='%s' AND startTime != 0;""" % (gearman_job.unique.__str__())
-        rows = databaseInterface.queryAllSQL(sql)
-        if len(rows):
+        task = Task.objects.get(taskuuid=gearman_job.unique)
+        if task.starttime is not None:
             exitCode = -1
             stdOut = ""
             stdError = """Detected this task has already started!
 Unable to determine if it completed successfully."""
             return cPickle.dumps({"exitCode" : exitCode, "stdOut": stdOut, "stdError": stdError})
-        
-        logTaskAssignedSQL(gearman_job.unique.__str__(), clientID, utcDate)
+        else:
+            task.client = clientID
+            task.starttime = utcDate
+            task.save()
 
         if execute not in supportedModules:
             output = ["Error!", "Error! - Tried to run and unsupported command." ]
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -45,7 +45,7 @@ import traceback
 os.environ['DJANGO_SETTINGS_MODULE'] = "settings.common"
 
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-import databaseInterface
+import databaseFunctions
 from executeOrRunSubProcess import executeOrRun
 printOutputLock = threading.Lock()
 
@@ -86,7 +86,7 @@ def executeCommand(gearman_worker, gearman_job):
         execute = gearman_job.task
         print "executing:", execute, "{", gearman_job.unique, "}"
         data = cPickle.loads(gearman_job.data)
-        utcDate = databaseInterface.getUTCDate()
+        utcDate = databaseFunctions.getUTCDate()
         arguments = data["arguments"]#.encode("utf-8")
         if isinstance(arguments, unicode):
             arguments = arguments.encode("utf-8")
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -48,29 +48,12 @@ import databaseInterface
 from databaseFunctions import logTaskAssignedSQL
 printOutputLock = threading.Lock()
 
-print "Attempting to read dashboard UUID..."
-
-dashboardUUID = False
-while dashboardUUID == False:
-    sql = "SELECT value FROM DashboardSettings WHERE name='dashboard_uuid';"
-    rows = databaseInterface.queryAllSQL(sql)
-    if not rows:
-        time.sleep(1)
-    else:
-        # This might be returned as a unicode, which may cause Problems later on
-        dashboardUUID = str(rows[0][0])
-
-print "Dashboard UUID: " + dashboardUUID
-
-os.environ['ARCHIVEMATICA_DASHBOARD_UUID'] = dashboardUUID
-
 databaseInterface.printSQL = True
 
 config = ConfigParser.SafeConfigParser({'MCPArchivematicaServerInterface': ""})
 config.read("/etc/archivematica/MCPClient/clientConfig.conf")
 
 replacementDic = {
-    "%dashboardUUID%": dashboardUUID, \
     "%sharedPath%":config.get('MCPClient', "sharedDirectoryMounted"), \
     "%clientScriptsDirectory%":config.get('MCPClient', "clientScriptsDirectory")
 }
--- a/src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
@@ -26,7 +26,9 @@ from optparse import OptionParser
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 from fileOperations import addFileToTransfer
 from fileOperations import addFileToSIP
-import databaseInterface
+
+sys.path.append("/usr/share/archivematica/dashboard")
+from main.models import File
 
 
 if __name__ == '__main__':
@@ -53,8 +55,7 @@ if __name__ == '__main__':
     else:
         print >>sys.stderr, "File already has UUID:", fileUUID
         if opts.update_use:
-            sql = """UPDATE Files SET fileGrpUse='%s' WHERE fileUUID = '%s';""" % (opts.use, fileUUID)
-            databaseInterface.runSQL(sql)
+            File.objects.filter(uuid=fileUUID).update(filegrpuse=opts.use)
         exit(0) 
 
 
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
@@ -24,12 +24,14 @@
 import os
 import sys
 import lxml.etree as etree
-import MySQLdb
 from xml.sax.saxutils import quoteattr
 
 import archivematicaXMLNamesSpace as ns
+
+sys.path.append("/usr/share/archivematica/dashboard")
+from main.models import File
+
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-import databaseInterface
 from archivematicaFunctions import escape
 
 
@@ -119,15 +121,17 @@ def createFileSec(path, parentBranch, structMapParent):
 
                     print "pathSTR", type(pathSTR), pathSTR
 
-                    sql = """SELECT fileUUID FROM Files WHERE removedTime = 0 AND %s = '%s' AND Files.currentLocation = '%s';""" % (fileGroupIdentifier, SIPUUID, MySQLdb.escape_string(pathSTR))
-                    c, sqlLock = databaseInterface.querySQL(sql)
-                    row = c.fetchone()
-                    if row == None:
-                        print >>sys.stderr, "No uuid for file: \"", pathSTR, "\""
-                    while row != None:
-                        myuuid = row[0]
-                        row = c.fetchone()
-                    sqlLock.release()
+                    kwargs = {
+                        'removedtime__isnull': True,
+                        fileGroupIdentifier: SIPUUID,
+                        'currentlocation': pathSTR
+                    }
+                    try:
+                        f = File.objects.get(**kwargs)
+                    except File.DoesNotExist:
+                        print >> sys.stderr, "No uuid for file: \"", pathSTR, "\""
+                    else:
+                        myuuid = f.uuid
 
                     if includeAmdSec:
                         createDigiprovMD(myuuid, itempath, myuuid)
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -44,10 +44,10 @@ from archivematicaCreateMETSTrim import getTrimFileDmdSec
 from archivematicaCreateMETSTrim import getTrimAmdSec
 from archivematicaCreateMETSTrim import getTrimFileAmdSec
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-import databaseInterface
 from archivematicaFunctions import escape
 from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import normalizeNonDcElementName
+from databaseFunctions import getUTCDate
 from sharedVariablesAcrossModules import sharedVariablesAcrossModules
 sharedVariablesAcrossModules.globalErrorCount = 0
 
@@ -849,7 +849,7 @@ if __name__ == '__main__':
         nsmap = rootNSMap,
         attrib = { "{" + ns.xsiNS + "}schemaLocation" : "http://www.loc.gov/METS/ http://www.loc.gov/standards/mets/version18/mets.xsd" },
     )
-    etree.SubElement(root, ns.metsBNS + "metsHdr").set("CREATEDATE", databaseInterface.getUTCDate().split(".")[0])
+    etree.SubElement(root, ns.metsBNS + "metsHdr").set("CREATEDATE", getUTCDate().split(".")[0])
 
 
 
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -24,13 +24,18 @@
 
 from glob import glob
 import lxml.etree as etree
-import MySQLdb
 import os
 import re
 import sys
 import traceback
 
 import archivematicaXMLNamesSpace as ns
+
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
+sys.path.append("/usr/share/archivematica/dashboard")
+from django.contrib.auth.models import User
+from main.models import Agent, Derivation, DublinCore, Event, File, FPCommandOutput, SIP, Transfer
+
 from archivematicaCreateMETSMetadataCSV import parseMetadata
 from archivematicaCreateMETSRights import archivematicaGetRights
 from archivematicaCreateMETSRightsDspaceMDRef import archivematicaCreateMETSRightsDspaceMDRef
@@ -129,27 +134,24 @@ def createAgent(agentIdentifierType, agentIdentifierValue, agentName, agentType)
 SIPMetadataAppliesToType = '3e48343d-e2d2-4956-aaa3-b54d26eb9761'
 TransferMetadataAppliesToType = '45696327-44c5-4e78-849b-e027a189bf4d'
 FileMetadataAppliesToType = '7f04d9d4-92c2-44a5-93dc-b7bfdf0c1f17'
+
+
 def getDublinCore(unit, id):
-    field_list = ["title", "creator", "subject", "description", "publisher", "contributor", "date", "type", "format", "identifier", "source", "relation", "language", "coverage", "rights", "isPartOf"]
-    sql = """SELECT {fields}
-    FROM Dublincore WHERE metadataAppliesToType = '{type}' AND metadataAppliesToidentifier = '{id}';""".format(
-            fields=', '.join(field_list),
-            type=unit,
-            id=id)
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    ret = None
-    if row is not None:
-        ret = etree.Element(ns.dctermsBNS + "dublincore", nsmap={"dc":ns.dctermsNS})
-        ret.set(ns.xsiBNS+"schemaLocation", ns.dctermsNS + " http://dublincore.org/schemas/xmls/qdc/2008/02/11/dcterms.xsd")
-    while row is not None:
-        #title, creator, subject, description, publisher, contributor, date, type, format, identifier, source, relation, language, coverage, rights = row
-        for i, term in enumerate(field_list):
-            txt = row[i] or ""
-            if txt:
-                newChild(ret, ns.dctermsBNS + term, text=txt)
-        row = c.fetchone()
-    sqlLock.release()
+    field_list = ["title", "creator", "subject", "description", "publisher", "contributor", "date", "type", "format", "identifier", "source", "relation", "language", "coverage", "rights", "is_part_of"]
+
+    try:
+        dc = DublinCore.objects.get(metadataappliestotype_id=unit,
+                                    metadataappliestoidentifier=id)
+    except DublinCore.DoesNotExist:
+        return
+
+    ret = etree.Element(ns.dctermsBNS + "dublincore", nsmap={"dc": ns.dctermsNS})
+    ret.set(ns.xsiBNS + "schemaLocation", ns.dctermsNS + " http://dublincore.org/schemas/xmls/qdc/2008/02/11/dcterms.xsd")
+
+    for term in field_list:
+        txt = getattr(dc, term)
+        if txt:
+            newChild(ret, ns.dctermsBNS + term, text=txt)
     return ret
 
 
@@ -287,14 +289,9 @@ def createTechMD(fileUUID):
 
     #premis = etree.SubElement( xmlData, "premis", attrib = {ns.xsiBNS+"type": "premis:file"})
 
-    sql = "SELECT fileSize, checksum FROM Files WHERE fileUUID = '%s';" % (fileUUID)
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
-        fileSize = row[0].__str__()
-        checksum = row[1].__str__()
-        row = c.fetchone()
-    sqlLock.release()
+    f = File.objects.get(uuid=fileUUID)
+    fileSize = str(f.size)
+    checksum = f.checksum
 
     #OBJECT
     object = etree.SubElement(xmlData, ns.premisBNS + "object", nsmap={'premis': ns.premisNS})
@@ -341,76 +338,59 @@ def createTechMD(fileUUID):
 
     objectCharacteristicsExtension = etree.SubElement(objectCharacteristics, ns.premisBNS + "objectCharacteristicsExtension")
 
-    sql = "SELECT FPCommandOutput.content FROM FPCommandOutput \
-           INNER JOIN fpr_fprule ON FPCommandOutput.ruleUUID = fpr_fprule.uuid \
-           WHERE fileUUID = '{file}' AND (purpose = 'characterization' OR purpose = 'default_characterization');".format(file=fileUUID)
-    c, sqlLock = databaseInterface.querySQL(sql)
     parser = etree.XMLParser(remove_blank_text=True)
-    for document, in c:
+    documents = FPCommandOutput.objects.filter(file_id=fileUUID, rule__purpose__in=['characterization', 'default_characterization']).values_list('content')
+    for document, in documents:
         # This needs to be converted into an str because lxml doesn't accept
         # XML documents in unicode strings if the document contains an
         # encoding declaration.
         output = etree.XML(document.encode("utf-8"), parser)
         objectCharacteristicsExtension.append(output)
-    sqlLock.release()
 
-    sql = "SELECT Files.originalLocation FROM Files WHERE Files.fileUUID = '" + fileUUID + "';"
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    if not row:
+    try:
+        f = File.objects.get(uuid=fileUUID)
+    except File.DoesNotExist:
         print >>sys.stderr, "Error: no location found."
-    while row != None:
-        etree.SubElement(object, ns.premisBNS + "originalName").text = escape(row[0])
-        row = c.fetchone()
-    sqlLock.release()
+    else:
+        etree.SubElement(object, ns.premisBNS + "originalName").text = escape(f.originallocation)
 
-    #Derivations
-    sql = "SELECT sourceFileUUID, derivedFileUUID, relatedEventUUID FROM Derivations WHERE sourceFileUUID = '" + fileUUID + "';"
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
+    # Derivations
+    derivations = Derivation.objects.filter(source_file_id=fileUUID)
+    for derivation in derivations:
         relationship = etree.SubElement(object, ns.premisBNS + "relationship")
         etree.SubElement(relationship, ns.premisBNS + "relationshipType").text = "derivation"
         etree.SubElement(relationship, ns.premisBNS + "relationshipSubType").text = "is source of"
 
         relatedObjectIdentification = etree.SubElement(relationship, ns.premisBNS + "relatedObjectIdentification")
         etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierType").text = "UUID"
-        etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierValue").text = row[1]
+        etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierValue").text = derivation.derived_file_id
 
         relatedEventIdentification = etree.SubElement(relationship, ns.premisBNS + "relatedEventIdentification")
         etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierType").text = "UUID"
-        etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierValue").text = row[2]
+        etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierValue").text = derivation.event_id
 
-        row = c.fetchone()
-    sqlLock.release()
-
-    sql = "SELECT sourceFileUUID, derivedFileUUID, relatedEventUUID FROM Derivations WHERE derivedFileUUID = '" + fileUUID + "';"
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
+    derivations = Derivation.objects.filter(derived_file_id=fileUUID)
+    for derivation in derivations:
         relationship = etree.SubElement(object, ns.premisBNS + "relationship")
         etree.SubElement(relationship, ns.premisBNS + "relationshipType").text = "derivation"
         etree.SubElement(relationship, ns.premisBNS + "relationshipSubType").text = "has source"
 
         relatedObjectIdentification = etree.SubElement(relationship, ns.premisBNS + "relatedObjectIdentification")
         etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierType").text = "UUID"
-        etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierValue").text = row[0]
+        etree.SubElement(relatedObjectIdentification, ns.premisBNS + "relatedObjectIdentifierValue").text = derivation.source_file_id
 
         relatedEventIdentification = etree.SubElement(relationship, ns.premisBNS + "relatedEventIdentification")
         etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierType").text = "UUID"
-        etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierValue").text = row[2]
+        etree.SubElement(relatedEventIdentification, ns.premisBNS + "relatedEventIdentifierValue").text = derivation.event_id
 
-        row = c.fetchone()
-    sqlLock.release()
     return ret
 
 def createDigiprovMD(fileUUID):
     ret = []
-    #EVENTS
+    # EVENTS
 
-    sql = "SELECT pk, fileUUID, eventIdentifierUUID, eventType, eventDateTime, eventDetail, eventOutcome, eventOutcomeDetailNote, linkingAgentIdentifier FROM Events WHERE fileUUID = '" + fileUUID + "';"
-    rows = databaseInterface.queryAllSQL(sql)
-    for row in rows:
+    events = Event.objects.filter(file_uuid_id=fileUUID)
+    for event_record in events:
         digiprovMD = etree.Element(ns.metsBNS + "digiprovMD")
         ret.append(digiprovMD) #newChild(amdSec, "digiprovMD")
         #digiprovMD.set("ID", "digiprov-"+ os.path.basename(filename) + "-" + fileUUID)
@@ -427,42 +407,34 @@ def createDigiprovMD(fileUUID):
 
         eventIdentifier = etree.SubElement(event, ns.premisBNS + "eventIdentifier")
         etree.SubElement(eventIdentifier, ns.premisBNS + "eventIdentifierType").text = "UUID"
-        etree.SubElement(eventIdentifier, ns.premisBNS + "eventIdentifierValue").text = row[2]
+        etree.SubElement(eventIdentifier, ns.premisBNS + "eventIdentifierValue").text = event_record.event_id
 
-        etree.SubElement(event, ns.premisBNS + "eventType").text = row[3]
-        etree.SubElement(event, ns.premisBNS + "eventDateTime").text = row[4].__str__().replace(" ", "T")
-        etree.SubElement(event, ns.premisBNS + "eventDetail").text = escape(row[5])
+        etree.SubElement(event, ns.premisBNS + "eventType").text = event_record.event_type
+        etree.SubElement(event, ns.premisBNS + "eventDateTime").text = event_record.event_datetime.isoformat()
+        etree.SubElement(event, ns.premisBNS + "eventDetail").text = escape(event_record.event_detail)
 
         eventOutcomeInformation  = etree.SubElement(event, ns.premisBNS + "eventOutcomeInformation")
-        etree.SubElement(eventOutcomeInformation, ns.premisBNS + "eventOutcome").text = row[6]
+        etree.SubElement(eventOutcomeInformation, ns.premisBNS + "eventOutcome").text = event_record.event_outcome
         eventOutcomeDetail = etree.SubElement(eventOutcomeInformation, ns.premisBNS + "eventOutcomeDetail")
-        etree.SubElement(eventOutcomeDetail, ns.premisBNS + "eventOutcomeDetailNote").text = escape(row[7])
+        etree.SubElement(eventOutcomeDetail, ns.premisBNS + "eventOutcomeDetailNote").text = escape(event_record.event_outcome_detail)
         
-        if row[8]:
+        if event_record.linking_agent:
             linkingAgentIdentifier = etree.SubElement(event, ns.premisBNS + "linkingAgentIdentifier")
             etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierType").text = "Archivematica user pk"
-            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = row[8].__str__()
+            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = str(event_record.linking_agent_id)
         
-        #linkingAgentIdentifier
-        sql = """SELECT agentIdentifierType, agentIdentifierValue, agentName, agentType FROM Agents;"""
-        c, sqlLock = databaseInterface.querySQL(sql)
-        row = c.fetchone()
-        while row != None:
+        # linkingAgentIdentifier
+        for agent in Agent.objects.all():
             linkingAgentIdentifier = etree.SubElement(event, ns.premisBNS + "linkingAgentIdentifier")
-            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierType").text = row[0]
-            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = row[1]
-            row = c.fetchone()
-        sqlLock.release()
+            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierType").text = agent.identifiertype
+            etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = agent.identifiervalue
     return ret
 
 def createDigiprovMDAgents():
     ret = []
     global globalDigiprovMDCounter
-    #AGENTS
-    sql = """SELECT agentIdentifierType, agentIdentifierValue, agentName, agentType FROM Agents;"""
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
+    # AGENTS
+    for agent in Agent.objects.all():
         globalDigiprovMDCounter += 1
         digiprovMD = etree.Element(ns.metsBNS + "digiprovMD")
         digiprovMD.set("ID", "digiprovMD_"+ globalDigiprovMDCounter.__str__())
@@ -471,14 +443,18 @@ def createDigiprovMDAgents():
         mdWrap.set("MDTYPE", "PREMIS:AGENT")
         xmlData = newChild(mdWrap, ns.metsBNS + "xmlData")
         #agents = etree.SubElement(xmlData, "agents")
-        xmlData.append(createAgent(row[0], row[1], row[2], row[3]))
-        row = c.fetchone()
-    sqlLock.release()
-    
-    sql = """SELECT auth_user.id, auth_user.username, auth_user.first_name, auth_user.last_name FROM SIPs JOIN Files ON SIPs.sipUUID = Files.sipUUID JOIN Events ON Files.fileUUID = Events.fileUUID JOIN auth_user ON Events.linkingAgentIdentifier = auth_user.id WHERE SIPs.sipUUID = '%s' GROUP BY auth_user.id;""" % (fileGroupIdentifier)
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
+        xmlData.append(createAgent(agent.identifiertype, agent.identifiervalue, agent.name, agent.agenttype))
+
+    # If this function is being called by other scripts, fileGroupIdentifier
+    # will not be defined; just return right away in that case.
+    try:
+        user_ids = SIP.objects.get(uuid=fileGroupIdentifier).file_set.filter(event__linking_agent__isnull=False).values_list('event__linking_agent').distinct()
+    except SIP.DoesNotExist:
+        return ret
+
+    for user_id, in user_ids:
+        user = User.objects.get(id=user_id)
+
         globalDigiprovMDCounter += 1
         digiprovMD = etree.Element(ns.metsBNS + "digiprovMD")
         digiprovMD.set("ID", "digiprovMD_"+ globalDigiprovMDCounter.__str__())
@@ -487,22 +463,12 @@ def createDigiprovMDAgents():
         mdWrap.set("MDTYPE", "PREMIS:AGENT")
         xmlData = newChild(mdWrap, ns.metsBNS + "xmlData")
         #agents = etree.SubElement(xmlData, "agents")
-        
-        id, username, first_name, last_name = row
-        id = id.__str__()
-        if not username:
-            username = ""
-        if not first_name:
-            first_name = ""
-        if not last_name:
-            last_name = ""
+
         agentIdentifierType = "Archivematica user pk"
-        agentIdentifierValue = id
-        agentName = 'username="%s", first_name="%s", last_name="%s"' % (username, first_name, last_name)       
+        agentIdentifierValue = str(user.id)
+        agentName = 'username="%s", first_name="%s", last_name="%s"' % (user.username, user.first_name, user.last_name)
         agentType = "Archivematica user"
         xmlData.append(createAgent(agentIdentifierType, agentIdentifierValue, agentName, agentType))
-        row = c.fetchone()
-    sqlLock.release()
     
     return ret
 
@@ -629,28 +595,26 @@ def createFileSec(directoryPath, parentDiv):
             DMDIDS=""
             directoryPathSTR = itemdirectoryPath.replace(baseDirectoryPath, baseDirectoryPathString, 1)
 
-            sql = """SELECT fileUUID, fileGrpUse, fileGrpUUID, Files.transferUUID, label, originalLocation, Transfers.type
-                    FROM Files
-                    LEFT OUTER JOIN Transfers ON Files.transferUUID = Transfers.transferUUID
-                    WHERE removedTime = 0 AND %s = '%s' AND Files.currentLocation = '%s';""" % (fileGroupType, fileGroupIdentifier, MySQLdb.escape_string(directoryPathSTR))
-            c, sqlLock = databaseInterface.querySQL(sql)
-            row = c.fetchone()
-            if row == None:
+            kwargs = {
+                "removedtime__isnull": True,
+                fileGroupType: fileGroupIdentifier,
+                "currentlocation": directoryPathSTR
+            }
+            try:
+                f = File.objects.get(**kwargs)
+            except File.DoesNotExist:
                 print >>sys.stderr, "No uuid for file: \"", directoryPathSTR, "\""
                 sharedVariablesAcrossModules.globalErrorCount += 1
-                sqlLock.release()
                 continue
-            while row != None:
-                myuuid = row[0]
-                use = row[1]
-                fileGrpUUID = row[2]
-                transferUUID = row[3]
-                label = row[4]
-                originalLocation = row[5]
-                typeOfTransfer = row[6]
-                row = c.fetchone()
-            sqlLock.release()
-            
+
+            myuuid = f.uuid
+            use = f.filegrpuse
+            fileGrpUUID = f.filegrpuuid
+            transferUUID = f.transfer_id
+            label = f.label
+            originalLocation = f.originallocation
+            typeOfTransfer = f.transfer.type if f.transfer else None
+
             directoryPathSTR = itemdirectoryPath.replace(baseDirectoryPath, "", 1)
 
             if typeOfTransfer == "TRIM" and trimStructMap is None:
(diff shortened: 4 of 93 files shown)


src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -114,25 +114,18 @@ def findOrCreateSipInDB(path, waitSleep=dbWaitSleep):
...
-        #if waitSleep != 0:
-            #time.sleep(waitSleep) #let db be updated by the microservice that moved it.
-        c, sqlLock = databaseInterface.querySQL(sql)
-        row = c.fetchone()
-        if not row:
-            print "Not opening existing SIP:", UUID, "-", path
-        while row != None:
-            UUID = row[0]
+        sips = SIP.objects.filter(currentpath=path)
+        if len(sips) > 1:
```
mistydemeo: In most places I've tried to not acknowledge the possibility of multiple results (unlike the original code) - but here there is a vague chance that this could happen.

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -114,25 +114,18 @@ def findOrCreateSipInDB(path, waitSleep=dbWaitSleep):
     #find UUID on end of SIP path
     UUID = fetchUUIDFromPath(path)
     if UUID:
-        sql = """SELECT sipUUID FROM SIPs WHERE sipUUID = '""" + UUID + "';"
-        rows = databaseInterface.queryAllSQL(sql)
-        if not rows:
+        if not SIP.objects.filter(uuid=UUID):
```
Hwesta: If we're not evaluating the queryset, [`.exists()`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#exists) seems to be good to use in the query.

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -114,25 +114,18 @@ def findOrCreateSipInDB(path, waitSleep=dbWaitSleep):
...
-        if not row:
-            print "Not opening existing SIP:", UUID, "-", path
-        while row != None:
-            UUID = row[0]
+        sips = SIP.objects.filter(currentpath=path)
+        if len(sips) > 1:
+            print "Warning: found more than one matching SIP for path", path, "- using first result"
+        if sips:
+            UUID = sips[0].uuid
             print "Opening existing SIP:", UUID, "-", path
```
Hwesta, line 124: I prefer not to use all caps for anything but constants - maybe sip_uuid for this variable?

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -196,17 +189,14 @@ def createUnitAndJobChainThreaded(path, config, terminate=True):
...
-        rows.append(row)
-        row = c.fetchone()
-    sqlLock.release()
-
-    for row in rows:
-        directory = row[0].replace("%watchDirectoryPath%", config.get('MCPServer', "watchDirectoryPath"), 1)
+    watched_directories = WatchedDirectory.objects.all()
+
+    for dir in watched_directories:
+        directory = dir.watched_directory_path.replace("%watchDirectoryPath%", config.get('MCPServer', "watchDirectoryPath"), 1)
```
Hwesta: Move querying the config for watchDirectoryPath outside the loop?  I don't know how efficient config.get is.

src/MCPServer/lib/jobChainLink.py
```diff
@@ -37,19 +37,21 @@
...
-constlinkTaskManagerGetMicroserviceGeneratedListInStdOut = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("Get microservice generated list in stdOut"))[0][0]
-constlinkTaskManagerGetUserChoiceFromMicroserviceGeneratedList = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("Get user choice from microservice generated list"))[0][0]
-constlinkTaskManagerSetUnitVariable = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("linkTaskManagerSetUnitVariable"))[0][0]
-constlinkTaskManagerUnitVariableLinkPull = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("linkTaskManagerUnitVariableLinkPull"))[0][0]
+# Constants
+constOneTask = TaskType.objects.get(description="one_instance").uuid
+constTaskForEachFile = TaskType.objects.get(description="for each file").uuid
+constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").uuid
+constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").uuid
```
Hwesta: Duplicate

src/MCPServer/lib/jobChainLink.py
```diff
@@ -62,27 +64,17 @@ def __init__(self, jobChain, jobChainLinkPK, unit, passVar=None, subJobOf=""):
         self.passVar=passVar
         self.createdDate = databaseInterface.getUTCDate()
         self.subJobOf = subJobOf
-        sql = """SELECT MicroServiceChainLinks.currentTask, MicroServiceChainLinks.defaultNextChainLink, TasksConfigs.taskType, TasksConfigs.taskTypePKReference, TasksConfigs.description, MicroServiceChainLinks.reloadFileList, Sounds.fileLocation, MicroServiceChainLinks.defaultExitMessage, MicroServiceChainLinks.microserviceGroup FROM MicroServiceChainLinks LEFT OUTER JOIN Sounds ON MicroServiceChainLinks.defaultPlaySound = Sounds.pk JOIN TasksConfigs on MicroServiceChainLinks.currentTask = TasksConfigs.pk WHERE MicroServiceChainLinks.pk = '%s'""" % (jobChainLinkPK.__str__())
-        c, sqlLock = databaseInterface.querySQL(sql)
-        row = c.fetchone()
-        if row == None:
```
Hwesta, line 68: Do we need the None handling?  Try/except that returns None around the get?

src/MCPServer/lib/jobChainLink.py
```diff
@@ -120,47 +112,35 @@ def createTasks(self, taskType, taskTypePKReference):
         else:
             print sys.stderr, "unsupported task type: ", taskType
 
+    # Deprecated, remove later
```
Hwesta, line 124: :+1:

src/MCPServer/lib/jobChainLink.py
```diff
@@ -120,47 +112,35 @@ def createTasks(self, taskType, taskTypePKReference):
...
-            row = c.fetchone()
-            if row != None:
-                ret = row[0]
-            sqlLock.release()
-            return ret
+        if exitCode is not None:
+            try:
+                link = MicroServiceChainLinkExitCode.objects.get(microservicechainlink=str(self.pk), exitcode=str(exitCode))
+                return link.pk
+            except MicroServiceChainLinkExitCode.DoesNotExist:
```
Hwesta: Should catch `MicroServiceChainLinkExitCode.MultipleObjectsReturned` as well in case someone messed up the MSCLEC table.

src/MCPServer/lib/archivematicaMCP.py
mistydemeo, line 124: Good point - seems like a decent chance to improve some of the variable naming.

src/MCPServer/lib/jobChainLink.py
```diff
@@ -120,47 +112,35 @@ def createTasks(self, taskType, taskTypePKReference):
...
+                return self.defaultNextChainLink
```
Hwesta, line 133: Add an explicit return at the bottom? As is I don't if None is correct if the 'if' block isn't hit (though I think None is correct in this case)

src/MCPServer/lib/jobChainLink.py
```diff
@@ -120,47 +112,35 @@ def createTasks(self, taskType, taskTypePKReference):
...
+                code = MicroServiceChainLinkExitCode.objects.get(microservicechainlink=str(self.pk), exitcode=str(exitCode))
+                message = code.exitmessage
+            except MicroServiceChainLinkExitCode.DoesNotExist:
+                pass
+        if message is not None:
+            self.setExitMessage(message)
         else:
             print "No exit message"
 
+    # Deprecated, remove later
```
Hwesta: This should be moved to the playSounds block, since other things in this function are important.

src/MCPServer/lib/linkTaskManagerChoice.py
```diff
@@ -48,16 +49,10 @@ def __init__(self, jobChainLink, pk, unit):
...
         self.delayTimerLock = threading.Lock()
         self.delayTimer = None
-        sql = """SELECT chainAvailable, description FROM MicroServiceChainChoice JOIN MicroServiceChains on chainAvailable = MicroServiceChains.pk WHERE choiceAvailableAtLink = '%s' ORDER BY MicroServiceChainChoice.pk;""" % (jobChainLink.pk.__str__())
-        c, sqlLock = databaseInterface.querySQL(sql)
-        row = c.fetchone()
-        while row != None:
-            print row
-            chainAvailable = row[0]
-            description = row[1]
-            self.choices.append((chainAvailable, description,))
```
Hwesta, line 58: This got dropped in changes - is it important?

src/MCPServer/lib/linkTaskManagerDirectories.py
```diff
@@ -28,41 +28,27 @@
...
 
 
 class linkTaskManagerDirectories(LinkTaskManager):
     def __init__(self, jobChainLink, pk, unit):
         super(linkTaskManagerDirectories, self).__init__(jobChainLink, pk, unit)
         self.tasks = []
-        sql = """SELECT * FROM StandardTasksConfigs where pk = '%s'""" % (pk.__str__())
-        c, sqlLock = databaseInterface.querySQL(sql)
-        row = c.fetchone()
-        if row == None:
```
Hwesta, line 44: Handle this in rewrite?  Or is the raised Exception from get enough?

src/MCPServer/lib/linkTaskManagerFiles.py
```diff
@@ -45,40 +45,39 @@ def __init__(self, jobChainLink, pk, unit):
...
-        except (SyntaxError, IndexError):
-            # IndexError = rows was empty
-            # SyntaxError = contents of variableValue weren't the expected dict
-            pass
-        else:
-            filterSubDir = variableValue['filterSubDir']
+            var = UnitVariable.objects.get(unittype=self.unit.unitType,
+                                           unituuid=self.unit.UUID,
+                                           variable=self.execute)
+        except UnitVariable.DoesNotExist:
```
Hwesta: Catch MultipleObjectsReturned as well?

src/MCPServer/lib/archivematicaMCP.py
mistydemeo: Fixed!

src/MCPServer/lib/archivematicaMCP.py
mistydemeo: Moved outside the loop.

src/MCPServer/lib/jobChainLink.py
mistydemeo, line 68: This is in the deleted part of the diff!

src/MCPServer/lib/jobChainLink.py
mistydemeo: Updated.

src/MCPServer/lib/jobChainLink.py
mistydemeo, line 133: It's the original behaviour, so I think it's fine for it to not return in this case?

src/MCPServer/lib/linkTaskManagerChoice.py
mistydemeo, line 58: Whoops, looks like I omitted the loop. Updated.

src/MCPServer/lib/linkTaskManagerDirectories.py
mistydemeo, line 44: I think the exception is probably fine, and more explicit. I don't know that this original condition ever happened.

src/MCPServer/lib/unit.py
```diff
@@ -27,6 +27,38 @@ def __init__(self, currentPath, UUID):
...
         self.UUID = UUID
 
+    def reloadFileList(self):
+        """Match files to their UUID's via their location and the File table's currentLocation"""
+        self.fileList = {}
+        currentPath = self.currentPath.replace("%sharedPath%", \
+                                               archivematicaMCP.config.get('MCPServer', "sharedDirectory"), 1) + "/"
+        try:
+            for directory, subDirectories, files in os.walk(currentPath):
+                directory = directory.replace( currentPath, self.pathString, 1)
```
Hwesta, line 49: unitDIP needs pathString set

src/MCPServer/lib/jobChainLink.py
mistydemeo, line 68: Dang, turns out this really can return no value. :| Adding a check.

src/MCPServer/lib/unit.py
```diff
@@ -34,7 +66,29 @@ def setMagicLink(self,link, exitStatus=""):
...
+                var.save()
+        else:
+            var = UnitVariable(
+                unittype=self.unitType, unituuid=self.UUID,
+                variable=variable, variablevalue=variableValue,
+                microservicechainlink=microServiceChainLink
+            )
+            var.save()
     
     def getmicroServiceChainLink(self, variable, variableValue, defaultMicroServiceChainLink):
```
Hwesta, line 100: `variableValue` is not used here (or in the original) - can it be removed? Or a comment reminding that work to happen?

src/MCPServer/lib/unitTransfer.py
```diff
@@ -49,143 +51,54 @@ def __init__(self, currentPath, UUID=""):
...
                        "%sharedPath%", 1)
 
-        if UUID == "":
-            sql = """SELECT transferUUID FROM Transfers WHERE currentLocation = '""" + MySQLdb.escape_string(currentPath2) + "'"
-            time.sleep(.5)
-            c, sqlLock = databaseInterface.querySQL(sql)
-            row = c.fetchone()
-            while row != None:
-                UUID = row[0]
-                print "Opening existing Transfer:", UUID, "-", currentPath2
```
Hwesta, line 59: Keep this print statement in the rewrite?

src/dashboard/src/main/models.py
```diff
@@ -586,7 +595,13 @@ class Meta:
 
 class StandardTaskConfig(models.Model):
     id = UUIDPkField()
-    execute = models.TextField(db_column='execute', blank=True)
+    filterfileend = models.CharField(max_length=50, null=True, blank=True, db_column='filterFileEnd')
+    filterfilestart = models.CharField(max_length=50, null=True, blank=True, db_column='filterFileStart')
+    filtersubdir = models.CharField(max_length=50, null=True, blank=True, db_column='filterSubDir')
+    requiresoutputlock = models.IntegerField(null=True, default=None, db_column='requiresOutputLock')
```
Hwesta: I think this is actually a BooleanField (or NullBooleanField).  Django's BooleanField [maps to tinyint](http://stackoverflow.com/questions/1778948/boolean-fields-in-mysql-django-models).

src/dashboard/src/main/models.py
```diff
@@ -605,6 +620,49 @@ class Meta:
...
+    execute = models.CharField(max_length=250, null=True, blank=True)
+    replaces = models.ForeignKey('TaskConfigAssignMagicLink', null=True)
+    lastmodified = models.DateTimeField(db_column='lastModified')
+
+    class Meta(object):
+        db_table = u'TasksConfigsAssignMagicLink'
+
+class TaskConfigSetUnitVariable(models.Model):
+    id = UUIDPkField()
+    variable = models.TextField(null=True, blank=True)
```
Hwesta: Include db_column.  Django will guess the correct value here, but the explicit is nice, especially since everything else has it.  Same on TaskConfigUnitVariableLinkPull.

src/dashboard/src/main/models.py
```diff
@@ -688,3 +746,24 @@ class Meta:
 
     def __unicode__(self):
         return self.term
+
+class WatchedDirectory(models.Model):
+    id = UUIDPkField()
+    watched_directory_path = models.TextField(null=True, blank=True, db_column='watchedDirectoryPath')
+    chain = models.ForeignKey('MicroServiceChain', null=True)
+    only_act_on_directories = models.IntegerField(default=1, db_column='onlyActOnDirectories')
```
Hwesta: Also a [Null]BooleanField.

src/MCPServer/lib/jobChainLink.py
```diff
@@ -37,19 +37,21 @@
...
-constlinkTaskManagerGetMicroserviceGeneratedListInStdOut = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("Get microservice generated list in stdOut"))[0][0]
-constlinkTaskManagerGetUserChoiceFromMicroserviceGeneratedList = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("Get user choice from microservice generated list"))[0][0]
-constlinkTaskManagerSetUnitVariable = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("linkTaskManagerSetUnitVariable"))[0][0]
-constlinkTaskManagerUnitVariableLinkPull = databaseInterface.queryAllSQL("SELECT pk FROM TaskTypes WHERE description = '%s';" % ("linkTaskManagerUnitVariableLinkPull"))[0][0]
+# Constants
+constOneTask = TaskType.objects.get(description="one instance").pk
+constTaskForEachFile = TaskType.objects.get(description="for each file").pk
+constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").pk
+constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").pk
```
Hwesta: Dupe of above line.

src/MCPServer/lib/jobChainLink.py
Hwesta, line 133: -shrug- Fine to leave it, but I prefer an explicit return, if elsewhere it does explicit return.

src/MCPServer/lib/linkTaskManagerDirectories.py
Hwesta, line 44: It shouldn't!  And the Exception is probably clearer anyway.

src/MCPServer/lib/unit.py
mistydemeo, line 49: Updated.

src/MCPServer/lib/unit.py
mistydemeo, line 100: I don't want to change the method signature (even though this is 100% internal), but I'll add a comment.

src/dashboard/src/main/models.py
mistydemeo: I noticed another tinyint boolean was mapped to an IntegerField, so I wasn't sure if we had a preference. I can move these over to BooleanFields though.

mistydemeo:
(Will squash down commits after review.)

src/dashboard/src/main/models.py
Hwesta: Since it's semantically a boolean, I think it makes sense to have the field reflect that.  I think BooleanField is more correct for this case, but since the DB allows NULL, then NullBooleanField may be better?  OTOH, if we want to eventually generate the tables from the models, BooleanField is what we're really want.

mistydemeo:
Updated to replace databaseInterface in archivematicaCommon too, and rebased on top of dev/issue-7303-update-main-models to incorporate all of those changes.

mistydemeo:
Squashed and rebased on qa/1.x. This is ready for review.

At this point databaseInterface has been replaced throughout MCPServer and MCPClient, along with localDevSetup; it remains in use in the following places:
- databaseInterface sqlLock in MCPServer and RCPServer - this can probably be removed?
- Elasticsearch backup
- Unused code (archivematicaFITS, archivematicaFIDO, Tika)
- Archivist's Toolkit integration
- Normalization report (uses django.db.connection, not databaseInterface, but uses raw SQL for a complex statement)
- Email fail report (uses MySQL views that aren't in Django models)
- Maildir
- PlantUMLgenerator

mistydemeo:
Rebased on top of qa/1.x.

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -111,17 +96,17 @@ def executeCommand(gearman_worker, gearman_job):
         sInput = ""
         clientID = gearman_worker.worker_client_id
 
-
-        sql = """SELECT Tasks.taskUUID FROM Tasks WHERE taskUUID='%s' AND startTime != 0;""" % (gearman_job.unique.__str__())
-        rows = databaseInterface.queryAllSQL(sql)
-        if len(rows):
+        task = Task.objects.get(taskuuid=gearman_job.unique)
+        if task.starttime is not None:
```
Hwesta, line 100: Should this be catching a `DoesNotExist` error instead of checking starttime??  It looks like the old code checked for existence.

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -53,8 +55,9 @@
     else:
         print >>sys.stderr, "File already has UUID:", fileUUID
         if opts.update_use:
-            sql = """UPDATE Files SET fileGrpUse='%s' WHERE fileUUID = '%s';""" % (opts.use, fileUUID)
-            databaseInterface.runSQL(sql)
+            f = File.objects.get(uuid=fileUUID)
+            f.filegrpuse = opts.use
+            f.save()
```
Hwesta: Use [update](https://docs.djangoproject.com/en/1.5/ref/models/querysets/#update) for this?

src/MCPClient/lib/archivematicaClient.py
mistydemeo, line 100: I think the old code checked for existence plus start time being non-zero; the logging below indicates it primarily cares about the latter (since the task should always exist here). The implicit assumption appeared to be that there would always be a Task with the requested taskUUID.

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
mistydemeo: Updated.

There's... probably quite a few places I could fix here. 😭

src/MCPClient/lib/archivematicaClient.py
Hwesta, line 100: Ok

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
```diff
@@ -119,15 +121,18 @@ def createFileSec(path, parentBranch, structMapParent):
...
+                    try:
+                        f = File.objects.get(**kwargs)
+                    except File.DoesNotExist:
+                        print >> sys.stderr, "No uuid for file: \"", pathSTR, "\""
+                        continue
+                    else:
+                        myuuid = f.uuid
 
                     if includeAmdSec:
                         createDigiprovMD(myuuid, itempath, myuuid)
```
Hwesta, line 137: It's a good thing `includeAmdSec` must never be true, since `createDigiprovMD` isn't and wasn't imported.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
```diff
@@ -119,15 +121,18 @@ def createFileSec(path, parentBranch, structMapParent):
...
+                    kwargs = {
+                        'removedtime__isnull': True,
+                        fileGroupIdentifier: SIPUUID,
+                        'currentlocation': pathSTR
+                    }
+                    try:
+                        f = File.objects.get(**kwargs)
+                    except File.DoesNotExist:
+                        print >> sys.stderr, "No uuid for file: \"", pathSTR, "\""
+                        continue
```
Hwesta: Should this `continue` be here?  Should it fail the microservice if it cannot find an associated file or continue with a default `myuuid` value or skip this file?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -130,25 +135,20 @@ def createAgent(agentIdentifierType, agentIdentifierValue, agentName, agentType)
 FileMetadataAppliesToType = '7f04d9d4-92c2-44a5-93dc-b7bfdf0c1f17'
 def getDublinCore(unit, id):
     field_list = ["title", "creator", "subject", "description", "publisher", "contributor", "date", "type", "format", "identifier", "source", "relation", "language", "coverage", "rights", "isPartOf"]
```
Hwesta: isPartOf should be fixed to is_part_of to reflect DB-SQL vs DB-Django.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -130,25 +135,20 @@ def createAgent(agentIdentifierType, agentIdentifierValue, agentName, agentType)
...
-    sqlLock.release()
+
+    try:
+        dc = DublinCore.objects.get(metadataappliestotype_id=unit,
+                                    metadataappliestoidentifier=id)
+    except DublinCore.DoesNotExist:
+        return
+
+    ret = etree.Element(ns.dctermsBNS + "dublincore", nsmap={"dc":ns.dctermsNS})
+    ret.set(ns.xsiBNS+"schemaLocation", ns.dctermsNS + " http://dublincore.org/schemas/xmls/qdc/2008/02/11/dcterms.xsd")
```
Hwesta: Fix linter errors on these lines?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -290,14 +290,9 @@ def createTechMD(fileUUID):
...
-    sql = "SELECT fileSize, checksum FROM Files WHERE fileUUID = '%s';" % (fileUUID)
-    c, sqlLock = databaseInterface.querySQL(sql)
-    row = c.fetchone()
-    while row != None:
-        fileSize = row[0].__str__()
-        checksum = row[1].__str__()
-        row = c.fetchone()
-    sqlLock.release()
+    f = File.objects.get(uuid=fileUUID)
+    fileSize = f.size
```
Hwesta: This needs to be converted to a string.

mistydemeo:
Made some style improvements in MCPClient.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
mistydemeo: Updated.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
mistydemeo: I guess for consistency with the old behaviour it shouldn't continue, just proceed with the empty string.

mistydemeo: Removed that line.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS.py
mistydemeo, line 137: 😱

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
```diff
@@ -27,10 +27,12 @@
 import uuid
 import lxml.etree as etree
 
+sys.path.append("/usr/share/archivematica/dashboard")
+from main.models import RightsStatement, RightsStatementCopyright, RightsStatementCopyrightDocumentationIdentifier, RightsStatementCopyrightNote, RightsStatementLicense, RightsStatementLicenseDocumentationIdentifier, RightsStatementLicenseNote, RightsStatementLinkingAgentIdentifier, RightsStatementOtherRightsDocumentationIdentifier, RightsStatementOtherRightsInformation, RightsStatementOtherRightsInformationNote, RightsStatementRightsGranted, RightsStatementRightsGrantedNote, RightsStatementRightsGrantedRestriction, RightsStatementStatuteDocumentationIdentifier, RightsStatementStatuteInformation, RightsStatementStatuteInformationNote
```
Hwesta: Only `RightsStatement` is actually used - the rest don't need to be imported.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
```diff
@@ -46,180 +48,147 @@ def archivematicaGetRights(metadataAppliesToList, fileUUID):
...
-                        etree.SubElement(licenseDocumentIdentifier, ns.premisBNS + "licenseDocumentationIdentifierValue").text = row2[4]
-                        etree.SubElement(licenseDocumentIdentifier, ns.premisBNS + "licenseDocumentationRole").text = row2[6]
-                        
-                        etree.SubElement(licenseInformation, ns.premisBNS + "licenseTerms").text = valueDic["licenseTerms"]
+        statements = RightsStatement.objects.filter(
+            metadataappliestoidentifier=metadataAppliesToidentifier,
+            metadataappliestotype_id=metadataAppliesToType
+        )
+        for statement in statements:
+            valueDic= {}
```
Hwesta: `valueDic` is never used.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
```diff
@@ -46,180 +48,147 @@ def archivematicaGetRights(metadataAppliesToList, fileUUID):
...
+                    copyrightApplicableDates = etree.SubElement(copyrightInformation, ns.premisBNS + "copyrightApplicableDates")
+                    if copyright.copyrightapplicablestartdate:
+                        etree.SubElement(copyrightApplicableDates, ns.premisBNS + "startDate").text = formatDate(copyright.copyrightapplicablestartdate)
+                    if copyright.copyrightenddateopen:
+                        etree.SubElement(copyrightApplicableDates, ns.premisBNS + "endDate").text = "OPEN"
+                    elif copyright.copyrightapplicableenddate:
+                        etree.SubElement(copyrightApplicableDates, ns.premisBNS + "endDate").text = formatDate(copyright.copyrightapplicableenddate)
+
+            elif statement.rightsbasis.lower() in ["license"]:
+                for license in rightsstatementlicense_set.all():
```
Hwesta: Should be `statement.rightsstatementlicense_set.all()`

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
```diff
@@ -46,180 +48,147 @@ def archivematicaGetRights(metadataAppliesToList, fileUUID):
...
+                # 4.1.5 statuteInformation (O, R)
+                getstatuteInformation(statement, rightsStatement)
+
+            elif statement.rightsbasis.lower() in ["donor", "policy", "other"]:
+                otherRightsInformation = etree.SubElement(rightsStatement, ns.premisBNS + "otherRightsInformation")
+                for info in statement.rightsstatementotherrightsinformation_set.all():
+                    # otherRightsDocumentationIdentifier
+                    for identifier in info.rightsstatementotherrightsdocumentationidentifier_set.all():
+                        otherRightsDocumentationIdentifier = etree.SubElement(otherRightsInformation, ns.premisBNS + "otherRightsDocumentationIdentifier")
+                        etree.SubElement(otherRightsDocumentationIdentifier, ns.premisBNS + "otherRightsDocumentationIdentifierType").text = identiifer.otherrightsdocumentationidentifiertype
```
Hwesta: Typo `identiifer` -> `identifier`

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
mistydemeo: Good catch - they were used in a previous version and I didn't update the import.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSRightsDspaceMDRef.py
```diff
@@ -52,20 +54,19 @@ def createMDRefDMDSec(LABEL, itemdirectoryPath, directoryPathSTR):
...
-            metsLoc = row[1].replace("%SIPDirectory%", "", 1)
-            metsLocation = os.path.join(os.path.dirname(itemdirectoryPath), "mets.xml")
-            LABEL = "mets.xml-%s" % (metsFileUUID)
-            ret.append(createMDRefDMDSec(LABEL, metsLocation, metsLoc))
+        # find the mets file
+        path = "%%SIPDirectory%%%s/mets.xml".format(os.path.dirname(filePath))
+        mets = File.objects.get(currentlocation=path,
+                                transfer_id=transferUUID)
+        metsFileUUID = mets.uuid
+        metsloc = mets.currentlocation.replace("%SIPDirectory%", "", 1)
```
Hwesta: `metsloc` vs `metsLoc` variable name

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -53,8 +55,7 @@
     else:
         print >>sys.stderr, "File already has UUID:", fileUUID
         if opts.update_use:
-            sql = """UPDATE Files SET fileGrpUse='%s' WHERE fileUUID = '%s';""" % (opts.use, fileUUID)
-            databaseInterface.runSQL(sql)
+            File.objects.filter(uuid=fileUUID).update(filegrpuse=use)
```
Hwesta: `use` -> `opts.use`

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -128,27 +133,24 @@ def createAgent(agentIdentifierType, agentIdentifierValue, agentName, agentType)
 SIPMetadataAppliesToType = '3e48343d-e2d2-4956-aaa3-b54d26eb9761'
 TransferMetadataAppliesToType = '45696327-44c5-4e78-849b-e027a189bf4d'
 FileMetadataAppliesToType = '7f04d9d4-92c2-44a5-93dc-b7bfdf0c1f17'
+
+
 def getDublinCore(unit, id):
     field_list = ["title", "creator", "subject", "description", "publisher", "contributor", "date", "type", "format", "identifier", "source", "relation", "language", "coverage", "rights", "isPartOf"]
```
Hwesta: `isPartOf` -> `is_part_of` to match Django model name not DB column name.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSTrim.py
```diff
@@ -83,47 +83,50 @@ def getTrimFileDmdSec(baseDirectoryPath, fileGroupIdentifier, fileUUID):
...
+                             filegrpuuid=fileUUID,
+                             filegrpuse="TRIM file metadata")
+    except File.DoesNotExist:
         print >>sys.stderr, "no metadata for original file: ", fileUUID
         return None
-    for row in rows:
-        label = os.path.basename(row[0])
+    else:
+        label = os.path.basename(f.currentlocation)
         attrib = {"LABEL":label, ns.xlinkBNS + "href":row[0].replace("%SIPDirectory%", "", 1), "MDTYPE":"OTHER", "OTHERMDTYPE":"CUSTOM", 'LOCTYPE':"OTHER", 'OTHERLOCTYPE':"SYSTEM"}
```
Hwesta: `row[0]` should probably be something else.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSTrim.py
```diff
@@ -83,47 +83,50 @@ def getTrimFileDmdSec(baseDirectoryPath, fileGroupIdentifier, fileUUID):
...
-    
-    
-    sql = "SELECT currentLocation FROM Files WHERE removedTime = 0 AND %s = '%s' AND fileGrpUse='TRIM file metadata' AND fileGrpUUID = '%s';" % ('sipUUID', fileGroupIdentifier, fileUUID)
-    rows = databaseInterface.queryAllSQL(sql)
-    if (len(rows) != 1):
+
+    try:
+        f = File.objects.get(removedtime__isnull=True,
+                             sip_id=fileGroupIdentifier,
+                             filegrpuuid=fileUUID)
```
Hwesta: This seems to have dropped the `fileGrpUse='TRIM file metadata'`

src/MCPClient/lib/clientScripts/backlogUpdatingTransferFileIndex.py
```diff
@@ -21,19 +21,17 @@
...
-from executeOrRunSubProcess import executeOrRun
 from fileOperations import renameAsSudo
 import elasticSearchFunctions
 
 def updateDB(dst, transferUUID):
-    sql =  """UPDATE Transfers SET currentLocation='""" + MySQLdb.escape_string(dst) + """' WHERE transferUUID='""" + transferUUID + """';"""
-    databaseInterface.runSQL(sql)
+    Transfer.objects.filter(uuid=transferUUID).update(currentlocation=dst)
 
 def moveSIP(src, dst, transferUUID, sharedDirectoryPath):
```
Hwesta: moveSIP and updateDB aren't actually called. >__>  Can probably be removed.

src/MCPClient/lib/clientScripts/checkForAccessDirectory.py
```diff
@@ -44,41 +47,30 @@ def something(SIPDirectory, accessDirectory, objectsDirectory, DIPDirectory, SIP
...
-                #ValueError: unsupported format character ''' (0x27) at index 76
-                sql = "SELECT fileUUID, currentLocation FROM Files WHERE currentLocation LIKE  '" + objectNameLike + "%' AND removedTime = 0 AND SIPUUID = '"+ SIPUUID + "'"
-                c, sqlLock = databaseInterface.querySQL(sql)
-                row = c.fetchone()
-                if not row:
+
+                files = File.objects.filter(removedtime__isnull=True,
+                                            currentlocation__startswith=objectNameLike,
+                                            sip_id=SIPUUID)
+                if files.count() < 1:
```
Hwesta: `if not files.exists()` probably better

src/MCPClient/lib/clientScripts/backlogUpdatingTransferFileIndex.py
mistydemeo: Added a commit that removes them, and removed this script from the MCPClient commit.

mistydemeo:
Tested with many of the alternate workflows, with some additional bugfixes and stylistic improvements.

mistydemeo:
At this point I've tested the following workflows to be working:
- Standard transfer
- Zipped bag
- Unzipped bag
- DSpace
- Manual normalization (via CSV, and via "normalize manually")
- CSV and JSON DC metadata
- Disk image with metadata
- Arbitrary non-Unicode filenames, and Unicode filenames

src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
```diff
@@ -76,14 +69,16 @@
 
     #get the database list of files in the objects directory
     #for each file, confirm it's in the SIP objects directory, and update the current location/ owning SIP'
-    sql = """SELECT  fileUUID, currentLocation FROM Files WHERE removedTime = 0 AND currentLocation LIKE '\%transferDirectory\%objects%' AND transferUUID =  '""" + transferUUID + "'"
-    for row in databaseInterface.queryAllSQL(sql):
-        fileUUID = row[0]
-        currentPath = databaseFunctions.deUnicode(row[1])
+    files = File.objects.filter(transfer_id=transferUUID,
+                                currentlocation__startswith='%transferDirectory%objects')
```
Hwesta: Should removedTime be in the query?

src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
```diff
@@ -76,14 +69,16 @@
...
+    for f in files:
+        fileUUID = f.uuid
```
Hwesta: fileUUID appears to be unused

src/MCPClient/lib/clientScripts/createSIPsfromTRIMTransferContainers.py
```diff
@@ -63,23 +57,26 @@
...
-            fileUUID = row[0]
-            currentPath = databaseFunctions.deUnicode(row[1]).replace('%transferDirectory%objects/' + container, '%transferDirectory%objects')
+        # get the database list of files in the objects directory
+        # for each file, confirm it's in the SIP objects directory, and update the current location/ owning SIP'
+        dir = '%transferDirectory%objects/' + container
+        files = File.objects.filter(removedtime__isnull=True,
+                                    currentlocation__startswith=dir,
+                                    transfer_id=transferUUID)
+        for f in files:
+            currentPath = databaseFunctions.deUnicode(f.currentlocation).replace('%transferDirectory%objects/' + container, '%transferDirectory%objects')
```
Hwesta: Replace `'%transferDirectory%objects/' + container` with variable defined above?  Rename `dir` so it doesn't conflict with builtins.

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -25,24 +25,21 @@
...
-    rows = databaseInterface.queryAllSQL(sql)
-    for row in rows:
-        email = row[0]
-        if email == "demo@example.com":
-            continue
-        if not email:
-            continue
-        ret.append(email)
-    return ret
+    return [u[0] for u in User.objects.filter(is_active=True).values_list('email').exclude(email__in=['demo@example.com', ''])]
```
Hwesta, line 41: :+1:  I love it when we can replace 11 lines with 1 that's just as clear.

src/MCPClient/lib/clientScripts/loadLabelsFromCSV.py
```diff
@@ -48,10 +48,4 @@
...
                 filePath = row[0]
             filePath = os.path.join("%transferDirectory%objects/", filePath)
-            sql = """UPDATE Files SET label = '%s' WHERE originalLocation = '%s' AND transferUUID = '%s';""" % (MySQLdb.escape_string(label), MySQLdb.escape_string(filePath), transferUUID)
-            databaseInterface.runSQL(sql)
-            
-
-
-    
-    
+            File.objects.filter(originallocation=label, transfer_id=transferUUID).update(label=label)
```
Hwesta: Should be `originallocation=filePath`

src/MCPClient/lib/clientScripts/manualNormalizationCheckForManualNormalizationDirectory.py
```diff
@@ -40,7 +41,11 @@
             # made and setting unit variables in client scripts is missed! This
             # should be eliminated (by having another way of setting when DIPs
             # are created) or moved to its own MSCL.
-            databaseInterface.runSQL("""UPDATE UnitVariables SET microServiceChainLink = 'f060d17f-2376-4c0b-a346-b486446e46ce' WHERE unitType='SIP' AND unitUUID = '%s' AND variable = 'returnFromManualNormalized' """ % (SIPUUID) )
+            uv = UnitVariable.objects.get(unittype="SIP",
+                                          unituuid=SIPUUID,
+                                          variable="returnFromManualNormalized")
+            uv.microservicechainlink_id = "f060d17f-2376-4c0b-a346-b486446e46ce"
+            uv.save()
```
Hwesta: Use `.update()`?

mistydemeo: Updated - but this idiom is used all over the place (not just in this PR). Do you care enough about changing it all now?

Hwesta: Kind of? It's better but depending on how many there are that might not justify the work involved.

src/MCPClient/lib/clientScripts/manualNormalizationCreateMetadataAndRestructure.py
```diff
@@ -44,17 +46,21 @@
 if i != -1 and k != -1:
     filePathLike = filePathLike[:i+1]
     # Matches "path/to/file/filename." Includes . so it doesn't false match foobar.txt when we wanted foo.txt
-    filePathLike1 = databaseInterface.MySQLdb.escape_string(filePathLike).replace("%", "\%") + "%"
+    filePathLike1 = filePathLike
     # Matches the exact filename.  For files with no extension.
-    filePathLike2 = databaseInterface.MySQLdb.escape_string(filePathLike)[:-1]
-unitIdentifierType = "sipUUID"
+    filePathLike2 = filePathLike[:-1]
+unitIdentifierType = "sip_id"
```
Hwesta: Can we replace `unitIdentifierType` with just `sip_id`?  It only seems to have one value.

mistydemeo: Good thought. I get the idea this might have been intended to be generalized like some of the other scripts, with this passed at the CLI, but since that didn't happen / won't happen I'd rather take it out.

mistydemeo: Inlined both `sip_id` and used `SIPUUID` directly instead of `unitidentifier`. This also lets me avoid using kwargs dicts in favour of directly passing kwargs, so that's nice.

src/MCPClient/lib/clientScripts/manualNormalizationMoveAccessFilesToDIP.py
```diff
@@ -39,21 +41,23 @@
...
-sql = "SELECT Files.fileUUID, Files.currentLocation FROM Files WHERE removedTime = 0 AND fileGrpUse='original' AND Files.currentLocation LIKE '" + filePathLike1 + "' AND " + unitIdentifierType + " = '" + unitIdentifier + "';"
-rows = databaseInterface.queryAllSQL(sql)
-if not len(rows):
-    #If not found try without extension
-    sql = "SELECT Files.fileUUID, Files.currentLocation FROM Files WHERE removedTime = 0 AND fileGrpUse='original' AND Files.currentLocation = '" + filePathLike2 + "' AND " + unitIdentifierType + " = '" + unitIdentifier + "';"
-rows = databaseInterface.queryAllSQL(sql)
-if len(rows) != 1:
+
+try:
+    path_condition = Q(currentlocation__startswith=filePathLike) | Q(currentlocation=filePathLike2)
```
Hwesta: While clever, I don't think that the Q statement does the same thing as the original code.

Given files `foo.txt` and `foobar.txt` with a search file of `foo.jpg`, the original code will only match `foo.txt` while the new code will match both and trigger the normalization.csv exception.  By searching for the path with the . first, and only searching without it if nothing was initially returned, we ensured that we'd find files that matched just the basename and reduced false positives.

Whether the old code was a _good_ way to accomplish our goal is another question...

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -638,28 +599,38 @@ def createFileSec(directoryPath, parentDiv):
...
-                fileGrpUUID = row[2]
-                transferUUID = row[3]
-                label = row[4]
-                originalLocation = row[5]
-                typeOfTransfer = row[6]
-                row = c.fetchone()
-            sqlLock.release()
-            
+
+            kwargs = {
```
mistydemeo: Oops, duplicated code in the merge.

src/MCPClient/lib/clientScripts/manualNormalizationCheckForManualNormalizationDirectory.py
mistydemeo: Turns out there was a much smaller surface area of stuff to change here than I thought, and it's definitely better to be consistent at this point, so made the changes everywhere there wasn't much restructuring to do so.

src/MCPClient/lib/clientScripts/manualNormalizationMoveAccessFilesToDIP.py
mistydemeo: Good catch - I copied the code here a bit too closely from the other tool that does something similar. Fixed.

src/MCPClient/lib/clientScripts/trimCreateRightsEntries.py
```diff
@@ -106,74 +109,68 @@ def getDateTimeFromDateClosed(dateClosed):
...
+        identifier = RightsStatementOtherRightsDocumentationIdentifier.objects.create(
+            rightsstatementotherrights=info
+        )
+
+        # RightsStatementRightsGranted
+        granted = RightsStatementRightsGranted.objects.create(
+            rightsstatement=statement,
+            act="Disseminate",
+            startdate = startTime,
+            enddate = endTime
```
Hwesta: Nitpick, but remove spaces around the =?

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -59,8 +58,8 @@
 from externals.singleInstance import singleinstance
 from archivematicaFunctions import unicodeToStr
 
-# Print SQL sent to the database interface 
-databaseInterface.printSQL = True
```
Hwesta, line 63: Hmm, this is going to remove most of the logging in archivematicaMCPClient-<hostname>.log

... that's probably a good thing, but we should look at adding enough logging to usefully replace that.

src/MCPServer/lib/jobChain.py
```diff
@@ -67,20 +64,11 @@ def __init__(self, unit, chainPK, notifyComplete=None, passVar=None, UUID=None,
...
-            return None
-        while row != None:
-            print "jobChain", row
-            #self.pk = row[0]
-            self.startingChainLink = row[1]
-            self.description = row[2]
-            row = c.fetchone()
-        sqlLock.release()
+
+        chain = MicroServiceChain.objects.get(id=str(chainPK))
```
Hwesta, line 68: The original code handles the case where there is no MicroServiceChain by returning None - this probably should as well.

src/MCPServer/lib/linkTaskManagerDirectories.py
```diff
@@ -28,41 +28,27 @@
...
-        sqlLock.release()
-        #if reloadFileList:
-        #    unit.reloadFileList()
-
-        #        "%taskUUID%": task.UUID.__str__(), \
+        stc = StandardTaskConfig.objects.get(id=str(pk))
+        filterFileEnd = stc.filter_file_end
+        filterFileStart = stc.filter_file_start
+        filterSubDir = stc.filter_subdir
+        self.requiresOutputLock = stc.requires_output_lock
```
Hwesta, line 46: This is now a bool, not a string.  Is this a problem?  Same question for other places that use StandardTaskConfig

src/MCPServer/lib/linkTaskManagerFiles.py
```diff
@@ -45,40 +45,39 @@ def __init__(self, jobChainLink, pk, unit):
...
-            row = c.fetchone()
-        sqlLock.release()
-        if requiresOutputLock:
+
+        stc = StandardTaskConfig.objects.get(id=str(pk))
+        # These three may be concatenated/compared with other strings,
+        # so they need to be bytestrings here
+        filterFileEnd = str(stc.filter_file_end) if stc.filter_file_end else ''
+        filterFileStart = str(stc.filter_file_start) if stc.filter_file_start else ''
+        filterSubDir = str(stc.filter_subdir) if stc.filter_subdir else ''
```
Hwesta, line 54: Should filterFoo get the same treatment in all the other places that query StandardTaskConfig?

src/MCPServer/lib/linkTaskManagerDirectories.py
mistydemeo, line 46: Should be fine! This was being fetched as an integer before, and Python treats `if 0/if 1` and `if False/if True` the same.

Hwesta, line 46: Ok.  Wasn't sure if it was getting fetched as `"0"` and `"1"`

src/MCPServer/lib/unit.py
```diff
@@ -34,7 +78,30 @@ def setMagicLink(self,link, exitStatus=""):
         return
 
     def setVariable(self, variable, variableValue, microServiceChainLink):
-        return
+        if not variableValue:
+            variableValue = ""
```
Hwesta, line 82: Set microServiceChainLink to None if not set?

src/MCPServer/lib/unit.py
```diff
@@ -34,7 +78,30 @@ def setMagicLink(self,link, exitStatus=""):
...
+        vars = UnitVariable.objects.filter(unittype=self.unitType,
+                                           unituuid=self.UUID,
```
Hwesta, line 84: Continuing with 'Why has Python taken all the good variable names?', `vars` is a builtin.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -20,84 +20,181 @@
...
+    ret = ""
+    for c in date:
+        if c in valid:
+            ret += c
+        #else:
+            #ret += replacementChar
+    return str("{:10.10f}".format(float(ret)))
+
+def insertIntoFiles(fileUUID, filePath, enteredSystem=getUTCDate(), transferUUID="", sipUUID="", use="original"):
+    """
```
Hwesta, line 48: Change `enteredSystem=getUTCDate` to `enteredSystem=None` and check for None to set enteredSystem.  Function signatures are evaluated when the file is loaded, so getUTCDate() will only be evaluated once when the file is loaded; likely not what we want.

Same for other functions that have getUTCDate() in the function declaration.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -20,84 +20,181 @@
...
     agent = getAgentForFileUUID(fileUUID)
-    if eventIdentifierUUID == "":
-        eventIdentifierUUID = uuid.uuid4().__str__()
-    databaseInterface.runSQL("""INSERT INTO Events (fileUUID, eventIdentifierUUID, eventType, eventDateTime, eventDetail, eventOutcome, eventOutcomeDetailNote, linkingAgentIdentifier)
-        VALUES (%s, %s, %s, %s, %s, %s, %s, %s)""",
-        (fileUUID, eventIdentifierUUID, eventType, eventDateTime,
-         eventDetail, eventOutcome, eventOutcomeDetailNote, agent)
-    )
+    if not eventIdentifierUUID:
+        eventIdentifierUUID = (uuid.uuid4())
```
Hwesta: ```str(uuid.uuid4())```?

Hwesta: Should be `str(uuid.uuid4())`?

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -20,84 +20,181 @@
...
+    Creates a new entry in the Derivations table using the supplied arguments. The two files in this relationship should already exist in the Files table.
+
+    :param str sourceFileUUID: The UUID of the original file.
+    :param str derivedFileUUID: The UUID of the derived file.
+    :param str relatedEventUUID: The UUID for an event describing the creation of the derived file. Can be blank.
+    """
+    if not sourceFileUUID:
+        raise ValueError("sourceFileUUID must be specified")
+    if not derivedFileUUID:
+        raise ValueError("derivedFileUUID must be specified")
```
Hwesta, line 153: If they're required, then don't make the optional in the function definition?

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -20,84 +20,181 @@
...
+
+    derivation = Derivation(source_file_id=sourceFileUUID,
+                            derived_file_id=derivedFileUUID,
+                            event_id=relatedEventUUID)
+    derivation.save()
```
Hwesta: Minor, but `Derivation.objects.create(source_file_id=sourceFileUUID, derived_file_id=derivedFileUUID, event_id=relatedEventUUID)` does the same thing without having to remember to save it.

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo, line 153: We already make heavy use of calling these as kwargs - rather than change the function signature and every place it's called, I thought I'd just verify the arguments instead. I could make that change in a separate commit though if you wanted.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -106,53 +203,72 @@ def logTaskCreatedSQL(taskManager, commandReplacementDic, taskUUID, arguments):
...
+    """
+    Fetches execution data from the completed task and logs it to the database.
+    Updates the entry in the Tasks table with data in the provided task.
+    Saves the following fields: exitCode, stdOut, stdError
+
+    :param task:
+    """
     print "Logging task output to db", task.UUID
     taskUUID = task.UUID.__str__()
     exitCode = task.results["exitCode"].__str__()
```
Hwesta, line 223: This is a BigIntegerField - should it be converted to a str?

src/archivematicaCommon/lib/databaseFunctions.py
Hwesta, line 153: All python functions can be called by specifying their args in order, or by naming them regardless of how the parameter is defined (http://www.diveintopython.net/power_of_introspection/optional_arguments.html) so it shouldn't affect the places this is called.

Probably worthwhile to keep the checks to catch things like `""` since those errors are almost certainly better than what Django generates if the constraint fails.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -106,53 +203,72 @@ def logTaskCreatedSQL(taskManager, commandReplacementDic, taskUUID, arguments):
...
-    #TODO -un hardcode executing exeCommand
-
-def logJobStepCompletedSQL(job):
-    databaseInterface.runSQL("""INSERT INTO jobStepCompleted (jobUUID, step, completedTime)
-        VALUES (%s, %s, %s)""",
-        (job.UUID, job.step, databaseInterface.getUTCDate())
-    )
-
-def fileWasRemoved(fileUUID, utcDate=databaseInterface.getUTCDate(), eventDetail = "", eventOutcomeDetailNote = "", eventOutcome=""):
+    job = Job(jobuuid=job.UUID,
```
Hwesta: Redefinition of variable name - job is both the parameter jobChainLink and the newly created Job.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -106,53 +203,72 @@ def logTaskCreatedSQL(taskManager, commandReplacementDic, taskUUID, arguments):
...
+              jobtype=job.description,
+              directory=job.unit.currentPath,
+              sipuuid=unitUUID,
+              currentstep="Executing command(s)",
+              unittype=job.unit.__class__.__name__,
+              microservicegroup=str(job.microserviceGroup),
```
Hwesta: Is this already a str?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -798,8 +797,9 @@ def connect_and_remove_transfer_files(uuid, unit_type=None):
     if unit_type == 'transfer':
         transfers = {uuid}
     else:
-        sql = "SELECT transferUUID FROM Files WHERE transferUUID=%s OR sipUUID=%s"
-        transfers = {f[0] for f in databaseInterface.queryAllSQL(sql, (uuid, uuid))}
+        condition = Q(transfer_id=uuid) | Q(sip_id=uuid)
+        File.objects.filter(condition).values_list('transfer_id')
```
Hwesta: Can be removed.

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo, line 153: I'll punt that to a new PR to avoid increasing the surface area of this one, but let's make those changes.

src/archivematicaCommon/lib/fileOperations.py
```diff
@@ -155,20 +155,21 @@ def updateFileLocation2(src, dst, unitPath, unitIdentifier, unitIdentifierType,
...
-        print >> sys.stderr, 'Rows returned:', rows
+    kwargs = {
+        "removedtime__isnull": True,
+        "currentlocation": srcDB,
+        unitIdentifierType: unitIdentifier
+    }
+    files = File.objects.filter(**kwargs)
+    count = files.count()
+    if count != 1:
+        print >> sys.stderr, 'ERROR: file information not found:', count, "rows for arguments:", repr(kwargs)
```
Hwesta: Use .get() and catch DoesNotExist and MultipleResults instead of counting?  Does mean you can't use .update below, but I think it's clearer.

src/MCPServer/lib/jobChain.py
mistydemeo, line 68: Actually, is that okay? Returning `None` has the effect of creating an instance with no attributes, which will blow up only later when something tries to access one of its standard attributes. It's probably better to raise something from within the constructor (even if `MicroServiceChain.DoesNotExist` is not the ideal exception to allow to bubble up)

src/MCPServer/lib/linkTaskManagerFiles.py
mistydemeo, line 54: We only use those properties here and in linkTaskManagerDirectories. I don't think ltmDirectories can get unusual paths like we can in ltmFiles.

(The properties are also assigned, but never used, in ltmGetMicroserviceGeneratedListInStdOut)

src/MCPServer/lib/unit.py
mistydemeo, line 82: I think it's valid to have an empty variable value, but not valid to have a variable with no microServiceChainLink. I see `null=True` in the model definition, but I'm not sure the original code allowed that?

src/MCPServer/lib/unit.py
mistydemeo, line 84: 😭

src/archivematicaCommon/tests/test_database_functions.py
```diff
@@ -0,0 +1,66 @@
...
+import databaseFunctions
+
+sys.path.append("/usr/share/archivematica/dashboard")
+from main.models import Event, File
+
+from django.test import TestCase
+import pytest
+
+
+class TestDatabaseFunctions(TestCase):
```
Hwesta, line 14: :D :clap:

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo, line 48: Yeah, I'm not happy this is a common thing. Mind if I make it a followup PR to make this change? I'm focusing on not changing functionality in this one.

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo: Good catch!

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo: Thanks, I thought I'd caught all of these

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo, line 223: I have a feeling it was always being coerced into a string and then back into an integer in the past; could adjust it in a followup PR to not do that?

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo: Ouch, good catch. I'll just not assign the new job to a variable since it's never used/returned.

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo: Not sure - the old code also always called `str()` on it though.

src/archivematicaCommon/lib/fileOperations.py
mistydemeo: Updated.

mistydemeo:
Okay - addressed the comments!

mistydemeo:
Rebased on top of qa/1.x.

src/dashboard/src/main/models.py
```diff
@@ -691,3 +837,45 @@ class Meta:
 
     def __unicode__(self):
         return self.term
+
+class WatchedDirectory(models.Model):
+    id = UUIDPkField()
+    watched_directory_path = models.TextField(null=True, blank=True, db_column='watchedDirectoryPath')
+    chain = models.ForeignKey('MicroServiceChain', null=True, db_column='chain')
+    only_act_on_directories = models.IntegerField(default=1, db_column='onlyActOnDirectories')
```
Hwesta: [Null]BooleanField?

Hwesta:
Whew!  I think this is ready to go in and be tested in reality. :+1:

src/dashboard/src/main/models.py
mistydemeo: I'll swap it to `BooleanField`; we've never used `NULL` values in this despite it being allowed in the schema.

mistydemeo:
Pushed a fix for the eraseElasticSearchIndexes script. Will give this a round of testing, then merge to qa/1.x.

src/archivematicaCommon/lib/databaseFunctions.py
mistydemeo, line 48: Addressed in https://github.com/artefactual/archivematica/pull/162.

PR #99, opened 2014-08-22
SplitTransfer

pbrantner:
Python script splitTransfer. Within a customized workflow it is possible
to split one transfer to multiple SIPs.

### DIFF
--- a/src/MCPClient/lib/clientScripts/splitTransfer.py
+++ b/src/MCPClient/lib/clientScripts/splitTransfer.py
@@ -1,0 +1,32 @@
+#!/usr/bin/python -OO
+
+# This file is part of a new Archivematica Workflow
+# It creates a new folder for each AIP that should get created
+# and moves the corresponding folders into the created ones
+#
+# @author Patrick Brantner
+
+import os
+import sys
+import shutil
+
+def getAipUid(folder):
+    return "SIP-" + folder
+		
+if __name__ == '__main__':
+    rootDir = sys.argv[1]
+    destination = sys.argv[2]
+    for dirname in os.listdir(rootDir):
+        src = os.path.join(rootDir,dirname)
+        if (os.path.isdir(src)):
+            uid = getAipUid(dirname)
+            des = os.path.join(rootDir,uid)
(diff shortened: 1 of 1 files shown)


mistydemeo:
Hi! Sorry for not replying sooner. Can you add a comment at the top with some documentation as to how this is used? For example, if you're adding it to the database as a microservice chain link, what variables does it use as arguments?

mistydemeo:
In the thread on the mailing list, you mentioned a new watched directory that it used. Could you also add a comment with some documentation about it?

src/MCPClient/lib/clientScripts/splitTransfer.py
```diff
@@ -0,0 +1,32 @@
...
+            des = os.path.join(rootDir,uid)
+            if not os.path.exists(os.path.join(rootDir,uid)):
+                os.makedirs(os.path.join(rootDir,uid))
+            shutil.move(src,des)
+            print "Moved " + dirname + " into SIP " + uid
+    for dirname in os.listdir(rootDir):
+        src = os.path.join(rootDir,dirname)
+        if (os.path.isdir(src)):
+            shutil.move(src,destination)
+            print "Starting transfer of " + dirname
```
mistydemeo, line 32: Just a note (I can fix it up when merging), we typically prefer that files end with a newline.

mistydemeo:
Ping!

pbrantner:
Steps made:
- add the script in the scripts-folder and of course the entry in the archivematicaClientModules file
- create a folder called "splittetTransfers" in watchedDirectories/activeTransfers
- database manipulation: "splittedTransfers" must be the watchedDirectory of the old ingest-workflow. the watchedDirectory that was used in the first place must start a setFilePermission-script and that must start the splitTransfer script

sevein:
Sadly, we haven't had time to tackle this. I've filed a new issue (feature request) capturing the idea: https://github.com/archivematica/Issues/issues/224. In the long term I think it's going to be easier to push this forward if we start from the issue where it's possible to discuss other alternatives.

PR #100, opened 2014-08-23
Send post store AIP callback to storage service (redux)

mistydemeo:
The last commit in the original PR (#82) was backed out since it was committed prematurely - opening a new PR for it until it gets recommitted. It's tiny and was already code reviewed, so the PR is more of a reminder at this point.


mistydemeo:
Updated to reflect the new URL pattern used when this was merged into the storage service.

PR #101, opened 2014-08-27
identifyFileFormat: quote arguments

mistydemeo:
The arguments were being misinterpreted in the case that `relativeLocation` contained a space.

I think it's actually a bug that directories with spaces were possible, but that's a separate issue.

Fixes #7150.


Hwesta:
:+1:

PR #102, opened 2014-08-28
Fix issue causing slow archival storage pageload.

mcantelon:
Fixed a minor issue causing the archival storage page to
load slowly when there are AIPs with extensive METS data.


Hwesta:
Should this change be made elsewhere to make other pages in archival storage load faster?

mistydemeo:
Going to merge this, but we should probably look at other places we could optimize too.

PR #103, opened 2014-08-29
Move upload-qubit, sanitizeNames, transcoder into MCPClient and eliminate their separate packages

mistydemeo:
These tools were basically internal to Archivematica to begin with, so it seemed strange to package and install them separately. This eliminates the extra packages and moves them into MCPClient.

In the process I updated sanitizeObjectNames so that it no longer shells out to an external script and instead calls `sanitizeNames.sanitizeRecursively`. That function has also been improved so it doesn't print directly, but instead returns `(old, new)` filename tuples to make it easier to process over the results.

UPDATE:

Added a new commit which removes the transcoder package too, and moves the one script still in use into MCPClient.


jhsimpson:
Nice :thumbsup:

Hwesta:
So happy to see this cleanup!

Are metaPackage-forensic-tools and metaPackage-shotgun actually unused?  I recall a discussion about Austin using them for packaging things?

Maybe update sanitize[Object]Names to use Python3 compatible syntax?  Out of scope, but if it's quick probably worthwhile.

jhsimpson:
The meta packages are not used.  They were experimental and that work has
moved to deployment.git

Hwesta:
Ok, great.

PR #104, opened 2014-09-03
Replace Python 2.5-style exception syntax with 2.6-style

mistydemeo:
Also fixes up an exception handling block which captured an exception under one name and then tried to use the object under a different name.

That exception handling block was the original inspiration, but decided to touch up the syntax everwhere. Came across it here: https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/databaseInterface.py#L52-L56 When a MySQL exception was raised but wasn't handled by the first case, and so blew up.

This should also help us with a hypothetical Python 3.x migration, since the old style of exception catching no longer exists there.


Hwesta:
:+1: :+1:

PR #105, opened 2014-09-03
Start transfer calls storage service

Hwesta:
Updated Archivematica to call the storage service to move files when starting a transfer, instead of assuming all files are locally available.

Part of this work cherry-picked from dev/issue-6488-aip-reingest, specifically the work to upload metadata files from a transfer source.  This will need to be cleaned up if it is merged before AIP reingest is.

JS browser now tracks the UUID of the Location a path was selected from, and POSTs that to the dashboard in the form `<UUID>:<full path>`.  Note the UUID is not displayed to the user - it is the 'id' of the element containing the path displayed to the user.

The dashboard POSTs to the storage service to move these to the pipeline's temp directory, then moves them to the correct watched directory when complete.

Some cleanup of dead code as well.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -145,17 +145,23 @@ def delete_arrange(request):
 
 def start_transfer(request):
     transfer_name = archivematicaFunctions.unicodeToStr(request.POST.get('name', ''))
-    # Note that the path may contain arbitrary, non-unicode characters,
-    # and hence is POSTed to the server base64-encoded
     transfer_type = archivematicaFunctions.unicodeToStr(request.POST.get('type', ''))
     accession = archivematicaFunctions.unicodeToStr(request.POST.get('accession', ''))
+    # Note that the path may contain arbitrary, non-unicode characters,
+    # and hence is POSTed to the server base64-encoded
     paths = request.POST.getlist('paths[]', [])
     paths = [base64.b64decode(path) for path in paths]
     row_ids = request.POST.getlist('row_ids[]', [])
 
+    if not transfer_name:
+        return helpers.json_response({'error': True, 'message': 'No transfer name provided.'})
+    if not paths:
+        return helpers.json_response({'error': True, 'message': 'No path provided.'})
+
     # Create temp directory that everything will be copied into
-    temp_base_dir = helpers.get_client_config_value('temp_dir') or None
+    temp_base_dir = os.path.join(SHARED_DIRECTORY_ROOT, 'tmp')
     temp_dir = tempfile.mkdtemp(dir=temp_base_dir)
+    os.chmod(temp_dir, 0o770)  # Needs to be writeable by the SS
 
     for i, path in enumerate(paths):
         index = i + 1  # so transfers start from 1, not 0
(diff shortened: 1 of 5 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -46,10 +46,9 @@
 logging.basicConfig(filename="/tmp/archivematicaDashboard.log",
     level=logging.INFO)
 
-SHARED_DIRECTORY_ROOT   = '/var/archivematica/sharedDirectory'
-ACTIVE_TRANSFER_DIR     = SHARED_DIRECTORY_ROOT + '/watchedDirectories/activeTransfers'
-STANDARD_TRANSFER_DIR   = ACTIVE_TRANSFER_DIR + '/standardTransfer'
-ORIGINAL_DIR            = SHARED_DIRECTORY_ROOT + '/www/AIPsStore/transferBacklog/originals'
+SHARED_DIRECTORY_ROOT = helpers.get_server_config_value('sharedDirectory')
```
mistydemeo, line 49: :thumbsup:

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -46,10 +46,9 @@
...
+ACTIVE_TRANSFER_DIR = SHARED_DIRECTORY_ROOT + '/watchedDirectories/activeTransfers'
```
mistydemeo: As long as we're updating these, switch to `os.path.join`?

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -188,88 +171,51 @@ def start_transfer(request):
...
-                               transfer_metadata_set_row_uuid=row_id)
 
+        transfer_relative = transfer_dir.replace(SHARED_DIRECTORY_ROOT, '', 1)
+        copy_from_transfer_sources([path], transfer_relative)
+
+        try:
+            copy_to_start_transfer(filepath=filepath,
+                type=transfer_type, accession=accession,
+                transfer_metadata_set_row_uuid=row_id)
+        except Exception:
```
mistydemeo, line 190: Catch a more limited set of exceptions here?

Hwesta, line 190: `copy_to_start_transfer` raises an `Exception` with whatever the error is, unfortunately.

mistydemeo, line 190: :crying_cat_face:

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta: Done

Hwesta:
Updated with a cleaned up copy of adding move from transfer source, from #109

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -651,6 +597,50 @@ def copy_to_arrange(request):
...
+    """
+    Helper to copy files from transfer source locations to the currently processing location.
+
+    Any files in locations not associated with this pipeline will be ignored.
+
+    :param list paths: List of paths.  Each path should be formatted <uuid of location>:<full path in location>
+    :param str relative_destination: Path relative to the currently processing space to move the files to.
+    :returns: Tuple of (boolean error, message)
+    """
+    processing_location = storage_service.get_location(purpose='CP')[0]
```
mistydemeo, line 610: `get_location` can return more than one result?

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -651,6 +597,50 @@ def copy_to_arrange(request):
...
+            logging.debug('Path %s cannot be split into location:path', p)
+            continue
+        if location not in files:
+            logging.debug('Location %s is not associated with this pipeline.', location)
+            continue
+
+        source = path.replace(files[location]['location']['path'], '', 1).lstrip('/')
+        # Use the last segment of the path for the destination - basename for a
+        # file, or the last folder if not. Keep the trailing / for folders.
+        last_segment = os.path.basename(source) or os.path.basename(source.rstrip('/')) + '/'
```
mistydemeo: Shouldn't the first condition here always be true? When would `os.path.basename(source)` not evaluate as true?

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -651,6 +597,50 @@ def copy_to_arrange(request):
...
+        files[location]['files'].append({'source': source, 'destination': destination})
+        logging.debug('source: %s, destination: %s', source, destination)
+
+    message = []
+    for pl in files.itervalues():
+        reply, error = storage_service.copy_files(pl['location'], processing_location, pl['files'])
+        if reply is None:
+            message.append(error)
+    if message:
+        return True, 'The following errors occured: %s' % ', '.join(message)
```
mistydemeo: Use `.format` instead of `%`?

mistydemeo:
:thumbsup:, beyond the comments I left. Really glad this took so little code!

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta, line 610: Correct - get_location returns a list of storage service locations, but we only ever want one currently processing location.

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta: If the path [ends with /](https://docs.python.org/2/library/os.path.html#os.path.basename), it returns ''. I want the basename if it's a file, or the last folder if it is a folder (and ends with /).

mistydemeo: How about making it more explicit?

``` python
last_segment = os.path.basename(source.rstrip('/')) + '/' if source.endswith('/') else os.path.basename(source)
```

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta: Done

src/dashboard/src/components/filesystem_ajax/views.py
Hwesta: Done

PR #106, opened 2014-09-05
Cleanup createStructuredDirectory

Hwesta:
Move createStructuredDirectory into archivematicaFunctions and rename create_structured_directory so it can take advantage of the directory constants and functions already there.  Add manual normalization constants to that file as well.  Update all callers, remove now deprecated archivematicaCreateStructuredDirectory.[py|sh], and do some minor cleanup.

Related to AIP Reingest.


src/MCPClient/lib/clientScripts/createSIPfromTransferObjects.py
```diff
@@ -47,11 +42,11 @@
 
     tmpSIPDir = os.path.join(processingDirectory, sipName) + "/"
     destSIPDir =  os.path.join(autoProcessSIPDirectory, sipName) + "/"
-    createStructuredDirectory(tmpSIPDir, createManualNormalizedDirectories=False)
+    archivematicaFunctions.create_structured_directory(tmpSIPDir, manual_normalization=False)
 
     #create row in SIPs table if one doesn't already exist
     lookup_path = destSIPDir.replace(sharedPath, '%sharedPath%')
-    sql = """SELECT sipUUID FROM SIPs WHERE currentPath = '""" + MySQLdb.escape_string(lookup_path) + "';"
+    sql = """SELECT sipUUID FROM SIPs WHERE currentPath = '""" + MySQLdb.escape_string(lookup_path) + """';"""
```
mistydemeo: Update these to use the SQL placeholders introduced in #110 instead of worrying about escaping?

mistydemeo:
:thumbsup:

Hwesta:
Updated SQL in both client scripts touched to use fixes from #110

mistydemeo:
Still :+1:

Hwesta:
Removed commit that added the SQL 'fixes' since they're unnecessary with #98

PR #107, opened 2014-09-05
MCPServer sets currentPath in watchedDirectories

Hwesta:
When starting a chain from a watched directory, ensure a SIP's current path matches the one found in the directory.  This can happen during AIP reingest, where the last know location is not the watched directory it is currently in. Do not change currentPath for DIPs, as they can run in parallel with the originating SIP and overwrite the SIP's information.

Also, related cleanup, mostly removing dead code.

Related to AIP Reingest.

### DIFF
--- a/src/MCPServer/lib/archivematicaMCP.py
+++ b/src/MCPServer/lib/archivematicaMCP.py
@@ -107,17 +107,23 @@ def fetchUUIDFromPath(path):
     if isUUID(path[uuidLen-1:-1]):
         return path[uuidLen-1:-1]
 
-def findOrCreateSipInDB(path, waitSleep=dbWaitSleep):
+def findOrCreateSipInDB(path, waitSleep=dbWaitSleep, unit_type='SIP'):
     """Matches a directory to a database sip by it's appended UUID, or path. If it doesn't find one, it will create one"""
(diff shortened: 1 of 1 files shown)


src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -107,22 +97,26 @@ def fetchUUIDFromPath(path):
...
+def findOrCreateSipInDB(path, waitSleep=dbWaitSleep, unit_type='SIP'):
     """Matches a directory to a database sip by it's appended UUID, or path. If it doesn't find one, it will create one"""
     path = path.replace(config.get('MCPServer', "sharedDirectory"), "%sharedPath%", 1)
 
-    #find UUID on end of SIP path
+    # Find UUID on end of SIP path
     UUID = fetchUUIDFromPath(path)
     if UUID:
-        sql = """SELECT sipUUID FROM SIPs WHERE sipUUID = '""" + UUID + "';"
+        sql = """SELECT sipUUID, currentPath FROM SIPs WHERE sipUUID = '""" + UUID + """';"""
```
mistydemeo: Replace this with the parameter support from #110, too?

mistydemeo:
:thumbsup:

Hwesta:
Updated for #110

mistydemeo:
Mind updating that last commit's message to clarify what part of the code it touches?

PR #108, opened 2014-09-05
Post API information to storage service

Hwesta:
If default setup requested, post the initial user and API key to the storage service when registering. If successful, add the SS hostname to the API whitelist.


mistydemeo:
Should we do this when the POST is happening unencrypted right now?

mistydemeo:
Are we planning this for 1.4.0?

PR #110, opened 2014-09-08
Allow MySQLdb to take parameter placeholders

mistydemeo:
These allow values to be passed into SQL statements without interpolating them into the string, preventing possible SQL injection issues. For example:

``` python
sql = """SELECT * FROM fpr_formatversion WHERE uuid='%s'""".format(uuid)
databaseInterface.queryAllSQL(sql)
```

Could be run as:

``` python
sql = """SELECT * FROM fpr_formatversion WHERE uuid=%s"""
databaseInterface.queryAllSQL(sql, (uuid,))
```

Since the values are inserted here without string interpolation, there is no risk of quoting issues or SQL injection regardless of the contents of the values.

### DIFF
--- a/src/MCPClient/lib/clientScripts/emailFailReport.py
+++ b/src/MCPClient/lib/clientScripts/emailFailReport.py
@@ -63,8 +63,8 @@ def sendEmail(subject, to, from_, content, server):
 
 def getUnitStatisticalDataHTML(unitIdentifier):
     fields = ["unitType", "Total time processing", "total file size", "number of files", "average file size KB", "average file size MB"]
-    sql = """SELECT `%s` FROM PDI_by_unit WHERE SIP_OR_TRANSFER_UUID = '%s';""" % ("`, `".join(fields), unitIdentifier)
-    rows = databaseInterface.queryAllSQL(sql)
+    sql = """SELECT `{fields}` FROM PDI_by_unit WHERE SIP_OR_TRANSFER_UUID = %s;""".format(fields="`, `".join(fields))
+    rows = databaseInterface.queryAllSQL(sql, (unitIdentifier,))
     return HTML.table(rows, header_row=fields)
 
 def getUnitJobLogHTML(unitIdentifier):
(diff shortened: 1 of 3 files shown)


Hwesta:
Why didn't we do this sooner? It was much easier than I thought it would be. :+1:

mistydemeo:
Updated the default value of `arguments` to be `None` instead of an empty tuple - with the empty tuple strings like '%sharedDirectoryStructure%'` were being mistaken for placeholders that couldn't be satisfied by the empty tuple.

mistydemeo:
Added a commit to convert `databaseFunctions` to use this instead of directly interpolating values.

Hwesta:
Semicolons at the end of SQL statements are inconsistently applied - do they need to be updated to all have ;'s? I noticed the old versions don't always have ;'s either.

Can we deleted `databaseInterface.separator` yet?  I only see it being used in assignFauxFileUUIDs.py (will be deleted soon) and databaseInterface.py.

How about `escapeForDB`?  I only see it's definition.

mistydemeo:
Semicolons are harmless but unnecessary; I left 'em in place where they already existed. They're used really inconsistently across the codebase.

`escapeForDB` can probably go, `databaseInterface.separator` should stay until it gets totally removed.

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -63,8 +63,8 @@ def sendEmail(subject, to, from_, content, server):
 
 def getUnitStatisticalDataHTML(unitIdentifier):
     fields = ["unitType", "Total time processing", "total file size", "number of files", "average file size KB", "average file size MB"]
-    sql = """SELECT `%s` FROM PDI_by_unit WHERE SIP_OR_TRANSFER_UUID = '%s';""" % ("`, `".join(fields), unitIdentifier)
-    rows = databaseInterface.queryAllSQL(sql)
+    sql = """SELECT `%s` FROM PDI_by_unit WHERE SIP_OR_TRANSFER_UUID = %s;""" % ("`, `".join(fields))
```
Hwesta: This fails because there are too many %s's for the % operator to fill.  Use string concatenation?

mistydemeo: Good catch, updated to use `{name}`-style `.format` formatting which ignores `%s`.

Hwesta:
:+1:

PR #111, opened 2014-09-08
Transfer: Hide transfer metadata link if none

Hwesta:
refs #7209

Previously, a TransferMetadataSet was created for every Transfer, even though it was only needed by Disk Image transfer types. It was updated to only create a TransferMetadataSet (and therefore transfermetadatasetrow) when needed, which broke rendering the link to the transfer metadata form. Update to only display the transfer metadata link if it exists.


mistydemeo:
:thumbsup:

PR #112, opened 2014-09-08
createMETSRights: namespace PREMIS elements

mistydemeo:
Fixes #7210.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSTrim.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSTrim.py
@@ -35,12 +35,12 @@ import databaseInterface
 
 def getTrimDmdSec(baseDirectoryPath, fileGroupIdentifier):
     #containerMetadata
-    ret = etree.Element("dmdSec") 
-    mdWrap = etree.SubElement(ret, "mdWrap")
+    ret = etree.Element(ns.metsBNS + "dmdSec") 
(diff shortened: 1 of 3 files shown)


Hwesta:
If the METS still validates, :+1:

mistydemeo:
Updated to also fix Trim and Dspace.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSTrim.py
```diff
@@ -93,7 +93,7 @@ def getTrimFileDmdSec(baseDirectoryPath, fileGroupIdentifier, fileUUID):
         return None
     for row in rows:
         xmlFilePath = row[0].replace('%SIPDirectory%', baseDirectoryPath, 1)
-        dublincore = etree.SubElement(xmlData, "dublincore", attrib=None, nsmap={None: ns.dctermsNS})
+        dublincore = etree.SubElement(xmlData, ns.dctermsBNS + "dublincore", attrib=None, nsmap={"ns": ns.dctermsNS})
```
Hwesta: Is `"ns"` supposed to be `"dc"` or  `"dcterms"`?

mistydemeo: `"dc"`, good catch. Fixed.

Hwesta:
:+1:

PR #113, opened 2014-09-09
Merge mysql files for dev and package builds`

jhsimpson:
This pull request merges the mysql and mysql_1.0.sql files into one, so that the debian packaging scripts and dev-helper both use the same mysql dump, rather than maintaining 2 dumps.


src/MCPClient/debian/copyright
```diff
@@ -11,10 +11,14 @@ Upstream Author(s):
     Austin Trask <austin@artefactual.com>
     Peter Van Garderen <peter@artefactual.com>
     Evelyn McLellan <evelyn@artefactual.com>
+    Mike Cantelon <mike@artefactual.com>
+    Justin Simpson <jsimpson@artefactual.com>
+    Holly Becker <hbecker@artefactual.com>
+    Misty DeMeo <mdemeo@artefactual.com>
```
Hwesta, line 17: Should be De Meo

Hwesta:
:+1:

PR #114, opened 2014-09-09
DBFunctions: Fix Event creation

Hwesta:
refs #4901

With passing parameters directly to the DB, do not use string 'NULL' instead of None/NULL.  Remove this fallback from insertIntoEvents, as getAgentForFileUUID return None by default. Actual agent values do not need to be double quoted, so just return the value instead of the string with quotes around it.


mistydemeo:
:thumbsup:

PR #115, opened 2014-09-10
Add validation microservice

Hwesta:
The validation event was removed when we stopped using FITS for everything (identification, characterization, validation, etc).  Validation event re-added using the FPR and a new FPR purpose 'validation'.

Add new client script after charaterize and extract metadata in transfer, and stock rules/command for Jhove.

Depends on artefactual/archivematica-fpr-admin#6


Hwesta:
New command and rules look good. Rebase to merge your two commits, @mistydemeo?

PR #116, opened 2014-09-10
Ingest: Normalization report links visible as links

Hwesta:
refs #7217

Use normal HTML to make links, instead of JS, so the links are visibly styled as links.  Remove duplicate links so that only the normalization attempted column is a link to reduce confusion.


mistydemeo:
:thumbsup:

PR #117, opened 2014-09-10
Maildir: Temporarily disable transfer type

Hwesta:
refs #7228

Remove maildir as a transfer type until it is better supported and doesn't always break.


mistydemeo:
:thumbsup:

PR #118, opened 2014-09-11
Correctly associate transfer metadata with transfer on creation

mistydemeo:
Fixes a pretty silly typo that prevented forensic disk image metadata from being associated with transfers.


Hwesta:
:+1:

PR #120, opened 2014-09-16
fix sql for Fido FPR command

jhsimpson:
As reported in issue  #7244

sql used to create FPR command for Fido on upgrades from 1.1.0
did not have spaces.


Hwesta:
:+1:  Has spaces now.

mistydemeo:
:shipit:

Hwesta:
:shipit:

PR #121, opened 2014-09-22
Config fixes for Ubuntu 14.04

Hwesta:
Apache now expects sites-[available|enabled] to end in .conf, so update dev creation to create 000-default.conf instead of [000-]default.

Gearman no longer creates a PID file that can be checked for. Instead, check the status directly by asking gearadmin for the PID.

refs #7293

### DIFF
--- a/src/MCPClient/init/archivematica-mcp-client.conf
+++ b/src/MCPClient/init/archivematica-mcp-client.conf
@@ -34,11 +34,14 @@ pre-start script
     # Check that $CONF directory exists
     [ -d $CONF ]
 
-    # Wait for Gearman service
-    while [[ "`gearadmin --getpid`" == '' ]]
+    # Wait for Gearman to start - timeout eventually
+    for i in $(seq 1 10)
     do
-        sleep 20
+        gearadmin --getpid && break
+        sleep 3
     done
+    gearadmin --getpid || exit 1
+    exit 0
 
 end script
 
--- a/src/MCPClient/init/archivematica-mcp-client.conf
+++ b/src/MCPClient/init/archivematica-mcp-client.conf
@@ -35,7 +35,7 @@ pre-start script
     [ -d $CONF ]
 
     # Wait for Gearman service
-    while [ ! -f /var/run/gearman/gearmand.pid ]
+    while [[ "`gearadmin --getpid`" == '' ]]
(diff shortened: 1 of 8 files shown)


src/MCPClient/init/archivematica-mcp-client.conf
```diff
@@ -34,11 +34,20 @@ pre-start script
     # Check that $CONF directory exists
     [ -d $CONF ]
 
-    # Wait for Gearman service
-    while [ ! -f /var/run/gearman/gearmand.pid ]
+    # Wait for Geraman to start - timeout eventually
```
mistydemeo: Typo!

src/MCPClient/init/archivematica-mcp-client.conf
```diff
@@ -34,11 +34,20 @@ pre-start script
...
+    for i in $(seq 1 10)
     do
-        sleep 20
+        if [ "`gearadmin --getpid`" != '' ]; then
```
mistydemeo: Check `$?` rather than length of string, maybe?

mistydemeo: Or structure as

``` sh
for i in $(seq 1 10)
do
    gearadmin --getpid && break
    sleep 3
done
```

src/MCPClient/init/archivematica-mcp-client.conf
```diff
@@ -34,11 +34,20 @@ pre-start script
...
+            break
+        else
+            sleep 3
+        fi
     done
+    if [ "`gearadmin --getpid`" != '' ]; then
```
mistydemeo: Similarly, can be shortened to:

``` sh
gearadmin --getpid || exit 1
```

src/MCPClient/init/archivematica-mcp-client.conf
Hwesta: Haha fixed.

src/MCPClient/init/archivematica-mcp-client.conf
Hwesta: That's a much better way of doing it. Updated for MCPClient and MCPServer.

mistydemeo:
:+1:

PR #122, opened 2014-09-22
MCPClient: remove unused rfc6266 dependency

mistydemeo:
This was once used in Archivematica's email code, but was removed in 6d86d6add9fe516e8a806eeaf4eef2cb389e9447 and never used anywhere else in the code.

I noticed this when setting up a 14.04 AM install. dev-helper complained about being unable to find the package, which we haven't packaged for 14.04. Since we're not using it, I'd rather delete it than rebuild it. ;)

We can probably remove the 12.04 package, too.


Hwesta:
:shipit:

PR #123, opened 2014-09-26
Transfer JS: Enable paths that don't start with /

Hwesta:
refs #7133

Duracloud 'paths' may not start with /. Update JS to not insert /'s at the start of every path.

### DIFF
--- a/src/dashboard/src/media/js/file-explorer.js
+++ b/src/dashboard/src/media/js/file-explorer.js
@@ -78,11 +78,12 @@
     // Provides a human-readable version of the path, suitable for display.
     // This is used by path() below.
     displaypath: function() {
-      parent = this.get('parent') || '';
-      parent = Base64.decode(parent);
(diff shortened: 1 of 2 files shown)


src/dashboard/src/media/js/file-explorer.js
```diff
@@ -78,11 +78,12 @@
     // Provides a human-readable version of the path, suitable for display.
     // This is used by path() below.
     displaypath: function() {
-      parent = this.get('parent') || '';
-      parent = Base64.decode(parent);
+      parent = this.get('parent');
       name = this.get('name') || '';
       name = Base64.decode(name);
-      return parent + '/' + name;
+      return (parent != undefined)
```
mistydemeo: Nitpick, but `!== undefined`? I don't _think_ this is likely to be coerced in weird ways, but...

mistydemeo:
:+1:

PR #124, opened 2014-10-07
Update to Django 1.7

Hwesta:
Update Archivematica to use Django 1.7.  Several changes made to update backwards incompatible changes, notably that ModelForms must specify their fields, and that standalone scripts (all the clientScripts) must setup Django before calling it.


mistydemeo:
:+1:

Hwesta:
In progress updating for #98 changes

src/MCPClient/lib/clientScripts/archivematicaTranscribeFile.py
```diff
@@ -8,18 +8,14 @@
 from tempfile import gettempdir
 from uuid import uuid4
 
-path = '/usr/share/archivematica/dashboard'
-if path not in sys.path:
-    sys.path.append(path)
+import django
+sys.path.append('/usr/share/archivematica/dashboard')
```
mistydemeo: I wonder if it's worth changing MCPClient to export `PYTHONPATH` and `DJANGO_SETTINGS_MODULE` when spinning up subprocesses? That way we could avoid requiring every client script to set its own `sys.path`.

mistydemeo: (Not that that needs to be in this PR, it just occurred to me.)

src/dashboard/src/components/rights/forms.py
```diff
@@ -81,12 +79,14 @@ class Meta:
 class RightsCopyrightNoteForm(ModelForm):
     class Meta:
         model = models.RightsStatementCopyrightNote
+        fields = ('copyrightnote',)  # TODO check this is correct
```
mistydemeo: Yep, `copyrightNote` is the only non-pk/non-fk field

src/dashboard/src/external/archivematica-fpr-admin
```diff
@@ -1 +1 @@
-Subproject commit c253d9e447344c3369c581025acaf016d8e114d4
+Subproject commit e87f9196d70eef1f5c1525c851859dd8bcd66568
```
mistydemeo: Was this intentional?

mistydemeo:
Still :+1:

src/MCPClient/lib/clientScripts/archivematicaTranscribeFile.py
Hwesta: That is a great idea, looking into it.  We use threads not subprocesses currently, so I dunno how different it'll be...

mistydemeo: We still run them as separate CLI scripts, so they're definitely being spun off as new processes at some level. IIRC our subprocess wrapper does support exporting the environment too.

src/dashboard/src/external/archivematica-fpr-admin
Hwesta: Yes - updates to match artefactual/archivematica-fpr-admin#8

Hwesta:
Includes #161

mistydemeo:
Still :+1:

src/MCPClient/lib/clientScripts/archivematicaTranscribeFile.py
Hwesta: This was fixed in #161 (after I forgot this conversation and did the work anyway)

Hwesta:
Added updates to postinst so the package upgrade should work, but needs to be tested.  Updated one of the databaseFunction tests, because updating to 1.7 broke it somehow. Responded to feedback.

mistydemeo:
:shipit:

PR #126, opened 2014-10-14
Transfer backlog deletion: remove all files from ES

mistydemeo:
When files are removed from transfer backlog after an AIP is stored, the current version of Archivematica will leave certain files dangling - any file that's not tracked in the Archivematica DB (such as logs or metadata) will be kept in the index even though they're gone from the disk.

This updates the function to avoid querying the database for files, and instead queries the database for transfers and asks ES for its list of files in those transfers.


Hwesta:
:+1: 

Oops, made comments on the commits instead of the PR again.

PR #127, opened 2014-10-21
characterizeFile: validate XML before inserting to DB

mistydemeo:
This ensures that if the characterization tool returns invalid XML, it will not be saved to the database in order to avoid later issues with createMETS2.

For instance, in some circumstances FITS may print error output to the stdout stream alongside actual XML content, resulting in a document that will not parse.

Fixes #7422.

### DIFF
--- a/src/MCPClient/lib/clientScripts/characterizeFile.py
+++ b/src/MCPClient/lib/clientScripts/characterizeFile.py
@@ -10,6 +10,8 @@ from __future__ import print_function
 import os
 import sys
 
+from lxml import etree
+
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 from executeOrRunSubProcess import executeOrRun
 from databaseFunctions import insertIntoFPCommandOutput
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/characterizeFile.py
```diff
@@ -75,8 +77,13 @@ def main(file_path, file_uuid, sip_uuid):
...
         if rule.command.output_format and rule.command.output_format.pronom_id == 'fmt/101':
-            insertIntoFPCommandOutput(file_uuid, stdout, rule.uuid)
-            print('Saved XML output for command "{}" ({})'.format(rule.command.description, rule.command.uuid))
+            try:
+                etree.fromstring(stdout)
+                insertIntoFPCommandOutput(file_uuid, stdout, rule.uuid)
+                print('Saved XML output for command "{}" ({})'.format(rule.command.description, rule.command.uuid))
+            except etree.XMLSyntaxError:
+                failed = True
+                print('XML output for command "{}" ({}) was not valid XML; not saving to database'.format(rule.command.description, rule.command.uuid), file=sys.stderr)
```
Hwesta, line 86: Print the XML output to stdout or stderr, like we do for non-XML output?

Hwesta:
:+1:

src/MCPClient/lib/clientScripts/characterizeFile.py
mistydemeo, line 86: Automatically printed by `executeOrRun` (it prints to stdout by default)

mistydemeo, line 86: I notice the line below duplicates it... which it probably shouldn't

PR #128, opened 2014-10-22
Index identifiers in ElasticSearch

mistydemeo:
This adds support for indexing document identifiers in ES, in addition to METS content. Document identifiers refer to identifiers used to reference the content in other systems. For example, if content originated in an Islandora system, the identifiers would contain the Islandora IDs of each item within the AIP.

A new `'identifiers'` key has been added to the ES document that is indexed, containing an array of zero or more strings. These can be returned via the "any" search in archival storage.

This adds support for retrieving these from one source, currently: identifiers will be extracted from MODS XML files located in the submissionDocumentation directory of each transfer contained in an AIP. This supports the Islandora workflow.

This also improves the existing indexing workflow by removing the need to extract the AIP's METS from the compressed AIP; the file was already contained at the root of the SIP and was being deleted unnecessarily. This leaves it in the SIP root and uses that copy. This should speed up indexing a bit. (Potentially a lot, for large solid archives.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,4 @@ dev (1.4.0)
 
 * Remove unused Elasticsearch backup code (#8076)
 * Improve performance of indexing AIP by saving uncompressed METS (#7424)
+* Index MODS identifiers in Elasticsearch (#7424)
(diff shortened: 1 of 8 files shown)


jhsimpson:
Looks good to me 
:+1:

Hwesta:
:+1:

mistydemeo:
Updated to move the MODS identifier helper into a new `identifier_functions` module (based on some IRL discussion we had on this).

PR #130, opened 2014-10-23
Hierarchical DIP support

mistydemeo:
This adds support for a new form of structMap, the logical structMap.

Directories created via SIP arrange can now be assigned a level of description label, in addition to a name. If directories with levels of description exist when the METS is created, then a logical structMap is created alongside the physical structMap. The structure of the logical structMap is identical, but the `type` value for directories is now the assigned level of description instead of `Directory`. The logical structMap can be used to assign levels of description to records on import to access systems like AtoM.

Major changes in this branch:
- AtoM upload configuration has been refactored, and now now form-based; the user no longer directly writes an argument string.
- If an AtoM system has been configured for upload, then the levels of description can be fetched from AtoM for use in SIP arrange.
- The hierarchy of levels of description can be configured in Archivematica. (For instance, in custom LOD hierarchies, to control what levels are higher than others.)
- Levels of description can be assigned to directories when they are created in SIP arrange.
- Logical structMap creation in createMETS2.


mistydemeo:
This history needs to be cleaned up before merging.

src/MCPServer/share/mysql_dev_hierarchical_dip_upload.sql
```diff
@@ -0,0 +1,11 @@
+-- Set AtoM DIP upload defaults
+INSERT INTO DashboardSettings (name, value) VALUES ('dip_upload_atom_url', 'http://localhost/ica-atom');
```
Hwesta: This should probably be 'http://localhost/atom' instead, as well as below.

mistydemeo: Updated.

Hwesta:
This depends on/works with #145

Hwesta:
This has been rebased on top of qa/1.x as dev/issue-6988-hierarchical-dip  Should we reset dev/issue-6988-atom-dip-config to be the same as dev/issue-6988-hierarchical-dip to update this PR, or open a new one?

mistydemeo:
Rebased, and squashed many of the fix commits into their original feature commits.

mistydemeo:
Tested and confirmed to be working again after the rebase.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -827,7 +829,59 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
             file_elem.set("DMDID", dspaceMetsDMDID)
     
     return structMapDiv
-        
+
+def build_arranged_structmap(original_structmap, fileGroupIdentifier):
```
Hwesta: Docstring?  Especially coming to this after a year I'm not sure what this function is supposed to do.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -827,7 +829,59 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
...
+    tag_dict = dict(SIPArrange.objects.filter(sip_id=fileGroupIdentifier).values_list('arrange_path', 'level_of_description'))
+    if len(tag_dict) < 1:
```
Hwesta: `if not tag_dict`?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
mistydemeo: Added.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -827,7 +829,66 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
...
     return structMapDiv
-        
+
+def build_arranged_structmap(original_structmap, fileGroupIdentifier):
+    """
+    Given a structMap, builds a new copy of the structMap with file and directory labels assigned according to their intellectual arrangement.
+    Logical arrangement is determined using the levels of description which were assigned to them during SIP arrange.
+
+    :param etree.Element original_structmap: the structMap on which the arranged structMap should be based.
+    :param str fileGroupIdentifier: the SIP's UUID
```
Hwesta: Especially since this is new code, maybe we can start referring to sip_uuid instead of the ever-confusing fileGroupIdentifier?

src/dashboard/src/components/helpers.py
```diff
@@ -204,6 +207,35 @@ def get_server_config_value(field):
...
+        raise Exception("AtoM URL not defined!")
+
+    auth = (
+        get_setting('dip_upload_atom_email'),
+        get_setting('dip_upload_atom_password'),
+    )
+    if not auth:
+        raise Exception("AtoM authentication settings not defined!")
+
+    dest = urljoin(url, 'api/taxonomies/34')
```
Hwesta, line 234: What's the significance of `api/taxonomies/34`?  Can we comment to that effect?

src/dashboard/src/components/helpers.py
```diff
@@ -204,6 +207,35 @@ def get_server_config_value(field):
     except:
         return ''
 
+def get_atom_levels_of_description(clear=True):
+    url = get_setting('dip_upload_atom_url')
+    if not url:
+        raise Exception("AtoM URL not defined!")
```
Hwesta: Would be nice if this raised a more specific exception, so we don't silence other things catching it.  Similarly, do we want `fetch_levels_of_description_from_atom` to catch all Exceptions, or just the AtoM-related ones we're raising here?

src/dashboard/src/components/helpers.py
mistydemeo, line 234: I believe the taxonomy at `34` is always the level of description in an AtoM app.

src/dashboard/src/main/models.py
```diff
@@ -909,3 +912,12 @@ class FileID(models.Model):
 
     class Meta:
         db_table = 'FilesIDs'
+
+class LevelOfDescription(models.Model):
+    id = UUIDPkField()
```
Hwesta, line 917: Does this have to be a UUID? Could we just use Django's built in synthetic key?

mistydemeo, line 917: I'm fine with that. I went with UUIDPKField only for consistency.

src/dashboard/src/components/helpers.py
mistydemeo: Do we want an `AtomException`? I guess there's no harm in adding a very specific exception.

Hwesta:
Do we still need 8d191e7903be5d7b2c85b7cd4b8e0fa5d65542d1 now that the LoD display is handled the same way as other metadata (more in Django)?

src/dashboard/src/components/helpers.py
```diff
@@ -204,6 +207,35 @@ def get_server_config_value(field):
     except:
         return ''
 
+def get_atom_levels_of_description(clear=True):
```
Hwesta, line 213: Docstring?

Hwesta:
Should we add tests for the new functionality now?

mistydemeo:
> Should we add tests for the new functionality now?

Probably a good idea.

src/dashboard/src/components/helpers.py
mistydemeo, line 213: Added!

src/dashboard/src/components/helpers.py
mistydemeo: Added an `AtomError` and converted all exceptions to use it.

src/dashboard/src/components/helpers.py
mistydemeo, line 234: Added a comment.

mistydemeo:
I will add tests, but it's going to take a bit for me to spin up an AtoM vagrant box from which I can record responses.

src/dashboard/src/components/administration/views.py
```diff
@@ -88,14 +89,40 @@ def failure_report_detail(request):
...
+        opts = []
+        char_fields = ['dip_upload_atom_url', 'dip_upload_atom_email',
+                       'dip_upload_atom_password', 'dip_upload_atom_rsync_target',
+                       'dip_upload_atom_rsync_command', 'dip_upload_atom_version']
+        for field in char_fields:
+            value = form.cleaned_data.get(field)
+            if not value:
+                continue
+            optname = field.replace('dip_upload_atom_', '').replace('_', '-')
+            opts.append('--{}={}'.format(optname, value))
```
Hwesta: Put quotes around the value passed as a parameter?  `'--{}="{}"'.format(optname, value)`

PR #131, opened 2014-10-23
characterizeFile: don't double-print stdout

mistydemeo:
stdout is already piped to the terminal as well as captured; the duplicate print here would result in duplicated output.


Hwesta:
Leave a print statement about how the output was not XML and not saved to DB (or METS)?

:+1:

mistydemeo:
Good thought!

mistydemeo:
Updated.

Hwesta:
Still :+1:

PR #133, opened 2014-10-23
Dev/issue xxxx 130 packaging

jhsimpson:
Updating files in the debian directories


src/MCPClient/debian/changelog
```diff
@@ -0,0 +1,11 @@
+archivematica-mcp-client (1.3.0.rc.4-1~trusty1) trusty; urgency=high
```
mistydemeo, line 1: It's not a big deal, but I've been using numeric versions for these - the alphabet will eventually wrap around but, hopefully, the year won't. ;)

mistydemeo:
Looks good, aside from the comment I left!

jhsimpson:
ok, let's fix the numeric version (using ~12.04 instead of ~trusty ) when the final changelog entry is made in the stable branch.

PR #134, opened 2014-10-29
Dashboard: Accession IDs associate with Transfers again

Hwesta:
Use os.path.join to add / to the end of SHARED_DIRECTORY_ROOT if needed before replacing it with %sharedPath%.

SHARED_DIRECTORY_ROOT was updated to read the shared directory path from a config file instead of being hardcoded in an earlier commit. The previous hardcoded value was not / terminated, but the config file path is / terminated.

When creating a new Transfer, a replace is done on the path to convert it from absolute to relative to %sharedPath%. The replace always added / to the end of SHARED_DIRECTORY_ROOT, which caused the replace to fail when SHARED_DIRECTORY_ROOT was already / terminated.

This caused the MCPServer to create a duplicate Transfer (without an accession ID) when it found the files on disk, as it could not find an associated Transfer by path in the DB.

refs #7442


mistydemeo:
:+1:

PR #135, opened 2014-10-31
Remove vendored pyes tests

mistydemeo:
We're not set up for these tests to run and succeed, and leaving them in this location causes py.test to find and run them.


Hwesta:
I like getting rid of the broken py.test things.  However, the version of pyes we're vendoring is still available on PyPI. Could we add pyes==0.19.0 to the archivematicaCommon requirements file and removed the entire vendored version instead?

mistydemeo:
Updated.

Looks like both archivematicaCommon and the dashboard install their pip deps (globally), so this should work okay.

localDevSetup/eraseElasticSearchIndexes.py
```diff
@@ -4,7 +4,7 @@
 import sys
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 import elasticSearchFunctions
-sys.path.append("/usr/lib/archivematica/archivematicaCommon/externals")
+
 import pyes
```
Hwesta, line 8: Move this above the `sys.path.append`s so we know it's installed globally and not at that path?  AM has an informal convention (or is trying to!) of only importing the things in the appended path immediately after we modify the sys.path.

src/dashboard/src/components/ingest/views.py
```diff
@@ -37,6 +37,7 @@
 from django.shortcuts import render, redirect
 from django.template import RequestContext
 from django.utils.text import slugify
+import pyes
 
 # External dependencies, alphabetical
```
Hwesta, line 42: Move import pyes to the External dependencies category?

src/archivematicaCommon/lib/utilities/restart-elasticsearch-if-down
```diff
@@ -7,7 +7,7 @@ import shutil
 
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 import elasticSearchFunctions
-sys.path.append("/usr/lib/archivematica/archivematicaCommon/externals")
+
```
Hwesta, line 10: Didn't this get moved to artefactual/archivematica-devtools?  I thought we'd deleted it already.

mistydemeo:
Closing in favour of #138.

PR #136, opened 2014-11-04
Insert "move to processing directory" before sanitizing transfer name

mistydemeo:
This fixes issue 7464, where name sanitization inside a watched directory could in some circumstances result in the transfer passing through the watched directory's microservice chain links twice. This moves the transfer to the processing directory before any changes to path occur.


Hwesta:
:+1:

PR #138, opened 2014-11-08
Migrate to Elasticsearch 1.x

mistydemeo:
This migrates Archivematica to Elasticsearch 1.x (tested with 1.3.4), and the official Elasticsearch Python library instead of pyes.

Known issues:
- [x] Number of files in an AIP does not display correctly in the advanced search view


src/dashboard/src/templates/archival_storage/archival_storage_search.html
```diff
@@ -71,7 +71,7 @@
       {% for term_usage in results %}
         <tr>
           <td>
-            <a href="{% url 'components.archival_storage.views.aip_download' term_usage.term %}">{{ term_usage.name }}</a>
+            <a href="{% url 'components.archival_storage.views.aip_download' term_usage.uuid %}">{{ term_usage.name }}</a>
```
mistydemeo, line 74: I decided to have the template use the same keys we use in our ES index, so the results don't need to be rewritten in the view.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -134,6 +141,7 @@ def check_server_status_and_create_indexes_if_needed():
     # no error!
     return 'OK'
 
+# TODO do we still need this?
```
Hwesta: I originally added this so that MAX_QUERY_SIZE would be added to all queries, since it was often forgotten and defaults to a low number (10, I think) which caused problems.  I think this is still relevant.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -142,29 +150,38 @@ def search_raw_wrapper(conn, query, indices=None, doc_types=None, **query_params
...
-        conn._send_request('GET', '/')
-    except:
+        conn.info()
+    except ConnectionError, TransportError:
         return 'Connection error'
 
     # no errors!
     return 'OK'
 
+# TODO look at third argument
```
Hwesta: Is this still needed?

src/dashboard/src/components/advanced_search.py
```diff
@@ -143,16 +166,16 @@ def query_clause(index, queries, ops, fields, types):
...
             else:
                 term_field = '_all'
-            return pyes.TermQuery(term_field, queries[index])
+            return {'term': {term_field, queries[index]}}
     else:
-        return pyes.StringQuery(queries[index], search_fields=search_fields)
+        return {'query_string': {'query': queries[index], 'fields': search_fields}}
 
 def indexed_count(index, types=None, query=None):
     aip_indexed_file_count = 0
```
Hwesta: Unused - delete.

src/dashboard/src/components/advanced_search.py
```diff
@@ -143,16 +166,16 @@ def query_clause(index, queries, ops, fields, types):
...
     try:
-        conn = pyes.ES(elasticSearchFunctions.getElasticsearchServerHostAndPort())
-        count_data = conn.count(indices=index, doc_types=types, query=query)
-        aip_indexed_file_count = count_data.count
+        conn = Elasticsearch(hosts=elasticSearchFunctions.getElasticsearchServerHostAndPort())
+        return conn.count(index=index, doc_type=types, body=query)['count']
     except:
-        pass
-    return aip_indexed_file_count
+        return
```
Hwesta: Return 0 to maintain compatibility?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -119,22 +118,23 @@ def search(request):
...
+            results = conn.search(
+                body=query,
+                index='aips',
+                doc_type='aipfile',
+                from_=start - 1,
                 size=items_per_page,
                 fields='AIPUUID,filePath,FILEUUID'
             )
     except:
+        logging.exception('Error accessing index.')
```
Hwesta: Use logger.exception, since we fetch one in this file anyway.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -163,40 +163,53 @@ def search(request):
...
-            aip_uuid.name = aip_info['fields']['name']
-            aip_uuid.size = '{0:.2f} MB'.format(aip_info['fields']['size'])
-            aip_uuid.date = aip_info['fields']['created']
-            aip_uuid.isPartOf = aip_info['fields'].get('isPartOf', '')
-            aip_uuid.AICID = aip_info['fields'].get('AICID', '')
-            aip_uuid.aips_in_aic = aip_info['fields'].get('countAIPsinAIC', '(unknown)')
+            aip = elasticSearchFunctions.normalize_results_dict(aip_info)
+            aip['uuid'] = aip_uuid['term']
+            aip['count'] = aip_uuid['count']
+            if not 'isPartOf' in aip:
```
Hwesta: `if 'isPartOf' not in aip:` is more idiomatic.  Also, add comment explaining this is adding values for things that weren't in the fetched values? I was confused what this was doing initially.

mistydemeo: Updated and commented. This section surprised me in the original, too!

src/dashboard/src/components/advanced_search.py
mistydemeo: Fixed.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -419,34 +449,53 @@ def list_display(request):
...
-        indices=['aips'],
+    query = {
+        "query": {
+            "match_all": {},
+        },
+        "from": start,
+        "size": items_per_page
+    }
+    results = conn.search(
+        body = query,
```
Hwesta: No spaces around parameters.

src/dashboard/src/components/ingest/views.py
```diff
@@ -450,6 +450,7 @@ def transfer_backlog(request):
             doc_types='transferfile',
         )
     except:
+        logging.exception('Error accessing index.')
```
Hwesta: Log to logger, not logging.  (Too high a percentage of logs in that sentence!)

mistydemeo: Already fixed, hit "refresh"!

src/dashboard/src/components/archival_storage/views.py
Hwesta: Also for AICID, countAIPsinAIC.

src/dashboard/src/components/ingest/views.py
Hwesta: You're too fast for me!

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: Okay, removed the comment. Since pyes has both a "raw" search and a more pythonic version, I wasn't sure if this was meant to be some kind of alternate wrapper to the raw search. Makes sense in that context.

Hwesta:
Can you add some docstrings (especially relating to the expected types of parameters and returns)?  Also, tests, especially for the elasticSearchFunctions?  (Or are we expecting to totally replace these so it's not worthwhile?

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: The third argument is (probably) not really needed in this version, but I always feel bad about changing function signatures in an incompatible way. That said, it's only used within this one file...

Hwesta: Add a _ in the front to mark as internal?

src/archivematicaCommon/lib/elasticSearchFunctions.py
Hwesta: Should other calls be updated to use this?  Maybe add my explanation to a docstring?

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: Right now I'm just trying to match the original internal API, so I'm avoiding renaming functions where necessary. Can look at a bigger refactor once we figure out what we're doing with search. :b

Meanwhile, reproduced the original behaviour, so I removed the comment.

mistydemeo:
>  (Or are we expecting to totally replace these so it's not worthwhile?

Possibly! I don't want to make bigger changes like that until we decide what we're doing with this longerterm.

mistydemeo:
Added a new commit which causes SIP arrange to both log and raise an exception if no file info is returned from ES. The old behaviour was to assign None as the file and transfer UUID, but that results in a SIP that will inevitably fail almost instantly upon being started.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -477,10 +477,16 @@ def _get_arrange_directory_tree(backlog_uuid, original_path, arrange_path):
...
-            file_uuid = file_info.get('fileuuid')
-            transfer_uuid = file_info.get('sipuuid')
+                'relative_path', relative_path)
+            try:
+                file_uuid = file_info['fileuuid']
+                transfer_uuid = file_info['sipuuid']
+            except KeyError:
+                message = 'No file information returned from Elasticsearch for file at relative_path: ' + relative_path
+                logging.warning(message)
+                raise Exception(message)
```
Hwesta: Is there a more specific Exception we can raise here?

mistydemeo: I don't mind defining a new one - say, `EmptySearchResultError` or such

Hwesta: Is there an ElasticSearchError or similar in the ES library that already exists that we can raise?

mistydemeo: Aha, yep, we do have `ElasticsearchError` in our own `elasticSearchFunctions` - could use or subclass that. (Subclassing might make sense, to let us catch empty search results independently of explicit "error" conditions?)

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -191,15 +206,14 @@ def aip_mapping_is_correct(conn):
...
-            # but didn't work with ES 0.19.0
-            pass
+        conn = Elasticsearch(hosts=getElasticsearchServerHostAndPort())
+
+        response = conn.indices.create(index, ignore=400)
+        if 'error' in response and 'IndexAlreadyExistsException' in response['error']:
+            return conn
+
+        set_up_mapping(conn, index)
+        conn = connect_and_create_index(index, attempt + 1)
```
Hwesta, line 219: This would make more sense iteratively than recursively, but I understand if you don't want to fix it.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -707,15 +750,15 @@ def get_transfer_file_info(field, value):
...
-    documents = search_raw_wrapper(
-        conn,
-        pyes.FieldQuery(pyes.FieldParameter(field, searchvalue)),
-        indicies=indicies,
-    )
+    indicies = 'transfers'
+    query = {
+        "query": {
+            "term": {
+                field: value
```
Hwesta, line 760: Do we still need to escape /'s?

mistydemeo, line 760: Not anymore! 🎉

Hwesta, line 760: Great!

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo, line 219: Yeah, I'd love to make improvements to this, but I don't want to spend too much time on it when this code may be replaced in the medium-to-long term.

Hwesta:
:+1:

PR #140, opened 2014-11-18
sanitizeNames: try to decode string before transliterating

mistydemeo:
refs #7478


src/MCPClient/lib/clientScripts/sanitizeNames.py
```diff
@@ -32,7 +32,12 @@
 replacementChar = "_"
 
 def transliterate(basename):
-    return unidecode(basename) 
+    # We get a more meaningful name sanitization if UTF-8 names
+    # are correctly decoded to unistrings instead of str
+    try:
+        unidecode(basename.decode('utf-8'))
+    except UnicodeDecodeError:
+        unidecode(basename)
```
Hwesta: These needs to be returned.

Hwesta:
:+1:

PR #142, opened 2014-11-20
ATK: Add quotes around parameters to handle whitespace

Hwesta:
This should also get cherry-picked to the 1.3.1 release.  ATK requires extra setup/external system, so have not tested yet.

refs #7606


mistydemeo:
👍 Needs a rebase though.

Hwesta:
Also cherry-picked to qa/1.3.1

PR #143, opened 2014-12-08
MCPClient control: specify libavcodec-extra-56 dependency

mistydemeo:
libvo_aacenc requires ffmpeg to be built as GPL3, and won't be available in the default libavcodec build. Since faac and libfdk_aac are both nonfree, this means that ffmpeg isn't installed with any AAC encoder other than the experimental internal one by default.


Hwesta:
:+1:

PR #144, opened 2014-12-12
Dashboard: replace Base64 with atob/btoa

mistydemeo:
This replaces the vendored Base64 code we're using with thin wrappers on top of the builtin `atob` and `btoa` functions that are in Firefox and Chrome, as well as IE 10 and newer. The wrappers are capable of safely roundtripping both Unicode strings and mojibake.

This resolves the issues with the Base64 code we were using misdecoding certain regions of Unicode characters.

### DIFF
--- a/src/dashboard/src/media/js/vendor/base64.js
+++ b/src/dashboard/src/media/js/vendor/base64.js
@@ -1,216 +1,22 @@
-/*
-Copyright (c) 2008 Fred Palmer fred.palmer_at_gmail.com
-
-Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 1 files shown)


mistydemeo:
Hm, I suppose I could reduce the size of the diff by simply naming the new functions `Base64.encode` and `Base64.decode`, now that I think about it.

Hwesta:
:+1:  either way.

Keeping Base64.[en|de]code is a smaller diff, base64_to_utf8 is more clear.

mistydemeo:
Updated to minimize the diff.

Hwesta:
Still :+1:

PR #145, opened 2014-12-16
SIP arrange: fix fetching results from relative_path

mistydemeo:
The `relative_path` field in the `transferfile` document type is indexed with analysis to allow for free-text searching; however, this can result in exact string matching failing. This PR adds a second, raw copy of that field, and uses that where an exact match is required.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -244,7 +244,10 @@ def set_up_mapping(conn, index):
     if index == 'transfers':
         mapping = {
             'filename'     : {'type': 'string'},
-            'relative_path': {'type': 'string'},
+            # Analyzed copy allows the stirng to be found via free-text search,
+            # but a raw unanalyzed copy is required for exact string matches.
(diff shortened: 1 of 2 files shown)


mistydemeo:
Since this changes the index, you'll have to erase the ES index to test this.

Hwesta:
:+1: 

Kind of related - add raising an EmptySearchResultError to https://github.com/artefactual/archivematica/blob/qa/1.x/src/dashboard/src/components/filesystem_ajax/views.py#L568 as well?

mistydemeo:
> Kind of related - add raising an EmptySearchResultError to https://github.com/artefactual/archivematica/blob/qa/1.x/src/dashboard/src/components/filesystem_ajax/views.py#L568 as well?

Sounds like a plan!

Since this makes a change to the index structure, I'll probably not merge this if we go ahead with the SS DB changes in the near future.

Hwesta:
Need to update to `relative_path.raw` on line 569

mistydemeo:
Incorporated the second fix, too.

mistydemeo:
Merged an alternate fix.

PR #146, opened 2014-12-17
FITS: remove validation events

mistydemeo:
FITS no longer produces JHOVE validation metadata, since this was moved to the new validation microservice. The lack of JHOVE output caused this script to fail.

Fixes #7719.


Hwesta:
archivematicaFITS is only used in to-be-revamped maildir, removed-in-dev-branch DIP from AIP and manual normalization, correct?  Is there an easy way to replace this with the FPR-based scripts?  I suppose choosing the file identification script is the problem again, especially since this is done before the identification for submission docs happens.

:+1:

PR #147, opened 2014-12-17
metadata.csv improvements

Hwesta:
metadata.csv can accept both file and directory metadata in the same file.  Also, clean up parsing code and createMETS2 to handle having multiple metadata.csvs better (as a result of SIP arrangement or adding them during ingest).


mistydemeo:
:+1:

PR #148, opened 2014-12-18
CONTENTdm: metadata-only project client DIP upload

Hwesta:
Changes to how CONTENTdm DIP upload works, specifically for project client in a metadata-only DIP 'upload'.

simple.txt and compound.txt are now generated soley from the METS file, instead of querying the CONTENTdm server for what fields are available for this collection.  The column names are based on the XML element names of associated dmdSecs, which are generated from the incoming metadata.csv

Both simple.txt and compound.txt have the AIP UUID and file UUID added as columns.  Bulk compound.txts (only directories) have Directory name as the first column, while the rest have Filename as the last column.

Converted restructureDIPForContentDMUPload to use lxml instead of xml.dom.minidom, as lxml is more powerful for searching.  All the project client code uses lxml, but not all the direct upload code uses it, so direct upload doesn't work currently.

This relies on and currently includes changes in #147 

refs #7239


src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -273,50 +270,22 @@ def getContentdmCollectionFieldInfo(contentdmServer, targetCollection):
...
-                # Assumes there is only one dublincore child element.
-                return node.getElementsByTagName('dublincore')[0]
-            else:
-                return node
+def getObjectDirectoryFiles(objectDir):
+    """
+    Get a list of all the files (recursive) in the provided directory.
 
+    "Even though there can be subdirectories in the objects directory,
+    assumes each file should have a unique name." What does this mean?
```
mistydemeo: I wonder if this means that filenames are (or should be) guaranteed unique across the SIP, regardless of the directory in which they're located? e.g. two subdirectories can't have objects with the same names?

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -490,139 +459,127 @@ def generateFullFileEntry(title, filename, extension):
...
-    if lenDmdSecs == 1:
-        mdWrap = dmdSecs[0].getElementsByTagName('mdWrap')[0]
-        if mdWrap.attributes['MDTYPE'].value == 'OTHER':
-            dmdSecPair['nonDc'] = parseDmdSec(dmdSecs[0])
-            dmdSecPair['dc'] = None
-        if mdWrap.attributes['MDTYPE'].value == 'DC':
-            dmdSecPair['dc'] = parseDmdSec(dmdSecs[0])
-            dmdSecPair['nonDc'] = None
+    dmdSecPair = {'dc': None, 'nonDc': None}
+    if len(dmdSecs) > 2:
```
mistydemeo: Use `lenDmdSecs` assigned above?

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataCSV.py
```diff
@@ -24,84 +24,103 @@
...
+    for each column.
+
+    Example CSV:
+    Filename,dc.title,dc.date,Other metadata
+    objects/foo.jpg,Foo,2000,Taken on a sunny day
+    objects/bar/,Bar,2000,All taken on a rainy day
+
+    Produces:
+    {
+        'objects/foo.jpg': OrderedDict(dc.title=Foo, dc.date=2000, Other metadata=Taken on a sunny day)
```
mistydemeo: Nitpick, but include comma at the end?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -869,7 +860,7 @@ def create_object_metadata(struct_map):
         import time
         time.sleep(10)
 
-    parseMetadata(baseDirectoryPath)
+    CSV_METADATA = parseMetadata(baseDirectoryPath)
```
mistydemeo, line 887: 👍👍👍👍👍 🎉

mistydemeo:
Looks good, aside from the comments I left!

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta: Maybe, but the paths returned in `fileList` are the full path, so that seems a weird qualifier.  Maybe it's a statement that the parts returned _are_ the full path, and are therefore unique?  It was in the original comments.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataCSV.py
Hwesta: Ooops, went in with #147

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta: Done

Hwesta:
Updated to current qa/1.x, fixed based on comments.

mistydemeo:
:+1:

Hwesta:
Holding off merging this until we can decide what to do with that currently-horribly-broken direct upload code path.

Hwesta:
Direct upload removed (MSCL, client scripts, admin config).

mistydemeo:
> instead of querying the CONTENTdm server for what fields are available for this collection

Out of curiosity, does this break any existing functionality?

mjordan:
The query to see what fields are available for a collection is comprehensive, i.e., it includes all fields that are configured for a collection. If I follow the provenance of the dmdSec values as described in the initial comment, one issue will be that if the incoming metadata.csv file doesn't contain columns corresponding to all fields for a collection (which it probably should not be trusted to do), the dmdSecs will not contain the missing fields, and consequently those fields will not get generated out into simple.txt and compound.txt.

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -22,614 +22,118 @@
...
+    mdType = dmdSec.xpath('mets:mdWrap/@MDTYPE', namespaces=ns.NSMAP)
+    if mdType == 'DC':
+        dcTitlesDom = dmdSec.findall('.//dcterms:title', namespaces=ns.NSMAP)
         if not dcTitlesDom:
-            return {'title' : '[Placeholder title]'} 
+            elementsDict['title'] = label
+
+    # Iterate over all descendants and put in the return dict
+    # Key is the element's tag name, value is a list of the element's text
+    # TODO check for problems with encoding & unicode vs str
```
mistydemeo: :scream: Here be dragons, I think...

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -22,614 +22,118 @@
...
-            
-    return filesInObjectDirectoryForThisDmdSecGroup
-
-
-# Add the AIP UUID to the DC metadata. We handle non-DC metadata within 
+# Add the AIP UUID to the DC metadata. We handle non-DC metadata within
 # each generateXXXProjectClientPackage function. Direct upload packages
 # have their DC metadata supplemented with the AIP UUID in generateDescFile().
 def addAipUuidToDcMetadata(dipUuid, dcMetadata):
+    if not dcMetadata:
```
mistydemeo, line 134: What's the intent of this clause?

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -640,569 +144,115 @@ def addAipUuidToDcMetadata(dipUuid, dcMetadata):
...
-                    # For each object file, output the object file. We need to find the full path
-                    # of the file identified in v['filename'].
-                    if (v['filename'] in fullPath):
-                        shutil.copy(fullPath, os.path.join(outputItemDir, accessFileBasenameName + accessFileBasenameExt))
+    :param output_dir: Path to directory for simple/compound.txt
+    :param structmap: structMap element from the METS (Preparse somehow?)
+    :param dmdsecs: Dict of {<DMDID>: OrderedDict{column name: value} or <dmdSec element>? }
+    :param dipuuid: UUID of the DIP
+    """
+    print 'dipuuid', dipuuid
```
mistydemeo: Just to check, is this logging meant to be here?

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
```diff
@@ -640,569 +144,115 @@ def addAipUuidToDcMetadata(dipUuid, dcMetadata):
...
+            if 'dirs' in package_type:
+                csv_header.insert(0, 'Directory name')
+                csv_values.insert(0, name)
+            else:
+                csv_header.append('Filename')
+                csv_values.append(name)
+
+            # Compare csv_header, if diff ERROR (first time set, write to file)
+            if csv_header_ref and csv_header_ref != csv_header:
+                print >> sys.stderr, 'ERROR headers differ,', csv_path, 'almost certainly invalid'
```
mistydemeo: Do we want to actually stop writing, or skip the row?

mistydemeo:
:+1: aise from comments

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta, line 134: If there's no DC metadata provided (None), trying to add an identifier to it will fail. This returns the falsey value as is.  Would it be clearer if it returned None?

mistydemeo, line 134: I think so, yeah. I was trying to figure out if there was some semantic meaning behind the specific falsey value that was passed in.

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta: Checked for problems and miraculously didn't find any. If the text is unicode it writes out just fine.

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta: I'd planned on leaving it so I could debug things if necessary. In general, I want more output in the client scripts. I could convert it to using the logging module if you think that's better.

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
Hwesta: Discussed with Sarah, and looks like it should fail.  This case shouldn't happen, anyway.

src/MCPClient/lib/clientScripts/restructureDIPForContentDMUpload.py
mistydemeo: Yeah, I'm glad to have logging - I'd just rather it be easy to understand. If the output were a bit clearer/more verbose I think this'd work for me :+1:

Hwesta:
> > instead of querying the CONTENTdm server for what fields are available for this collection
> 
> Out of curiosity, does this break any existing functionality?

Yes, but by design. Previously, this queried the CDM server to get the list of fields and performed a matching between the columns in metadata.csv to the fields in CDM.  This has been changed to read from the dmdSecs (preferring OTHER over DC), and trusting that the metadata.csv matches the columns. CDM has a matching interface to fix things up as needed, and we keep the order of the columns the same as the metadata.csv.

mistydemeo:
:shipit:

PR #149, opened 2014-12-19
Modifies storeAIP.py to calculate the size of uncompressed packages.

andrewjbtw:
Checks if package is a file or a directory. If it's a directory, the size is calculated by walking through the sub-directories. No change is made to how the size is calculated for compressed AIPs. This code is based on the code in moveToBacklog.py for calculating the size of transfers sent to the backlog.

This should fix issue #7175.

### DIFF
--- a/src/MCPClient/lib/clientScripts/storeAIP.py
+++ b/src/MCPClient/lib/clientScripts/storeAIP.py
@@ -83,6 +83,16 @@ def store_aip(aip_destination_uri, aip_path, sip_uuid, sip_name, sip_type):
     else:
         uuid = sip_uuid
 
+    # If AIP is a directory, calculate size recursively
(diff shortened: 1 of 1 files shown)


jhsimpson:
Looks great, thanks for contributing this.  We created a qa/1.3.1 branch, which is about to get merged into stable/1.3.x.  I'd like to merge this change into qa/1.3.1, rather than stable/1.3.x directly.

andrewjbtw:
No problem, that makes sense. Do I need to close this and then open a new pull request on qa/1.3.1 or is there a way to redirect this request?

mistydemeo:
The patch should apply cleanly either way, so no worries!

andrewjbtw:
Ok, new request opened on qa/1.3.1, so I'm closing this one.

PR #151, opened 2015-01-06
Ingest file browser: improve empty path check

mistydemeo:
Fixes an issue where hitting "cancel" at the directory creation prompt would create a directory named "null".

`!!path` will filter out both null paths (from hitting "cancel") and empty strings.

Fixes #7766.

### DIFF
--- a/src/dashboard/src/media/js/ingest/ingest_file_browser.js
+++ b/src/dashboard/src/media/js/ingest/ingest_file_browser.js
@@ -192,7 +192,7 @@ $(document).ready(function() {
   $('#arrange_create_directory_button').click(function() {
     var path = prompt('Name of new directory?');
 
-    if (path != '') {
+    if (path) {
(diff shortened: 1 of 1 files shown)


src/dashboard/src/media/js/ingest/ingest_file_browser.js
```diff
@@ -192,7 +192,7 @@ $(document).ready(function() {
   $('#arrange_create_directory_button').click(function() {
     var path = prompt('Name of new directory?');
 
-    if (path != '') {
+    if (!!path) {
```
Hwesta: Can we just do `if (path)`?

mistydemeo: Yeah, good point. I'll fix that.

mistydemeo:
Fixed!

Hwesta:
:+1:

PR #152, opened 2015-01-07
Install: put sudoers in sudoers.d, check validity

Hwesta:
Instead of modifying /etc/sudoers directly, add an archivematica entry to /etc/sudoers.d/  Also, use visudo to validate the file first.

refs #7774


mistydemeo:
:+1:

PR #154, opened 2015-01-13
Dspace: Add link to parent collection in dmdSec

Hwesta:
Change dmdSec generated for Dspace METS files to contain DublinCore with the identifier and parent collection (isPartOf) instead of a mdRef to the METS.

refs #6273

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -183,7 +183,6 @@ def createDmdSecsFromCSVParsedMetadata(metadata):
 
     for key, value in metadata.iteritems():
         if key.startswith("dc.") or key.startswith("dcterms."):
-            #print "dc item: ", key, value
             if dc is None:
                 globalDmdSecCounter += 1
                 ID = "dmdSec_" + globalDmdSecCounter.__str__()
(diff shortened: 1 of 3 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -254,19 +254,28 @@ def createDublincoreDMDSecFromDBData(type, id):
...
+    if identifier is None or part_of is None:
+        print >> sys.stderr, 'Unable to parse identifer and isPartOf from', dspace_mets_path
+        return (None, None)
+
+    metadata = {
+        'dc.identifier': identifier,
+        'dc.isPartOf': part_of,
+    }
+    dmdsecs = createDmdSecsFromCSVParsedMetadata(metadata)
+    dmdSec = dmdsecs[0]  # Should only be one dmdSec
```
mistydemeo: Worth logging if more than one dmdSec occurs?

mistydemeo:
:thumbsup:

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
Hwesta: I don't think so, since I know exactly what metadata was passed into `createDmdSecsFromCSVParsedMetadata`.

Hwesta:
Updated to contain both the old mdRef dmdSec and the new identfier dmdSec.

DubinCore metadata is now namespaced differently depending on if it's dc or dcterms.

mistydemeo:
Still :+1:

mistydemeo:
:+1: on METS change

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -255,20 +258,48 @@ def createDublincoreDMDSecFromDBData(type, id):
...
-    for item in root.findall("{http://www.loc.gov/METS/}dmdSec"):
-        XPTR = "%s %s" % (XPTR, item.get("ID"))
-    XPTR = XPTR.replace(" ", "'", 1) + "'))"
-    newChild(dmdSec, ns.metsBNS + "mdRef", text=None, sets=[("LABEL", LABEL), (ns.xlinkBNS +"href", directoryPathSTR), ("MDTYPE", "OTHER"), ("LOCTYPE","OTHER"), ("OTHERLOCTYPE", "SYSTEM"), ("XPTR", XPTR)])
-    return (dmdSec, ID)
+    dmdid = "dmdSec_" + str(globalDmdSecCounter)
+    dmdsec = etree.Element(ns.metsBNS + "dmdSec", ID=dmdid)
+    dmdsecs[dmdid] = dmdsec
+    xptr_dmdids = [i.get('ID') for i in root.findall("{http://www.loc.gov/METS/}dmdSec")]
+    xptr = "xpointer(id('{}'))".format(' '.join(xptr_dmdids))
```
Hwesta: This will bail if the dmdSec does not contain an ID attribute. Easy fix is `i.get('ID', '')`, but do we want to suppress that error?

mistydemeo: Sounds like we probably want to fail with useful information in that case?

mistydemeo:
:thumbsup: to catching and logging the ID error

PR #157, opened 2015-01-20
Fix syntax errors

Hwesta:
Used flake8 to find syntax errors, primarily variables used before being defined. Remove unused removeEmptyDirectories.py rather than fixing the syntax error.

Didn't fix repeated problems in databaseInterface because that will be fixed in #98


mistydemeo:
:thumbsup:

mistydemeo:
Still :+1:

PR #158, opened 2015-01-21
Tests: Update tests to run from root

Hwesta:
Several tests only run from certain directories - update to use absolute paths instead of relative paths.  Also, updated versions of pytest etc to be a range rather than a specific version (especially since we already specified different specific versions!)

All of the FPRClient tests fail, because they're trying to query a non-existent server.  They need to be updated to use VCR.py

Tests can be run with `~/archivematica$ py.test` after setting `$PYTHONPATH` to  `archivematica/src/dashboard/src/` and `$DJANGO_SETTINGS_MODULE` to  `settings.local`


mistydemeo:
:+1:

PR #159, opened 2015-01-21
advanced_search: fix query generation

mistydemeo:
Due to a typo, certain dicts were being created as sets instead.

refs #6273


Hwesta:
:+1:

mistydemeo:
Pushed an additional query structure fix.

Hwesta:
:+1:  again

PR #160, opened 2015-01-22
launchSubprocess: don't mutate own environment

mistydemeo:
Since assigning to a new variable just assigns a new reference to the same object, launchSubprocess has been mutating its own process's environment instead of creating a new environment for the subprocess.

Fixes #5652.


Hwesta:
Really that easy? :+1:

PR #161, opened 2015-01-22
MCPClient: Add dashboard & archivematicaCommon to sys.path

Hwesta:
MCPClient adds the dashboard and archivematicaCommon to the sys.path for all client scripts.  Remove all changes to sys.path from client scripts.

Feedback on this both as a feature and implementation approach welcome.

### DIFF
--- a/src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
+++ b/src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
@@ -20,10 +20,10 @@
 # @package Archivematica
 # @subpackage archivematicaClientScript
 # @author Mike Cantelon <mike@artefactual.com>
-import sys, os, time, ConfigParser
-os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
-sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-sys.path.append("/usr/lib/archivematica/archivematicaCommon/externals")
+import ConfigParser
+import sys
+
+# archivematicaCommon
 import elasticSearchFunctions
 
 exitCode = 0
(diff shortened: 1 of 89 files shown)


src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
```diff
@@ -21,9 +21,9 @@
 # @subpackage archivematicaClientScript
 # @author Mike Cantelon <mike@artefactual.com>
 import sys, os, time, ConfigParser
-os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
-sys.path.append("/usr/lib/archivematica/archivematicaCommon")
-sys.path.append("/usr/lib/archivematica/archivematicaCommon/externals")
+
+# archivematicaCommon
+# externals
```
mistydemeo: I actually think `externals` is no longer required here.

mistydemeo: Ignore that, I forgot about `xmltodict`. Probably don't need to comment it here though?

mistydemeo:
:+1:

mistydemeo:
Instead of mutating the parent thread's own environment, maybe it's worth pushing this into the `executeOrRun` call in archivematicaClient.py? We could provide a generic way to specify a dict that will be merged into the subprocess's environment, and make the call of `executeOrRun` pass an environment dict with the `PYTHONPATH` and `DJANGO_SETTINGS_MODULE`.

Hwesta:
A-ha! I hadn't spotted the executeOrRun call, and thought everything had to be done in the threading calls and startThread.  I like that idea better.

DJANGO_SETTINGS_MODULE will still have to be set in archivematicaClient.py, because it imports from the dashboard.

mistydemeo:
Looks good! :+1:  I'm vaguely iffy on `env_updates` as the kwarg name, but I don't have a concrete suggestion. Going to merge this as part of #124?

Hwesta:
No, it just came out of that work but was conceptually separate.  Make the boilerplate for setting up Django in each client script shorter if I don't have to set the path and settings module every time.

PR #163, opened 2015-01-26
Add automation endpoints

Hwesta:
Supports automating transfers in automation-tools.
- Add endpoint to return status of a SIP or transfer, one of `FAILED`, `REJECTED`, `USER_INPUT`, `COMPLETE` or `PROCESSING`.  Attention to how these statuses are generated appreciated
- Add endpoint to start a transfer that takes API key instead of requiring the caller to be logged in
- Commit: "Dashboard API: Endpoint for all waiting units" may be dropped, as nothing currently uses that endpoint (it was part of an earlier implementation).
- PEP8 and restructuring cleanup

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -21,7 +21,7 @@ import os
 
 # Core Django, alphabetical
 from django.db.models import Q
-from django.http import Http404, HttpResponseForbidden, HttpResponseServerError
+import django.http
 
 # External dependencies, alphabetical
 from tastypie.authentication import ApiKeyAuthentication
@@ -115,7 +115,7 @@ def unapproved_transfers(request):
                 response['message'] = 'Fetched unapproved transfers successfully.'
 
                 if error is not None:
-                    return HttpResponseServerError(
+                    return django.http.HttpResponseServerError(
                         json.dumps(response),
                         mimetype='application/json'
                     )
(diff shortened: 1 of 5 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -48,26 +55,190 @@ def authenticate_request(request):
...
+    COMPLETE.  Otherwise, it is None.
+
+    :param str unit_uuid: UUID of the SIP or Transfer
+    :param str unit_type: unitSIP or unitTransfer
+    :return: Dict with status info.
+    """
+    ret = {}
+    job = models.Job.objects.filter(sipuuid=unit_uuid).filter(unittype=unit_type).order_by('-createdtime', '-createdtimedec')[0]
+    ret['microservice'] = job.jobtype
+    if job.currentstep == 'Awaiting decision':
```
mistydemeo, line 78: My kingdom for a case statement!

src/dashboard/src/components/api/views.py
```diff
@@ -48,26 +55,190 @@ def authenticate_request(request):
...
+        unit_uuid = job.sipuuid
+        directory = os.path.basename(os.path.normpath(job.directory))
+        unit_name = directory.replace('-' + unit_uuid, '', 1)
+
+        waiting_units.append({
+            'sip_directory': directory,
+            'sip_uuid': unit_uuid,
+            'sip_name': unit_name,
+            'microservice': job.jobtype,
+            # 'choices': []  # TODO? Return list of choices, see ingest.views.ingest_status
```
mistydemeo, line 179: Is the plan to do this before merging?

Hwesta, line 179: Probably not.  I thought having the list of choices would be a useful tidbit to send back, but ingest_status does ridiculous contortions to get the info, and I decided it wasn't worthwhile until something needed it.

PR #164, opened 2015-01-27
Extract packages recursively

Hwesta:
Packages may contain other packages. Check if unextracted packages still exist in the transfer after extraction happens and loop back if so. Also check if the file has been deleted before attempting extraction.

### DIFF
--- a/src/MCPClient/lib/clientScripts/hasPackages.py
+++ b/src/MCPClient/lib/clientScripts/hasPackages.py
@@ -1,13 +1,18 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
 import sys
 
 # dashboard
 from fpr.models import FPRule
-from main.models import FileFormatVersion, Transfer
+from main.models import FileFormatVersion, Transfer, File, Event
 
 
 def is_extractable(f):
+    """
+    Returns True if this file can be extracted, False otherwise.
+    """
+    # Check if an extract FPRule exists
     try:
         format = f.fileformatversion_set.get().format_version
     except FileFormatVersion.DoesNotExist:
@@ -19,13 +24,25 @@ def is_extractable(f):
     else:
         return False
 
+def already_extracted(f):
+    """
+    Returns True if this package has already been extracted, False otherwise.
+    """
+    # Look for files in a directory that starts with the package name
+    files = File.objects.filter(transfer=f.transfer, currentlocation__startswith=f.currentlocation, removedtime__isnull=True).exclude(uuid=f.uuid)
+    # Check for unpacking events that reference the package
+    if Event.objects.filter(file_uuid__in=files, event_type='unpacking', event_detail__contains=f.currentlocation).exists():
+        return True
+    return False
+
 
 def main(sip_uuid):
     transfer = Transfer.objects.get(uuid=sip_uuid)
-    for f in transfer.file_set.iterator():
-        if is_extractable(f):
+    for f in transfer.file_set.filter(removedtime__isnull=True).iterator():
+        if is_extractable(f) and not already_extracted(f):
+            print(f.currentlocation, 'is extractable and has not yet been extracted.')
             return 0
-
+    print('No extractable files found.')
     return 1
 
 if __name__ == '__main__':
--- a/src/MCPClient/lib/clientScripts/extractContents.py
+++ b/src/MCPClient/lib/clientScripts/extractContents.py
@@ -14,6 +14,9 @@ from fileOperations import addFileToTransfer, updateSizeAndChecksum
 from fpr.models import FPCommand
 from main.models import FileFormatVersion, File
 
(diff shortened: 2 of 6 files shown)


src/MCPClient/lib/clientScripts/hasPackages.py
```diff
@@ -19,13 +24,24 @@ def is_extractable(f):
     else:
         return False
 
+def already_extracted(f):
+    """
+    Returns True if this package has already been extracted, False otherwise.
+    """
+    # Look for files in a directory named like <zip file path>-<datetime of extraction>
+    # Don't know exact datetime of extraction, so suffix with '-20' - should be good until year 2100
+    if File.objects.filter(currentlocation__startswith=f.currentlocation + '-20', transfer=f.transfer).exists():
```
mistydemeo: I'm a little iffy about basing it on the name like this - was it not feasible to base it on the events instead?

Hwesta: Converted to a query that gets all files that start with <packagename>, and looks for unpacking events that contain <packagename> for those files.

src/MCPClient/lib/clientScripts/hasPackages.py
```diff
@@ -19,13 +24,25 @@ def is_extractable(f):
...
         return False
 
+def already_extracted(f):
+    """
+    Returns True if this package has already been extracted, False otherwise.
+    """
+    # Look for files in a directory that starts with the package name
+    files = File.objects.filter(transfer=f.transfer, currentlocation__startswith=f.currentlocation, removedtime__isnull=True).exclude(uuid=f.uuid)
+    # Check for unpacking events that reference the package
+    if Event.objects.filter(file_uuid__in=files, event_type='unpacking', event_detail__contains=f.currentlocation).exists():
```
mistydemeo, line 34: Just to check, then, the `f` in question is not a File object for the file to be extracted? Is it possible to query for Events using the given `f` object?

Hwesta, line 34: I'm assuming with already_extracted that f is_extractable and is a package.  Events aren't generated for the package being extracted from, just for each of the files extracted (since the package itself could be deleted).

mistydemeo, line 34: If we don't already, let's start tracking parentage information and use that here - e.g. have a way to filter on "unpacked from X", which we could use to determine if anything had been extracted from it.

Hwesta, line 34: We don't explicitly track that anywhere, and I'm not sure how often it would be used except here. The Events table implicitly has it, but it's subject to the eventDetail text.  Also, the source package may have been deleted.

mistydemeo:
Justin noticed an issue where existing files in the transfer get sent through identification again on every loop through the transfer; this results in an arbitrary number of identical identification events being written to the METS.

We don't currently guard against multiple identification events showing up for the same file because we intentionally allow users to choose to reidentify files in ingest, but we didn't have a workflow where the same files could get sent through multiple times like this.

Maybe we should add a flag to the identifyFileFormat script that suppresses writing the ID event if one already exists - leave the current behaviour the default, but allow that option to be passed to the microservice that's identifying files in this workflow.

Example:

``` xml
<premis:format>
  <premis:formatDesignation>
    <premis:formatName>ZIP file</premis:formatName>
    <premis:formatVersion></premis:formatVersion>
  </premis:formatDesignation>
  <premis:formatRegistry>
    <premis:formatRegistryName>PRONOM</premis:formatRegistryName>
    <premis:formatRegistryKey>x-fmt/263</premis:formatRegistryKey>
  </premis:formatRegistry>
</premis:format>
<premis:format>
  <premis:formatDesignation>
    <premis:formatName>ZIP file</premis:formatName>
    <premis:formatVersion></premis:formatVersion>
  </premis:formatDesignation>
  <premis:formatRegistry>
    <premis:formatRegistryName>PRONOM</premis:formatRegistryName>
    <premis:formatRegistryKey>x-fmt/263</premis:formatRegistryKey>
  </premis:formatRegistry>
</premis:format>
<premis:format>
  <premis:formatDesignation>
    <premis:formatName>ZIP file</premis:formatName>
    <premis:formatVersion></premis:formatVersion>
  </premis:formatDesignation>
  <premis:formatRegistry>
    <premis:formatRegistryName>PRONOM</premis:formatRegistryName>
    <premis:formatRegistryKey>x-fmt/263</premis:formatRegistryKey>
  </premis:formatRegistry>
</premis:format>
```

Hwesta:
Updated file identification to add a --disable-reidentify switch and added that to extraction identification.

mistydemeo:
Looks good - so long as this still works okay with the new processed_structmap stuff that follows extraction.

PR #166, opened 2015-01-30
Metadata.csv: Strip whitespace from headers

Hwesta:
Remove whitespace from the element names in metadata.csv, to reduced XML errors when trying to create elements from them.


jhsimpson:
:shipit:

PR #168, opened 2015-02-04
AIC: misc fixes

Hwesta:
DublinCore elements are namespaced correctly now, but in older versions the 'title' element may be dc or dcterms.  Update parsing to handle either.

AIC METS structMap LABEL will fall back to AIP name if no title is found.

AIC metadata form doesn't inherit widget info, so copy that.

Fix template reading of # of AIPs in an AIC.


mistydemeo:
:thumbsup:

PR #170, opened 2015-02-05
taskStandard.writeOutputs: encode strings before concatenating

mistydemeo:
This was presumably broken by the model branch. The `standardOut` and `standardError` strings are now as unicode strings, which can contain unencoded Unicode characters, in which case the string concatenation fails.


Hwesta:
:+1:

PR #171, opened 2015-02-18
Transfer/SIP creation feedback

Hwesta:
Add feedback so users can tell easier if Transfer/SIP creation will be slow.

This relies on artefactual/archivematica-storage-service#64 to provide additional information, specifically a properties dict with keys including 'object count' and 'size'
- Adds a display_string to the properties dict, which is displayed next to the entry name in the file browser/explorer
- Converts SIP Arrange's not_draggable attribute to use the properties dict
- Add an activity indicator when creating a SIP. Renamed to be more generic
- Remove unused templates.

This shouldn't have many conflicts with #130, since I was inspired by how display of extra text was done there, but they will need extra testing together.


mistydemeo:
:thumbsup:

PR #172, opened 2015-02-18
Dashboard: Remove transfer components

Hwesta:
Fix finding the ID of the transfer component to delete. Display path of to-be-removed transfer component in popup. Fixes #7482


mistydemeo:
:thumbsup:

PR #153, opened 2015-01-07
Transfer METS: namespace all elements, include additional metadata

mistydemeo:

PR #173, opened 2015-02-19
DublinCore: Namespace properly, update indexing

Hwesta:
Namespace DublinCore generated from the DB correctly as dcterms or dc. Fix is_part_of to isPartOf. When indexing, search for both dc and dcterms versions of the DublinCore elements.

Add tests for DC changes.  To support importing createMETS2, cherry-picked commits from #153

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -118,7 +118,24 @@ FileMetadataAppliesToType = '7f04d9d4-92c2-44a5-93dc-b7bfdf0c1f17'
 
 
 def getDublinCore(unit, id):
-    field_list = ["title", "creator", "subject", "description", "publisher", "contributor", "date", "type", "format", "identifier", "source", "relation", "language", "coverage", "rights", "is_part_of"]
+    db_field_mapping = collections.OrderedDict([
+        ("title", "title"),
+        ("creator", "creator"),
+        ("subject", "subject"),
+        ("description", "description"),
+        ("publisher", "publisher"),
+        ("contributor", "contributor"),
+        ("date", "date"),
+        ("type", "type"),
+        ("format", "format"),
+        ("identifier", "identifier"),
+        ("source", "source"),
+        ("relation", "relation"),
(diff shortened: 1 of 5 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -145,13 +143,19 @@ def getDublinCore(unit, id):
...
 
-    for term in field_list:
-        txt = getattr(dc, term)
+    for dbname, term in db_field_mapping.iteritems():
+        txt = getattr(dc, dbname)
+        elem_ns = ''
+        # See http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=elements for which elements are which namespace
+        if term in ('contributor', 'coverage', 'creator', 'date', 'description', 'format', 'identifier', 'language', 'publisher', 'relation', 'rights', 'source', 'subject', 'title', 'type'):
+            elem_ns = ns.dcBNS
+        elif term in ('isPartOf'):
```
mistydemeo: Nitpick, but just check for string equality? Otherwise, this needs an extra comma (it's being evaluated as `'isPartOf`, not as a one-length tuple)

Hwesta: Good catch. I'd like to leave it as `term in (...)` because we may have multiple dcterms terms in the future. Not a strong preference.

mistydemeo:
_Love_ the test text.

:+1: aside from the one comment!

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
mistydemeo: Works for me.

Hwesta: Added tuple-comma.

Hwesta:
Thanks! I got bored of 'test', 'testy', 'foo' etc as testing values and set up something more entertaining and referencing one of my favorite book series (Tamora Pierce's Tortall books). :)

mistydemeo:
:shipit:

PR #175, opened 2015-02-27
Dashboard: Accession IDs associate with Transfers again

Hwesta:
For packages, we still want to add the package name to the transfer path. Only append a / if it is a directory. Otherwise, a Transfer is created in the dashboard that does not match the path of the transfer on disk, so
MCPClient creates another Transfer when it sees the new file.


mistydemeo:
:+1:

PR #176, opened 2015-02-27
Tests: Setup default testing config in pytest.ini

Hwesta:
Add Django settings and a useful PYTHONPATH to the pytest.ini file.

Please test this and tell me if it actually makes running tests easier.


mistydemeo:
:+1:

PR #177, opened 2015-03-04
sanitize: Don't write events for directory name sanitization

mistydemeo:
These events were associated with files (not directories), were repeated several times over in the METS file, and were highly misleading.

We don't currently treat directories as intellectual entities for the sake of PREMIS, but this might change in the future. Meanwhile, the sanitization log file (which is stored in the AIP) has accurate information on the sanitization process and should be preferred for looking up directory names.

Fixes #8033.

### DIFF
--- a/src/archivematicaCommon/lib/fileOperations.py
+++ b/src/archivematicaCommon/lib/fileOperations.py
@@ -178,8 +178,14 @@ def updateFileLocation2(src, dst, unitPath, unitIdentifier, unitIdentifierType, 
     f.currentlocation = dstDB
     f.save()
 
-def updateFileLocation(src, dst, eventType, eventDateTime, eventDetail, eventIdentifierUUID = uuid.uuid4().__str__(), fileUUID="None", sipUUID = None, transferUUID=None, eventOutcomeDetailNote = ""):
-    """If the file uuid is not provided, will use the sip uuid and old path to find the file uuid"""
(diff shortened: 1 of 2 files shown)


Hwesta:
:+1:

mistydemeo:
Updated; had to alter the approach, but should be fine now.

src/archivematicaCommon/lib/fileOperations.py
```diff
@@ -178,8 +178,12 @@ def updateFileLocation2(src, dst, unitPath, unitIdentifier, unitIdentifierType,
     f.currentlocation = dstDB
     f.save()
 
-def updateFileLocation(src, dst, eventType, eventDateTime, eventDetail, eventIdentifierUUID = uuid.uuid4().__str__(), fileUUID="None", sipUUID = None, transferUUID=None, eventOutcomeDetailNote = ""):
-    """If the file uuid is not provided, will use the sip uuid and old path to find the file uuid"""
+def updateFileLocation(src, dst, eventType, eventDateTime, eventDetail, eventIdentifierUUID = uuid.uuid4().__str__(), fileUUID="None", sipUUID = None, transferUUID=None, eventOutcomeDetailNote = "", createEvent=True):
+    """
+    If the file uuid is not provided, will use the sip uuid and old path to find the file uuid.
```
Hwesta: Add a one-line summary to the beginning of the docstring?  "Updates file location in the database and optionally generates an Event.  NOTE does not actually move file. """ ?

Hwesta:
:+1:

mistydemeo:
Updated docstring, and updated the function to make the event-related arguments optional; they're no longer passed when creating a directory.

PR #179, opened 2015-03-09
Remove unused Elasticsearch backup code

mistydemeo:
Closes #8076.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+dev (1.4.0)
+===========
+
+* Remove unused Elasticsearch backup code (#8076)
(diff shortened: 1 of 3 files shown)


mistydemeo:
The script to restore from this backup is in archivematica-devtools; PR there at artefactual/archivematica-devtools#5.

Hwesta:
Update SQL to drop ElasticsearchIndexBackup?

mistydemeo:
I'm fine with leaving the empty table personally, especially since I don't want to drop data users might have actually stored (assuming this was ever used).

Hwesta:
I guess there's no model for it, so when we switch to using the ORM to create things it'll stop existing then.

:+1:

PR #180, opened 2015-03-11
upload-archivistsToolkit: use placeholders instead of string interpolation

mistydemeo:
Fixes #7771


jhsimpson:
This is a better solution, thanks, looks good to me.

:+1:

Hwesta:
:+1:

PR #182, opened 2015-03-13
Dashboard: Display Transfer Source descriptions instead of path

Hwesta:
Instead of always displaying the path, display the user-provided description of transfer source locations. This disambiguates identically named folders on different systems. The description defaults to the path if none provided.


mistydemeo:
Haven't tested but looks good. :+1:

Hwesta:
It works for display but doesn't work for actually fetching the files (the path I changed is what's sent to browse).  Requires more work, unfortunately.

PR #183, opened 2015-03-18
start_transfer: log exceptions

mistydemeo:
Fixes #8117

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -255,6 +255,7 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
                 type=transfer_type, accession=accession,
                 transfer_metadata_set_row_uuid=row_id)
         except Exception:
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -255,6 +255,7 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
                 type=transfer_type, accession=accession,
                 transfer_metadata_set_row_uuid=row_id)
         except Exception:
+            logging.exception("Error copying %s to start of transfer", filepath)
```
Hwesta, line 258: Do we want this to show up as ERROR level?  Is `logging.info(..., exc_info=True)` enough?

mistydemeo, line 258: I think we'd want this to usually be printed; at `info` there's a good chance that many installations would be configured to not show it.

Hwesta, line 258: Good point, it is a pretty important error.

Hwesta:
:+1:

PR #184, opened 2015-03-26
Duplicate columns in metadata.csv

Hwesta:
metadata.csvs may have duplicate columns. Store values in a list to handle this case.  Also, fix some typos.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataCSV.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataCSV.py
@@ -33,6 +33,15 @@ import traceback
 from sharedVariablesAcrossModules import sharedVariablesAcrossModules
 
 
+class OrderedListsDict(collections.OrderedDict):
+    def __setitem__(self, key, value):
+        # When inserting, insert into a list of items with the same key
+        try:
+            self[key]
+        except KeyError:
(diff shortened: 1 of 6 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataCSV.py
```diff
@@ -43,7 +51,7 @@ def parseMetadata(SIPPath):
     See parseMetadataCSV for details on parsing.
 
     :param SIPPath: Path to the SIP
-    :return: {<filename>: OrderedDict{<metadata name>=<metadata value>} }
+    :return: {<filename>: OrderedDict(key: [values]) }
```
mistydemeo, line 55: Nitpick, but document that it returns an `OrderedListsDict`?

Hwesta, line 55: I mostly used the OrderedListsDict so I didn't have to handle duplicates myself when inserting. Should I `return OrderedDict(metadata)` instead?

mistydemeo, line 55: Yeah, I'd say either document the actual return type or convert to an `OrderedDict` before returning.

mistydemeo:
:thumbsup:

PR #185, opened 2015-03-30
ArchivesSpace integration

mistydemeo:
This makes quite a number of changes to the Archivists' Toolkit workflow and introduces ArchivesSpace as an option.
- Breaks out the existing ATK code from multiple locations into a single client class
- Introduces a new ArchivesSpace class with an identical interface
- Adds an ArchivesSpace upload script
- Separates the generic pair matching code from ATK-specific code; the new `pair_matcher` module is used by the `views_atk` module and the new `views_as` module
- Allows re-entering the pair matching UI
- Display existing matches when loading the pair matching UI
- Streamline the pair matcher by POSTing and DELETEing matches as soon as the "pair" button is hit, and remove the "save" button


src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/python
```
Hwesta: Source compatible? :D

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
...
+            yield os.path.join(root, file)
+
+
+def get_files_from_dip(dip_location, dip_name, dip_uuid):
+    # need to find files in objects dir of dip:
+    # go to dipLocation/dipName/objects
+    # get a directory listing
+    # for each item, set fileName and go
+    try:
+        mydir = dip_location + "objects/"
```
Hwesta, line 33: os.path.join?  Or is this a URL?

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
...
+        mylist = list(recursive_file_gen(mydir))
+
+        if len(mylist) > 0:
+            return mylist
+        else:
+            logger.error("no files in " + mydir)
+            raise ValueError("cannot find dip")
+            exit(2)
```
Hwesta, line 41: Will exit ever get called? Won't the raise trigger first and end the function?  Same below.

src/MCPClient/lib/clientScripts/upload-archivesspace.py
mistydemeo: Probably not! This is the same shebang as upload-atk though... which probably wasn't intentional there

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
...
+def get_pairs(dip_uuid):
+    return {pair.fileuuid: pair.resourceid for pair in ArchivesSpaceDIPObjectResourcePairing.objects.filter(dipuuid=dip_uuid)}
+
+
+def delete_pairs(dip_uuid):
+    ArchivesSpaceDIPObjectResourcePairing.objects.filter(dipuuid=dip_uuid).delete()
+
+
+def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_type, use_statement, uri_prefix, dip_uuid, access_conditions, use_conditions, restrictions, dip_location):
+
```
Hwesta, line 56: Docstring?  Also elsewhere, but holy cow this has a lot of params.

mistydemeo, line 56: Yeah, a _lot_ of the structure of this is copied from upload-atk. :cold_sweat:

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/python
+
+import argparse
+import logging
+import os
+from urlparse import urljoin
+
+from main.models import ArchivesSpaceDIPObjectResourcePairing
```
Hwesta, line 7: Depending on the order that #124 comes in this might have to be updated.

src/MCPClient/lib/clientScripts/upload-archivesspace.py
Hwesta, line 56: Ahh, I didn't realize that (I didn't see a matching removal diff). Docstring extra useful then, if extra painful to write. :P

mistydemeo, line 56: The other script still exists; I used it as a template, since this does mostly the same things with some customizations. Eventually might want to refactor into one script with two backends but that wasn't feasible at this point in time.

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -0,0 +1,169 @@
...
+
+    uri = get_uri(uri_prefix, dip_uuid)
+    pairs = get_pairs(dip_uuid)
+    dashboard_uuid = getDashboardUUID()
+
+    # get mets object if needed
+    mets = None
+    if restrictions == 'premis' or len(access_conditions) == 0 or len(use_conditions) == 0:
+        logger.debug("Looking for mets: {}".format(dip_uuid))
+        mets_source = dip_location + 'METS.' + dip_uuid + '.xml'
```
Hwesta, line 66: os.path.join?  Are there any guarantees on dip_location?

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+
+class ConnectionError(Exception):
+    pass
+
+
+class AuthenticationError(Exception):
+    pass
+
+
+class CommunicationError(Exception):
```
Hwesta: Should these inherit from `ArchivesSpaceError`? So it's easier to catch anything generated by this module?

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+    def _login(self):
+        try:
+            response = requests.post(self.host + '/users/' + self.user + '/login',
+                                     data={'password': self.passwd})
+        except requests.ConnectionError as e:
+            raise ConnectionError("Unable to connect to ArchivesSpace server: " + str(e))
+
+        try:
+            output = response.json()
+        except:
```
Hwesta: Catch the specific error(s)? ValueError?

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+        message = "ArchivesSpace server responded {}".format(status_code)
+        self.response = response
+        super(CommunicationError, self).__init__(message)
+
+
+class ArchivesSpaceError(Exception):
+    pass
+
+
+class ArchivesSpaceClient(object):
```
Hwesta, line 26: Tests for this?

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+    def _request(self, method, url, params, expected_response, data=None, retry=True):
+        if not url.startswith('/'):
+            url = '/' + url
+
+        response = method(self.host + url, params=params, data=data)
+        if response.status_code != expected_response:
+            # session has expired; acquire a new session, then retry once
+            if retry and response.status_code == 412 and "SESSION_GONE" in response.text:
+                self._login()
+                return self._request(method, url, params, expected_response, data=data, retry=False)
```
Hwesta, line 80: Could this infinitely recurse if the session isn't created each time? (Is that out of scope?)

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+            raise CommunicationError(response.status_code, response)
+
+        output = response.json()
```
Hwesta: Is this likely to have a JSON parsing error?

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo, line 80: It recurses with the `retry` kwarg set to `False`, and L78 checks that first - it will never retry more than once.

Hwesta, line 80: Ah, I didn't catch that.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+                'dates': dates,
+                'levelOfDescription': record['level']
+            }
+            if record['children'] and descend:
+                result['children'] = [format_record(child, level) for child in record['children']]
+                if sort_by is not None:
+                    kwargs = {'reverse': True} if sort_by == 'desc' else {}
+                    result['children'] = sorted(result['children'], key=lambda c: c['title'], **kwargs)
+            else:
+                result['children'] = False
```
Hwesta, line 186: None is a better indicator of no-data.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def find_parent_id_for_component(self, component_id):
+        """
+        Given the URL to a component, returns the parent component's URL.
+
+        :param string component_id: The URL of the component.
+        :return tuple: A tuple containing:
+            * The type of the parent record; valid values are ArchivesSpaceClient.RESOURCE and ArchivesSpaceClient.RESOURCE_COMPONENT.
+            * The URL of the parent record.
+            If the provided URL fragment references a resource, this method will simpl return the same URL.
```
Hwesta: Typo: s/simpl/simply/

src/archivematicaCommon/lib/lazy_paged_sequence.py
```diff
@@ -0,0 +1,48 @@
+class LazyPagedSequence(object):
```
Hwesta: Do we want to vendor this or use the dependency lazy-paged-sequence?

mistydemeo: Let's use the dependency; this is only vendored because I hadn't published it yet the last time this was touched.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo, line 186: This is identical to the behaviour used by the previous Archivist's Toolkit code - I decided not to make any changes to the data format.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo: It looks like the raised exception is a `simplejson.scanner.JSONDecodeError`, but I'm not 100% sure if requests always uses that as a JSON decoder?

mistydemeo: Looks like the surface area where this could come up is limited, but it's possible for the JSON decoder to not be simplejson (and hence for the exception to be a different class). Think I'll leave this alone.

mistydemeo: https://github.com/kennethreitz/requests/blob/8b5e457b756b2ab4c02473f7a42c2e0201ecc7e9/requests/compat.py#L24-L29

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo: Good thought.

mistydemeo:
Added tests to `ArchivesSpaceClient`. They aren't 100% comprehensive but they cover most functionality.

mistydemeo:
Pushed a fix to the ATK path.

I think this is ready to be merged?

src/MCPClient/lib/clientScripts/upload-archivesspace.py
Hwesta: Update this to /usr/bin/env python2 to match #255 ?

src/archivematicaCommon/lib/archivesspace/client.py
Hwesta: Then how about except Exception:  so we don't catch SysExit etc

mistydemeo: Updated!

src/archivematicaCommon/tests/test_archivesspace_client.py
```diff
@@ -0,0 +1,124 @@
+# -*- coding: UTF-8 -*-
```
Hwesta, line 1: :tada:  tests!

Hwesta:
:+1:

mistydemeo:
Merged! :fireworks:

PR #186, opened 2015-04-02
Manual Normalization: Misc fixes

Hwesta:
Fix typo.

Update the preservation file's path in the DB instead of the original file's path. Minor cleanup.

Generate expected exception when looking for a normalization event, instead of using .filter().update() which always succeeds.


mistydemeo:
CRed as a part of #187.

mistydemeo:
:+1:

PR #187, opened 2015-04-03
Ingest MSCL cleanup

Hwesta:
Numerous small fixes. Notable ones include:
- Merging the Approve Normalization nodes. ProcessingMCP.xml now works to automate Approve Normalization in all cases
- DIP creation is based on the existence of a DIP directory, which is created in all access normalization paths
- Cleanup of microservice groups in Normalization. Added Process manually normalized files, Generate AIP METS and Verify checksums, and ensure that microserviceGroups are contiguous
- Removed dead DIPfromAIP MSCL & scripts
- Delete all TasksConfigs and StandardTasksConfigs with no foreign keys pointing at them.

Recommend checking this out and reviewing the graphlinks and normalization behaviours.

Currently includes #186 but will be merged after that.


src/MCPClient/lib/clientScripts/manualNormalizationCheckForManualNormalizationDirectory.py
```diff
@@ -20,31 +20,25 @@
...
-            exit(179)
-    manualNormalizationPreservationPath = os.path.join(manualNormalizationPath, "preservation")
-    if os.path.isdir(manualNormalizationPreservationPath):
-        if len(os.listdir(manualNormalizationPreservationPath)):
-            exit(179)
+    mn_access_path = os.path.join(manualNormalizationPath, "access")
+    mn_preserve_path = os.path.join(manualNormalizationPath, "preservation")
+    # Return to indicate manually normalized files exist
+    if os.path.isdir(mn_access_path) and os.listdir(mn_access_path):
+        print('Manually normalized files found')
```
mistydemeo, line 38: Maybe specify which files were found when returning?

Hwesta, line 38: Thought about it and decided not to. Partly because it could be a very long list, and because this returns true if access _or_ preservation copies found - if both were found only one list would be printed because we exit immediately (or doing the other check just so we could print the other set of files)

mistydemeo:
Need to test, but looks good. :+1:

src/MCPServer/share/mysql_dev_6488_aip_reingest.sql
```diff
@@ -1,3 +1,125 @@
...
+SET @verifychecksumMSCL = '88807d68-062e-4d1a-a2d5-2d198c88d8ca' COLLATE utf8_unicode_ci;
+UPDATE MicroServiceChainLinks SET microserviceGroup='Verify checksums' WHERE pk IN (@verifychecksumMSCL);
+--  Skip verify
+UPDATE MicroServiceChainLinks SET defaultNextChainLink='ee438694-815f-4b74-97e1-8e7dde2cc6d5' WHERE defaultNextChainLink=@verifychecksumMSCL;
+UPDATE MicroServiceChainLinksExitCodes SET nextMicroServiceChainLink='ee438694-815f-4b74-97e1-8e7dde2cc6d5' WHERE nextMicroServiceChainLink=@verifychecksumMSCL;
+--   New predecessor point to verify
+UPDATE MicroServiceChainLinks SET defaultNextChainLink=@verifychecksumMSCL WHERE defaultNextChainLink='54b73077-a062-41cc-882c-4df1eba447d9';
+UPDATE MicroServiceChainLinksExitCodes SET nextMicroServiceChainLink=@verifychecksumMSCL WHERE nextMicroServiceChainLink='54b73077-a062-41cc-882c-4df1eba447d9';
+--   verify point to new successor
+UPDATE MicroServiceChainLinks SET defaultNextChainLink='54b73077-a062-41cc-882c-4df1eba447d9' WHERE pk=@verifychecksumMSCL;
```
Hwesta: This should be removed to not revert #190

PR #188, opened 2015-04-06
DSpace transfers from files, automation fixes

Hwesta:
Updates to support automation of DSpace transfers
- Handle folders & deleted files in get status
- DSpace transfer type handles files or folders when creating a transfer

Works with artefactual/automation-tools#2 and artefactual/archivematica-storage-service#73

### DIFF
--- a/src/MCPClient/lib/clientScripts/file_to_folder.py
+++ b/src/MCPClient/lib/clientScripts/file_to_folder.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import os
+import sys
+
+# dashboard
+from main.models import Transfer
+
+def main(transfer_path, transfer_uuid, shared_path):
+    """
+    Move a Transfer that is a file into a directory.
+
+    Given a transfer that is one file, move the file into a directory with the same basename and updated the DB for the new path.
+    No action taken if the transfer is already a directory.
+
+    WARNING This should not be run inside a watched directory - may result in duplicated transfers.
+
+    :param transfer_path: Path to the current transfer, file or folder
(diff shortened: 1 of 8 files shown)


src/MCPClient/lib/clientScripts/file_to_folder.py
```diff
@@ -0,0 +1,44 @@
...
+    :param transfer_path: Path to the current transfer, file or folder
+    :param transfer_uuid: UUID of the transfer to update
+    :param shared_path: Value of the %sharedPath% variable
+    """
+    # If directory, return unchanged
+    if os.path.isdir(transfer_path):
+        print(transfer_path, 'is a folder, no action needed. Exiting.')
+        return 0
+    # If file, move into directory and update transfer
+    dirpath = os.path.splitext(transfer_path)[0]
```
mistydemeo, line 28: I'm guessing this isn't possible with this workflow, but would this blow up if given a file with no extension?

mistydemeo:
:+1:

mistydemeo:
:shipit:

src/MCPClient/lib/clientScripts/file_to_folder.py
Hwesta, line 28: If given a file with no extension, then `dirpath` would be the same as `transfer_path`, so `os.mkdir(dirpath)` below would fail.  It will also have weird results if `dirpath` already exists.  I think I'll change this to exit 1 if something weird happens.

mistydemeo, line 28: Sounds good!

mistydemeo:
:shipit:

PR #191, opened 2015-04-07
Ingest, index bag-info metadata

mistydemeo:
This adds support for ingesting transfer metadata from bags using the data contained within the `bag_info.txt`. `bag-info.txt` contains [simple key-value pairs](https://tools.ietf.org/html/draft-kunze-bagit-10#section-2.2.2).

Changes in this branch:
- createMETS2 now uses python-bagit to parse each originating transfer's bag-info.txt. Each key/value pair is translated into XML tags containing the value as text, within a `mets:sourceMD/xmlData/transfer_metadata` element; the overall structure is essentially identical to disk image transfer metadata. Any key/value pairs whose keys are not XML tag-compatible are skipped.
- All `transfer_metadata` elements are now indexed in Elasticsearch.
- "Transfer metadata" has been added as a searchable value in archival storage, and the values of any BagIt tags can be searched on.

### DIFF
--- a/src/archivematicaCommon/requirements/base.txt
+++ b/src/archivematicaCommon/requirements/base.txt
@@ -1,2 +1,3 @@
+bagit==1.5.2
 Django==1.5.4
 elasticsearch>=1.0.0,<2.0.0
--- a/src/dashboard/src/components/advanced_search.py
+++ b/src/dashboard/src/components/advanced_search.py
@@ -16,10 +16,10 @@
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
 from django.http import HttpResponse
-from datetime import datetime
 import logging
 import sys
 
+import dateutil.parser
 from elasticsearch import Elasticsearch
 
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
@@ -195,6 +195,12 @@ def _parse_date_range(field):
 
     return field.split(':')[:2]
 
+def _normalize_date(date):
+    try:
+        return dateutil.parser.parse(date).strftime('%Y-%m-%d')
+    except ValueError:
+        raise ValueError("Invalid date received ({}); ignoring date query".format(date))
+
 def query_clause(index, queries, ops, fields, types):
     if fields[index] == '':
         search_fields = []
@@ -219,12 +225,12 @@ def query_clause(index, queries, ops, fields, types):
         return {'query_string': {'query': queries[index], 'fields': search_fields}}
     elif types[index] == 'range':
         start, end = _parse_date_range(queries[index])
-        for date in (start, end):
-            try:
-                datetime.strptime(date, '%Y-%m-%d')
-            except ValueError:
-                logger.info("Invalid date received (%s); ignoring date query", (date))
-                return
+        try:
+            start = _normalize_date(start)
+            end = _normalize_date(end)
+        except ValueError as e:
+            logger.info(str(e))
+            return
         return {'range': {fields[index]: {'gte': start, 'lte': end}}}
 
 def indexed_count(index, types=None, query=None):
--- a/src/dashboard/src/components/advanced_search.py
+++ b/src/dashboard/src/components/advanced_search.py
@@ -16,6 +16,8 @@
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
 from django.http import HttpResponse
+from datetime import datetime
+import logging
 import sys
 
 from elasticsearch import Elasticsearch
@@ -23,9 +25,15 @@ from elasticsearch import Elasticsearch
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 import elasticSearchFunctions
 
+logger = logging.getLogger("archivematica.dashboard.advanced_search")
+
 OBJECT_FIELDS = (
     "mets",
     "transferMetadata",
+)
+
+OTHER_FIELDS = (
+    "transferMetadataOther"
 )
 
 def search_parameter_prep(request):
@@ -33,6 +41,7 @@ def search_parameter_prep(request):
     ops     = request.GET.getlist('op')
     fields  = request.GET.getlist('field')
     types   = request.GET.getlist('type')
+    other_fields = request.GET.getlist('fieldName')
 
     # prepend default op arg as first op can't be set manually
     ops.insert(0, 'or')
@@ -41,10 +50,8 @@ def search_parameter_prep(request):
         queries = ['*']
         fields  = ['']
     else:
-        index = 0
-
         # make sure each query has field/ops set
-        for query in queries:
+        for index, query in enumerate(queries):
             # a blank query makes ES error
             if queries[index] == '':
                 queries[index] = '*'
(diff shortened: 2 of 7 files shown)


src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -49,14 +49,15 @@ $(document).ready(function() {
...
-    'AICID'        : 'AIC Identifier',
+    ''                : 'Any',
+    'FILEUUID'        : 'File UUID',
+    'filePath'        : 'File path',
+    'fileExtension'   : 'File extension',
+    'AIPUUID'         : 'AIP UUID',
+    'sipName'         : 'AIP name',
+    'isPartOf'        : 'Part of AIC',
+    'AICID'           : 'AIC Identifier',
+    'transferMetadata': 'Transfer metadata',
```
Hwesta, line 69: Nitpick, but why not remove all the extra spaces and be PEP8-happy instead of re-aligning them?

mistydemeo, line 69: I disagree with PEP8 on this point to be honest :b But I can switch the style!

mistydemeo, line 69: (I guess this kind of change requiring touching extra lines of code is why though...)

src/archivematicaCommon/requirements/base.txt
```diff
@@ -1,2 +1,3 @@
+bagit==1.5.2
```
Hwesta, line 1: Should we add a MCPClient requirements file?  The dependency is in MCPClient not archivematicaCommon.  Or do we want to let archivematicaCommon manage requirements for MCPClient and have MCPClient require archivematicaCommon?

src/dashboard/src/components/advanced_search.py
```diff
@@ -147,11 +152,21 @@ def assemble_query(queries, ops, fields, types, **kwargs):
...
 
+def _fix_object_fields(fields):
+    """
+    Adjusts field names for nested object fields.
+
+    Elasticsearch is able to search through nested object fields, provided that the field name is specified appropriately in the query.
+    Appending .* to the field name (for example, transferMetadata.*) causes Elasticsearch to consider any of the values within key/value pairs nested in the object being searched.
+    Without doing this, Elasticsearch will attempt to match the value of transferMetadata itself, which will always fail since it's an object and not a string.
+    """
+    return [field + '.*' if field in OBJECT_FIELDS else field for field in fields]
```
Hwesta, line 184: Your Clojure is showing. :P

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
Hwesta, line 69: Yeah, re-adjusting a bunch of lines because you added a long one is annoying, and having one be misaligned is also annoying. Not a strong preference.

src/dashboard/src/components/advanced_search.py
mistydemeo, line 184: :b I originally wrote this as a long loop, then thought better.

Hwesta:
:+1:

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
mistydemeo, line 69: I'll switch to PEP8-style.

mistydemeo:
We're waiting on client feedback so there may be more changes, but will merge otherwise!

mistydemeo:
Pushed a couple more commits to meet new requirements:
- Dates are autodetected in documents that are pushed, allowing dates to be found in arbitrary transferMetadata (e.g. bag-info) fields
- A new date range query option has been added to archival storage
- A new field option has been added to archival storage, allowing specific transfer metadata fields to be specified by name. The input box is only displayed for "Transfer Metadata other".

src/dashboard/src/components/advanced_search.py
```diff
@@ -59,7 +71,9 @@ def search_parameter_prep(request):
             except:
                 types.insert(index, '')
 
-            index = index + 1
+        for index, field in enumerate(fields):
```
Hwesta, line 86: Is this supposed to be inserting into `fields` after the above for loop?  Are we overwriting entries in `fields`?   `enumerate` takes a second param of where to start indexing from.

mistydemeo, line 86: >  Are we overwriting entries in fields?

Yep! The idea is that there is a second array of "other fields", populated when the user has manually typed in a field. If the field name indicates that an "other field" has been chosen, the placeholder value is overwritten with the actual name of the field from the other array.

Because there can be multiple searches, this enumerates from the beginning of the array to the end.

src/dashboard/src/components/advanced_search.py
```diff
@@ -167,8 +197,17 @@ def query_clause(index, queries, ops, fields, types):
...
                 term_field = '_all'
             return {'term': {term_field: queries[index]}}
-    else:
+    elif types[index] == 'string':
         return {'query_string': {'query': queries[index], 'fields': search_fields}}
+    elif types[index] == 'range':
+        start, end = _parse_date_range(queries[index])
+        for date in (start, end):
+            try:
+                datetime.strptime(date, '%Y-%m-%d')
```
Hwesta: If you're parsing user-inputted dates, [dateutil.parse](http://labix.org/python-dateutil#head-c0e81a473b647dfa787dc11e8c69557ec2c3ecd2) may be useful.  It does fuzzy matching.

src/dashboard/src/components/advanced_search.py
```diff
@@ -147,13 +161,29 @@ def assemble_query(queries, ops, fields, types, **kwargs):
...
+    """
+    Adjusts field names for nested object fields.
+
+    Elasticsearch is able to search through nested object fields, provided that the field name is specified appropriately in the query.
+    Appending .* to the field name (for example, transferMetadata.*) causes Elasticsearch to consider any of the values within key/value pairs nested in the object being searched.
+    Without doing this, Elasticsearch will attempt to match the value of transferMetadata itself, which will always fail since it's an object and not a string.
+    """
+    return [field + '.*' if field in OBJECT_FIELDS else field for field in fields]
+
+def _parse_date_range(field):
```
Hwesta, line 186: Docstring? I'm not sure what this is supposed to be doing.

src/dashboard/src/components/advanced_search.py
Hwesta, line 86: Ahh, okay.  Can you add that as a comment?  Otherwise it looks like a subtle overwrite bug.

Hwesta:
:+1:

src/dashboard/src/components/advanced_search.py
mistydemeo, line 186: Added a docstring.

src/dashboard/src/components/advanced_search.py
mistydemeo, line 86: Added!

Hwesta:
:shipit:

src/dashboard/src/components/advanced_search.py
mistydemeo: Oh wow, this looks nifty. Maybe worth switching.

mistydemeo:
Added a commit to use dateutil to normalize dates.

Hwesta:
Speaking of dates & date search, how does this handle timezones?

mistydemeo:
It doesn't. The supplied data doesn't contain time zones, so the result is ¯_(ツ)_/¯

Hwesta:
:shipit:

PR #192, opened 2015-04-08
Event: linking_agent is an integer, not foreign key

mistydemeo:
linkingAgentIdentifier is an integer, which can be a reference either to a
value in the Agents table or to a user in the auth_user table. Rather than
enforce the foreign key at the Django level, just return it as an integer.

Fixes #8230.


Hwesta:
:+1:

PR #193, opened 2015-04-09
createMETS2: Fix Dspace dmdSec

Hwesta:
createDmdSecsFromCSVParsedMetadata expects dicts whose values are lists.


mistydemeo:
:crying_cat_face: :shipit:

PR #194, opened 2015-04-10
AIC: Continue if submission docs don't exist to be copied

Hwesta:
In an AIC, submission docs don't exist, so if that microservice fails it shouldn't fail the SIP.

I have two implementations here: copySubmissionDocs always returns success (which silences all failures), and continue if that MSCL fails (which shows as a fail for AICs even though that's expected). Thoughts on which is better, or if I should continue to do both?

refs #8258

### DIFF
--- a/src/MCPClient/lib/clientScripts/copySubmissionDocs.sh
+++ b/src/MCPClient/lib/clientScripts/copySubmissionDocs.sh
@@ -7,4 +7,4 @@
 # Copy submission documentation from the AIP back into the SIP
 subdir="$1/submissionDocumentation"
 mkdir -p $subdir
-cp -R "$1/$2/data/objects/submissionDocumentation" $subdir
+cp -R "$1/$2/data/objects/submissionDocumentation" $subdir || true


src/MCPClient/lib/clientScripts/copySubmissionDocs.sh
```diff
@@ -8,3 +8,4 @@
 subdir="$1/submissionDocumentation"
 mkdir -p $subdir
 cp -R "$1/$2/data/objects/submissionDocumentation" $subdir
```
mistydemeo: Nitpick, but maybe `cp -R "$1/$2/data/objects/submissionDocumentation" $subdir || true`?

mistydemeo:
Making `copySubmissionDocs` always exit 0 works for me!

src/MCPClient/lib/clientScripts/copySubmissionDocs.sh
mistydemeo: (The reason I wonder about that specifically is because I'm interested at some point in adopting the `set -e` behaviour [described here](http://redsymbol.net/articles/unofficial-bash-strict-mode/) - so all non-zero exit commands fail the script even if they're not the final line.)

Hwesta: Not a bash expert, but `exit 0` seems clearer that it's always returning 0 (and silencing the error from cp), rather than the `|| true`.

mistydemeo: Yep, but if we adopt `set -e` then the script fails if any individual line exits non-1 - failures get treated like unhandled exceptions in Python. We don't have to do that right away, but it seems like something I'd like to do in the future, and getting scripts ready for that where we touch them would make it easier to handle that in the future.

Hwesta: Ahh, that makes sense.  I'll update it.

PR #195, opened 2015-04-10
Index AIP: Merge index AIP files & index AIP

Hwesta:
Indexing AIP files before the AIP is stored causes the Archival Storage tab to error when the AIP is partially indexed but not stored. Move indexing AIP files to where the rest of the AIP is indexed (after storing) and merge with existing index AIP.

refs #7922


Hwesta:
This needs to be updated to handle changes in #196

mistydemeo:
:shipit:

PR #196, opened 2015-04-13
Index MODS identifiers in aips/aipfile

mistydemeo:
This data was previously only indexed in aips/aip. Even though we treat MODS identifiers as SIP-level, we need to index them here because the archival storage tab actually searches on the aips/aipfile index.

Fixes #8266.

### DIFF
--- a/src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
+++ b/src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
@@ -21,13 +21,19 @@
 # @subpackage archivematicaClientScript
 # @author Mike Cantelon <mike@artefactual.com>
 import ConfigParser
+from glob import glob
 import sys
 
 # archivematicaCommon
 from custom_handlers import get_script_logger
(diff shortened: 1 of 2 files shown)


Hwesta:
:+1:

src/MCPClient/lib/clientScripts/elasticSearchIndexProcessAIP.py
```diff
@@ -51,12 +57,20 @@
...
+            identifiers.extend(extract_identifiers_from_mods(mods))
+
         exitCode = elasticSearchFunctions.connect_and_index_files(
             'aips',
             'aipfile',
             uuid,
             pathToAIP,
-            sipName
+            sipName,
+            identifiers=identifiers
```
Hwesta, line 73: `identifiers` is the last param here, but the second-last in the definition.  sipName needs to be `sipName=sipName` or moved to the end.

mistydemeo, line 73: Fix in #197

PR #198, opened 2015-04-13
dev-helper: remove restart services

mistydemeo:
This also exists in the archivematica-devtools repository; suggest removing this extra copy to avoid maintaining two versions. If memory serves, the archivematica-devtools version is also slightly more up-to-date.

### DIFF
--- a/dev-helper
+++ b/dev-helper
@@ -150,38 +150,6 @@ function clear-storage() {
   fi
 }
 
-function restart() {
-  part="restart archivematica services"
-  echo -n "\"Would you like to ${part}?\" (y/N) "
-  read a
-  if [[ $a == "Y" || $a == "y" ]]; then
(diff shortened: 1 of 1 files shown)


Hwesta:
:+1: 

Alternately, have dev-helper call am restart-services.  Either works for me, since I use a fish function most of the time.

mistydemeo:
Oh yeah, I didn't think of that. Do we want to make archivematica-devtools a requirement of running dev-helper?

Hwesta:
Hmm, I don't really want AM to require devtools, but it's convenient that dev-helper restarts things to pick up all the updates that likely just happened, especially if you don't already know to do that.

Hwesta:
Since we're moving to ansible for all deployments, am-devtools is pretty easy to install (and dev-helper is mostly deprecated).  I vote this gets merged.

PR #199, opened 2015-04-13
Added processing directory usage page. refs #7921

mcantelon:
Adds an admin page where processing space usage is shown and these processing areas can be purged.

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -280,12 +280,28 @@ def storage(request):
     system_directory_description = 'Available storage'
     return render(request, 'administration/locations.html', locals())
 
+"""
+Usage summary page
+
+Return page summarizing storage usage
+
+:param HttpRequest request
+:returns HttpResponse
+"""
 def usage(request):
     usage_dirs = _usage_dirs()
 
     context = {'usage_dirs': usage_dirs}
     return render(request, 'administration/usage.html', context)
 
+"""
+Provide usage data
+
+Return maximum size and, optionally, current usage of a number of directories
+
+:param bool calculate_usage: whether or not to calculate usage 
+:returns OrderedDict: data structure defining directories/usage/size
+"""
 def _usage_dirs(calculate_usage=True):
     # Put spaces before directories contained by the spaces
     #
@@ -341,6 +357,12 @@ def _usage_dirs(calculate_usage=True):
 
     return dirs
 
+"""
+Check the size of the volume containing a given path
+
+:param string path: path to check
+:returns int: size in bytes
+"""
 def _usage_check_directory_volume_size(path):
     # Get volume size (in 512 byte blocks)
     try:
@@ -353,25 +375,49 @@ def _usage_check_directory_volume_size(path):
         size = usage_summary.split()[1]
 
         return int(size) * 512
-    except Exception, e:
-        logger.exception(str(e))
+    except OSError:
+        logger.exception('No such directory: {}'.format(path))
+        return 0
+    except subprocess.CalledProcessError:
+        logger.exception('Unable to determine size of {}.'.format(path))
         return 0
 
+"""
+Check the spaced used at a given path
+
+:param string path: path to check
+:returns int: usage in bytes
+"""
 def _usage_get_directory_used_bytes(path):
     """ Get total usage in bytes """
     try:
         output = subprocess.check_output(["du", "--bytes", "--summarize", path])
         return output.split("\t")[0]
-    except Exception, e:
-        logger.exception(str(e))
+    except OSError:
+        logger.exception('No such directory: {}'.format(path))
+        return 0
+    except subprocess.CalledProcessError:
+        logger.exception('Unable to determine usage of {}.'.format(path))
         return 0
 
+"""
+Confirmation context for emptying a directory
+
+:param HttpRequest request
+:returns RequestContext
+"""
 def clear_context(request, dir_id):
     usage_dirs = _usage_dirs(False)
     prompt = 'Clear ' + usage_dirs[dir_id]['description'] + '?'
     cancel_url = reverse("components.administration.views.usage")
     return RequestContext(request, {'action': 'Delete', 'prompt': prompt, 'cancel_url': cancel_url})
 
+"""
+Empty a directory
+
+:param HttpRequest request
+:returns HttpResponse
+"""
 @user_passes_test(lambda u: u.is_superuser, login_url='/forbidden/')
 @decorators.confirm_required('simple_confirm.html', clear_context)
 def usage_clear(request, dir_id):
@@ -403,9 +449,10 @@ def usage_clear(request, dir_id):
                     else:
                         shutil.rmtree(entry_path)
                 successes.append(directory)
-            except Exception, e:
-                logger.exception(str(e))
-                errors.append(str(e))
+            except OSError:
+                message = 'No such file or directory: {}'.format(directory)
+                logger.exception(message)
+                errors.append(message)
 
         # If any deletion attempts successed, summarize in flash message
         if len(successes):
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -279,29 +279,24 @@ def storage(request):
     system_directory_description = 'Available storage'
     return render(request, 'administration/locations.html', locals())
 
-"""
-Usage summary page
-
-Return page summarizing storage usage
-
-:param HttpRequest request
-:returns HttpResponse
-"""
 def usage(request):
+    """
+    Return page summarizing storage usage
+    """
     usage_dirs = _usage_dirs()
 
     context = {'usage_dirs': usage_dirs}
     return render(request, 'administration/usage.html', context)
 
-"""
-Provide usage data
-
-Return maximum size and, optionally, current usage of a number of directories
-
-:param bool calculate_usage: whether or not to calculate usage 
-:returns OrderedDict: data structure defining directories/usage/size
-"""
 def _usage_dirs(calculate_usage=True):
+    """
+    Provide usage data
+
+    Return maximum size and, optionally, current usage of a number of directories
+
+    :param bool calculate_usage: True if usage should be calculated.
+    :returns OrderedDict: Dict where key is a descriptive handle and value is a dict with the path, description, parent directory ID, and optionally size and usage.
+    """
     # Put spaces before directories contained by the spaces
     #
     # Description is optional, but either a path or a location purpose (used to
@@ -356,13 +351,13 @@ def _usage_dirs(calculate_usage=True):
 
     return dirs
 
-"""
-Check the size of the volume containing a given path
-
-:param string path: path to check
-:returns int: size in bytes
-"""
 def _usage_check_directory_volume_size(path):
+    """
+    Check the size of the volume containing a given path
+
+    :param string path: path to check
+    :returns int: size in bytes
+    """
     # Get volume size (in 512 byte blocks)
     try:
         output = subprocess.check_output(["df", path])
@@ -375,51 +370,48 @@ def _usage_check_directory_volume_size(path):
 
         return int(size) * 512
     except OSError:
-        logger.exception('No such directory: {}'.format(path))
+        logger.exception('No such directory: %s', path)
         return 0
     except subprocess.CalledProcessError:
-        logger.exception('Unable to determine size of {}.'.format(path))
+        logger.exception('Unable to determine size of %s', path)
         return 0
 
-"""
-Check the spaced used at a given path
-
-:param string path: path to check
-:returns int: usage in bytes
-"""
 def _usage_get_directory_used_bytes(path):
-    """ Get total usage in bytes """
+    """
+    Check the spaced used at a given path
+
+    :param string path: path to check
+    :returns: usage in bytes
+    """
     try:
         output = subprocess.check_output(["du", "--bytes", "--summarize", path])
         return output.split("\t")[0]
     except OSError:
-        logger.exception('No such directory: {}'.format(path))
+        logger.exception('No such directory: %s', path)
         return 0
     except subprocess.CalledProcessError:
-        logger.exception('Unable to determine usage of {}.'.format(path))
+        logger.exception('Unable to determine usage of %s.', path)
         return 0
 
-"""
-Confirmation context for emptying a directory
-
-:param HttpRequest request
-:returns RequestContext
-"""
 def clear_context(request, dir_id):
+    """
+    Confirmation context for emptying a directory
+
+    :param dir_id: Key for the directory in _usage_dirs
+    """
     usage_dirs = _usage_dirs(False)
     prompt = 'Clear ' + usage_dirs[dir_id]['description'] + '?'
     cancel_url = reverse("components.administration.views.usage")
     return RequestContext(request, {'action': 'Delete', 'prompt': prompt, 'cancel_url': cancel_url})
 
-"""
-Empty a directory
-
-:param HttpRequest request
-:returns HttpResponse
-"""
 @user_passes_test(lambda u: u.is_superuser, login_url='/forbidden/')
 @decorators.confirm_required('simple_confirm.html', clear_context)
 def usage_clear(request, dir_id):
+    """
+    Empty a directory
+
+    :param dir_id: Descriptive shorthand for the directory, key for _usage_dirs
+    """
     if request.method == 'POST':
         usage_dirs = _usage_dirs(False)
         dir_info = usage_dirs[dir_id]
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -16,7 +16,6 @@
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
 import collections
-import ConfigParser
 import logging
 import os
 import shutil
@@ -28,7 +27,9 @@ from django.contrib import messages
 from django.contrib.auth.decorators import user_passes_test
 from django.db.models import Max, Min
 from django.forms.models import modelformset_factory
+from django.http import Http404, HttpResponseNotAllowed, HttpResponseRedirect
 from django.shortcuts import redirect, render
+from django.template import RequestContext
 
 from main import forms
 from main import models
@@ -38,12 +39,10 @@ from components.administration.forms import ArchivesSpaceConfigForm
 from components.administration.forms import ArchivistsToolkitConfigForm
 from components.administration.forms import SettingsForm
 from components.administration.forms import StorageSettingsForm, ChecksumSettingsForm
-from components.administration.models import ArchivesSpaceConfig, ArchivistsToolkitConfig
 from components.administration.forms import TaxonomyTermForm
-from django.http import Http404, HttpResponseNotAllowed, HttpResponseRedirect
-from django.template import RequestContext
+from components.administration.models import ArchivesSpaceConfig, ArchivistsToolkitConfig
+import components.administration.views_processing as processing_views
 import components.decorators as decorators
-from django.template import RequestContext
 import components.helpers as helpers
 import storageService as storage_service
 


src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+def _usage_dirs(calculate_usage = True):
+    # Get shared directory path
+    clientConfigFilePath = '/etc/archivematica/MCPClient/clientConfig.conf'
+    config = ConfigParser.SafeConfigParser()
+    config.read(clientConfigFilePath)
+
+    shared_dir = config.get('MCPClient', 'sharedDirectoryMounted')
+
+    # Get DIP uploads directory
+    dip_upload_dir = os.path.join(shared_dir, 'watchedDirectories/uploadDIP')
```
Hwesta: This should be uploadedDIPs.  uploadDIP is the watched directory where we ask where the DIP should be uploaded to.  uploadedDIPs is where DIPs that have already been uploaded go to.  Also, splitting the path components given to os.path.join is preferred.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
     system_directory_description = 'Available storage'
     return render(request, 'administration/locations.html', locals())
 
+def usage(request):
+    usage_dirs = _usage_dirs()
+
+    return render(request, 'administration/usage.html', locals())
+
+def _usage_dirs(calculate_usage = True):
```
Hwesta: No spaces around default params in a function definition please.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+def _usage_check_directory_volume_size(path):
+    # Get volume size (in 512 byte blocks)
+    df = subprocess.Popen(["df", path], stdout=subprocess.PIPE)
+    output = df.communicate()[0]
+    device, size, used, available, percent, mountpoint = output.split("\n")[1].split()
+    return int(size) * 512
+
+def _usage_check_directory_used(path):
+    # Get total usage in bytes
+    du = subprocess.Popen(["du", "-b", "-c", path], stdout=subprocess.PIPE)
```
Hwesta: `du --bytes --summarize` returns a one-line summary of the folder, so you don't have to pipe it to tail.  Also, I prefer long options in code so it's clearer what's happening.

src/dashboard/src/components/administration/views.py
```diff
@@ -33,7 +37,7 @@
 from components.administration.forms import StorageSettingsForm
 from components.administration.models import ArchivistsToolkitConfig
 from components.administration.forms import TaxonomyTermForm
-from django.http import HttpResponseRedirect
+from django.http import Http404, HttpResponseBadRequest, HttpResponseRedirect
```
Hwesta: Nitpick, but don't add trailing whitespace?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+        },
+        'backlog': _usage_location_info('BL', shared_size)
+    }
+
+def _usage_check_directory_volume_size(path):
+    # Get volume size (in 512 byte blocks)
+    df = subprocess.Popen(["df", path], stdout=subprocess.PIPE)
+    output = df.communicate()[0]
+    device, size, used, available, percent, mountpoint = output.split("\n")[1].split()
+    return int(size) * 512
```
Hwesta: [`subprocess.check_output`](https://docs.python.org/2/library/subprocess.html#subprocess.check_output) may be useful here, especially since you can catch the `CalledProcessError` and return a default if the command doesn't run properly (eg path doesn't exist).  Same advice for `_usage_check_directory_used`

src/dashboard/src/templates/administration/usage.html
```diff
@@ -0,0 +1,49 @@
...
+        <tr>
+          <td>{{ dir.description }}</td>
+          <td>{{ dir.path }}</td>
+          <td>{{ dir.used|filesizeformat }}
+            ({% if dir.used %}{% widthratio dir.used dir.size 100 %}{% else %}0{% endif %}%)
+          </td>
+          <td>{{ dir.size|filesizeformat }}</td>
+          <td>
+            {% if dir.description %}
+            <form method='POST' action='/administration/usage/clear/{{ id }}/'>
```
Hwesta: Use `{% url %}` for this instead of hard coding the path.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+            'size': shared_size,
+            'used': shared_used
+        },
+        'dips': {
+            'description': 'DIP uploads',
+            'path': dip_upload_dir,
+            'size': shared_size,
+            'used': dip_upload_used
+        },
+        'backlog': _usage_location_info('BL', shared_size)
```
Hwesta: It would be nice if it was easier to add more directories to this function.  Maybe have a dict (or an OrderedDict?) with the static info for all the directories you want to display, and iterate over them and populate the size/used if calculate_usage=True?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+
+    # Parse type total from command output
+    output = subprocess.check_output(["tail", "-n", "1"], stdin=du.stdout)
+    return output.split("\t")[0]
+
+def _usage_location_info(purpose, space_size):
+    # Get currently processing location
+    locations = storage_service.get_location(purpose=purpose)
+    location = locations[0]
+    if location['quota'] == None:
```
Hwesta: Nitpick, but `is None`

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+                    else:
+                        shutil.rmtree(entry_path)
+                except Exception, e:
+                    messages.error(request, str(e))
+                    return redirect('components.administration.views.usage')
+
+        messages.info(request, 'Cleared.')
+        return redirect('components.administration.views.usage')
+    else:
+        return HttpResponseBadRequest()
```
Hwesta: I think you want [HttpResponseNotAllowed](https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.HttpResponseNotAllowed)

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+        'purpose': purpose,
+        'description': purpose_labels[purpose],
+        'path': location['path'],
+        'size': location_size,
+        'used': used}
+
+def usage_clear(request, dir_id):
+    if request.method == 'POST':
+        usage_dirs = _usage_dirs(False)
+        dir = usage_dirs[dir_id]
```
Hwesta: Don't shadow builtins like `dir`

src/dashboard/src/components/administration/urls.py
```diff
@@ -28,6 +28,8 @@
     (r'dips/contentdm/$', 'contentdm_dips'),
     (r'sources/$', 'sources'),
     (r'storage/$', 'storage'),
+    (r'usage/$', 'usage'),
+    (r'usage/clear/(?P<dir_id>\w+)/$', 'usage_clear'),
```
Hwesta: Should this be in api/ instead of administration?  Do we care about login-required vs api-key?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+
+        # Prevent shared directory from being cleared
+        if dir_id == 'shared' or not dir:
+            raise Http404
+
+        # Determine which directories need to be cleared
+        if 'purpose' in dir and dir['purpose'] == 'BL':
```
Hwesta: `if dir_info.get('purpose') == 'BL':`?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+            # Get location data
+            locations = storage_service.get_location(purpose=dir['purpose'])
```
Hwesta: Isn't the path already in usage_dirs?  Why re-fetch it?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+            for entry in os.listdir(dir):
+                entry_path = os.path.join(dir, entry)
+                try:
+                    if os.path.isfile(entry_path):
+                        os.unlink(entry_path)
+                    else:
+                        shutil.rmtree(entry_path)
+                except Exception, e:
+                    messages.error(request, str(e))
+                    return redirect('components.administration.views.usage')
```
Hwesta: Do we want to bail if _any_ file/folder fails?  Or should we just log it and continue?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +227,127 @@ def storage(request):
...
+
+        messages.info(request, 'Cleared.')
```
Hwesta: Would be nice if this displayed the path(s) that were deleted, for better usability.

Hwesta:
Do we want to be deleting the backlog?  I'm not sure that should be so easily cleared, especially since it may be the only copy of archival files halfway through being processed.

We probably also want to display info for and clear the failed, rejected, tmp and staging dirs.

src/dashboard/src/components/administration/urls.py
mcantelon: Maybe it should be available both in the UI and as an API? It's in the UI so folks using AM in the cloud can do some basic housekeeping without SSH access, but it definitely would make sense to have accessible via the API for automation.

src/dashboard/src/components/administration/views.py
mcantelon: So, for the splitting, something like this?

```
path_parts = [os.path.sep] + shared_dir.split('/') + 'watchedDirectories/uploadDIP'.split('/')
dip_upload_dir = os.path.join(*path_parts)
```

Hwesta: No, I meant that having a literal `/` in a parameter to os.path.join kinda defeats the purpose, since os.path.join behaves differently depending on the OS and `/` pins it to a particular OS.  Should be `os.path.join(shared_dir, 'watchedDirectories', 'uploadedDIPs')`.  Here I'm also assuming shared_dir is constructed with os.path.join (or read from a config file with an OS-correct value).

mcantelon: Ah, makes sense!

mcantelon:
I think I've implemented all recommendations (except adding a REST API interface for clearing which I can do once the admin UI way seems good to you).

mcantelon:
I also did some other things to get rid of hardcoding so specifying what can be cleared is fully configuration-driven. We could conceivably change things so this is read from an /etc file (YAML or something) to give sys admins the ability to customize their installation in this respect.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+
+def _usage_check_directory_used(path):
+    # Get total usage in bytes
+    try:
+        output = subprocess.check_output(["du", "--bytes", "--summarize", path])
+        return output.split("\t")[0]
+    except subprocess.CalledProcessError:
+        return 0
+
+def _usage_shared_dir_path():
```
Hwesta: Remove this and replace with a call to `helpers.get_client_config_value`?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+    # Resolve location paths and make relative paths absolute
+    for directory in dirs:
+        if dirs[directory].get('location_purpose') is not None:
+            # If a location, determine path
+            dirs[directory]['path'] = _usage_location_path(dirs[directory]['location_purpose'])
+        elif dirs[directory].get('contained_by') is not None:
+            # If contained, make path absolute
+            space = dirs[directory]['contained_by']
+            absolute_path = os.path.join(dirs[space]['path'], dirs[directory]['path'])
+            dirs[directory]['path'] = absolute_path
```
Hwesta: More stylistic, but why do this in a loop, instead of in the definition? The shared directory path should be the same for all of them, so another option is to define a shared_dir variable, and os.path.join it to everything.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+        })
+    )
+
+    dirs = collections.OrderedDict(dir_defs)
+
+    # Resolve location paths and make relative paths absolute
+    for directory in dirs:
+        if dirs[directory].get('location_purpose') is not None:
+            # If a location, determine path
+            dirs[directory]['path'] = _usage_location_path(dirs[directory]['location_purpose'])
```
Hwesta: What's the use case for this?  None of the above directories have 'location_purpose' defined.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+        elif dirs[directory].get('contained_by') is not None:
+            # If contained, make path absolute
+            space = dirs[directory]['contained_by']
+            absolute_path = os.path.join(dirs[space]['path'], dirs[directory]['path'])
+            dirs[directory]['path'] = absolute_path
+
+    # Calculate usage/size, if requested
+    if calculate_usage:
+        for directory in dirs:
```
Hwesta: You may find [`.iteritems()`](https://docs.python.org/2/library/stdtypes.html#dict.iteritems) useful instead of `dirs[directory]`  Modifying the value returned by the iterator modifies it in the dict.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+            if dirs[directory].get('contained_by') is None:
```
Hwesta: Do you care if it's specifically None?  Or can you do `if 'contained_by' in dirs[directory]`?  Same advice applies elsewhere.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+
+    return dirs
+
+def _usage_check_directory_volume_size(path):
+    # Get volume size (in 512 byte blocks)
+    try:
+        output = subprocess.check_output(["df", path])
+        device, size, used, available, percent, mountpoint = output.split("\n")[1].split()
+        return int(size) * 512
+    except subprocess.CalledProcessError:
```
Hwesta, line 375: Can this also generate other types of exceptions?  Do we want to log something if an exception is generated?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+        return 0
+
+def _usage_check_directory_used(path):
```
Hwesta: Python convention uses [docstrings](https://www.python.org/dev/peps/pep-0257/) to summarize what a function does at a high level, and comments for descriptions inside a function.  Since this looks like a summary, should be triple quoted.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+                dirs[directory]['size'] = dirs[space]['size']
+                dirs[directory]['used'] = _usage_check_directory_used(dirs[directory]['path'])
+
+    return dirs
+
+def _usage_check_directory_volume_size(path):
+    # Get volume size (in 512 byte blocks)
+    try:
+        output = subprocess.check_output(["df", path])
+        device, size, used, available, percent, mountpoint = output.split("\n")[1].split()
```
Hwesta: Python convention uses `_` to indicated [a variable we don't care about](http://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python).  If that gets ugly, you could also index into the list for the value you want (and use a comment to explain why).  Or leave it the way it is as a form of explanation.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+    return render(request, 'administration/usage.html', locals())
+
+def _usage_dirs(calculate_usage=True):
+    # Put spaces before directories contained by the spaces
+    #
+    # Description is optional, but either a path or a location purpose (used to
+    # look up the path) should be specified
+    #
+    # If only certain sudirectories within a path should be deleted, set
+    # 'subdirectories' to a list of them
```
Hwesta: Is this still true?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
     system_directory_description = 'Available storage'
     return render(request, 'administration/locations.html', locals())
 
+def usage(request):
+    usage_dirs = _usage_dirs()
+
+    return render(request, 'administration/usage.html', locals())
```
Hwesta: I'd like to start moving away from using `locals()` as it's bad security, and instead of explicitly name the variables passed to the template. Can you use `{'usage_dirs': usage_dirs}`?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+        # Prevent shared directory from being cleared
+        if dir_id == 'shared' or not dir_info:
+            raise Http404
+
+        # Determine if specific subdirectories need to be cleared, rather than
+        # whole directory
+        if dir_info.get('subdirectories') is not None:
+            dirs_to_empty = []
+            for subdirectory in dir_info['subdirectories']:
+                dirs_to_empty.append(os.path.join(dir_info['path'], subdirectory))
```
Hwesta: Have you discovered the magic of [list comprehensions](https://docs.python.org/2/tutorial/datastructures.html#list-comprehensions)?  They're fantastic.  This could be `dirs_to_empty = [os.path.join(dir_info['path'], subdir) for subdir in dir_info['subdirectories']]`

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+            try:
+                for entry in os.listdir(directory):
+                    entry_path = os.path.join(directory, entry)
+                    if os.path.isfile(entry_path):
+                        os.unlink(entry_path)
+                    else:
+                        shutil.rmtree(entry_path)
+                successes.append(directory)
+            except Exception, e:
+                errors.append(str(e))
```
Hwesta: Do we want to log these as well?  logger.exception or logger.foo(..., exc_info=True) may be useful.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
     system_directory_description = 'Available storage'
     return render(request, 'administration/locations.html', locals())
 
+def usage(request):
+    usage_dirs = _usage_dirs()
+
+    return render(request, 'administration/usage.html', locals())
+
+def _usage_dirs(calculate_usage=True):
```
Hwesta, line 291: Nice touch putting the leading _ on all the internal-only functions.

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,160 @@ def storage(request):
...
+        usage_dirs = _usage_dirs(False)
+        dir_info = usage_dirs[dir_id]
+
+        # Prevent shared directory from being cleared
+        if dir_id == 'shared' or not dir_info:
+            raise Http404
+
+        # Determine if specific subdirectories need to be cleared, rather than
+        # whole directory
+        if 'subdirectories' in dir_info:
```
mcantelon, line 425: This functionality was originally added to accommodate Transfer Backlog clearing where we needed to clear the contents of two subdirectories, rather than the whole directory's contents. I'm wondering if I should leave this in in case it's useful or if I should remove it?

src/dashboard/src/components/administration/views.py
```diff
@@ -223,6 +229,146 @@ def storage(request):
...
+            'path': 'tmp',
+            'contained_by': 'shared'
+        })
+    )
+
+    dirs = collections.OrderedDict(dir_defs)
+
+    # Resolve location paths and make relative paths absolute
+    for _, dir_spec in dirs.iteritems():
+        if 'contained_by' in dir_spec:
```
mcantelon, line 337: I do the contained_by thing, rather than always path.join'ing the shared directory, in case in the future we want to support having things on multiple volumes, etc. Maybe it's overkill for now?

src/dashboard/src/components/administration/views.py
mcantelon, line 425: (Probably removing it makes sense as it's not hard to re-add.)

mcantelon:
Merged.

src/dashboard/src/components/administration/views.py
```diff
@@ -184,6 +190,146 @@ def storage(request):
...
+        output = subprocess.check_output(["df", path])
+
+        # Second line returns disk usage-related values
+        usage_summary = output.split("\n")[1]
+
+        # Split value by whitespace and size (in blocks)
+        size = usage_summary.split()[1]
+
+        return int(size) * 512
+    except Exception, e:
```
Hwesta: The modern syntax for providing a name for an exception is [`except Exception as e`](http://stackoverflow.com/questions/5119751/in-python-whats-the-difference-between-except-exception-as-e-and-except-exc).  Also, catching `Exception` is generally a [bad practice](http://ischenko.blogspot.ca/2005/01/exception-based-code-antipatterns.html), as it silences all errors, not just the ones you're expecting.

I was thinking something like

``` python
except (ValueError, IndexError, subprocess.CalledProcessError):  # Known errors that could be raised
    logger.exception('Error checking directory size for %s', path)  # logger.exception already logs the exception raised
    return 0
```

src/dashboard/src/components/administration/views.py
Hwesta, line 337: Seems reasonable

src/dashboard/src/components/administration/views.py
```diff
@@ -184,6 +190,146 @@ def storage(request):
...
+        logger.exception(str(e))
+        return 0
+
+def _usage_get_directory_used_bytes(path):
+    """ Get total usage in bytes """
+    try:
+        output = subprocess.check_output(["du", "--bytes", "--summarize", path])
+        return output.split("\t")[0]
+    except Exception, e:
```
Hwesta: Same comment here as above.

src/dashboard/src/components/administration/urls.py
Hwesta: I agree - being able to do automation would be useful, but do we want to give remote access to deleting things?  Especially for uploadedDIPs, which may contain DIPs that (despite the name :P) haven't actually been sent elsewhere and are awaiting manual download.

Hwesta:
Can you update all your `except Exception, e: logger.exception(str(e))` to name the Exception(s) that are being caught, and provide a useful logging message? logger.exception already provides the exception it caught along with a traceback, so a human-useful error message is more appropriate.  Also, logger.exception always generates an ERROR level message - a WARNING level might be better in some cases (`logger.warning('message', exc_info=True)`) - your call.

Also, can you add docstrings to your functions? For most of them a one-liner summary is fine, but for the longer ones an explanation of what the function is trying to accomplish, restrictions on the params and return value is very useful when reading later.  We're using Sphinx's reST style, here's a sample:

``` python
"""
Short summary

Optional longer summary

:param optional-type-of-param name-of-param: Description of what param does
:param str path: Path to get the size of
:param flags: Bool for something in this function
:returns: Size of the path in bytes
"""
```

mcantelon:
Okay, I've added docstrings and made the exception handling less generic.

PR #200, opened 2015-04-16
Create new logging location before running django

jhsimpson:
In the debian/postinst script of the dashboard, this change moves the creation of the new logging directory before any manage.py commands are run.


mistydemeo:
Would you mind reopening this based on qa/1.x? This pulled in additional commits, since it contains a lot of stuff that's not in stable/1.3.x.

jhsimpson:
ok, I will redo the PR on qa/1.x

PR #201, opened 2015-04-16
create new logging directory before running syncdb

jhsimpson:
refs #8267

Create the new logging directory before issuing any manage.py
commands in the debian/postinst script, as django will try to log
to the new logging directory, and it needs to exist for that to work.


mistydemeo:
:+1:

PR #202, opened 2015-04-16
Add Siegfried as an ID tool for testing purposes

mistydemeo:
Tool source can be seen in artefactual/archivematica-fpr-tools#3.


jhsimpson:
I think this PR should also include the change to MCPClient/debian/control that makes siegfried an os dependency?

mistydemeo:
Added!

PR #203, opened 2015-04-16
Archival storage: handle items with no 'fields' key

mistydemeo:
This was reported in #8290, where a few items in the aipfile index are empty. Because they have no `fields` key, `search_augment_file_results` raised a KeyError when trying to access their contents.

``` python
{u'_id': u'AUy_JM2bZTVg0eDcbrB9',
 u'_index': u'aips',
 u'_score': 1.0,
 u'_type': u'aipfile'},
```


Hwesta:
:+1:

PR #204, opened 2015-04-16
Archival storage: skip empty items

mistydemeo:
After further testing of #203, I realized that archival storage couldn't properly operate on empty items - so it's better just to skip them.


Hwesta:
:+1:

PR #206, opened 2015-04-17
createMETS2: pass correct identifier to createDigiprovMDAgents

mistydemeo:
Fixes #8293.


Hwesta:
Since this is the second oops-wrong-id PR, have you searched createMETS(2) for more bugs like this?

(Using `id` makes me sad, but I understand that it's a much larger change to fix that)

:+1:

mistydemeo:
Looks like there was one other occurrence in the same function, so updated.

Hwesta:
:+1:

PR #207, opened 2015-04-17
Backlog search: run status query as a filter

mistydemeo:
When evaluating a bool query, Elasticsearch may ignore "should" values if a "must" or "must not" clause has been specified. Since we were previously specifying "status: backlog" as a "must" clause to the query, this meant that the actual search preferences were being ignored.

Fixes #8292.


mistydemeo:
Relevant part of the bool query documentation: http://www.elastic.co/guide/en/elasticsearch/reference/1.3/query-dsl-bool-query.html

Hwesta:
:+1:

PR #209, opened 2015-04-18
Copy preconfigured choice XML to the DIP

mistydemeo:
The DIP chain now has one preconfigured choice (DIP store location) and may gain more in the future.

Fixes #7321.


Hwesta:
Do any DIP types care if they have the processingMCP in the DIP? Is it getting sent to the access system?

mistydemeo:
Hm, depends on the script. It looks like `upload-qubit` would receive the processingMCP.xml (since the entire DIP is uploaded, not just the objects directory), while `upload-archivistsToolkit` and `upload-archivesspace` wouldn't. Not sure if that would mess up the access system, though I suspect not. I'll ask the AtoM people.

mistydemeo:
AtoM people say this shouldn't be a problem.

Hwesta:
:+1:

PR #210, opened 2015-04-20
createMETS2: skip metadata.csv values that aren't UTF-8

mistydemeo:
Previously, any non-ASCII/non-UTF-8 value would cause the entire METS generation process to fail and kill the AIP. This will skip invalid values.

Fixes #7277.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -219,7 +219,10 @@ def createDmdSecsFromCSVParsedMetadata(metadata):
             if match:
                 key, = match.groups()
             for v in value:
-                etree.SubElement(dc, elem_namespace + key).text = v.decode('utf-8')
+                try:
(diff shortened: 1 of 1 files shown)


Hwesta:
I like being able to skip-but-notify errors, but we'd discussed (planned on?) making a failed METS creation fail the SIP, on the premise that if the METS is bad we don't want the AIP.  This would fail the SIP for a skipped error.  createMETS2 doesn't have an easy way to mark fail-errors from warning-errors.

mistydemeo:
In this case the METS is fine, it's just missing some of the CSV metadata. I think that counts as "good METS" rather than "bad METS"!

Hwesta:
I agree, but by incrementing `globalErrorCount` it will exit an error case, which looks the same as exiting an error because everything failed.  I'm worried this will interact poorly with making an error case from createMETS2 fail the SIP (which is generally a good idea).

mistydemeo:
Hmm, right. I'll disable the error reporting since we don't really do that elsewhere.

Hwesta:
:+1:

PR #211, opened 2015-04-20
Don't move zipped bags to processing directory

mistydemeo:
Zipped bags are already extracted to the processing directory; sending zipped bag transfers to this chainlink results in an attempt to move them to the location at which they are already located, failing the transfer.

Fixes #8318


Hwesta:
:+1:

PR #212, opened 2015-04-20
Search: Handle empty searches better

Hwesta:
Set a default value for 'type' in search queries to match what's displayed, and handle '' and '*' as empty search queries for type=term.


mistydemeo:
Works! :+1: :shipit:

PR #213, opened 2015-04-21
Archival Storage: Fix pagination

Hwesta:
Fix pagination of AIPs and AIP files using Django's Paginator.  To avoid fetching all results (to get correct pagination information), use LazyPagedSequence that fetches from ElasticSearch as needed.

Other cleanup:
- Return deep copy of MATCH_ALL_QUERY so changes don't persist across requests
- Update search_raw_wrapper to search_all_results and rename parameters to be a drop-in replacement for conn.search.  Since this didn't end up being needed I may remove it
- Use _pager.html for paging in archival storage, instead of a custom one
- Submit archival storage search on Enter
- Remove unused templates


src/dashboard/src/components/archival_storage/views.py
```diff
@@ -538,7 +555,7 @@ def list_display(request):
 
             try:
                 size = '{0:.2f} MB'.format(float(aip['size']))
-            except TypeError, ValueError:
+            except (TypeError, ValueError):
```
mistydemeo, line 558: 😱

mistydemeo:
:+1: 

Should I go ahead and publish LazyPagedSequence to pypi so we can use it here and in the pair matcher?

Hwesta:
Yes, LazyPagedSequence being in PyPI would be great.

mistydemeo:
:+1:

PR #214, opened 2015-04-23
Archival storage: avoid searching nested date fields as strings

mistydemeo:
Searching nested documents using wildcard field names and the query_string query type can fail in unexpected ways when the nested document contains mixed types. For example, if the nested document contains both strings and dates, a search for "foo" will fail because Elasticsearch will attempt to parse it into a date for use with the date fields.

Since the "phrase" search in archival storage is only meant to search string fields, this adds a check to query_string generation to search only string-type nested document fields.

Fixes #8338.

### DIFF
--- a/src/dashboard/src/components/advanced_search.py
+++ b/src/dashboard/src/components/advanced_search.py
@@ -101,7 +101,7 @@ def extract_url_search_params_from_request(request):
         pass
     return search_params
 
-def assemble_query(queries, ops, fields, types, **kwargs):
+def assemble_query(queries, ops, fields, types, search_index=None, doc_type=None, **kwargs):
     must_haves     = kwargs.get('must_haves', [])
     filters        = kwargs.get('filters', {})
     should_haves   = []
@@ -110,7 +110,7 @@ def assemble_query(queries, ops, fields, types, **kwargs):
 
     for query in queries:
         if queries[index] != '':
-            clause = query_clause(index, queries, ops, fields, types)
+            clause = query_clause(index, queries, ops, fields, types, search_index=search_index, doc_type=doc_type)
             if clause:
                 if ops[index] == 'not':
                     must_not_haves.append(clause)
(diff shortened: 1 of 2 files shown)


src/dashboard/src/components/advanced_search.py
```diff
@@ -179,7 +179,42 @@ def _normalize_date(date):
...
+    """
+    Given search fields which search nested documents with wildcards (such as "transferMetadata.*"), returns a list of subfields filtered to contain only string-type fields.
+
+    When searching all fields of nested documents of mixed types using query_string queries, query_string queries may fail because the way the query string is interpreted depends on the type of the field being searched.
+    For example, given a nested document containing a string field and a date field, a query_string of "foo" would fail when Elasticsearch attempts to parse it as a date to match it against the date field.
+    This function uses the actual current mapping, so it supports automatically-mapped fields.
+
+    Sample input and output, given a nested document containing three fields, "Bagging-Date" (date), "Bag-Name" (string), and "Bag-Type" (string):
+    ["transferMetadata.*"] #=> ["transferMetadata.Bag-Name", "transferMetadata.Bag-Type"]
+    """
```
Hwesta, line 198: Sphinx-style docs of params?  Specifically, is search_fields a list or something else? Code implies it's a list, but nice to know explicitly it's not supposed to be a str.

src/dashboard/src/components/advanced_search.py
```diff
@@ -179,7 +179,42 @@ def _normalize_date(date):
...
+        return search_fields
+
+    new_fields = []
+    for field in search_fields:
+        # Not a wildcard nested document search, so just add to the list as-is
+        if not field.endswith('.*'):
+            new_fields.append(field)
+            continue
+        try:
+            field_name = field.split('.')[0]
```
Hwesta: What if the field is `foo.bar.*`? Should you use `rsplit`?

mistydemeo:
Updated docstring, switched to use `.rsplit('.', 1)`.

Hwesta:
:+1:

PR #216, opened 2015-05-01
Advanced search: fix row generation for rows 2+

mistydemeo:
Commit fe101c585d89d9b2182e7eb5e2d6e3f39bd36970 fixed a bug where the initial value for the "keyword" dropdown, "term", would be sent as an empty string. However, the bug recurs for any subsequent rows that are added using the "add new" button. This fixes this issue by adding a row template as a parameter to the advanced search javascript, and creating both initial and subsequent rows from that template rather than naively.

Refs #8292.


Hwesta:
:+1:

PR #218, opened 2015-05-04
WIP contribution guidelines

mistydemeo:
These are very, very WIP (and possibly controversial).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,3 +40,17 @@ When in doubt: document!
 Python's [PEP 257](https://www.python.org/dev/peps/pep-0257/) document provides a useful guideline for docstring style.
 Generally, prefer using [Sphinx-compatible docstrings](http://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions).
 More [examples](http://sphinx-doc.org/domains.html#info-field-lists) and [attributes to use](http://sphinx-doc.org/domains.html#the-python-domain) can be found on the Sphinx website.
+
+## Tests
+
+New code should also have unit tests.
+Tests are written in [unittest](https://docs.python.org/2/library/unittest.html) style and run with [py.test](http://pytest.org).
+For tests requiring the Django ORM, we use the Django-provided [TestCase](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.TestCase), which extends `unittest.TestCase`.
+
+Tests are found in the `tests` directory, a sibling of the directory containing the code.
+`test_foo.py` contains tests for `foo.py`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,56 @@
...
+
+from main.models import File
+
+f = File.objects.get(uuid="something")
+
+# Bad
+
+import main.models
+
+f = models.File.objects.get(uuid="something")
```
Hwesta: Interesting - I prefer the opposite. I'd rather know when I'm looking at a line if it's from another module (especially stdlib) or defined in this file.  It also makes it easier to use more things from the same module.

I care less for Django models though, probably because you're more likely to import multiple models.py and doing `from main import models; from fpr import models as fpr_models` is a pain.

CONTRIBUTING.md
```diff
@@ -0,0 +1,56 @@
...
+When working in sections of code that don't conform to PEP8, it's okay to relax a few PEP8 rules in order to match existing code.
+In particular:
+
+* When modifying an existing function which uses camelCase variables or parameters, it's okay to make your new variables/parameters camelCase to match.
+* When adding new functions to a module or class that uses camelCase naming, it's okay to make your new function and its parameters camelCase to match.
+  Try to use snake_case internally, however.
+
+## Documentation
+
+New classes and functions should generally be documented using [docstrings](https://en.wikipedia.org/wiki/Docstring#Python); these help in providing clarity, and can also be used to generate API documentation later.
```
Hwesta, line 37: Link to [PEP257](https://www.python.org/dev/peps/pep-0257/) instead/in addition?

mistydemeo, line 37: I didn't want to overload with detail! Probably better to be thorough though?

CONTRIBUTING.md
mistydemeo: Haha, I expected this to be a point of contention. I really like being able to glance at the top of a file and see where names come from, but I totally get that perspective too.

CONTRIBUTING.md
Hwesta, line 37: I'm likely to cite that PEP when I comment on docstrings, so I think it's fair to link it initially.  

...am I the only person that reads PEPs for fun?

mistydemeo, line 37: Linked to PEP257.

Hwesta:
Added a section on tests, and removed the bikesheding issue.

Hwesta:
:+1: 

Maybe we can combine this with code_review.md later.

Hwesta:
This contained commits 67f1fa0f13ae64b3668a2c6c8d17669c48acae6a through d1eca8befe2e407b1c19bd283b9f75b93cea358f

PR #219, opened 2015-05-05
mysql_dev: touch a file to indicate completion

mistydemeo:
This assists the Ansible deployment scripts in determining if mysql_dev has already been run, since it can't be safely rerun on the same database.


jhsimpson:
I think this is a good idea, and could be included in 1.4.0

PR #220, opened 2015-05-05
CreateMETS2: Don't create Derivations if no Event

Hwesta:
Events are only created for preservation Derivations, not access or thumbnail. Do not create PREMIS relationships in the METS file if there is no Event.

ref #8382


mistydemeo:
:+1:

PR #224, opened 2015-05-07
AIC: Pass full search results

Hwesta:
AIC creation form needs the whole list of AIP names and UUIDs that match the search, not just this page's results. The search only returned the UUID. Modify search to return the UUID and name. Since we're searching aip/aipfile (not aip/aips), dedupe the results and add a count of the number of files in the AIP.

An alternate implementation is to call `search_augment_aip_results` on the entire results list, but that seemed inefficient, and the aipfiles -> aip conversion still needed to happen.

Also:
- Change to use logger instead of logging
- Add namespacing to create AIC METS
- Remove now-unused facets from assemble_query


mistydemeo:
:+1:

PR #226, opened 2015-05-07
taskStandard: stdout/stderr values are already normalized

mistydemeo:
Due to the changes to ReplacementDict, the stdout/stderr values are now being expanded before being passed to taskStandard.

Fortunately, this allows for a general simplification/cleanup of the validation code in taskStandard; it can now perform a straightforward check whether the files are writable, instead of relying on heuristics based on the presence of a variable string in the text.

Fixes #8404.

### DIFF
--- a/src/MCPServer/lib/taskStandard.py
+++ b/src/MCPServer/lib/taskStandard.py
@@ -27,6 +27,7 @@ import gearman
 import cPickle
 import datetime
 import archivematicaMCP
+import os
 import sys
 sys.path.append("/usr/lib/archivematica/archivematicaCommon")
 from fileOperations import writeToFile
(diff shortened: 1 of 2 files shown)


src/MCPServer/lib/taskStandard.py
```diff
@@ -110,40 +111,54 @@ def check_request_status(self, job_request):
...
-                ret = "<^Not allowed to write to file^> " + ret
-        return ret
+    def outputFileIsWritable(self, fileName):
+        """
+        Validates whether a given file is writeable or, if the file does not exist, whether its parent directory is writeable.
+        """
+        if os.path.exists(fileName):
+            target = fileName
+        else:
+            target = os.path.pardir(fileName)
```
Hwesta: `os.path.pardir` is not a function. Did you mean `os.path.dirname(fileName)`?

mistydemeo: Good catch, thanks. I trusted SublimePythonIDE's autocomplete too much.

src/MCPServer/lib/taskStandard.py
```diff
@@ -110,40 +111,54 @@ def check_request_status(self, job_request):
...
 
         if a:
-            self.stdError = "Failed to write to file{" + standardOut.encode('utf-8') + "}\r\n" + self.results["stdOut"]
+            self.stdError = "Failed to write to file{" + self.standardOutputFile.encode('utf-8') + "}\r\n" + self.results["stdOut"]
         if b:
-            self.stdError = "Failed to write to file{" + standardError.encode('utf-8') + "}\r\n" + self.results["stdError"]
+            self.stdError = "Failed to write to file{" + self.standardOutputFile.encode('utf-8') + "}\r\n" + self.results["stdError"]
         if  self.results['exitCode']:
             return self.results['exitCode']
         return a + b
```
Hwesta: Maybe rename `a` and `b` to something with actual meaning?

Hwesta:
:+1:

src/MCPServer/lib/taskStandard.py
```diff
@@ -110,40 +111,54 @@ def check_request_status(self, job_request):
...
 
         if self.outputLock != None:
             self.outputLock.release()
 
         if a:
-            self.stdError = "Failed to write to file{" + standardOut.encode('utf-8') + "}\r\n" + self.results["stdOut"]
+            self.stdError = "Failed to write to file{" + self.standardOutputFile.encode('utf-8') + "}\r\n" + self.results["stdOut"]
         if b:
-            self.stdError = "Failed to write to file{" + standardError.encode('utf-8') + "}\r\n" + self.results["stdError"]
+            self.stdError = "Failed to write to file{" + self.standardOutputFile.encode('utf-8') + "}\r\n" + self.results["stdError"]
```
Hwesta: This should be `self.standardErrorFile`

Also, do we want to be writing over whatever was in self.stdError if both a and b are true?

mistydemeo:
Updated the variable names, tested this has fixed the logs.

Hwesta:
:+1:

PR #227, opened 2015-05-08
METSRights: fix model attribute name

mistydemeo:


jhsimpson:
tested  and works

:shipit:

PR #228, opened 2015-05-08
do not delete MicroServiceChainLinks

jhsimpson:
refs #8414
refs #8039
refs #6488
refs #7239
refs #7922
refs #8019

updated 4 mysql_dev*sql scripts, removing lines that
delete from MicroServiceChainLinks.  These delete statements
can fail when run on an Archivematica server that already has
Jobs in its database that reference the MCSL's.

The fail prevents upgrading from completing properly.

### DIFF
--- a/src/MCPServer/share/mysql_dev.sh
+++ b/src/MCPServer/share/mysql_dev.sh
@@ -22,5 +22,7 @@ mysql -u root "${dbpassword}" --execute="USE ${databaseName}; SOURCE $currentDir
 mysql -u root "${dbpassword}" --execute="USE ${databaseName}; SOURCE $currentDir/mysql_dev_8287_siegfried.sql;"
 mysql -u root "${dbpassword}" --execute="USE ${databaseName}; SOURCE $currentDir/mysql_dev_7321_dip_processing_xml.sql;"
 # ...
+# optional delete unused MCSL's
+mysql -u root "${dbpassword}" --execute="USE ${databaseName}; SOURCE $currentDir/mysql_dev_delete_links.sql;"
 
 touch $currentDir/mysql_dev.complete
--- a/src/MCPServer/share/mysql_dev_6488_aip_reingest.sql
+++ b/src/MCPServer/share/mysql_dev_6488_aip_reingest.sql
@@ -13,7 +13,6 @@ SET @d2 = '26cf64e2-21b5-4935-a52b-71695870f1f2' COLLATE utf8_unicode_ci;
 SET @d3 = '65916156-41a5-4ed2-9472-7dca11e6bc08' COLLATE utf8_unicode_ci;
 SET @d4 = '14a0678f-9c2a-4995-a6bd-5acd141eeef1' COLLATE utf8_unicode_ci;
 DELETE FROM MicroServiceChainLinksExitCodes WHERE microServiceChainLink IN (@d1, @d2, @d3, @d4);
-DELETE FROM MicroServiceChainLinks WHERE pk IN (@d1, @d2, @d3, @d4);
 
 -- Update microserviceGroups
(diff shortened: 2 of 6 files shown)


mistydemeo:
:+1:

Hwesta:
Can you comment the lines out instead of deleting them?

I wish we could have a try-catch around the deletes, but I know SQL makes that hard.

jhsimpson:
What about collecting the deleted deletions and storing them in a separate sql file?  We can try to delete them, in a separate step, but if that fails, then at least it won't bork the upgrade.

Hwesta:
I really like that idea.  I don't want the MSCLs cluttering up the DB if they don't have to, and the separate file for deletions would handle that on upgrade.

mistydemeo:
How about making the deletions a separate script entirely? `mysql_dev_cleanup.sh` or some such.

src/MCPServer/share/dev_mysql_delete_links.sql
```diff
@@ -0,0 +1,92 @@
...
+SET @d7 = '2d751fc6-dc9d-4c52-b0d9-a4454cefb359' COLLATE utf8_unicode_ci;
+SET @d8 = 'b063c4ce-ada1-4e72-a137-800f1c10905c' COLLATE utf8_unicode_ci;
+DELETE FROM MicroServiceChainLinks WHERE defaultNextChainLink in (@d0, @d1, @d2, @d3, @d4, @d5, @d6, @d7, @d8);
+DELETE FROM MicroServiceChainLinks WHERE pk IN (@d0, @d1, @d2, @d3, @d4, @d5, @d6, @d7, @d8);
+
+SET @del = '58fcd2fd-bcdf-4e49-ad99-7e24cc8c3ba5' COLLATE utf8_unicode_ci;
+DELETE FROM MicroServiceChainLinks WHERE pk IN (@del);
+
+SET @d1 = '2307b24a-a019-4b5b-a520-a6fff270a852' COLLATE utf8_unicode_ci;
+SET @d2 = 'c4e109d6-38ee-4c92-b83d-bc4d360f6f2e' COLLATE utf8_unicode_ci;
```
Hwesta: These two should be

```
SET @d1 = '4df4cc06-3b03-4c6f-b5c4-bec12a97dc90' COLLATE utf8_unicode_ci;
SET @d2 = '5e4f7467-8637-49b2-a584-bae83dabf762' COLLATE utf8_unicode_ci;
```

instead of duplicating the ones below.

src/MCPServer/share/dev_mysql_delete_links.sql
```diff
@@ -0,0 +1,92 @@
...
+
+SET @d1='0745a713-c7dc-451d-87c1-ec3dc28568b8' COLLATE utf8_unicode_ci;
+SET @d2='f7323418-9987-46ce-aac5-1fe0913c753a' COLLATE utf8_unicode_ci;
+SET @d3='9304d028-8387-4ab5-9539-0aab9ac5bdb1' COLLATE utf8_unicode_ci;
+SET @d4='45f01e11-47c7-45a3-a99b-48677eb321a5' COLLATE utf8_unicode_ci;
+SET @d5='6fe4678a-b3fb-4144-a8a3-7386eb87247d' COLLATE utf8_unicode_ci;
+DELETE FROM MicroServiceChainLinks WHERE pk IN (@d1, @d2, @d3, @d4, @d5);
+
+SET @removeIndexedAIPFiles='bfade79c-ab7b-11e2-bace-08002742f837' COLLATE utf8_unicode_ci;
+DELETE FROM MicroServiceChainLinks WHERE pk=@removeIndexedAIPFiles;
```
Hwesta: Also needs

```
SET @indexAIPfiles='3ba518ab-fc47-4cba-9b5c-79629adac10b' COLLATE utf8_unicode_ci;
DELETE FROM MicroServiceChainLinks WHERE pk=@indexAIPfiles;
```

mistydemeo:
A reason I'd be interested in doing that is because I want to start migrating our bash scripts to use [`set -e`](http://redsymbol.net/articles/unofficial-bash-strict-mode/), which will cause the entire script to fail if an expression returns non-zero. If we do that, it might be better to keep something expected to fail in another file where we can treat it specially.

jhsimpson:
I added the missing delete statements, and fixed the name of the file holding the delete statements, so it matches how it is called in mysql_dev.sh, and squashed those changes together, so this PR now has only 2 commits.

Hwesta:
:+1:

PR #229, opened 2015-05-09
Fixes for verifyMD5

mistydemeo:
- Hardcode "checksum" as the filename; this was technically customizeable, but it was never used, and that feature was broken.
- Fix the `groupType` passed to the Django ORM


mistydemeo:
Updated to change the database and remove the unused variable entirely.

Hwesta:
:+1:

jhsimpson:
:+1:

mistydemeo:
Merged with an additional fix, e82d3afb9698395c390d3c894bad6584cfbfea01.

PR #231, opened 2015-05-11
Never interpolate values directly into SQL

mistydemeo:
The vast majority of SQL statements in Archivematica have either been converted to use the Django ORM or updated to use SQL placeholders, but these files still interpolated values directly into SQL.

Fixes #4901.


mistydemeo:
Most of these look like they're unexploitable sections of code, since bad `fileUUID`s could only be inserted into the database manually - and someone who has access to the database already has the keys to the castle. But fixed up everything for consistency, along with a few places user-supplied data was still being interpolated into strings.

src/testingTools/btree test/btreeTest1.py
```diff
@@ -9,7 +9,7 @@
 database=MySQLdb.connect(**DB_CONNECTION_OPTS)
```
Hwesta: Do we want to just delete this whole directory instead?

jhsimpson:
+1 to deleting the whole testingTools directory, and migrating anything needed to the dev-tools repo.

Hwesta:
upload-archivistsToolkit-fcd.py lines 107, 119 are missing the safe interpolation, or is that file unused and can be removed?

jhsimpson:
upload-archivistsToolkit-fcd.py is a client specific file, that should be removed.

mistydemeo:
Pushed a commit that removes the entire set of `testingTools`.

Hwesta:
:+1:

jhsimpson:
#4901 is re-opened, there is a small change required still in views_NormalizationReport.py

PR #233, opened 2015-05-11
Archival Storage: Sorting persists across pages

Hwesta:
Use the 'search_params' to specify the sorting order in the pager for the default archival storage view.


mistydemeo:
:+1:

PR #234, opened 2015-05-12
METSRights: handle missing license identifier

mistydemeo:
Fixes #8425.


jhsimpson:
:+1: 

This has been tested on one dev server, and works.

PR #235, opened 2015-05-12
xml2obj: not every amdSec contains a techMD

mistydemeo:
Fixes #8431.

### DIFF
--- a/src/archivematicaCommon/lib/xml2obj.py
+++ b/src/archivematicaCommon/lib/xml2obj.py
@@ -13,6 +13,11 @@ def mets_file(src):
     
     file_uuid = None
     for amd in raw.mets_amdSec:
+        # Some amdSecs are other types of metadata (for instance sourceMD),
+        # and we don't care about those; skip them.
+        if amd.mets_techMD is None:
+            continue
+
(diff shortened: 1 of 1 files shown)


src/archivematicaCommon/lib/xml2obj.py
```diff
@@ -13,6 +13,11 @@ def mets_file(src):
     
     file_uuid = None
     for amd in raw.mets_amdSec:
+        # Some amdSecs are other types of metadata (for instance sourceMD),
+        # and we don't care about those; skip them.
+        if amd.mets_techMD is None:
```
Hwesta, line 18: Is this specifically None?  `if not amd.mets_techMD`?

Hwesta:
:+1:

src/archivematicaCommon/lib/xml2obj.py
mistydemeo, line 18: It is, yes! `DataNode.__getattr__` returns `None` specifically if the requested attribute isn't present.

Hwesta, line 18: Then `is None` is probably better. :)

PR #236, opened 2015-05-12
Remove unused audio playback support

mistydemeo:


Hwesta:
Also remove the Sounds table/model and foreign keys to it?

mistydemeo:
Wow, I didn't even know we'd made a model for it.

src/MCPServer/lib/plantUMLTextGenerator/main.py
```diff
@@ -53,7 +53,7 @@ def jobChainLinkExitCodesTextGet(indent, exitCode, nextMicroServiceChainLink, ex
...
+    sql = """SELECT MicroServiceChainLinks.currentTask, MicroServiceChainLinks.defaultNextChainLink, TasksConfigs.taskType, TasksConfigs.taskTypePKReference, TasksConfigs.description, MicroServiceChainLinks.reloadFileList, Sounds.fileLocation, MicroServiceChainLinks.defaultExitMessage, MicroServiceChainLinks.microserviceGroup, StandardTasksConfigs.execute FROM MicroServiceChainLinks JOIN TasksConfigs on MicroServiceChainLinks.currentTask = TasksConfigs.pk LEFT OUTER JOIN StandardTasksConfigs ON TasksConfigs.taskTypePKReference = StandardTasksConfigs.pk WHERE MicroServiceChainLinks.pk = %s;"""
```
Hwesta: Drop `Sounds.fileLocation` from the fields list

Hwesta:
Yeah - I added it so when we start creating DB things from the models everything would still work.

mistydemeo:
It looks like the plantUML generator is not working, and I don't think we've used it in a long time, so I'm tempted to drop it entirely in another commit.

Hwesta:
Sounds good to me

PR #240, opened 2015-05-13
WIP:  1.4.0 packaging and versioning changes

jhsimpson:
This Pull Request includes packaging changes, and changes to the sql scripts, required in advance of a 1.4.0 release.

This has been testing on a new install only.  Additional testing with upgrades is ongoing.


src/MCPClient/debian/copyright
```diff
@@ -1,6 +1,6 @@
 This work was packaged for Ubuntu by:
 
-    Justin Simpson <jsimpson@artefactual.com>
+    Artefactual Systems Inc. <sysadmin@artefactual.com>
```
Hwesta, line 3: Do we want to use ops@artefactual.com (as in the control file) or sysadmin@artefactual.com (as here).  Are they supposed to be the same?

jhsimpson, line 3: It should be sysadmin@ in all places, if there are individual emails left or ops@ anywhere, it is because I missed it, I will update the PR.

src/MCPServer/debian/postinst
```diff
@@ -13,9 +21,14 @@ else
   rsync -a /usr/share/archivematica/sharedDirectoryStructure/* /var/archivematica/sharedDirectory/.
 fi
 
-chown -R archivematica:archivematica /var/archivematica/
+chown -R archivematica:archivematica /var/archivematica/sharedDirectory
```
Hwesta: Add the trailing `/` - if `/var/archivematica/sharedDirectory` is a symlink this will only change permissions on the symlink not the directory.

src/MCPServer/share/postinstSharedWithDev
```diff
@@ -7,17 +7,10 @@ else
...
-sudo chmod -R g+s "/var/archivematica/"
-sudo chmod -R 664 "/var/archivematica/sharedDirectory" 
-sudo find -L /var/archivematica/ -type d  | sudo xargs -IF chmod u+rwx,g+rwxt,o-rwx F
-
-logdir=/var/log/archivematica/MCPServer
-sudo mkdir -p $logdir
-sudo chown -R archivematica:archivematica $logdir
-sudo chmod -R g+s $logdir
+sudo chmod -R g+s "/var/archivematica/sharedDirectory"
+sudo chmod -R 664 "/var/archivematica/sharedDirectory"
```
Hwesta: These should also have a trailing / if we want to affect the directory not the (possible) symlink. Or do we want to affect the symlink, given the line below?

src/MCPrpcCLI/debian/control
```diff
@@ -2,12 +2,14 @@ Source: archivematica-mcp-rpccli
 Section: utils
 Priority: extra
 Maintainer: Austin Trask <austin@artefactual.com>
-Build-Depends: debhelper (>= 9)
-Standards-Version: 3.9.3
+Build-Depends:
+  debhelper (>= 9)
+Standards-Version: 3.9.5
+X-Python-Version: 2.7
```
Hwesta, line 8: Glad to see the python version being specified now.

src/dashboard/debian/changelog
```diff
@@ -1,3 +1,10 @@
+archivematica-dashboard (1:1.3.2+4SNAPSHOT20150512174342-828eb-qa140~14.04) trusty; urgency=high
+
+  * commit: 828eb90be1a4bfe860b67a4398e8018456695dce
+  * 1.4 qa rc3 build
+
+ -- Justin Simpson <jsimpson@artefactual.com>  Tue, 12 May 2015 17:43:42 -0800
```
Hwesta, line 6: Should this be Artefactual instead of Justin?

jhsimpson, line 6: in the changelogs,I think it might have to be the person who signed the package.  It is my gpg key that signed the files being uploaded to launchpad, that is based on my email address.

For copyright and control files, it should be Artefactual.

We could create an Artefactual gpg key but I am not sure if that is a good idea or not.  So far each individual uses their own gpg key.

Hwesta, line 6: Ah, makes sense. Easiest to leave it as individual keys for now.

jhsimpson:
edits made and commits in the branch rebased.

jhsimpson:
I just merged the 8252 branch into qa/1.4.0, and made some changes to it.  Specfically, I updated the mysql and 1.4.0 files in MCPServer/share, with the new sql from 8252, and then updated mysql_dev.sh and removed the 8252 update script.

Not sure this is the cleanest way to get this change into the commit histor.

mistydemeo:
:+1:

mistydemeo:
The release has been made, do we want to merge this into qa/1.x?

Hwesta:
These changes are now in the packaging branch for 1.5 #380

PR #242, opened 2015-05-14
updating sql fix for Normalization Report

jhsimpson:
Previous pull request #231 changed how
sql parameters are handled, one example
was missed in the Normalization Report, now fixed.

### DIFF
--- a/src/dashboard/src/components/ingest/views_NormalizationReport.py
+++ b/src/dashboard/src/components/ingest/views_NormalizationReport.py
@@ -29,24 +29,24 @@
 #databaseInterface.printSQL = True
 from components import helpers
 from django.db import connection
-    
+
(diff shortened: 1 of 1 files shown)


Hwesta:
Is this different from #241 ?

jhsimpson:
Yes, this is different thatn #241 , I made a typo in the comment, when I said #231, I meant #241 .

#241 replaced '{0}' with %s in views_NormalizationReport.py, but it did it in only one place, there were 2 '{0}' strings to replace.  This PR fixes the 2nd one.

Looking at the diff, my text editor appears to have done something with whitespace that is a bit annoying, should this change be redone?  The diff should only be 2 lines.

jhsimpson:
this is a duplicate, turns out, of 241

PR #243, opened 2015-05-15
Update PRONOM v82

Hwesta:
Update for PRONOM v82.  Also, remove deprecated FPR update script, and sneak in an update to .gitignore


mistydemeo:
:+1:, though we probably want to get these categorized before we ship them.

jhsimpson:
I have been working on categorizing, I will add an update to this PR today.

PR #244, opened 2015-05-19
MCPServer: Improve logging

Hwesta:
When we stopped using print statements of SQL logged to the tmp directory, most of the logging for MCPServer disappeared. Convert it to use Python's logging framework, and add logging statements to replace the print statements.


mistydemeo:
:heart_eyes_cat:

mistydemeo:
> most of the logging for MCPServer disappeared

These should still be logged, however - the location of that log has simply moved. These statements were logged in `/var/log/archivematica/MCPServer/hostname-date.log`

mistydemeo:
:+1: Can also remove `logapp` from the dependencies in the control file.

PR #245, opened 2015-05-19
Reduce DB queries

Hwesta:
Convert several places that fetch and object, update and save it to use filter, which generates better SQL.

Try to ensure that unitFile replacement dicts only query the parent object once.  Unsure why this required changes in ReplacementDict for unicode.

### DIFF
--- a/src/MCPServer/lib/unit.py
+++ b/src/MCPServer/lib/unit.py
@@ -85,10 +85,7 @@ class unit:
                                            variable=variable)
         if variables:
             LOGGER.info('Existing UnitVariables for %s updated to %s (MSCL %s)', variable, variableValue, microServiceChainLink)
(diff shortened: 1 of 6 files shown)


mistydemeo:
:+1:

Hwesta:
Update to Django 1.6 to use [CONN_MAX_AGE](https://docs.djangoproject.com/en/1.6/ref/databases/#persistent-connections)

mistydemeo:
Do the changes you made to Units to query for the model only once still have the issue with caching you'd encountered previously? e.g., that they could end up not reflecting changes made by MCPClient?

Hwesta:
Rebased on top of #124.  Need to re-add and re-test changes around CONN_MAX_AGE

PR #246, opened 2015-05-22
Fix timestamps

Hwesta:
Fix timestamps to be timezone aware and display is localtime.

It may also be necessary to set the mysql default timezone to UTC (add `default_time_zone='+00:00'` to `/etc/mysql/my.conf`) - please verify.

### DIFF
--- a/src/dashboard/src/media/js/misc.js
+++ b/src/dashboard/src/media/js/misc.js
@@ -137,17 +137,31 @@ Date.prototype.getArchivematicaDateString = function()
     return dateText;
   };
 
-function localizeTimestampElements() {
-  $('.timestamp').each(function() {
-    $(this).text(timestampToLocal($(this).text()));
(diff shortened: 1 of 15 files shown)


src/dashboard/src/media/js/misc.js
```diff
@@ -137,19 +137,33 @@ Date.prototype.getArchivematicaDateString = function()
...
 
-function localizeTimestampElements() {
-  $('.timestamp').each(function() {
-    $(this).text(timestampToLocal($(this).text()));
-  });
-}
-
 function timestampToLocal(timestamp) {
   // convert to milliseconds
+  'use strict';
```
mistydemeo, line 142: 👍🏻

mistydemeo:
I'll test on mysql w/o setting timezone settings, but looks good!

PR #247, opened 2015-06-08
MCPServer: Log all exceptions, use ReplacementDict.replace

Hwesta:
Add a new except handler that logs the exception before continuing with default behavior.

Wrap all functions that are the start of a thread in a decorator that logs any exception raised

Update: Use ReplacementDict.replace for all cases when applying %var%s, since it handles encodings properly.

dev/issue-8509-log-server-exceptions-1.4 is this branch but based on qa/1.4.1

### DIFF
--- a/src/MCPServer/lib/utils.py
+++ b/src/MCPServer/lib/utils.py
@@ -1,0 +1,18 @@
+
+import logging
+
+LOGGER = logging.getLogger('archivematica.mcp.server')
+
+def log_exceptions(fn):
+    """
+    Decorator to wrap a function in a try-catch that logs the exception.
+
+    Useful for catching exceptions in threads, which do not normally report back to the parent thread.
+    """
+    def wrapped(*args,  **kwargs):
+        try:
+            return fn(*args, **kwargs)
+        except Exception:
+            LOGGER.exception('Uncaught exception')
+            raise
+    return wrapped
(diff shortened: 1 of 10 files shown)


src/MCPServer/lib/utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+LOGGER = logging.getLogger('archivematica.mcp.server')
+
+def log_exceptions(fn):
+    """
+    Decorator to wrap a function in a try-catch that logs the exception.
+
+    Useful for catching exceptions in threads, which do not normally report back to the parent thread.
+    """
+    def wrapped(*args,  **kwargs):
```
mistydemeo, line 12: Do we want to use `functools.wraps` to give the wrapper the original function's name? Not sure what this will do to tracebacks.

mistydemeo:
:+1:

src/MCPServer/lib/utils.py
Hwesta, line 12: Good idea! functool.wraps doesn't seem to affect tracebacks though.  I can still add it but I don't know if it'll do anything useful.

The error looks like

``` python
ERROR     2015-06-10 22:27:53  archivematica.mcp.server:utils:wrapped:18:  Uncaught exception
  File "~/archivematica/src/MCPServer/lib/utils.py", line 16, in wrapped
    return fn(*args, **kwargs)
  File "~/archivematica/src/MCPServer/lib/watchDirectory.py", line 74, in start
    raise Exception('uncaught')
Exception: uncaught
```

PR #248, opened 2015-06-09
templatetags: fix active tag when unhandled exception raised

mistydemeo:
If Django calls handle_uncaught_exception, the passed "request" object will be an empty string instead of a Request object.

This is the cause of the "AttributeError: 'str' object has no attribute 'path'" exception that happens when a generic Apache 500 page is seen.


Hwesta:
:+1:

PR #249, opened 2015-06-09
AIP deletion request: fix updating Elasticsearch

mistydemeo:
The structure of the query body in the "update" query was incorrect, causing Elasticsearch to return an error.

Also move this to elasticsearchFunctions at the same time to thin out the logic occurring in the view.

Fixes #8533.


Hwesta:
:+1:

PR #250, opened 2015-06-10
Dependencies: Improve SSL dependencies

Hwesta:
Add better openSSL support. See https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning


mistydemeo:
:+1:

mistydemeo:
Tentatively tagged this for 1.5.0, if that makes sense to you.

PR #251, opened 2015-06-15
Fix archival storage display of deletion requests

mistydemeo:
- "term" query no longer matches `DEL_REQ` in Elasticsearch 1.x; use a "match" query instead
- Fix `elasticSearchFunctions.delete_matching_documents`, which used the old pyes-style function name/parameters and had an invalid query


Hwesta:
Oops, I should have made a PR with this earlier. https://github.com/artefactual/archivematica/commit/0961e58760e1e0ee9d3a7f2b05da7b381034c84c maybe be useful re deleting documents.

mistydemeo:
Cherry-picked!

Hwesta:
:+1:

PR #254, opened 2015-06-22
Jobs: Always show arguments

Hwesta:
All troubleshooting requires seeing the command & argumens, so always display.

Also, there are only 2 columns to span.


mistydemeo:
:+1: Did you run this past Evelyn/Sarah/Courtney?

Hwesta:
Neither Evelyn nor Courtney were aware of a historical reason to hide the arguments.

PR #255, opened 2015-06-24
Shebangs use /usr/bin/env python2

Hwesta:
Update shebangs to use the environment's python2, instead of the system python. This will make it easier to run in virtualenvs, especially on systems where `/usr/bin/python` is not `python2`.

`/usr/bin/env` doesn't take arguments, so removing `-00` where it exists.

Any thoughts on changing `executeOrRunSubProcess.executeOrRun` to interpret 'pythonScript' as 'python2' ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ dev (1.5.0)
 * MCPServer: remove duplicate ReplacementDict code (#8509)
 * Dashboard: use GroupWriteRotatingFileHandler to ensure group-writeability of rotated logs (#8587)
 * Dashboard: always show arguments in job detail view
+* Tests: Use an in-memory database
+* Specify python2 in shebangs.  Python version choice respects path.
 
 1.4.0
 =====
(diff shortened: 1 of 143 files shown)


mistydemeo:
> /usr/bin/env doesn't take arguments, so removing -00 where it exists.

Huh, interesting - this behaviour seems different between GNU coreutils' env and BSD env. BSD env has no problem with a shebang of `#/usr/bin/env python2 -OO`.

mistydemeo:
Looks fine to me! I assume the performance implications of dropping `-OO` are minimal?

mistydemeo:
I'm wrong: apparently this is a Linux vs BSD thing, not a GNU env vs BSD env thing. https://twitter.com/jckarter/status/614583546921906177

Hwesta:
Interesting!  I just noticed it erroring on my machine with -OO and SO suggested because env didn't take params.  Didn't realize it was different in BSD.

`-OO` [doesn't do much](http://stackoverflow.com/questions/4777113/what-does-python-optimization-o-or-pythonoptimize-do), so I'm not very concerned.

PR #256, opened 2015-06-24
Tests use in memory DB

Hwesta:
Added a testing-focused settings file that uses an in-memory sqlite DB.

Pro: Much much faster than using MySQL, requires less setup
Con: MySQL and sqlite enforce constraints differently, so testing vs production are different

Another testing-vs-production difference: because Django is creating the DB instead of our hand-rolled SQL, the production and testing environments are different, especially around foreign keys.  I think this is true regardless of DB backend, though, just a different set of errors show up depending on backend.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 django_find_projects=false
 python_paths=/usr/share/archivematica/dashboard/;/usr/lib/archivematica/archivematicaCommon/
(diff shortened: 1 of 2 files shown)


mistydemeo:
As long as test and prod were _already_ difference, the minor extra differences don't bother me too much. This should make testing much easier, so I'm +1 on this.

PR #258, opened 2015-07-09
ArchivesSpace: Automate matching GUI

Hwesta:
Automatically parse an `archivesspacesids.csv` and populate the matching GUI for archivesspace.
- Do not automatically open the matching GUI
- Add DIP generation helper to Prepare DIP that parses `archivesspaceids.csv` for `filename,reference ID`, queries ArchivesSpace for the resource ID associated with the reference identifier, and populates ArchivesSpaceDIPObjectResourcePairing

### DIFF
--- a/src/MCPClient/lib/clientScripts/dip_generation_helper.py
+++ b/src/MCPClient/lib/clientScripts/dip_generation_helper.py
@@ -1,8 +1,123 @@
 #!/usr/bin/env python2
 from __future__ import print_function
 import argparse
+import ast
+import csv
 import sys
 
+# dashboard
+from django.db.models import Q
+from main import models
+
+# archivematicaCommon
+import archivesspace
+import archivematicaFunctions
+
+def create_archivesspace_client():
+    """
+    Create an ArchivesSpace client instance.
+    """
+    # TODO use same code as views_as.py?
+    repl_dict = models.MicroServiceChoiceReplacementDic.objects.get(description='ArchivesSpace Config')
+    config = ast.literal_eval(repl_dict.replacementdic)
+
+    try:
+        client = archivesspace.ArchivesSpaceClient(
+            host=config['%host%'],
+            port=config['%port%'],
+            user=config['%user%'],
+            passwd=config['%passwd%']
+        )
+    except archivesspace.AuthenticationError:
+        print("Unable to authenticate to ArchivesSpace server using the default user! Check administrative settings.")
+        return None
+    except archivesspace.ConnectionError:
+        print("Unable to connect to ArchivesSpace server at the default location! Check administrative settings.")
+        return None
+    return client
+
+def parse_archivesspaceids_csv(files):
+    """
+    Parse filename and reference ID from archivesspaceids.csv files
+
+    :param files: List of paths to archivesspaceids.csv files
+    :return: Dict with {filename: reference ID}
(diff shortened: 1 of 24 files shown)


src/MCPClient/lib/clientScripts/dip_generation_helper.py
```diff
@@ -0,0 +1,132 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Parse metadata for DIP helpers')
+    parser.add_argument('--sipUUID', required=True, help='%SIPUUID%')
+    parser.add_argument('--sipPath', required=True, help='%SIPDirectory%')
+    args = parser.parse_args()
+
+    rc = 0
+
+    rc = rc or parse_archivesspace_ids(args.sipPath, args.sipUUID)
```
mistydemeo, line 130: Since this exits 0 on success, can you just `sys.exit(parse_archivesspace_ids(args.sipPath, args.sipUUID))`?

src/dashboard/src/media/js/ingest.js
```diff
@@ -331,13 +331,6 @@ $(function()
             window.location.href = '/ingest/' + this.model.sip.get('uuid') + '/upload/atk/';
           }
 
-          // redirect to object/resource mapping pages
```
mistydemeo, line 334: Will the user need to manually click through in this case?

src/MCPClient/lib/clientScripts/dip_generation_helper.py
Hwesta, line 130: I was thinking this script would be used for multiple DIP helpers, and the script should return failure if any one of them failed, so the RC should be the or'd results of all the helpers. Does this seem reasonable?

mistydemeo:
:+1: Tests are nice to see too :D

src/MCPClient/lib/clientScripts/dip_generation_helper.py
mistydemeo, line 130: That makes sense. Maybe leave a comment to that effect (since it's not clear when there aren't yet multiple values to `or`), or do that when there is more than one exit code?

src/dashboard/src/media/js/ingest.js
Hwesta, line 334: Yes.  Since this PR adds automating the matching GUI (for ArchivesSpace), I removed starting up the matching GUI in JS. I'd rather it checked if it was necessary first, but that's non-trivial.  Open to discussion on that one.

mistydemeo, line 334: It'd probably be good to ask Justin/Sarah about this, but given the circumstances this is probably the best compromise. Maybe we could make the way you open the matching GUI manually more obvious to compensate?

src/MCPClient/lib/clientScripts/dip_generation_helper.py
Hwesta, line 130: Updated

Hwesta:
Note: this is in the process of getting updated other ArchivesSpace enhancements (tickets #8697 and #8696 in Redmine)

Hwesta:
Now based on top of #261 

Added 0d065ecd2af7bb1c575ff13bd99246ceccdc9177

PR #260, opened 2015-07-15
ArchivesSpaceClient: dispatch on resource type properly

mistydemeo:
There were a few issues with the original code; several places depended on an external resource_type parameter being passed in, and that parameter inconsistently used `"collection"` (for historical reasons) and `"resource"` to refer to a resource-level record. This patch instead determines the resource_type automatically, and renders the parameter a no-op. (It's retained for compatibility with the ATK client, though that should probably also get adjusted at some point.)


Hwesta:
:+1:

PR #261, opened 2015-07-28
Add identifier search to ArchivesSpace/ATK upload

mistydemeo:


src/archivematicaCommon/lib/archivistsToolkit/client.py
```diff
@@ -235,12 +235,24 @@ def find_collection_ids(self, search_pattern='', page=None, page_size=30):
...
+                clause = 'title LIKE %s OR' + clause
+                params.insert(0, '%' + search_pattern + '%')
+
+            if identifier != '':
+                clause = 'resourceIdentifier1 LIKE %s OR ' + clause
+                params.insert(0, '%' + identifier + '%')
+
+            params = tuple(params)
+
+            sql = "SELECT resourceId FROM Resources WHERE ({}) AND resourceLevel in ('recordgrp', 'collection') ORDER BY title".format(clause)
```
Hwesta, line 255: Why format in the clause, instead of adding to params?

mistydemeo, line 255: The params are SQL placeholders; they're not string interpolation, but rather passing parameters to stored procedures. Since this isn't a single value it can't be passed in the params that way.

Hwesta, line 255: Ahh, okay. And because this is being considered for an external library you don't want to use an ORM?

Hwesta:
Tests for new functionality?

src/archivematicaCommon/lib/archivistsToolkit/client.py
mistydemeo, line 255: Yeah, I think so. Converting this to use the ORM would be some extra work, especially since we don't have control over the ATK database structure (and the ATK stuff is happening in a separate database from Archivematica).

mistydemeo:
Added ASpace tests.

Hwesta:
:+1:

mistydemeo:
Updated to make the search UI clearer.

PR #262, opened 2015-07-29
Removed vendored requests library

Hwesta:
Remove `requests` and `requests_1_20` from `archivematicaCommon/lib/externals`.  Add requests to requirements.txt


jhsimpson:
+1

PR #263, opened 2015-07-29
Fix FPRClient tests

Hwesta:
Update so they don't always fail.  Now just running `py.test` should work from the root of the repo.


mistydemeo:
Didn't have the chance to test, but looks good :thumbsup:

PR #264, opened 2015-08-04
Libffi is a runtime not build dependency

Hwesta:


jhsimpson:
I tested this change and it didn't work as expected.  libffi is a build dependency, needed on launchpad to build a package.  I think the original issue was related to the ansible deployment script used, which was not fetching this dependency at all.  

I think this PR should be closed, once testing for 1.4.2 release is complete, which should confirm that the original issue is resolved.

Hwesta:
Ok.  Closing this.

PR #265, opened 2015-08-07
Forms: normalize and validate dates

mistydemeo:
These were previously treated as pure text, so any value was permissible. However, Elasticsearch may now opportunistically parse these into dates, which means that we have to be more careful about validating them. Use dateutil to try to normalize these values; if we can't, raise a ValidationError to provide the user the chance to fix it.

We probably want to get this into 1.5.x instead of 1.4.x since this introduces new constraints into the date field, even though this fixes a bug that's present in the 1.4.x series. (cc @sromkey for opinions on that.)

### DIFF
--- a/src/dashboard/src/components/helpers.py
+++ b/src/dashboard/src/components/helpers.py
@@ -29,10 +29,13 @@ from django.utils.dateformat import format
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger, InvalidPage
 from django.core.urlresolvers import reverse
 from django.db.models import Max
+from django import forms
 from django.http import HttpResponse, HttpResponseRedirect
 from django.core.servers.basehttp import FileWrapper
 from django.shortcuts import render
 from main import models
+
+import dateutil.parser
 
 logger = logging.getLogger('archivematica.dashboard')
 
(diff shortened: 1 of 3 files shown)


src/dashboard/src/components/helpers.py
```diff
@@ -334,3 +337,11 @@ def default_processing_config_path():
...
         'sharedMicroServiceTasksConfigs/processingMCPConfigs/defaultProcessingMCP.xml'
     )
+
+def clean_date(date):
+    if date:
+        try:
+            return dateutil.parser.parse(date).strftime('%Y-%m-%d')
+        except ValueError:
+            raise forms.ValidationError('Unable to interpret as a date: %(date)s',
+                                        params={'date': date})
```
Hwesta, line 358: `return date` in all cases?  Django clean functions say to always return data, and it may be something Falsey that isn't `None`.

src/dashboard/src/components/helpers.py
```diff
@@ -334,3 +337,11 @@ def default_processing_config_path():
         get_server_config_value('sharedDirectory'),
         'sharedMicroServiceTasksConfigs/processingMCPConfigs/defaultProcessingMCP.xml'
     )
+
+def clean_date(date):
```
Hwesta, line 341: Docstring?

src/dashboard/src/components/helpers.py
mistydemeo, line 358: Updated to return the original value in this case.

src/dashboard/src/components/helpers.py
mistydemeo, line 341: Added!

Hwesta:
:+1: 

I added it to 1.5.0 milestone because it'll definitely get in there, and if it's untagged or 1.4.0 we might miss it when we collect what went into 1.5.0

sromkey:
After more internal discussion, and realizing that we can't express ranges of dates in one field in ES, it's been concluded that instead we'll revert the date field back to a string rather than a date. See: https://projects.artefactual.com/issues/8812

Hwesta:
Should we close this?

jhsimpson:
I think this PR needs to be closed?

mistydemeo:
Should I delete the branch or leave it for reference?

PR #267, opened 2015-08-26
Initialize Django in MCP client scripts ref. #8879

hakamine:
Standalone scripts require to initialize Django to avoid
an AppRegistryNotReady exception

So far identified 3 scripts, others may also need to be fixed.


Hwesta:
Thanks for catching these! I've been putting the Django setup stuff after the system imports (eg `import os` but before any AM specific imports (like from archivematicaCommon or models) in case those files also need Django to be set up before they're imported.  If it works like this, though, don't worry about changing it.

jhsimpson:
I tested this with the upload_archivessspace.py and  it works, 

:+1:

Hwesta:
Cherry-picked to qa/1.5.x as bf50761651d5b388f9ffbf714186c5f82048c52a

PR #268, opened 2015-08-28
arrange_contents: fix calling without the `path` parameter

mistydemeo:
`DEFAULT_ARRANGE_PATH` is not base64-encoded; trying to decode it raises an exception.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -114,13 +114,16 @@ def contents(request):
     return helpers.json_response(response)
 
 def arrange_contents(request):
-    base_path = base64.b64decode(request.GET.get('path', DEFAULT_ARRANGE_PATH))
+    path = request.GET.get('path')
+    if path is not None:
+        base_path = base64.b64decode(path)
+        # Must indicate that base_path is a folder by ending with /
+        if not base_path.endswith('/'):
+            base_path += '/'
 
-    # Must indicate that base_path is a folder by ending with /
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -114,13 +114,23 @@ def contents(request):
...
-    if not base_path.endswith('/'):
-        base_path += '/'
+    path = request.GET.get('path')
+    if path is not None:
+        try:
+            base_path = base64.b64decode(path)
+        except TypeError:
+            response = {
+                'success': False,
+                'message': 'Could not base64-decoded provided path: ' + path,
```
Hwesta: Use % or .format or str(), since `+` will fail if path is not a basestring.

mistydemeo: Fixed! Too much time in JS land...

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -114,13 +114,23 @@ def contents(request):
...
+            }
+            helpers.json_response(response, status_code=400)
```
Hwesta, line 126: https://docs.djangoproject.com/en/1.8/ref/request-response/#jsonresponse-objects :o Probably not for this PR but I'd like to replace `helpers.json_response` with `JsonResponse`

mistydemeo, line 126: Yeah, in the future let's make that change all at once.

Hwesta:
:+1:

PR #269, opened 2015-08-31
Add new /file/ API

mistydemeo:
This API is focused on retrieval of file metadata; it is currently used by the appraisal tab.

Supports:
- Retrieval of file record data
- Add/list/remove file tags
- Retrieve file's bulk_extractor reports

### DIFF
--- a/src/dashboard/src/components/file/views.py
+++ b/src/dashboard/src/components/file/views.py
@@ -1,0 +1,217 @@
+# Standard library, alphabetical by import source
+import json
+import logging
+import os
+import sys
+
+# Django Core, alphabetical by import source
+from django.views.generic import View
+
+# External dependencies, alphabetical
+import requests
+
+# This project, alphabetical by import source
+from components import helpers
+from main import models
+
+sys.path.append("/usr/lib/archivematica/archivematicaCommon")
+import elasticSearchFunctions
+import storageService as storage_service
+
+logger = logging.getLogger('archivematica.dashboard')
+
+class TransferFileTags(View):
+    """
+    Fetch, set or remove tags from a record.
+    This view's methods operate on the file data indexed in Elasticsearch, and do not affect the database.
+    All methods return 404 if a record can't be retrieved for the given UUID, and 400 if more than one file was found for a given UUID.
+    """
+    def get(self, request, fileuuid):
+        """
+        Requires a file UUID.
+        Returns a JSON-encoded list of the file's tags on success.
+        """
+        try:
+            tags = elasticSearchFunctions.connect_and_get_file_tags(fileuuid)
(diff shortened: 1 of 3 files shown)


mistydemeo:
This depends on #271.

src/dashboard/src/components/file/views.py
```diff
@@ -0,0 +1,210 @@
...
+from components import helpers
+from main import models
+
+sys.path.append("/usr/lib/archivematica/archivematicaCommon")
+import elasticSearchFunctions
+import storageService as storage_service
+
+logger = logging.getLogger('archivematica.dashboard')
+
+def transfer_file_tags(request, uuid):
```
Hwesta: Rename `uuid` to not maybe-shadow the `uuid` module?  Also to be more specific.

src/dashboard/src/components/file/views.py
```diff
@@ -0,0 +1,210 @@
...
+        GET: requires a file UUID.
+            Returns a JSON-encoded list of the file's tags on success.
+        PUT: requires a file UUID, and document body must be a JSON-encoded list.
+            Replaces the list of tags in the record with the provided list.
+            Returns {"success": true} on success.
+            Returns 400 if no JSON document is provided in the request body, if the body can't be decoded, or if the body is any JSON object other than a list.
+        DELETE: requiest a file UUID.
+            Deletes all tags for the given file.
+    """
+    if request.method == 'GET':
```
Hwesta: This may be a good fit for Django's [class-based views](https://docs.djangoproject.com/en/dev/topics/class-based-views/) since they have different functions for different verbs.  Take a look at it?

src/dashboard/src/components/file/views.py
```diff
@@ -0,0 +1,210 @@
...
+            if isinstance(e, elasticSearchFunctions.EmptySearchResultError):
+                status_code = 404
+            else:
+                status_code = 400
+            return helpers.json_response(response, status_code=status_code)
+
+        return helpers.json_response(tags)
+    elif request.method == 'PUT':
+        try:
+            tags = json.load(request)
```
Hwesta, line 57: Should this be `json.loads(request.body)`?  Or does `request` expose a file-like object of the body for `json.load`?

mistydemeo, line 57: `request` does expose a file-like interface!

Hwesta, line 57: Neat! I didn't know that!

src/dashboard/src/components/file/views.py
mistydemeo: Worked, so I switched to that!

mistydemeo:
Any more comments?

Hwesta:
:+1:

PR #270, opened 2015-09-02
Make `filepath` parameter to /filesystem/delete/ APIs mandatory

mistydemeo:
Defaulting to an empty string means that this deletes the entire contents of arrange. This is very easy to do by accident when calling this API manually, for example if the user accidentally passes the parameters as URL parameters or JSON data instead of form-encoded data.


Hwesta:
:+1:

PR #271, opened 2015-09-02
Elasticsearch: update transfer index format, add tag functions

mistydemeo:
This is required for the appraisal tab backend APIs, such as #269.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -594,6 +594,34 @@ def normalize_list_dict_elements(data):
             data[index] =  normalize_dict_values(value)
     return data
 
+def _get_file_formats(f):
+    formats = []
+    fields = ['format_version__pronom_id',
+              'format_version__description',
+              'format_version__format__group__description']
+    for puid, format, group in f.fileformatversion_set.all().values_list(*fields):
+        formats.append({
+            'puid': puid,
+            'format': format,
+            'group': group,
+        })
+
+    return formats
+
+def _list_bulk_extractor_reports(transfer_path, file_uuid):
+    reports = []
+    log_path = os.path.join(transfer_path, 'logs', 'bulk-' + file_uuid)
+
+    if not os.path.isdir(log_path):
+        return reports
+    for report in ['telephone', 'ccn', 'ccn_track2', 'pii']:
+        path = os.path.join(log_path, report + '.txt')
+        if os.path.isfile(path) and os.path.getsize(path) > 0:
+            reports.append(report)
+
+    return reports
+
+
 def index_transfer_files(conn, uuid, pathToTransfer, index, type):
     """
     Indexes files in the Transfer with UUID `uuid` at path `pathToTransfer`.
@@ -608,7 +636,6 @@ def index_transfer_files(conn, uuid, pathToTransfer, index, type):
     """
     filesIndexed = 0
     ingest_date  = str(datetime.datetime.today())[0:10]
-    create_time  = time.time()
 
     # Some files should not be indexed
     # This should match the basename of the file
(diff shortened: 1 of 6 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -736,6 +789,71 @@ def connect_and_change_transfer_file_status(uuid, status):
     return len(document_ids)
 
 
+def connect_and_get_file_tags(uuid):
+    """
+    Retrieve the complete set of tags for the file with the fileuuid `uuid`.
+
+    :param str uuid: A file UUID.
+    """
```
Hwesta, line 798: Document return value structure?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -736,6 +789,71 @@ def connect_and_change_transfer_file_status(uuid, status):
...
+
+    result = results['hits']['hits'][0]
+    if 'fields' not in result:  # file has no tags
+        return []
+    return result['fields']['tags']
+
+
+def connect_and_set_file_tags(uuid, tags):
+    """
+    Updates the file(s) with the fileuuid `uuid to the provided value(s).
```
Hwesta: Matching ` missing.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -736,6 +789,71 @@ def connect_and_change_transfer_file_status(uuid, status):
     return len(document_ids)
 
 
+def connect_and_get_file_tags(uuid):
```
Hwesta, line 792: Is `uuid` shadowing the `uuid` module? Maybe change the parameter name?

src/archivematicaCommon/tests/test_elasticsearch_functions.py
```diff
@@ -67,3 +68,22 @@ def test_delete_aip_files(self):
             fields='AIPUUID,FILEUUID',
         )
         assert results['hits']['total'] == 0
+
+    @vcr.use_cassette(os.path.join(THIS_DIR, 'fixtures', 'test_list_tags.yaml'))
```
Hwesta, line 72: :100:

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo, line 792: Unfortunately this is already a pretty commonly used param name in this module.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -583,6 +597,34 @@ def normalize_list_dict_elements(data):
...
+            'puid': puid,
+            'format': format,
+            'group': group,
+        })
+
+    return formats
+
+def _list_bulk_extractor_reports(transfer_path, file_uuid):
+    reports = []
+    log_path = os.path.join(transfer_path, 'logs', 'bulk-' + file_uuid)
```
Hwesta, line 616: Is there a way to get this info without assuming the backlogged transfer is locally accessible?  In the future, it would be nice if it could be in Duracloud or something.  Does the storage service browse endpoint work here?  Unfortunately it looks like you need size & created time, which the storage service doesn't currently provide.

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: Fixed!

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo, line 798: Documented.

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo, line 616: This is used at the time the transfer's indexed, so it hasn't actually been stored yet. It's still available locally.

src/archivematicaCommon/lib/elasticSearchFunctions.py
Hwesta, line 792: Unfortunate, but fair.

src/archivematicaCommon/lib/elasticSearchFunctions.py
Hwesta, line 616: Yay!

Hwesta:
:+1:

PR #272, opened 2015-09-04
AS/ATK clients: add method to get levels of description

mistydemeo:


src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -130,6 +130,17 @@ def resource_type(self, resource_id):
     def get_record(self, record_id):
         return self._get(record_id).json()
 
+    def get_levels_of_description(self):
+        """
+        Returns an array of all levels of description defined in this ArchivesSpace instance.
+        """
+        if not hasattr(self, 'levels_of_description'):
+            # TODO: * fetch human-formatted strings
+            #       * is hardcoding this ID okay?
```
Hwesta, line 139: Is this still a TODO?

mistydemeo, line 139: Yeah, I need to talk to the AS people about this. Don't think they'll be around until Monday though.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -130,6 +130,17 @@ def resource_type(self, resource_id):
     def get_record(self, record_id):
         return self._get(record_id).json()
 
+    def get_levels_of_description(self):
+        """
+        Returns an array of all levels of description defined in this ArchivesSpace instance.
+        """
+        if not hasattr(self, 'levels_of_description'):
```
Hwesta, line 137: Why check `hasattr` instead setting `self.levels_of_description = None` in the constructor?  I can't find a source for why this looks odd to me, though.  [SO](http://stackoverflow.com/a/9748715) has some other approaches for handling `hasattr`.

src/archivematicaCommon/lib/archivistsToolkit/client.py
```diff
@@ -22,6 +22,23 @@ def __init__(self, host, user, passwd, db):
...
             raise
 
+    def get_levels_of_description(self):
+        """
+        Returns an array of all levels of description defined in this Archivist's Toolkit instance.
+        """
+        if not hasattr(self, 'levels_of_description'):
+            cursor = self.db.cursor()
+            levels = set()
+            cursor.execute("SELECT distinct(resourceLevel) FROM Resources")
```
Hwesta, line 32: :( No models, eh?

mistydemeo, line 32: Nope. The rest of the file uses raw SQL too; this was ported pretty straight from the old code.

Didn't seem appropriate to use Django models for this, since Django assumes a greater degree of ownership over the schema and this is an external app's DB.

Hwesta, line 32: Reasonable.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo, line 139: Haven't heard back yet; I think it may take a bit to get a response. Maybe merge now and update later with the info we get?

Hwesta, line 139: Seems reasonable.

Hwesta:
:+1:

PR #273, opened 2015-09-08
renameAsSudo: run command as an array, not a string

mistydemeo:
This ensures that this has no chance of failing if spaces or quotes are in command names, since shell splitting won't take place.


src/archivematicaCommon/lib/fileOperations.py
```diff
@@ -116,9 +116,7 @@ def writeToFile(output, fileName, writeWhite=False):
 
 def renameAsSudo(source, destination):
     """Used to move/rename Directories that the archivematica user may or may not have writes to move"""
-    command = "sudo mv \"" + source + "\"   \"" + destination + "\""
-    if isinstance(command, unicode):
-        command = command.encode("utf-8")
+    command = ["sudo", "mv", source, destination]
```
Hwesta, line 115: Does this still need to be encoded?

Hwesta:
:+1:

src/archivematicaCommon/lib/fileOperations.py
mistydemeo, line 115: Testing, no - mixed encodings are fine if no shell splitting is taking place. e.g. this is valid, with one unistring and one bytestring that's not valid Unicode:

``` python
subprocess.check_output(['echo', u'é', '\xFF'])
```

PR #274, opened 2015-09-08
Elasticsearch: index only needed fields

mistydemeo:
This is very WIP and here for discussion right now.

We've been running into various issues from indexing the entire METS file in ES; this was based on a discussion Justin and I had about indexing only the fields that we're explicitly making searchable.

As a part of this we're also likely to find fields that we may have expected to be searchable that were only actually searchable by accident because we indexed the entire METS file, even though they were never exposed in the search interface.

This is connected to https://github.com/artefactual/archivematica/issues/1199.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -386,9 +386,6 @@ def set_up_mapping(client, index):
 def index_aip(client, uuid, name, filePath, pathToMETS, size=None, aips_in_aic=None, identifiers=[]):
     tree = ElementTree.parse(pathToMETS)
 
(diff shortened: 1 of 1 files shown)


mcantelon:
This looks good and makes sense. I think when we decided to index the entire METS file we didn't realize how big some of them would end up being.

mistydemeo:
Maybe we should set up a demo server using this and have the archivists perform some searches that they would expect to work, to see what breaks?

sevein:
I'm closing this PR because it's not being actively developed and we also have a new issue where other solutions are being discussed (see https://github.com/artefactual/archivematica/issues/1199). I'm not removing the branch pushed by Misty because it could be definitely useful during future work.

PR #275, opened 2015-09-08
Rename endpoint, s/ransfer/transfer

sevein:
As @mistydemeo said:

> At one time in the past a bad regex meant that `/filesystem/transfer` would be interpreted as `/transfer`. So the API got named `ransfer` and it’s never been changed even after the regex ambiguity was fixed.


Hwesta:
:shipit:  :sparkles:

mistydemeo:
:stars:

jhsimpson:
:+1:

sevein:
Merged, 23b0ba4.

Hwesta:
Cherry picked into qa/1.5.x as 487cd68cb7397c85ccb57cba299c3e1764229342

PR #277, opened 2015-09-09
Add new /access/ API

mistydemeo:
This API allows retrieving Archivist's Toolkit or ArchivesSpace records, using the same instances configured for use with the upload scripts.

### DIFF
--- a/src/archivematicaCommon/lib/archivesspace/client.py
+++ b/src/archivematicaCommon/lib/archivesspace/client.py
@@ -213,11 +213,16 @@ class ArchivesSpaceClient(object):
             }
             if record['children'] and descend:
                 result['children'] = [format_record(child, level) for child in record['children']]
+                result['has_children'] = True
                 if sort_by is not None:
                     kwargs = {'reverse': True} if sort_by == 'desc' else {}
                     result['children'] = sorted(result['children'], key=lambda c: c['title'], **kwargs)
+            elif record['children']:
+                result['children'] = []
+                result['has_children'] = True
             else:
                 result['children'] = False
+                result['has_children'] = False
 
             return result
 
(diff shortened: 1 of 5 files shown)


src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -345,17 +350,42 @@ def count_collections(self, search_pattern='', identifier=''):
...
+            for example, "text" will match "this title has this text in it".
+            If omitted, then all resources will be fetched.
+        :param string identifier: Restrict records to only those with this identifier.
+            This refers to the human-assigned record identifier, not the automatically generated internal ID.
+            This value can contain wildcards.
+        :param string accession: Restrict records to only those associated with this accession.
+            This should be provided using ArchivesSpace's internal format, for instance /repositories/2/accessions/1; it does not refer to human-assigned accession identifiers.
+            Use ArchivesSpaceClient.find_collections_by_accession to locate records via a human-assigned accession identifier.
+
+        :return list: A list containing every matched resource's ID.
```
Hwesta: Sphinx doesn't support :return type: unfortunately.  It does support :return: and :returntype: http://sphinx-doc.org/domains.html#info-field-lists

src/dashboard/src/components/access/views.py
```diff
@@ -0,0 +1,112 @@
...
+        records = client.find_collections_by_accession(accession, page=page, page_size=page_size)
+        return helpers.json_response(records)
+    except ArchivesSpaceError as e:
+        return helpers.json_response({"error": True, "message": str(e)})
+
+
+@_authenticate_to_archivesspace
+def record_children(client, request, system='', record_id=''):
+    if request.method == 'PUT':
+        pass
```
Hwesta, line 110: Why explicitly accept but do nothing on PUT?

Hwesta:
:+1:

src/dashboard/src/components/access/views.py
mistydemeo, line 110: It's just stubbed in right now; the intention is for this to allow new records to be added. Now that the edit form is in, I'm going to have a followup PR to add the real functionality.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo: I'll fix this, and add a new commit that changes the rest.

mistydemeo: Fixed!

PR #278, opened 2015-09-09
Add new appraisal tab APIs, add appraisal tab as a submodule

mistydemeo:


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -809,3 +809,15 @@ def download_fs(request):
...
+        f = models.File.objects.get(uuid=uuid)
+    except models.File.DoesNotExist:
+        response = {
+            'success': False,
+            'message': 'File with UUID ' + uuid + ' could not be found',
+        }
+        return helpers.json_response(response, status_code=404)
+    relative_path = f.currentlocation.replace('%transferDirectory%', '')
+    redirect_url = storage_service.extract_file_url(f.transfer_id, relative_path)
+    return django.http.HttpResponseRedirect(redirect_url)
```
mistydemeo: Redirect may not be the best approach, since it's possible the Storage Service may not actually be reachable by the user's machine.

mistydemeo: Updated - this now proxies the response between the storage service and the appraisal tab, ensuring this works even if the user doesn't have access to the storage service's server.

mistydemeo:
I wonder if it's worth, in a followup PR, converting `download_ss` from filesystem_ajax to stream the file too?

src/dashboard/src/components/ingest/views.py
```diff
@@ -425,6 +425,79 @@ def _es_results_to_directory_tree(path, return_list, not_draggable=False):
...
+                'children': [],
+            }
+            record_map[node] = dir_record
+            # directory_list should consist only of top-level records
+            if is_transfer:
+                directory_list.append(dir_record)
+        else:
+            dir_record = record_map[node]
+
+        if parent is not None and not dir_record in parent['children']:
```
Hwesta: Nitpick `dir_record not in parent['children']`

src/dashboard/src/components/ingest/views.py
```diff
@@ -425,6 +425,79 @@ def _es_results_to_directory_tree(path, return_list, not_draggable=False):
...
+            parent['children'].append(dir_record)
+            object_count = parent['object_count'] + 1
+            parent['object_count'] = object_count
```
Hwesta: Why not `parent['object_count'] += 1`?  Is `parent['object_count']` always the length of `parent['children']` or am I missing something?

src/dashboard/src/components/ingest/views.py
```diff
@@ -425,6 +425,79 @@ def _es_results_to_directory_tree(path, return_list, not_draggable=False):
...
+        'not_draggable': not_draggable
+    }
+    if record['format']:
+        format = record['format'][0]  # TODO handle multiple format identifications
+        child['format'] = format['format']
+        child['group'] = format['group']
+        child['puid'] = format['puid']
+
+    record_map[dir]['children'].append(child)
+    record_map[dir]['object_count'] += 1
```
Hwesta, line 499: Is `object_count` supposed to be the count of children or the count of all descendants?

src/dashboard/src/components/ingest/views.py
```diff
@@ -500,6 +573,86 @@ def transfer_backlog(request):
...
+                queries,
+                ops,
+                fields,
+                types,
+                # Specify this as a filter, not a must_have, for performance,
+                # and so that it doesn't cause the "should" queries in a
+                # should-only query to be ignored.
+                filters={'term': {'status': 'backlog'}},
+            )
+        except:
```
Hwesta: Only catch `Exception`, or figure out what exceptions are being generated and list them.

src/dashboard/src/components/ingest/views.py
```diff
@@ -500,6 +573,86 @@ def transfer_backlog(request):
...
+            logger.exception('Error accessing index.')
```
Hwesta: Is this ERROR-level?  If not, then `logger.foo(..., exc_info=True)` might be better.

src/dashboard/src/components/ingest/views.py
```diff
@@ -500,6 +573,86 @@ def transfer_backlog(request):
...
+            return HttpResponse('Error accessing index.')
```
Hwesta: Should this have a status code of 200?

src/dashboard/src/components/ingest/views.py
```diff
@@ -500,6 +573,86 @@ def transfer_backlog(request):
     return helpers.json_response(return_list)
 
 
+@decorators.elasticsearch_required()
+def transfer_appraisal(request):
```
Hwesta: Instead of almost duplicating transfer_backlog, maybe make it a parameter that selects between the different functions?  Could be done in the urls.py.  A bit messier now, but then we don't have to worry about applying fixes to both, and we're planning on removing transfer_backlog eventually anyway.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -813,3 +815,25 @@ def download_fs(request):
     except ValueError:
         raise django.http.Http404
 
+def download(request, uuid):
```
Hwesta: More specific name and/or docstring? Otherwise we have download_fs, download_ss and download with nothing clear to differentiate them.

It's probably out of scope to merge all the download_foos into one function, but it would be nice.

src/dashboard/src/components/ingest/views.py
mistydemeo, line 499: Count of all descendants.

src/dashboard/src/components/ingest/views.py
mistydemeo: Merged the two.

src/dashboard/src/components/filesystem_ajax/views.py
mistydemeo: Updated the name and added a docstring.

I'm thinking `download_ss` should probably stream the response as well, since the bug I added streaming for likely affects some current Archivematica installations.

Hwesta: download_ss (and download_fs?) also streaming seems like a good idea.

mistydemeo: I'll do that in a followup PR.

src/dashboard/src/components/ingest/views.py
```diff
@@ -494,10 +564,34 @@ def transfer_backlog(request):
...
+    """
+    return transfer_request(request, 'legacy')
+
+
+@decorators.elasticsearch_required()
+def transfer_appraisal(request):
+    """
+    AJAX endpoint to query for and return transfer backlog items in the appraisal tab format.
+    """
+    return transfer_request(request, 'appraisal')
```
Hwesta: Probably cleaner to use urls.py to pass the extra parameter and not have the wrapping functions. https://docs.djangoproject.com/en/1.7/topics/http/urls/#passing-extra-options-to-view-functions

mistydemeo: Oh, neat - that is cleaner. Changed!

src/dashboard/src/components/ingest/views.py
```diff
@@ -425,12 +425,81 @@ def _es_results_to_directory_tree(path, return_list, not_draggable=False):
...
+            record_map[node] = dir_record
+            # directory_list should consist only of top-level records
+            if is_transfer:
+                directory_list.append(dir_record)
+        else:
+            dir_record = record_map[node]
+
+        if parent is not None and dir_record not in parent['children']:
+            parent['children'].append(dir_record)
+            parent['object_count'] += 1
```
Hwesta, line 478: Directories are counted in the object_count intentionally?

src/dashboard/src/components/ingest/views.py
```diff
@@ -494,10 +564,34 @@ def transfer_backlog(request):
...
             original_path__endswith=path['relative_path']).exists():
             not_draggable = True
-        _es_results_to_directory_tree(path['relative_path'], return_list, not_draggable=not_draggable)
+        if ui == 'legacy':
+            _es_results_to_directory_tree(path['relative_path'], return_list, not_draggable=not_draggable)
+            response = return_list
+        else:
+            _es_results_to_appraisal_tab_format(path, directory_map, return_list, not_draggable=not_draggable)
+            response = {
+                "formats": [],  # TODO populate this
```
Hwesta, line 577: Is this TODO for a future PR?

src/dashboard/src/components/ingest/views.py
mistydemeo, line 478: Yeah, it's consistent with the existing format.

Hwesta:
:+1:

src/dashboard/src/components/ingest/views.py
mistydemeo, line 577: It is, yeah; not having this in right now is making testing future PRs a bit harder so I'd like to get it in sooner.

PR #280, opened 2015-09-10
AS: add new API to add a child to an existing resource

mistydemeo:


src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -601,3 +601,36 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
             "digital_object": {"ref": new_object_uri}
         })
         self._post(parent_archival_object, data=json.dumps(parent_record))
+
+    def add_child(self, parent, title="", level=""):
```
Hwesta, line 642: Does this need to be added to ATK to maintain API compatibility?

Hwesta:
:+1:

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo, line 642: It does; I'll look into doing the equivalent there.

mistydemeo:
Thinking about this, maybe this should be `PATCH` and `PUT` should be reserved for new records? Or should `PUT` handle both?

mistydemeo:
Updated to use POST for new records.

PR #282, opened 2015-09-12
copy_to_arrange: "objects" shouldn't be dragged onto arrange root

mistydemeo:
We normally disallow dragging individual files into arrange root, and only allow dragging directories. However, if the "objects" directory is dragged, we have special behaviour to skip that directory and instead copy over its paths as if each of those had been dragged instead. That special-casing happens _after_ we check whether the path being dragged over is a file or directory, though, which means "objects" passes the check - even though that would result in individual files being copied into the root of arrange.


Hwesta:
:+1:

Hwesta:
Cherry-picked to qa/1.5.x as 2817ee62e571391cc9c6a77dc43469ab5db40e69

PR #283, opened 2015-09-12
Correct dump file names

michal-ruzicka:
Filenames src/MCPServer/share/1:1.4.0 and src/MCPServer/share/1:1.4.1 seems to
be breaking dump files naming conventions and were renamed to
src/MCPServer/share/1.4.0 and src/MCPServer/share/1.4.1 respectively.

Moreover, colons (‘:’) in filenames are prohibited characters on Windows and as
such it is not possible to fully checkout working copy on Windows filesystem.


jhsimpson:
Thanks for this Pull Request. I tried checking out the code on a windows machine and I see what you are saying, these 2 sql dump files can't be named with a colon on a windows filesystem.

These files are used by the Ubuntu installer, dpkg, when installing Archivematica from .deb packages on an Ubuntu machine.  The colon in the file names is deliberate, it is required to match the package version number, as listed in the debian changelog.  There is an explanation of how this works [here](http://askubuntu.com/questions/441879/why-do-some-packages-have-extra-numbers-on-the-front-of-their-version-string).

Before version 1.4.0 of Archivematica, the numbering of Ubuntu packages did not include the epoch number (i.e. no colon), so this did not come up.

As a work around right now, on windows, I found that when unzipping the files, I click 'skip' and then these two files show up in my windows filesystem named 1_1.4.0 and 1_1.4.1.  That should work to let you at least check out the code on Windows in order to examine it, and as long as you don't try to check those 2 files back in, there should not be a problem.

I assume you are aware that Archivematica is not designed to actually run in a Windows environment, but this issue you have pointed out is still a problem, it should be possible to check out the code and work on it in a Windows environment, so a solution will need to be found.

I can think of a couple ways of fixing this, to resolve the issue you have identified without breaking the  Ubuntu packaging code.  One is to rename these two files as you have suggested, and also modify this file: https://github.com/artefactual/archivematica/blob/stable/1.4.x/src/MCPServer/debian/archivematica-mcp-server.install - that is the place where the sql dump files are copied from the .deb package, to the machine where Archivematica is being installed.  If this script was modified, to rename the files (in the .deb they are called 1.4.0 and on the target machine they become 1:1.4.0) then we might be in business.  

Another option is to move the code that does the debian packaging entirely outside of the Archivematica codebase.  This is a much bigger change, but it could be the first step towards making Archivematica work on other operating systems.  

hakamine has a repo that automates the process of building Archivematica packages for Ubuntu:
https://github.com/hakamine/am-packbuild  .  It might be possible to do the file renaming there instead. 

I will try testing the first solution I suggested, if you have other suggestions, feel free to post them.

michal-ruzicka:
Thanks for the explanation. Is the commit 6ddc794c593d3fccfbf0aaa74891270d436bfb04 added to the push request the correct solution?

jhsimpson:
Thanks for contributing that fix.  It does not quite work, the debian packaging system has its own Way of doing things.   

I tried a test in this commit, which works https://github.com/artefactual/archivematica/commit/1e4e4146041da2c2f5de520d364f1a498b0b5e6a

The install script has to be executed using dh-exec, which means making the install script executable, and modifying the syntax a bit.

This is documented in a link in that commit.

michal-ruzicka:
Thank you very much! I just used your solution in the pull request.

Hwesta:
These changes were in the packaging branch for 1.4.x (#240) and are now in the packaging branch for 1.5 (#380).  Thanks for the contribution!

PR #286, opened 2015-09-15
Upload-ArchivesSpace: Only send ArchivesSpace formats

Hwesta:
ArchivesSpace only accepts a set list of file formats. Attempt to match between FPR formats and the ones ArchivesSpace accepts, or send no format information.


jhsimpson:
I tested, this works, but only in conjuction with PR267 
https://github.com/artefactual/archivematica/pull/267

:+1:

Hwesta:
Cherry-picked into qa/1.5.x as 45e39190903ed7b995274ca18dab9738e8310772

PR #289, opened 2015-09-16
Decouple renderBacklogSearchForm from originals_browser

mistydemeo:
This will be used by the appraisal tab, so it needs to be able to

a) Look up arbitrary URLs,
b) Send the data to arbitrary places.

Note that the openInNewTab parameter I removed hasn't done anything since d5b83fa6.


Hwesta:
:+1:

PR #293, opened 2015-09-18
Add APIs to associate arranged SIPs with access system records

mistydemeo:
This supports a workflow where users can arrange and start SIPs from within an access system tree, instead of via the traditional SIP arrange panel.

When a user first drags records into a given access system record, Archivematica creates a backing SIP arrange directory, which is associated with the record in a new table in the database. All files dragged into that directory are copied into this directory using the normal SIP arrange APIs. The new APIs here sit in front of the SIP arrange APIs, allowing operations to be specified in terms of the access system identifier instead of the directory name.


mistydemeo:
Slightly updated, so that `copy_to_arrange` now also creates the arrange directory if it doesn't already exist.

The intention is for the arrange version of `copy_to_arrange` to be called whenever records are copied to the root of a tree, while the regular version of `copy_to_arrange` to be called when records are copied to subdirectories of an access tree.

Hwesta:
Tests for the new endpoints? test_filesystem_ajax.py has some structure to borrow.

mistydemeo:
I can add new tests! What's the timeframe on needing this merged for future dev/demoing, though? Do we need it merged before I get back?

Hwesta:
I was planning on making a qa/appraisal-tab to consolidate things into once branch until we can merge everything properly.

mistydemeo:
Added some basic tests. I was going to use VCR to handle the Storage Service ones, but realized that since they depend on database state for Storage Service configuration, it might be a bit iffy to handle those right now.

Hwesta:
:+1:

mistydemeo:
Minor change: added `move_within_arrange`.

Hwesta:
:+1:

PR #295, opened 2015-09-23
ArchivesSpace: find_by_field uses new endpoint

Hwesta:
find_by_field renamed to find_by_id and uses new ArchivesSpace plugin endpoint. This should return no more than 1 result, and no false positives.


mistydemeo:
Rename the ArchivistsToolkitClient at the same time, maybe?

:+1: otherwise

jhsimpson:
I think the ArchivistsToolkit change should happen in its own pull request.

mistydemeo:
Sounds good. :+1:

Hwesta:
> I think the ArchivistsToolkit change should happen in its own pull request.

The ATK change was made initially to keep the APIs of the ATK and ArchivesSpace classes in sync.  The change I amended into the commit does the same thing - renames the function to keep the APIs the same. The ATK version still raises a NotImplementedError.  @jhsimpson do you still want it in a separate PR?

mistydemeo:
:shipit: 

It's probably easier for us to actually implement this in ATK now, but we can reserve _that_ for a followup PR.

jhsimpson:
yes, this PR looks good to me as is.  I tested it and the new endpoint is working as expected.

:+1:

Hwesta:
Double-committed to qa/1.5.x as d12f3f1d351ae4a23824500c485b5fa7450ea181

PR #296, opened 2015-09-24
Django ORM fixes

Hwesta:
- Init Django in file_to_folder (only used in DSpace path)
- Make FPCommandOutput use legacy table, instead of creating its own. (@mistydemeo was this done intentionally? My problem was FPCommandOutput seem to disappear when queried at MySQL command line)

### DIFF
--- a/src/MCPClient/lib/clientScripts/file_to_folder.py
+++ b/src/MCPClient/lib/clientScripts/file_to_folder.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 import os
 import sys
 
+import django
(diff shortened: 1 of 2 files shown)


jhsimpson:
I think we need to keep the legacy table name, so existing records are not orphaned, so 

:+1:

Hwesta:
Double-committed to qa/1.5.x as 0a1152e10b215c0c86666a45e399368d99b205e6 and 21d223d06d6d9edf05552e48d9004f5a73134ac1

PR #297, opened 2015-09-24
Reduce MCP Server database usage

jhsimpson:
This branch adds a new dependency to the MCP Server, a python module called django-mysqlpool.  This module provides the ability to pool database connections (it uses SQLAlchemy to do that).  

The implementation here only adds connection pooling to the MCP Server.  Probably it should/could be added to the dashboard and MCP Client packages as well.  90% of the database load in a typical Archivematica installation seems to come from the MCPServer, so this is the low hanging fruit.

This particular solution is mysql specific, which is not a factor at the moment, but admittedly it does sort of defeat the purpose of using an ORM if you are still tightly coupled to one db backend.  

Basic testing has shown a reduction in time to process a transfer of maybe 8 to 10%.  The size of the machine used is a big variable here, ymmv.  Mostly this reduces the number of database connections required by the MCPServer, from dozens or hundreds down to less than 10.

This pull request adds some changes which could be considered optional.  Instead of changing the dashboards django settings, a new settings module is added to the MCPServer, which includes the smallest possible settings file that still seems to work.  This seems important because the MCP Server is using the Django ORM, in order to be able to access Models, but it is not a django app.  MCP Server does not need most of Django.

### DIFF
--- a/src/MCPServer/lib/archivematicaMCP.py
+++ b/src/MCPServer/lib/archivematicaMCP.py
@@ -45,7 +45,7 @@ import time
 import uuid
 
 import django
-sys.path.append('/usr/share/archivematica/dashboard')
+sys.path.append('/usr/lib/archivematica/MCPServer')
 os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
 django.setup()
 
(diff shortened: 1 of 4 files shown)


mistydemeo:
> This particular solution is mysql specific, which is not a factor at the moment, but admittedly it does sort of defeat the purpose of using an ORM if you are still tightly coupled to one db backend. 

I think this is an important first step, and we can build on this with our own decorators in the future to handle it in DB-specific ways for other DBs if we need to. This doesn't entangle the code too tightly with mysql, I don't think.

sevein:
Replaced by #305.

PR #298, opened 2015-09-24
Logging: silence verbose libraries in client script logs

Hwesta:
Convert get_script_logger to use a dict for config. Keep formatters/handler and loggers for passed-in root ('archivematica') and script logger.  Add a root logger that only logs WARNING and higher from other libraries. Notably this silences elasticsearch's verbose INFO logs.

Also removes duplicate logs (from both root_logger and logger).


mistydemeo:
:+1:

Hwesta:
Double-committed to qa/1.5.x as e40f2099813f6374b5717836c289e2dc34bf0479

PR #300, opened 2015-09-26
Fix Archivists Toolkit DIP Upload for Unix

jhsimpson:
When Archivists Toolkit is run on a unix server, the table names are case sensitive.  This PR fixes all sql statements in the upload-archiviststoolkit.py client script to work on both Windows and Unix servers.

This also adds empty values to a couple of columns in database inserts into Archivists Toolkit, because in some ATK deployments, those columns are not-nullable.  Archivematica doesn't have a value to put in, so an empty string ('') is used.

Small refactor to re-use one of the sql statements, combining 3 separate statements into 1.


Hwesta:
:+1:

jhsimpson:
merged into stable/1.4.x, and then cherry-picked into qa/1.5.x as:
5aa36e55f761b08c9a1c2be22fa87f9e8898e596
and
2ae9678a8d11ecb1c9850ec421855d7523da1923

Also cherry-picked into qa/1.x as:
ee1718087eda28c137f1436fd2b8d9e2360abe49
and
c39b618c4c6c673b9bea5d155e97535254aa7f0d

PR #301, opened 2015-09-28
SIP arrange APIs: various bugfixes

mistydemeo:
- Handle `storage_service.ResourceNotFound` exception gracefully, returning JSON-formatted responses instead of 500ing
- Return 404 when listing the contents of arrange if the requested path doesn't exist, instead of reporting that there are no files within the requested path


mistydemeo:
Also added a test for the new behaviour in the second commit.

Hwesta:
:+1:

PR #302, opened 2015-09-30
change uuid used in determing file format

jhsimpson:
the access file does not have file identification info in the database
use the original file uuid instead.

I am not 100% sure this is going to be correct in all circumstances.  I tested on a client site and it works as I want it to.


Hwesta:
:+1: Very client-specific but will be modified by in-progress work anyway.

PR #303, opened 2015-10-01
Logging (again/still)

Hwesta:
Tweak the logging dictConfig for the dashboard to catch or silence things better.  This should silence archivematica.mcp logs, log all other archivematica logs at DEBUG, and elasticsearch/archivesspace/archiviststoolkit at INFO. Everything else gets logged at WARNING.  Adds a dashboard.debug.log.

MCPClient also gets a MCPClient.debug.log file, and logs all archivematica logs, and everything else at WARNING

Tweaked MCPServer logging to log all archivematica logs and everything else at WARNING. No longer silences py.warnings.

Tweaked client script logging to remove duplicate handles.

This may get further work to read logging config from /etc/ instead of from code.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -211,10 +211,46 @@ def startThreads(t=1):
         t.daemon = True
         t.start()
 
+LOGGING_CONFIG = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'formatters': {
+        'detailed': {
+            'format': '%(levelname)-8s  %(asctime)s  %(name)s:%(module)s:%(funcName)s:%(lineno)d:  %(message)s',
+            'datefmt': '%Y-%m-%d %H:%M:%S'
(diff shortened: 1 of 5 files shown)


sevein:
Looking very good! Precisely, I'd love the logging to be configurable. If you ever do that, please also consider environment variables support in addition to /etc, e.g. I'd like to see something like `os.environ['MCPSERVER_LOG_LEVEL']` and `os.environ['MCPSERVER_LOG_HANDLER']` being prioritized over the same options in `/etc` or the application defaults.
We could have a number of predefined handlers that the user could choose via `MCPSERVER_LOG_HANDLER`, e.g. I want to redirect to `stderr` or `stdout` because that's ideal in containers.

mistydemeo:
Haven't tested, but looks good. :+1:

Hwesta:
Also merged into qa/1.5.x as 27e55620f7035311feca9e238e65a61b209ee5f4

PR #304, opened 2015-10-06
Close unusable database connections, refs #9021

sevein:
Wrap threaded functions to ensure that their database connections are only
recycled if they haven't outlived its maximum age or are in a recoverable
state, like Django does before and after HTTP requests.

It relies on `BaseDatabaseWrapper.close_if_unusable_or_obsolete()` in
`django.db.backends.base`.

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -26,6 +26,9 @@ import lxml.etree as etree
 import os
 import re
 
+from django.db import close_old_connections
+
+
 REQUIRED_DIRECTORIES = [
     "logs",
     "logs/fileMeta",
@@ -211,9 +214,7 @@ def close_db_connections(fn):
     Similar to Django's behavior within HTTP request cycles.
     """
     def wrapped(*args,  **kwargs):
-        from django.db import close_old_connections
         try:
-            close_old_connections()
             return fn(*args, **kwargs)
         finally:
             close_old_connections()
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -200,3 +200,21 @@ def create_structured_directory(basepath, manual_normalization=False, printing=F
     create_directories(REQUIRED_DIRECTORIES, basepath=basepath, printing=printing)
     if manual_normalization:
         create_directories(MANUAL_NORMALIZATION_DIRECTORIES, basepath=basepath, printing=printing)
+
+
+def close_db_connections(fn):
+    """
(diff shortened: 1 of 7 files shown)


src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -200,3 +200,21 @@ def create_structured_directory(basepath, manual_normalization=False, printing=F
...
+def close_db_connections(fn):
+    """
+    Decorator to wrap a threaded function in a try-catch ensuring that the
+    database connections are closed if unrecoverable errors have occurred or if
+    it outlived its maximum age.
+
+    Similar to Django's behavior within HTTP request cycles.
+    """
+    def wrapped(*args,  **kwargs):
+        from django.db import close_old_connections
```
Hwesta: Imports should all be at the top of the file.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -200,3 +200,21 @@ def create_structured_directory(basepath, manual_normalization=False, printing=F
...
+        try:
+            close_old_connections()
```
Hwesta: Why call this before the wrapped function? Isn't after enough?

src/archivematicaCommon/lib/archivematicaFunctions.py
sevein: That was intentional because I was trying to make sure that the module is not being cached by Python across multiple threads... but maybe that's going to happen regardless where the import statement is or just doesn't make any difference. I don't have much understanding on the internals.
But I tried with the import on the top and it seems to be working too.

src/archivematicaCommon/lib/archivematicaFunctions.py
sevein: I was trying to follow [Django's cleaning process before and after HTTP requests](https://github.com/django/django/blob/master/django/db/__init__.py#L60-L66) to the letter, but on a second thought I think you are totally right as we are not reusing threads in the same way that uWSGI is reusing processes. I'm guessing Django has its own mechanism to avoid sharing connections across threads.
I've tried without the first call and it seemed to work. More testing would be ideal.

sevein:
:+1: thanks

Hwesta:
:+1: Code changes look fine, and the investigation around them looks solid.

sevein:
@Hwesta, I'll spend some time tomorrow making sure that in #297 the connections are returned to the pool at the end of each thread using django_mysqlpool's `@auto_close_db` in the same places where I used `@close_db_connections` here. That may leave us with Justin's pool solution and no exhaustion of the pool because of threads not freeing the connections.

PR #306, opened 2015-10-13
Init Django before using ORM, refs #9040

sevein:
Redmine ticket: https://projects.artefactual.com/issues/9040

### DIFF
--- a/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
+++ b/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
@@ -9,13 +9,15 @@ import uuid
 
 import archivematicaXMLNamesSpace as ns
 
-# archivematicaCommon
-import fileOperations
-import databaseFunctions
-
+import django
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/parse_mets_to_db.py
```diff
@@ -13,6 +13,8 @@
 import fileOperations
 import databaseFunctions
 
+import django
+django.setup()
```
Hwesta, line 13: Does this need to happen before `fileOperations`/`databaseFunctions` is imported?  I usually moved the Django imports & init above the archivematicaCommon imports.

Hwesta:
:+1:

src/MCPClient/lib/clientScripts/parse_mets_to_db.py
sevein, line 13: That makes sense, I'll do it.

PR #307, opened 2015-10-20
Enable use of repository in ASpace config

jhsimpson:
refs #9075

The ArchivesSpace DIP Upload settings page,
in the admistration tab of the dashboard,
allows a user to set a value for the repository
to connect to in ArchivesSpace.

This value is now saved to the ReplacementDic
and used in the matching gui.

### DIFF
--- a/src/MCPClient/lib/clientScripts/dip_generation_helper.py
+++ b/src/MCPClient/lib/clientScripts/dip_generation_helper.py
@@ -30,7 +30,8 @@ def create_archivesspace_client():
             host=config['%host%'],
             port=config['%port%'],
             user=config['%user%'],
-            passwd=config['%passwd%']
+            passwd=config['%passwd%'],
+            repository=config['%repository%']
         )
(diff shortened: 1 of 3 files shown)


jhsimpson:
I discovered that this pull request does not address the use case where the archives space matching gui is not used (i.e. when a csv file is supplied with a transfer that includes matches between digital objects and archival objects).  The dip_generation_helper.py script in MCP Client needs to be changed to honour the repository parameter still.

mistydemeo:
Good catch! :+1:

> The dip_generation_helper.py script in MCP Client needs to be changed to honour the repository parameter still.

Do we want to handle it by having it fetch that setting from the dashboard, or by having the `arguments` string of StandardsTasksConfig be rewritten by the form in the same way as with the AtoM settings form?

jhsimpson:
added repository_id to dip_generation_helper.py, tested, it is working now.

mistydemeo:
:+1:

PR #308, opened 2015-10-23
ArchivesSpace DIP Upload enhancements

jhsimpson:
based on Dev/issue 9093 aspace restrictions.

This commit changes the logic slightly in setting restrictions on digital objects in ArchivesSpace, when based on PREMIS.
It also sets the publish and restrictions flags on both digital objects and file versions.
Finally, notes are copied from the parent archival object to the digital object.

### DIFF
--- a/src/MCPClient/lib/clientScripts/upload-archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload-archivesspace.py
@@ -123,13 +123,33 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
 
         # Get file & format info
         # Client wants access copy info
+
+        original_name = ''
+        # Get file & format info
+        # Client wants access copy info
+        try:
+            original_file = File.objects.get(
+                filegrpuse='original',
+                uuid=uuid
+            )
+        except (File.DoesNotExist, File.MultipleObjectsReturned):
+            original_name=''
+            size = format_name = format_version = None
+        else:
+            # Set some variables based on the original, we will override most
+            # of these if there is an access derivative
+            size = os.path.getsize(f)
+            fv = FormatVersion.objects.get(fileformatversion__file_uuid=uuid)
+            format_version = fv.description
+            format_name = fv.format.description
+            original_name = os.path.basename(original_file.originallocation)
         try:
             access_file = File.objects.get(
                 filegrpuse='access',
                 original_file_set__source_file=uuid
             )
         except (File.DoesNotExist, File.MultipleObjectsReturned):
-            size = format_name = format_version = None
+            pass  # Just use original file info
         else:
             # HACK remove DIP from the path because create DIP doesn't
             access_file_path = access_file.currentlocation.replace('%SIPDirectory%DIP/', dip_location)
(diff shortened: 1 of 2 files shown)


src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -486,27 +487,46 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
...
-                "content": [dashboard_uuid]
+                "content": [dashboard_uuid],
+                "publish": True,
             }],
             "restrictions": restricted,
             "subjects": parent_record['subjects'],
             "linked_agents": parent_record['linked_agents'],
         }
 
+        i = 0
```
mistydemeo: Is this count used?

If you want to get a count when iterating over an enumerable, you can use the `enumerate` function:

``` python
for i, pnote in enumerate(parent_record["notes"]):
    ...
```

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -486,27 +487,46 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
...
+        for pnote in parent_record["notes"]:
+            content = []
+            for subnote in pnote["subnotes"]:
+                content.append(subnote["content"])
+
+            new_object["notes"].append({
+                "jsonmodel_type": "note_digital_object",
+                "type": pnote["type"],
```
mistydemeo: Like we discussed in person today, not all note types supported by `note_multipart` are supported by `note_digital_object`, so ArchivesSpace is likely to reject some notes copied to records this way. The relevant types are in the `note_digital_object_type` and `note_multipart_type` enums, which are probably fetchable using the API.

You can list all enumerations (and their values) via the `/config/enumerations` API. `note_digital_object_type` is located at `/config/enumerations/44`, and `note_multipart_type` is located at `/config/enumerations/45`.

src/archivematicaCommon/lib/archivesspace/client.py
jhsimpson: That i variable is not actually needed.  In an rough version of the code, that I did not commit, the count was being used, but I realized it wasn't necessary, but forgot to remove it.

I have removed it in a new commit now

src/archivematicaCommon/lib/archivesspace/client.py
jhsimpson: In d8f2090 , I have added a hardcoded array, copied from the ArchivesSpace code, rather than connect to the ArchivesSpace api and check.  I did this in order for testing, I'll need to go back now and change this to an api call.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -486,27 +487,50 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
...
+                content.append(subnote["content"])
+
+            new_object["notes"].append({
+                "jsonmodel_type": "note_digital_object",
+                "type": dnote,
+                "label": pnote["label"],
+                "content": content,
+                "publish": pnote["publish"],
+            })
+
```
mistydemeo: Nitpick, but it looks like this line has some whitespace on it.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo: It's probably faster with the hardcoded array, but I wonder if it's possible for that value to change... maybe it is better to fetch it.

jhsimpson:
I think I removed the whitespace you were referring to.

src/archivematicaCommon/lib/archivesspace/client.py
jhsimpson: It would be better to fetch the array, I think, but the funding to work on this feature has run out.  I will ask the client if they are interested in optimizing this, for now I vote for it to stay as a hardcoded array.

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -123,13 +123,33 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
...
+        try:
+            original_file = File.objects.get(
+                filegrpuse='original',
+                uuid=uuid
+            )
+        except (File.DoesNotExist, File.MultipleObjectsReturned):
+            original_name=''
+            size = format_name = format_version = None
+        else:
+            #set some variables based on the original, we will override most ofthese
```
mistydemeo: Minor nitpick, but comments should have a space between the hash and teh text. (It's worth configuring a linter, it'll catch stuff like this.)

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -123,13 +123,33 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
...
+            format_name = fv.format.description
+            original_name = os.path.basename(original_file.originallocation)
         try:
             access_file = File.objects.get(
                 filegrpuse='access',
                 original_file_set__source_file=uuid
             )
         except (File.DoesNotExist, File.MultipleObjectsReturned):
-            size = format_name = format_version = None
+            pass #just use original file info
```
mistydemeo: Similarly, two spaces after text before the comment for an inline comment.

mistydemeo:
:+1: aside from comments!

jhsimpson:
fixed spacing in comments, will merge into qa/1.5.x and qa/1.x now

jhsimpson:
merged into qa/1.5.x and cherry-picked into qa/1.x as 81fb134c79ddf10ac3615393350bf86110bf614d

PR #309, opened 2015-10-26
Appraisal tab supporting dev

Hwesta:
Add support for artefactual-labs/appraisal-tab#93 and artefactual-labs/appraisal-tab#94
- Return paths with ArchivesSpace nodes & on directory creation
- Add DELETE record to ArchivesSpace, and access endpoint


src/dashboard/src/media/js/file_browser.js
```diff
@@ -45,7 +45,7 @@ var FileExplorer = fileBrowser.FileExplorer.extend({
     }
 
     // the file explorer base class doesn't handle deletion
-    this.ajaxDeleteUrl = this.options.ajaxDeleteUrl || '/filesystem/delete/';
+    this.ajaxDeleteUrl = this.options.ajaxDeleteUrl || undefined;
```
mistydemeo, line 48: Won't the default be `undefined` if this is falsy?

mistydemeo:
Can you add an unimplemented version of the new ArchivesSpaceClient method to ArchivistsToolkitClient so we remember to add it later?

src/dashboard/src/components/access/views.py
```diff
@@ -178,8 +195,22 @@ def record_children(client, request, system='', record_id=''):
...
+def _get_sip_arrange_children(record, system):
+    """ Recursively check for SIPArrange associations. """
+    try:
+        mapping = SIPArrangeAccessMapping.objects.get(system=system, identifier=record['id'])
+        record['path'] = mapping.arrange_path
+        record['children'] = record['children'] or []
+        record['has_children'] = True
+    except (SIPArrangeAccessMapping.MultipleObjectsReturned, SIPArrangeAccessMapping.DoesNotExist):
+        pass
+    if record['children']:  # record['children'] may be False
```
mistydemeo, line 210: Maybe unconditionally perform `record['children'] = record['children'] or []` to skip the `False` check below?

mistydemeo, line 210: This'd also fix #9098.

Hwesta, line 210: That would make all ArchivesSpace children directories, not files, which would probably be a good thing, so sure.

src/dashboard/src/components/access/views.py
```diff
@@ -94,10 +94,17 @@ def all_records(client, request, system=''):
...
-                                                         page_size=page_size))
+    records = client.find_collections(
+        search_pattern=search_pattern,
+        identifier=identifier,
+        page=page,
+        page_size=page_size,
+    )
+
+    for r in records:
+        r = _get_sip_arrange_children(r, system)
```
mistydemeo: Does this actually do anything? It's just reassigning the loop variable, not changing the entry in `records`, right?

src/dashboard/src/components/access/views.py
```diff
@@ -178,8 +195,22 @@ def record_children(client, request, system='', record_id=''):
...
+    try:
+        mapping = SIPArrangeAccessMapping.objects.get(system=system, identifier=record['id'])
+        record['path'] = mapping.arrange_path
+        record['children'] = record['children'] or []
+        record['has_children'] = True
+    except (SIPArrangeAccessMapping.MultipleObjectsReturned, SIPArrangeAccessMapping.DoesNotExist):
+        pass
+    if record['children']:  # record['children'] may be False
+        for r in record['children']:
+            r = _get_sip_arrange_children(r, system)
```
mistydemeo: Ditto.

src/dashboard/src/components/access/views.py
mistydemeo: Should this be:

``` python
for i, r in enumerate(records):
    records[i] = _get_sip_arrange_children(r, system)
```

?

mistydemeo: I think this worked by accident because `_get_sip_arrange_children` mutates the passed record, but I'll update it to reassign the contents of `record` for clarity.

PR #310, opened 2015-10-26
Improvements to ArchivesSpaceClient suggested by Hudson Molonglo

mistydemeo:
- Don't allow sessions to timeout, and remove timeout-related code
- Properly restrict results returned via accession number search
- Use `/tree` instead of `/search` to determine existence of children, which avoids a potential place that we could end up with outdated information


Hwesta:
:+1:

PR #312, opened 2015-10-27
Access move_within_arrange: Don't strip leading slash from filepath

mistydemeo:


mistydemeo:
This accidentally removed leading slashes for consistency with `copy_to_arrange`; however, the `filesystem_ajax` version of this function doesn't perform stripping of leading slashes and rejected values being passed in.

Hwesta:
:+1:

PR #313, opened 2015-10-28
Support multiple processing configs, refs #9105

sevein:
Redmine ticket: https://projects.artefactual.com/issues/9105

The processing configuration form is now a subclass of Django's `Form`.

An upcoming commit should expose the different configurations available via the API.

### DIFF
--- a/src/dashboard/src/components/administration/forms.py
+++ b/src/dashboard/src/components/administration/forms.py
@@ -15,14 +15,23 @@
 # You should have received a copy of the GNU General Public License
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
+import sys
+from lxml import etree
+from collections import OrderedDict
+
 from django import forms
 from django.conf import settings
-from django.forms import ModelForm
 from django.forms.widgets import TextInput, RadioSelect, CheckboxInput
+from django.utils.translation import ugettext as _
 
 from components import helpers
 from main import models
 from components.administration.models import ArchivistsToolkitConfig, ArchivesSpaceConfig
+
+sys.path.append('/usr/lib/archivematica/archivematicaCommon')
+import storageService as storage_service
+
 
 class AgentForm(forms.ModelForm):
     class Meta:
@@ -32,6 +41,7 @@ class AgentForm(forms.ModelForm):
             'identifiervalue': TextInput(attrs=settings.INPUT_ATTRS),
             'name': TextInput(attrs=settings.INPUT_ATTRS),
         }
+
 
 class SettingsForm(forms.Form):
     def __init__(self, *args, **kwargs):
@@ -64,7 +74,7 @@ class StorageSettingsForm(SettingsForm):
         label="Full URL of the storage service")
 
 
-class ArchivistsToolkitConfigForm(ModelForm):
+class ArchivistsToolkitConfigForm(forms.ModelForm):
     class Meta:
         model = ArchivistsToolkitConfig
         fields = ('host', 'port', 'dbname', 'dbuser', 'dbpass', 'atuser', 'premis', 'ead_actuate', 'ead_show', 'object_type', 'use_statement', 'uri_prefix', 'access_conditions', 'use_conditions')
@@ -86,7 +96,7 @@ class ArchivistsToolkitConfigForm(ModelForm):
         }
 
 
-class ArchivesSpaceConfigForm(ModelForm):
+class ArchivesSpaceConfigForm(forms.ModelForm):
     class Meta:
         model = ArchivesSpaceConfig
         fields = ('host', 'port', 'user', 'passwd', 'premis', 'xlink_actuate', 'xlink_show', 'use_statement', 'object_type', 'access_conditions', 'use_conditions', 'uri_prefix', 'repository')
@@ -105,6 +115,7 @@ class ArchivesSpaceConfigForm(ModelForm):
             'uri_prefix': TextInput(attrs=settings.INPUT_ATTRS),
             'repository': TextInput(attrs=settings.INPUT_ATTRS),
         }
+
 
 class AtomDipUploadSettingsForm(SettingsForm):
     dip_upload_atom_url = forms.CharField(required=True,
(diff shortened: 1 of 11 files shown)


src/dashboard/src/components/administration/forms.py
```diff
@@ -15,15 +15,23 @@
...
 
+import os
+import sys
+from lxml import etree
+
 from django import forms
 from django.conf import settings
-from django.forms import ModelForm
 from django.forms.widgets import TextInput, RadioSelect, CheckboxInput
+from django.utils.translation import ugettext as _
```
Hwesta, line 26: :sparkling_heart:

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+    try:
+        dirs = storage_service.get_location(purpose=purpose)
+        if len(dirs) == 0:
+            raise Exception('Storage server improperly configured.')
+    except Exception:
+        dirs = []
+    return dirs
+
+
+class PreconfiguredChoices:
```
Hwesta: Classes should inherit from object: `class PreconfiguredChoices(object):`

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+    """
+    Encode processing configuration XML documents and optionally write to disk.
+    """
+    def __init__(self, config_path):
+        self.xml = etree.Element('processingMCP')
+        self.choices = etree.Element('preconfiguredChoices')
```
Hwesta: Tip: etree.SubElement(parent, ...) may make this easier/cleaner.

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+        self.config_path = config_path
+
+    def add_choice(self, applies_to_text, go_to_chain_text, delay_duration=None):
+        choice = etree.Element('preconfiguredChoice')
+
+        applies_to = etree.Element('appliesTo')
+        applies_to.text = applies_to_text
```
Hwesta: Tip: `etree.Element('foo').text = 'text'` works, though you may not think it's cleaner. :)

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -15,404 +15,64 @@
...
-
-        return redirect('components.administration.views.processing')
+        form = ProcessingConfigurationForm(request.POST)
+        if form.is_valid():
+            try:
+                form.save_config()
+            except Exception as e:
+                logger.exception(str(e))
+                messages.error(request, 'Processing configuraiton could not be saved!')
+            finally:
```
Hwesta: `finally:` is always run, even if an exception is raised. This should be `else:` or move into the `try` block.

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -15,404 +15,64 @@
...
-
-        messages.info(request, 'Saved!')
-
-        return redirect('components.administration.views.processing')
+        form = ProcessingConfigurationForm(request.POST)
+        if form.is_valid():
+            try:
+                form.save_config()
+            except Exception as e:
+                logger.exception(str(e))
```
Hwesta: `logger.exception` already includes a traceback for the exception. Maybe make the error message more useful, like 'Failed to save processing config' ie why this traceback happened.

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+            else:
+                choices = opts['choices'] = list(self.EMPTY_CHOICES)
+                if ftype == 'boolean':
+                    if 'yes_option' in field:
+                        choices.append((field['yes_option'], 'Yes'))
+                    if 'no_option' in field:
+                        choices.append((field['no_option'], 'No'))
+                elif ftype == 'chain_choice':
+                    link = models.MicroServiceChainLink.objects.get(pk=choice_uuid)
+                    chain_choices = models.MicroServiceChainChoice.objects.filter(choiceavailableatlink_id=link.pk)
```
Hwesta: Tip: Don't need to fetch `link`, can just be `models.MicroServiceChainChoice.objects.filter(choiceavailableatlink_id=choice_uuid)`

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+                    ignored_choices = field.get('ignored_choices')
+                    for item in chain_choices:
+                        chain = item.chainavailable
+                        if ignored_choices is not None:
```
Hwesta: Tip: If above is `ignored_choices = field.get('ignored_choices', [])` then you don't need to check 'if ignore_choices is None'.  Second parameter `.get()` is default value.

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+    def load_config(self, name):
+        """
+        Bound the choices found in the XML document to the form fields.
+        """
+        self.fields['name'].initial = name
+        self.fields['name'].widget.attrs['readonly'] = 'readonly'
+        config_path = os.path.join(helpers.processing_config_path(), '{}ProcessingMCP.xml'.format(name))
+        with open(config_path, 'r') as f:
+            xml = f.read()
+            root = etree.fromstring(xml)
```
Hwesta: Tip: Can combine these 2 lines into `root = etree.parse(f)`  The rest of the block doesn't have to be within the `with` block, since the file's already been read from.

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+            for choice in root.findall('.//preconfiguredChoice'):
+                applies_to = choice.find('appliesTo').text
+                go_to_chain = choice.find('goToChain').text
```
Hwesta: Tip: [.findtext()](http://lxml.de/tutorial.html#elementpath) is pretty cool.

src/dashboard/src/templates/administration/processing_edit.html
```diff
@@ -0,0 +1,51 @@
...
+
+    <div class="span12">
+
+      <h3>{% trans "Processing configuration" %}</h3>
+      
+      {% if new %}
+        <form class="form-stacked" method="post">
+      {% else %}
+        <form class="form-stacked" method="post">
+      {% endif %}
```
Hwesta: These two cases appear to be the same?

src/dashboard/src/templates/administration/processing_edit.html
```diff
@@ -0,0 +1,51 @@
...
+            <tr>
+              <td>
+                {{ field.label_tag }}
+                {{ field.errors }}
+              </td>
+              <td>{{ field }}</td>
+            </tr>
+          {% endfor %}
+
+        </table>
```
Hwesta, line 34: Other options for display include using the _form template ,or just {{form}}, but this look fine too.

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +139,306 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
         }
+
+
+class ProcessingConfigurationForm(forms.Form):
+    """
+    Build processing configuration form bounded to a processingMCP document.
+    """
+
+    # The available processing fields indexed by choice_uuid.
+    processing_fields = {
```
Hwesta: Idea: It would be nice if the processing config options were in the same order they came up in processing.  If this was an [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) and the options were reordered, would that happen?

Hwesta:
Nicely done!  Most of my feedback is tips on how to make it cleaner.

Tests for this?

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +140,277 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+    except Exception:
+        dirs = []
+    return dirs
+
+
+class PreconfiguredChoices(object):
+    """
+    Encode processing configuration XML documents and optionally write to disk.
+    """
+    def __init__(self, config_path):
```
mistydemeo: Nitpick, but since this is only used by `save`, maybe defer asking for a `config_path` until the document actually gets saved?

src/dashboard/src/components/administration/forms.py
```diff
@@ -129,10 +140,277 @@ class AtomDipUploadSettingsForm(SettingsForm):
...
+            self.choices.append(comment)
+        choice = etree.SubElement(self.choices, 'preconfiguredChoice')
+        etree.SubElement(choice, 'appliesTo').text = applies_to_text
+        etree.SubElement(choice, 'goToChain').text = go_to_chain_text
+        if delay_duration is not None:
+            etree.SubElement(choice, 'delay', {'unitCtime': 'yes'}).text = delay_duration
+
+    def save(self):
+        with open(self.config_path, 'w') as f:
+            f.write(etree.tostring(self.xml, pretty_print=True))
```
mistydemeo: Newline at the end of the file?

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -21,12 +21,14 @@
...
+    FULL = 'full'
     REINGEST_CHOICES = (
         (METADATA_ONLY, 'Re-ingest metadata only'),
-        (OBJECTS, 'Re-ingest metadata and objects')
+        (OBJECTS, 'Re-ingest metadata and objects for DIP generation'),
+        (FULL, 'Full re-ingest'),
     )
-    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES,  widget=forms.RadioSelect, required=True)
-
+    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES,  widget=forms.RadioSelect, required=True)
```
mistydemeo: Newline here too?

mistydemeo:
Looks good! :+1: Tests for some of the new stuff would be nice if you're able.

PR #314, opened 2015-10-28
Return notes from ArchivesSpaceClient and allow them to be edited

mistydemeo:
This currently supports a subset of ArchivesSpace's notes functionality, supporting a single block of text in a single note. (ArchivesSpace allows arbitrary numbers of notes, and each note may contain arbitrary numbers of blocks of text.)

### DIFF
--- a/src/archivematicaCommon/lib/archivesspace/client.py
+++ b/src/archivematicaCommon/lib/archivesspace/client.py
@@ -158,6 +158,7 @@ class ArchivesSpaceClient(object):
         Currently supported fields are:
             * title
             * targetfield
+            * notes
 
         :raises ValueError: if the 'id' field isn't specified, or no fields to edit were specified.
         """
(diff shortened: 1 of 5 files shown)


mistydemeo:
Updated to allow start and end dates, and date expressions, to be edited.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -118,6 +118,26 @@ def _delete(self, url, params={}, expected_response=200):
                              params=params,
                              expected_response=expected_response)
 
+    def _format_notes(self, record):
+        """
+        Extracts notes from a record and reformats them in a simplified format.
+        """
+        if len(record['notes']) == 0:
```
Hwesta: Are notes guaranteed to exist? Should this be `.get('notes', [])`? Can we check for truthiness?

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -165,6 +189,47 @@ def edit_record(self, new_record):
...
+            # If the record already has at least one note, then replace the first note
+            # within that record with this one.
+            if len(record['notes']) == 0:
+                record['notes'] = [new_note]
+            else:
+                record['notes'][0] = new_note
+
+            fields_updated = True
+
+        if 'start_date' in new_record:
```
Hwesta, line 208: end_date and date_expression are meaningless if there's no start_date?

mistydemeo, line 208: That's right. `start_date` is mandatory.

src/archivematicaCommon/lib/archivesspace/client.py
mistydemeo: Records are guaranteed to contain a `notes` property. Could check truthiness though.

Hwesta:
Tests for new functionality?

:+1:

mistydemeo:
Added a test!

Hwesta:
:+1:

PR #315, opened 2015-10-30
Backlog tab & enhancements

MikeFE:
- Add a backlog tab
- Allow the user to search for backlog transfers
- Allow user to view/download specific files in backlog transfers
- Allow user to download entire backlog transfer package
- Allow user to make delete requests for backlog transfers


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -234,74 +239,91 @@ def _sortable_string_field_specification(field_name):
         }
     }
 
-def set_up_mapping(conn, index):
-    machine_readable_field_spec = {
-        'type':  'string',
-        'index': 'not_analyzed'
+def set_up_mapping_aip_index(conn):
```
mistydemeo, line 242: Really appreciate splitting these up into separate functions.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -991,3 +1069,20 @@ def normalize_results_dict(d):
     This normalizes the dict by replacing the arrays with their first value.
     """
     return {k: v[0] for k, v in d['fields'].iteritems()}
+
+def augment_raw_search_results(raw_results):
+    """
+    This function takes JSON returned by an ES query and removes some superfluous data from the structure.
```
mistydemeo: Maybe be more specific and say that it returns the source document for each result?

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,216 @@
...
+    but it appears necessary as the storage service doesn't talk directly to ES.
+
+    :return: None
+    """
+    conn = elasticSearchFunctions.connect_and_create_index('transfers')
+
+    query = {
+        'query': {
+            'bool': {
+                'should': {
```
mistydemeo: Should this be a `must` clause instead of a `should`?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -234,74 +239,91 @@ def _sortable_string_field_specification(field_name):
...
-                'type'      : 'nested',
-                'properties': {
-                    'puid'  : machine_readable_field_spec,
-                    'format': {'type': 'string'},
-                    'group' : {'type': 'string'},
-                }
+    LOGGER.info('Creating AIP mapping...')
+    conn.indices.put_mapping(
+        doc_type='aip',
+        body={'aip': {'date_detection': True, 'properties': mapping}},
```
Hwesta, line 252: This looks like it duplicates the index creation a couple lines earlier. What's the significance of creating the aip/aips index sans mapping, and then with mapping?

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    A JSON end point that returns results for various backlog transfers and their files.
+
+    :param request: The Django request object
+    :return: A JSON object including required metadata for the datatable and the backlog search results.
+    """
+    # get search parameters from request
+    queries, ops, fields, types = advanced_search.search_parameter_prep(request)
+
+    file_mode = True if request.GET.get('file_mode') == 'true' else False
```
Hwesta: Tip: Same as `file_mode = request.GET.get('file_mode') == 'true'`

Could this be replace by just using `doc_type = 'transferfile' if request.GET.get('file_mode') == 'true' else 'transfer'` from below? `table_columns` in the above function could be a dict.

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,216 @@
...
+    :param file_mode: Whether we're looking at transfers or transfer files
+    :return: The ES document property name corresponding to the column index in the data table.
+    """
+    table_columns = (
+        ('name', 'uuid', 'file_count', 'ingest_date', None),     # Transfers are being displayed
+        ('filename', 'sipuuid', None)                            # Transfer files are being displayed
+    )
+
+    if index < 0 or index >= len(table_columns[file_mode]):
+        raise IndexError('Column index specified is invalid, got {}'.format(index))
```
Hwesta: Do we want to handle this case? Perhaps by setting it to the max/min value? Or is that confusing behavior?

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,216 @@
...
+            index='transfers',
+            doc_type=doc_type,
+            body=query,
+            from_=start,
+            size=page_size,
+            sort=order_by + ':' + sort_direction if order_by else ''
+        )
+
+    except Exception as e:
+        logger.exception('Error accessing index: {}'.format(e))
```
Hwesta: `logger.exception` prints the full stacktrace for the exception, so printing it in the message is unnecessary.  The descriptive message of what was happening when the exception happened is useful.

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,216 @@
...
+
+def download(request, uuid):
+    """
+    Download a package from a requested backlog transfer.
+
+    :param request: The Django request object
+    :param uuid: UUID for the transfer we're downloading the package from
+    :return: Respond with a TAR'd version of the requested package
+    """
+    return HttpResponseRedirect(storage_service.download_file_url(uuid))
```
Hwesta: Point this at `filesystem_ajax.views.download_ss` instead since that one does streaming, and keeps all the SS-downloading in one place. Perhaps remove this and just use the filesystem_ajax one?

src/dashboard/src/templates/backlog/delete_request.html
```diff
@@ -0,0 +1,21 @@
+{% extends "layout_fluid.html" %}
```
Hwesta: This is almost identical to `archival_storage/delete_request.html`.  Can `archival_storage/delete_request.html` be moved to the `templates` dir and reused? Seems like a useful common template.

Hwesta:
Nicely done, especially with the docstrings and cleanup. 

Is this still WIP, or can those tags be removed?

MikeFE:
Thanks for the suggestions. I'll work on implementing them today and will remove the tag then.

src/dashboard/src/components/backlog/views.py
MikeFE: That's funny, I'm not sure why I was so verbose with that line

src/dashboard/src/components/backlog/views.py
MikeFE: Good call.

src/archivematicaCommon/lib/elasticSearchFunctions.py
MikeFE, line 252: Nice catch, it was apparently like this before too and I just didn't notice it when I was refactoring/moving code around. I don't think we'll need the first create mapping call... I'll remove it and find out

src/dashboard/src/components/backlog/views.py
MikeFE: Yeah sure, I suppose just setting it to use the name/filename column by default if the index is out of bounds and logging a warning would be more friendly.

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    A JSON end point that returns results for various backlog transfers and their files.
+
+    :param request: The Django request object
+    :return: A JSON object including required metadata for the datatable and the backlog search results.
+    """
+    # get search parameters from request
+    queries, ops, fields, types = advanced_search.search_parameter_prep(request)
+
+    file_mode = request.GET.get('file_mode') == 'true'
```
MikeFE, line 127: I use file_mode twice so I think it's best to keep it in a variable instead of using request.GET.get('file_mode') == 'true' multiple times.

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    When the user clicks a column header in the data table, we'll receive info in the ajax request
+    telling us which column # we're supposed to sort across in our query. This function will translate
+    the column index to the corresponding property name we'll tell ES to sort on.
+
+    :param index: The column index that the data table says we're sorting on
+    :param file_mode: Whether we're looking at transfers or transfer files
+    :return: The ES document property name corresponding to the column index in the data table.
+    """
+    table_columns = (
```
MikeFE, line 104: Wasn't sure what you meant by make this a dict

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,218 @@
...
+
+def download(request, uuid):
+    """
+    Download a package from a requested backlog transfer.
+
+    :param request: The Django request object
+    :param uuid: UUID for the transfer we're downloading the package from
+    :return: Respond with a TAR'd version of the requested package
+    """
+    return helpers.stream_file_from_storage_service(storage_service.download_file_url(uuid))
```
MikeFE, line 218: The methods in the archival storage views file all had extra code that broke things for my use case. I can just call stream_file_from_storage_service directly and it seems to work great.

src/dashboard/src/components/backlog/views.py
```diff
@@ -0,0 +1,218 @@
...
+    :param file_mode: Whether we're looking at transfers or transfer files
+    :return: The ES document property name corresponding to the column index in the data table.
+    """
+    table_columns = (
+        ('name', 'uuid', 'file_count', 'ingest_date', None),     # Transfers are being displayed
+        ('filename', 'sipuuid', None)                            # Transfer files are being displayed
+    )
+
+    if index < 0 or index >= len(table_columns[file_mode]):
+        logger.warning('Backlog column index specified is invalid for sorting, got {}'.format(index))
```
Hwesta: Tip: logging also allows `logger.warning('Backlog column index specified is invalid for sorting, got %s', index)`

src/dashboard/src/components/backlog/views.py
Hwesta, line 127: Reasonable

src/dashboard/src/components/backlog/views.py
Hwesta, line 104: Stylistic choice - this is also fine.

Hwesta:
:+1:

Hwesta:
Clean up commit history before you rebase & merge, please

PR #316, opened 2015-10-31
Fix test integrity errors

Hwesta:
SQLite doesn't enforce referential integrity, so tests fail when run against a real DB.  Add fixtures/clean up code to fix this. Some updates relate to Django's new transaction approach.

### DIFF
--- a/src/archivematicaCommon/lib/utilities/FPRClient/test_fprclient.py
+++ b/src/archivematicaCommon/lib/utilities/FPRClient/test_fprclient.py
@@ -40,6 +40,9 @@ class TestFPRClient(TestCase):
     Test fetching and updating rules.
     """
 
+    fixture_files = ['identification_links.json', ]
+    fixtures = [os.path.join(THIS_DIR, p) for p in fixture_files]
+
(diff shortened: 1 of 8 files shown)


src/archivematicaCommon/lib/utilities/FPRClient/test_fprclient.py
```diff
@@ -100,6 +103,9 @@ class TestFPRClient(TestCase):
     def setUp(self):
         self.fprclient = client.FPRClient(fprserver=FPRSERVER)
 
+    def test_fixtures(self):
+        assert main.models.MicroServiceChainLink.objects.all().count() == 3
```
mistydemeo: Don't need the `.all()` here.

mistydemeo:
:+1:

src/archivematicaCommon/lib/utilities/FPRClient/test_fprclient.py
Hwesta: Fixed

mistydemeo:
:shipit:

PR #317, opened 2015-11-02
Add checksum type setting

MikeFE:
Allows an admin to set which checksum algorithm to use when verifying files and calculating checksums.

### DIFF
--- a/src/MCPClient/lib/clientScripts/trimVerifyChecksums.py
+++ b/src/MCPClient/lib/clientScripts/trimVerifyChecksums.py
@@ -20,16 +20,15 @@
 # @package Archivematica
 # @subpackage archivematicaClientScript
 # @author Joseph Perry <joseph@artefactual.com>
-import re
 import os
 from lxml import etree as etree
 import sys
-import traceback
 import uuid
 
 # fileOperations, databaseFunctions requires Django to be set up
 import django
 django.setup()
+
 # archivematicaCommon
 from externals.checksummingTools import get_file_checksum
 from fileOperations import getFileUUIDLike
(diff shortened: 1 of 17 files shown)


src/dashboard/src/components/administration/forms.py
```diff
@@ -64,6 +64,16 @@ class StorageSettingsForm(SettingsForm):
         label="Full URL of the storage service")
 
 
+class ChecksumSettingsForm(SettingsForm):
+    CHOICES = (
+        ('md5', 'MD5'),
```
mistydemeo, line 79: Was MD5 specifically requested? Both it and SHA1 feel a bit insecure at this point.

src/archivematicaCommon/lib/externals/checksummingTools.py
```diff
@@ -1,26 +1,37 @@
...
 
-#Borrowed from http://stackoverflow.com/questions/1131220/get-md5-hash-of-a-files-without-open-it-in-python
-def md5_for_file(filename):
-    md5 = hashlib.md5()
-    with open(filename, 'rb') as f:
-        for chunk in iter(lambda: f.read(2**10 * md5.block_size), b''):
-            md5.update(chunk)
-    return md5.hexdigest()
 
+def get_file_checksum(filename, algorithm='sha256'):
```
mistydemeo, line 9: Nice to have this genericized!

src/dashboard/src/components/administration/forms.py
MikeFE, line 79: It was requested (I confirmed with Justin). We don't think the security is a huge issue in this case since we're just using it as a checksum to detect corruption in files.

mistydemeo, line 79: Okay, sounds good!

src/MCPClient/lib/clientScripts/trimVerifyChecksums.py
```diff
@@ -64,21 +64,21 @@ def callWithException(exception):
...
-            print >>sys.stderr, "Error parsing: ", xmlFilePath 
+            print >>sys.stderr, "Error parsing: ", xmlFilePath
             exitCode += 1
             continue
         #if extension.lower() != file[i+1:].lower():
-        #    print >>sys.stderr, "Warning, extension mismatch(file/xml): ", file[:i], extension , file[i+1:] 
-        
-        objectMD5 = md5_for_file(filePath)
-        
+        #    print >>sys.stderr, "Warning, extension mismatch(file/xml): ", file[:i], extension , file[i+1:]
```
Hwesta: Since you're doing cleanup anyway, dead code like this can be removed if it's not still relevant.

src/MCPClient/lib/clientScripts/verifyChecksumsInFileSecOfDspaceMETSFiles.py
```diff
@@ -46,11 +45,11 @@ def verifyMetsFileSecChecksums(metsFile, date, taskUUID, relativeDirectory="./")
         #print "%s - %s - %s " % (checksumType, checksum, fileLocation)
         fileFullPath = os.path.join(relativeDirectory, fileLocation)
         if checksumType == "MD5":
-            checksum2 = md5_for_file(fileFullPath)
-            eventDetail = "program=\"python\"; module=\"hashlib.sha256()\""
```
Hwesta, line 50: Nice catch on the md5/sha256 mixup.

src/archivematicaCommon/lib/externals/checksummingTools.py
```diff
@@ -1,26 +1,37 @@
...
+    Perform a checksum on the specified file.
+
+    This function reads in files incrementally to avoid memory exhaustion.
+    See: http://stackoverflow.com/questions/1131220/get-md5-hash-of-a-files-without-open-it-in-python
+
+    :param filename: The path to the file we want to check
+    :param algorithm: Which algorithm to use for hashing, e.g. 'md5'
+    :return: Returns a checksum string for the specified file.
+    """
+    h = hashlib.new(algorithm)
```
Hwesta, line 20: Do we want to handle the case where `algorithm` isn't valid, or assume that we're controlling the input?

MikeFE, line 20: Sure, I can check `if algorithm and algorithm in hashlib.algorithms` or something like that. Do you think we should instantiate a logger here to log the error or just throw an exception?

Hwesta, line 20: Hashlib will throw an exception here, so it's probably not a big deal.

Hwesta:
:+1:

PR #319, opened 2015-11-05
Remove unused init.d scripts

mistydemeo:


Hwesta:
:+1: 

This removes support for 12.04, right? I guess we're only officially supporting 14.04 now.

mistydemeo:
12.04 has Upstart, too; I think it's only older versions than that, for which we haven't been building packages for awhile. These scripts haven't actually been installed by the packages for years, as far as I can tell!

PR #321, opened 2015-11-07
Backlog: omit metadata/log folders

mistydemeo:


src/dashboard/src/components/ingest/views.py
```diff
@@ -520,20 +520,23 @@ def transfer_backlog(request, ui):
...
+        query = elasticSearchFunctions.MATCH_ALL_QUERY.copy()
+        query['filter'] = elasticSearchFunctions.LOG_FILE_FILTER
     else:
         queries, ops, fields, types = advanced_search.search_parameter_prep(request)
 
         try:
+            # Specify this as a filter, not a must_have, for performance,
+            # and so that it doesn't cause the "should" queries in a
+            # should-only query to be ignored.
+            filter = {'term': {'status': 'backlog'}}
```
Hwesta: `filter` is a builtin...

Hwesta:
:+1:

src/dashboard/src/components/ingest/views.py
mistydemeo: arg, right. I'll rename this "filters"

Hwesta:
Still :+1:

PR #322, opened 2015-11-10
Fix AIP data pagination

mistydemeo:
This considerably reworks archival storage's search view in order to fix paging of AIP searches.

AIP searches are very broken in the current iteration. For historical reasons, searches are performed against the `aips/aipfile` index, which is limited to returning only AIP-related fields. This is done because `aips/aipfile` contains relevant information which is not present in the `aips/aip` index, which contains actual AIP data.

The issue arises when paging is applied. Because these are _file_ results, there are many files to one AIP; these are incorrectly paged as file results. This means that each page, instead of containing ten AIPs as intended, contains ten files' worth of AIPs - which could be a single AIP, if the first returned AIP contains more than one file.

In addition to fetching metadata from `aips/aipfile`, the view later uses the fetched UUIDs to fetch records from `aips/aip` - in order to have enough information to seed the AIC form, and some other additional metadata that `aips/aipfile` is missing. It currently does this by performing one query for every fetched UUID, which is fairly wasteful.

This patch simplifies the current workflow to the following:
- Run the query against `aips/aipfile`, with an aggregation which returns all unique UUIDs in the requested set and the count of files within that UUID. This query is set to return no results - only the aggregation is used.
- Uses that set of UUIDs to fetch pages of AIP data from `aips/aip` in a single query, which feeds the template.

Because the paging is now based on AIPs, not files within an AIP, a page in archival storage can display the appropriate number of AIPs. In addition, because AIPs are fetched in pages and the total number of queries performed has been reduced substantially, performance is substantially better.

This currently breaks the AIC input form, which requires information on every single AIP in storage that matches the search (not just the current page). I'd like to rethink if that's the appropriate way to do this - especially in this view, which is not actually rendering the form.

The code here is still messy; it unifies the codepaths much more than before, but nonetheless introduces a lot of `if` statements. Could be worth refactoring, though I don't know if that's a priority at this stage.

### DIFF
--- a/src/dashboard/src/components/archival_storage/views.py
+++ b/src/dashboard/src/components/archival_storage/views.py
@@ -106,73 +106,78 @@ def search(request):
         # pulling only one field (we don't need field data as we augment
         # the results using separate queries)
         if not file_mode:
+            # Fetch all unique AIP UUIDs in the returned set of files
+            query['aggs'] = {'aip_uuids': {'terms': {'field': 'AIPUUID', 'size': 0}}}
+            # Don't return results, just the aggregation
+            query['size'] = 0
             # Searching for AIPs still actually searches type 'aipfile', and
             # returns the UUID of the AIP the files are a part of.  To search
             # for an attribute of an AIP, the aipfile must index that
             # information about their AIP in
             # elasticSearchFunctions.index_mets_file_metadata
-            # Because we're searching aips/aipfile and not aips/aip,
-            # cannot using LazyPagedSequence
             results = conn.search(
                 body=query,
                 index='aips',
                 doc_type='aipfile',
-                fields='AIPUUID,sipName',
                 sort='sipName:desc',
             )
+            # Given these AIP UUIDs, now fetch the actual information we want from aips/aip
+            buckets = results['aggregations']['aip_uuids']['buckets']
+            uuids = [bucket['key'] for bucket in buckets]
+            uuid_file_counts = {bucket['key']: bucket['doc_count'] for bucket in buckets}
+            query = {
+                'query': {
+                    'terms': {
+                        'uuid': uuids,
+                    },
+                },
+            }
+            index = 'aips'
+            doc_type = 'aip'
+            fields = 'name,uuid,size,created,status,AICID,isPartOf,countAIPsinAIC'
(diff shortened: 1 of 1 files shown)


mistydemeo:
Added a fix for the AIC issue. The AIC form is now passed a list of UUIDs instead of the full set of AIP data; actually fetching a full, unpaged set of AIP data is deferred until the AIC form is actually rendered, at which point it does a query to fetch the appropriate data.

mistydemeo:
This still attempts to fetch a full set of queries in a single query, which may be inefficient, so in sets with large numbers of AIPs we might want to look at using [a scan query with scrolling](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html#scroll-scan).

mistydemeo:
I hvaven't actually profiled anything but this feels a lot snappier now, at least with the size of my index. Not sure how it behaves in a larger index.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -106,74 +106,82 @@ def search(request):
...
+            :return: List of dicts for each entry with additional information
+            """
+            start = (page - 1) * page_size
+            results = conn.search(
+                body=query,
+                from_=start,
+                size=page_size,
+                index=index,
+                doc_type=doc_type,
+                **kwargs
```
Hwesta: Why `**kwargs` instead of passing `fields` and `sort` directly?

mistydemeo: They differ between the two code paths; rather than assign to two variables I stuck them in a dictionary.

Hwesta: Seems a bit odd when query/start/page_size/index/doc_type are passed directly, but -shrug-

Hwesta:
:+1:

src/dashboard/src/components/archival_storage/views.py
mistydemeo: Updated to just make them variables.

Hwesta:
Still :+1:

Hwesta:
Cherry-picked to 1.5 as 01b7e6500a162ff693d5b230a03c936b099336fb

PR #323, opened 2015-11-11
Dev-helper: Update pip dependencies

Hwesta:
Update pip dependencies. Don't find tests in share because /var/archivematica/sharedDirectory is sometimes symlinked there and has permissions problems.


mistydemeo:
:+1:

PR #325, opened 2015-11-16
Processing config: small fix in form

sevein:
PreconfiguredChoices was changed so config_path is passed to save() instead of
to the constructor. This commit updates its only consumer accordingly.


Hwesta:
:+1:

PR #326, opened 2015-11-16
Processing config: Reorder to match appearance in processing

Hwesta:
Small addition on top of #313 

Processing config options displayed in same order they appear during processing.


sevein:
:+1:

PR #327, opened 2015-11-17
Unzipped Bag: Restructure doesn't move processingMCP

Hwesta:
When verifying and restructuring a bag, do not move OPTIONAL_FILES (currently just processingMCP).  This fixes a problem where the automation-tools processingMCP was being moved and then not used.


mistydemeo:
:+1:

PR #328, opened 2015-11-19
Transfer type identification, refs #9171

sevein:
Redmine ticket: https://projects.artefactual.com/issues/9171

The [Transfer](https://github.com/artefactual/archivematica/blob/8b859d75111607eb0b31a9e3c2db85750f1e9041/src/dashboard/src/main/models.py#L252-L272) model has two fields related to the type of the transfer:
- `Transfer.typeoftransfer`: currently unused.
- `Transfer.transfer`
  - Client scripts making use of it:
    - `archivematicaSetTransferType.py`
    - `archivematicaCreateMETS2.py`

`archivematicaSetTransferType.py` sets a value for `Transfer.type` so it can be used later by `archivematicaCreateMETS2.py`. This doesn't happen in only some cases:
- Set transfer type: DSpace
- Set transfer type: Standard
- Set transfer type: TRIM
- Set transfer type: Maildir

The concept of transfer type is also used in the dashboard when it comes to start a new transfer:
- `components.api.views.approve_transfer`
- `components.filesystem_ajax.views.start_transfer`

The type is used to determine what [watchedDirectory/activeTransfers](https://github.com/artefactual/archivematica/tree/qa/1.x/src/MCPServer/share/sharedDirectoryStructure/watchedDirectories/activeTransfers) the transfer is going to start in, which determines the starting chain. One of the following values is expected:

```
'standard' => 'standardTransfer',
'unzipped bag' => 'baggitDirectory',
'zipped bag' => 'baggitZippedDirectory',
'dspace' => 'Dspace',
'maildir' => 'maildir',
'TRIM' => 'TRIM'
```

This map is hard-coded in three different places:
- [src/dashboard/src/templates/transfer/grid.html](https://github.com/artefactual/archivematica/blob/36dda44e8ad8fea2506f3dca83194738dc70108c/src/dashboard/src/templates/transfer/grid.html#L175-L186)
- [components.api.views.approve_transfer_via_mcp](https://github.com/artefactual/archivematica/blob/36dda44e8ad8fea2506f3dca83194738dc70108c/src/dashboard/src/components/api/views.py#L370-L377)
- [components.filesystem_ajax.views](https://github.com/artefactual/archivematica/blob/8b859d75111607eb0b31a9e3c2db85750f1e9041/src/dashboard/src/components/filesystem_ajax/views.py#L58-L65)
### Actions
- Delete `Transfer.typeOfTransfer`
- Investigate why the dropdown in the Ingest tab is missing the option `maildir`?

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaSetTransferType.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaSetTransferType.py
@@ -1,5 +1,18 @@
 #!/usr/bin/env python2
 
+"""
+mysql> SELECT pk, arguments FROM StandardTasksConfigs WHERE execute = 'archivematicaSetTransferType_v0.0';
++--------------------------------------+------------------------+
+| pk                                   | arguments              |
++--------------------------------------+------------------------+
+| 353f326a-c599-4e66-8e1c-6262316e3729 | "%SIPUUID%" "TRIM"     |
+| 36ad6300-5a2c-491b-867b-c202541749e8 | "%SIPUUID%" "Standard" |
+| 6c261f8f-17ce-4b58-86c2-ac3bfb0d2850 | "%SIPUUID%" "Dspace"   |
+| 7b455fc5-b201-4233-ba1c-e05be059b279 | "%SIPUUID%" "Maildir"  |
++--------------------------------------+------------------------+
+4 rows in set (0.00 sec)
+"""
+
 # This file is part of Archivematica.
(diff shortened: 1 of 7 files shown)


mistydemeo:
> Investigate why the dropdown in the Ingest tab is missing the option maildir?

We removed that about a year ago because maildir transfers currently don't work: d74479413fa22e42b5849de42259372eba282f5e

PR #329, opened 2015-11-19
Elasticsearch: set type for dc:date

mistydemeo:
Our Elasticsearch configuration now opportunistically detects date fields. This is normally fine, but it causes problems with the `<dc:date>` field in the parsed METS data since that field can contain arbitrary user-entered dates. Specifically, if a user enters a date that Elasticsearch detects as a date, this will prevent any further documents from being indexed if they contain date values that cannot be parsed as dates.

This adds `dc:date` to the mapping with an explicit string type.


jhsimpson:
:+1:

Hwesta:
:+1:

PR #330, opened 2015-11-24
ArchivesSpace digital object component support

mistydemeo:
This adds a few pieces of digital object-related functionality.

The Bentley/appraisal tab workflow for ArchivesSpace digital objects differs significantly from the existing DIP upload. This workflow is based around AIP storage, _not_ DIP upload, and handles the creation of file records differently.

This workflow begins in the appraisal tab, rather than at DIP storage.
- In the appraisal tab, the user creates digital object components associated with a given archival object
- The user drags one or more files into those components.
- Archivematica stores information about the digital object components and the files which are stored within those components in its database. (This is to ATK pairing.)
- When the AIP is stored, Archivematica creates one digital object associated with the archival object, and creates the components as children of that digital object. No ArchivesSpace records are created for individual files.
- The Storage Service is sent information on the division of the AIP into digital object components at the time the AIP is stored, and repackages the AIP into multiple ZIP files which are stored in DSpace or Fedora in records associated with those digital object components.

Changes in this PR:
- ~~Adds support for creating digital object components in ArchivesSpace to ArchivesSpaceClient. These are small records containing only a label and/or title which are children of digital objects.~~ This was merged in the agentarchives repo, and removed as these no longer exist here.
- ~~Add stub digital object methods to ArchivistsToolkitClient. (ATK doesn't have digital object components, and we haven't yet added DO support.)~~
- Adds models to represent DO components in the AM database~~, and to pair paths within a SIP with those components.~~
- Adds new access views to support adding/listing/editing DO components and their files.

### DIFF
--- a/src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
+++ b/src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python2
+
+from argparse import ArgumentParser
+from collections import defaultdict
+from csv import DictWriter
+from glob import iglob
+import os
+import sys
+from uuid import uuid4
+
+from main.models import ArchivesSpaceDOComponent
+
+# archivematicaCommon
+from elasticSearchFunctions import getDashboardUUID
+
+from agentarchives.archivesspace import ArchivesSpaceClient
+
+
+def process_digital_object_components(client, writer, current_location):
+    dashboard_uuid = getDashboardUUID()
+
+    component_ids = [os.path.basename(path).split('digital_object_component_')[1] for path in
+                     iglob(os.path.join(current_location, 'digital_object_component_*'))]
+
+    components = ArchivesSpaceDOComponent.objects.filter(id__in=component_ids)
+    # exit early to avoid the expense of hitting up ASpace
+    if components.count() == 0:
+        return
+
+    # In most circumstances an AIP being processed will be associated
+    # with only a single archival object, but since that isn't guaranteed at
+    # the database level let's operate with the assumption this could happen.
+    archival_object_map = defaultdict(list)
+    for component in components:
+        archival_object_map[component.resourceid].append(component)
+
+    for resource_id, components in archival_object_map.iteritems():
+        do_id = client.add_digital_object(resource_id, dashboard_uuid, str(uuid4))['id']
+        for component in components:
+            if component.digitalobjectid:
+                continue  # already created
+
+            doc = client.add_digital_object_component(do_id,
+                                                      label=component.label,
+                                                      title=component.title)
+            # Track status about this component having been created for future use
+            component.digitalobjectid = do_id
+            component.remoteid = doc['id']
+            component.save()
+
+            writer.writerow({
+                'component_id': component.id,
+                'digital_object_component_id': component.remoteid,
+            })
+
+            yield 'Created digital object component "{}" associated with record "{}"'.format(doc['id'], resource_id)
+
+
+def main(current_location, host, port, user, passwd):
+    with open(os.path.join(current_location, 'metadata', 'digital_object_components.csv'), 'w') as csvfile:
+        writer = DictWriter(csvfile, ['component_id', 'digital_object_component_id'])
+        writer.writeheader()
+
+        client = ArchivesSpaceClient(host=host, port=port, user=user, passwd=passwd)
+
+        for message in process_digital_object_components(client, writer, current_location):
+            print(message)
+
+        return 0
+
+if __name__ == '__main__':
+    parser = ArgumentParser()
+    parser.add_argument('--current-location',
+                        type=str, help='%SIPDirectory%')
+    parser.add_argument('--host', default="localhost",
+                        metavar="host", help="hostname of ArchivesSpace")
+    parser.add_argument('--port', type=int, default=8089,
+                        metavar="port", help="Port used by ArchivesSpace backend API")
+    parser.add_argument('--user', metavar="Administrative user")
+    parser.add_argument('--passwd', metavar="Administrative user password")
+
+    args = parser.parse_args()
+
+    sys.exit(main(args.current_location, args.host, args.port, args.user, args.passwd))
--- a/src/MCPClient/etc/archivematicaClientModules
+++ b/src/MCPClient/etc/archivematicaClientModules
@@ -131,4 +131,5 @@ verifyMD5_v0.0 = %clientScriptsDirectory%verifyMD5.sh
 verifyPREMISChecksums_v0.0 = %clientScriptsDirectory%verifyPREMISChecksums.py
 verifySIPCompliance_v0.0 = %clientScriptsDirectory%verifySIPCompliance.py
 verifyTransferCompliance_v0.0 = %clientScriptsDirectory%verifyTransferCompliance.py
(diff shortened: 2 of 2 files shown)


mistydemeo:
I opened a new PR in the agentarchives repo with the ArchivesSpaceClient stuff; will remove it from this PR after #343 lands.

mistydemeo:
Rebased to remove the ArchivesSpaceClient and ArchivistsToolkitClient commits, which have been merged into agentarchives.

src/dashboard/src/main/models.py
```diff
@@ -838,6 +838,26 @@ class Meta:
...
+class ArchivesSpaceDOComponent(models.Model):
+    """
+    Represents a digital object component to be created in ArchivesSpace at the time an AIP is stored by Archivematica.
+
+    In ArchivesSpace, a digital object component is meant to be parented to a digital object record.
+    The workflow in use by the appraisal tab doesn't expose digital objects to the user, just components;
+    one digital object should be created as a parent for these components before creating the
+    components themselves.
+    """
+    id = models.AutoField(primary_key=True, db_column='pk')
```
Hwesta: Django will create the ID field automatically with db_column='id' https://docs.djangoproject.com/en/1.7/topics/db/models/#automatic-primary-key-fields  Is there a reason to call the column 'pk' instead of letting Django's defaults handle it?

src/dashboard/src/components/access/views.py
```diff
@@ -201,6 +212,97 @@ def record_children(client, request, system='', record_id=''):
...
+    mapping = create_arranged_directory(system, record_id)
+    component_path = os.path.join(mapping.arrange_path, 'digital_object_component_{}'.format(component_id), '')
+    if create:
+        filesystem_views.create_arrange_directory(component_path)
+
+    return component_path
+
+
+@_authenticate_to_archivesspace
+def digital_object_components(client, request, system='archivesspace', record_id=''):
```
Hwesta, line 226: This looks like a good candidate for class-based views?

src/dashboard/src/main/models.py
mistydemeo: I just copied this from another table in the models; I can remove it.

src/dashboard/src/components/access/views.py
mistydemeo, line 226: Doesn't work well with the decorator, unfortunately.

Hwesta, line 226: Ah, fair. Unfortunate.

src/dashboard/src/components/access/views.py
```diff
@@ -201,6 +212,97 @@ def record_children(client, request, system='', record_id=''):
...
+            component = ArchivesSpaceDOComponent.objects.get(id=component_id)
+        except ArchivesSpaceDOComponent.DoesNotExist:
+            response = {
+                'success': False,
+                'message': 'No digital object component exists with the specified ID: {}'.format(component_id),
+            }
+            return helpers.json_response(response, status_code=404)
+        else:
+            component.label = record.get('label', '')
+            component.title = record.get('title', '')
```
Hwesta, line 295: If a 'label' or 'title' isn't specified, it will be blanked - is this intentional?

mistydemeo, line 295: It is, yes. Both fields are optional, omitting them feels like it should mean "blank" to me.

src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
```diff
@@ -0,0 +1,80 @@
...
+if __name__ == '__main__':
+    parser = ArgumentParser()
+    parser.add_argument('--current-location', dest='current_location',
+                        type=str, help='%SIPDirectory%')
+    parser.add_argument('--host', default="localhost", dest="host",
+                        metavar="host", help="hostname of ArchivesSpace")
+    parser.add_argument('--port', type=int, default=8089, dest='port',
+                        metavar="port", help="Port used by ArchivesSpace backend API")
+    parser.add_argument('--user', dest='user', metavar="Administrative user")
+    parser.add_argument('--passwd', dest='passwd', metavar="Administrative user password")
```
Hwesta: Tip: `dest` can be inferred from the first argument and the inferred one matches your specified one in every case. https://docs.python.org/2/library/argparse.html#dest

src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
```diff
@@ -0,0 +1,80 @@
...
+# archivematicaCommon
+from elasticSearchFunctions import getDashboardUUID
+
+from agentarchives.archivesspace import ArchivesSpaceClient
+
+
+def main(current_location, host, port, user, passwd):
+    dashboard_uuid = getDashboardUUID()
+
+    csvfile = open(os.path.join(current_location, 'metadata', 'digital_object_components.csv'), 'w')
```
Hwesta: This should use a `with` or be explicitly closed (but a `with` is better)

mistydemeo: It'd need to remain open throughout the remaining code, though I guess I could pull that into a different function.

src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
```diff
@@ -0,0 +1,80 @@
...
+        return 0
+
+    client = ArchivesSpaceClient(host=host, port=port, user=user, passwd=passwd)
+
+    # In most circumstances an AIP being processed will be associated
+    # with only a single archival object, but since that isn't guaranteed at
+    # the database level let's operate with the assumption this could happen.
+    archival_object_map = {}
+    for component in components:
+        if component.resourceid not in archival_object_map:
```
Hwesta: Tip: `archival_object_map` could be a DefaultDict https://docs.python.org/2/library/collections.html#collections.defaultdict

mistydemeo: Oh nice, that's handy! Switched to using it.

src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
```diff
@@ -0,0 +1,84 @@
...
+
+    components = ArchivesSpaceDOComponent.objects.filter(id__in=component_ids)
+    # exit early to avoid the expense of hitting up ASpace
+    if components.count() == 0:
+        return
+
+    # In most circumstances an AIP being processed will be associated
+    # with only a single archival object, but since that isn't guaranteed at
+    # the database level let's operate with the assumption this could happen.
+    archival_object_map = defaultdict(lambda: [])
```
Hwesta: Can be `archival_object_map = defaultdict(list)` but it doesn't matter

Hwesta:
:+1:

src/MCPClient/lib/clientScripts/writeDigitalObjectComponentsToArchivesSpace.py
mistydemeo: Oh right, classes are callables in python 😅

mistydemeo:
When I merge I'll omit the last commit, which isn't ready for use yet.

Hwesta:
This isn't how we ended up integrating with ArchivesSpace for the appraisal tab. See #502

PR #331, opened 2015-11-25
Added a configuration option to use a different databse name

pieterdp:
Added a new variable to the dbsettings-file "db" that contains the name of the database. Updated database interface code to support this. This allows users to choose a different database name from the default (MCP).

Tested on a development environment.

### DIFF
--- a/src/archivematicaCommon/etc/dbsettings
+++ b/src/archivematicaCommon/etc/dbsettings
@@ -2,4 +2,4 @@
 user=archivematica
 password=demo
 host=localhost
-
+db=archivematica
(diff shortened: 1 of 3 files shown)


src/archivematicaCommon/etc/dbsettings
```diff
@@ -2,4 +2,4 @@
 user=archivematica
 password=demo
 host=localhost
-
+db=archivematica
```
mistydemeo: Do you mind setting this to the current default of `MCP`? I think we'd like to keep the existing default.

pieterdp: Fixed

Hwesta:
Squashed and merged into qa/1.5.x as d518a774b656812afeb835ea5d16a701be664084 and qa/1.x as 769b2612fe90b08815fef170ebe1e2b783502056 I also took the liberty of making the whitespace in common.py match the lines next to it, I hope that's okay.

Thank you for the fix!

PR #332, opened 2015-11-26
filesystem_ajax: Unify copy_to_arrange and move_within_arrange

mistydemeo:
Right now, filesystem_ajax has two APIs for moving files around in arrange and the backlog: `copy_to_arrange`, for moving files from backlog to arrange; and `move_within_arrange`, for moving files from one location within arrange to another.

Having two APIs for this seems unnecessary: `copy_to_arrange` will behave erratically if passed any sourcepath that does not originate within backlog, and `move_within_arrange` will error out if passed any sourcepath that does not originate within arrange. Instead of requiring the consumer to pick the API to use, it makes more sense to me to have a single API that chooses its functionality based on the location of the sourcepath it is given. This PR unifies them into a single view, under the `copy_to_arrange` name, and changes every location that uses `move_within_arrange` to use that name.


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -628,87 +588,103 @@ def copy_to_arrange(request, sourcepath=None, destination=None):
...
+            copy_files_to_arrange(sourcepath, destination)
+            response = {'message': 'Files added to the SIP.'}
+            status_code = 201
+        elif sourcepath.startswith(DEFAULT_ARRANGE_PATH):
+            move_files_within_arrange(sourcepath, destination)
+            response = {'message': 'SIP files successfully moved.'}
+            status_code = 200
+        else:
+            raise ValueError('filepath is in neither the base backlog path nor the arrange path ({} provided)'.format(sourcepath))
+    except ValueError as e:
```
Hwesta, line 691: Log the exception as well?

Also, we've figured out more about logging since this was written and it should be `logger.foo` instead of `logging.foo` for all logging calls so it doesn't go to the root logger.

mistydemeo, line 691: Added logs.

mistydemeo:
Added a new commit which changes all calls to `logging` to go to the logger.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -628,87 +588,104 @@ def copy_to_arrange(request, sourcepath=None, destination=None):
...
+        destination = base64.b64decode(request.POST.get('destination', ''))
+    logger.info('copy_to_arrange: sourcepath: {}'.format(sourcepath))
+    logger.info('copy_to_arrange: destination: {}'.format(destination))
+
+    try:
+        if sourcepath.startswith('/' + DEFAULT_BACKLOG_PATH):
+            copy_files_to_arrange(sourcepath, destination)
+            response = {'message': 'Files added to the SIP.'}
+            status_code = 201
+        elif sourcepath.startswith(DEFAULT_ARRANGE_PATH):
```
Hwesta, line 685: The leading vs not leading / here should get a comment, if it's too hard to fix in this PR.

Hwesta:
:+1:

mistydemeo:
Pushed a commit that comments the leading slash/no leading slash discussion.

PR #333, opened 2015-11-27
Supporting work for Dataverse

Hwesta:
Generally-useful work that came out of Dataverse development
- Clean up archivematicaFunctions, RPCServer, createMETS2's createFileSec
- settings.test logs to console, not files
- `find_metadata_files` can only search transfers' metadata folders, excluding SIP
- `File.currentlocation` can be null in the model, eg if the file is deleted.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -652,8 +652,8 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
         if os.path.isdir(itemdirectoryPath):
             createFileSec(itemdirectoryPath, structMapDiv, baseDirectoryPath, baseDirectoryName, fileGroupIdentifier, fileGroupType, includeAmdSec)
         elif os.path.isfile(itemdirectoryPath):
-            myuuid=""
-            DMDIDS=""
+            myuuid = ""
+            DMDIDS = ""
             directoryPathSTR = itemdirectoryPath.replace(baseDirectoryPath, baseDirectoryName, 1)
 
             kwargs = {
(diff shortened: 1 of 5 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -697,38 +705,43 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
 
                 digiprovMD = getTrimAmdSec(baseDirectoryPath, fileGroupIdentifier)
                 globalDigiprovMDCounter += 1
-                digiprovMD.set("ID", "digiprovMD_"+ globalDigiprovMDCounter.__str__())
+                digiprovMD.set("ID", "digiprovMD_" + globalDigiprovMDCounter.__str__())
```
mistydemeo: As long as you're changing this line anyway, switch over to an `str()` call?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -697,38 +705,43 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
...
                 trimAmdSec.append(digiprovMD)
 
                 trimStructMapObjects.set("ADMID", ID)
 
-            fileId="file-%s" % (myuuid, )
-
-            #<fptr FILEID="file-<UUID>" LABEL="filename.ext">
+            # Create <div TYPE="Item"> and child <fptr>
+            # <fptr FILEID="file-<UUID>" LABEL="filename.ext">
+            fileId = "file-%s" % (f.uuid, )
```
mistydemeo: If you're changing this, change it to a `.format()` call?

mistydemeo:
:+1:

PR #334, opened 2015-11-27
AS/ATK clients: include resource/resource_component type in records

mistydemeo:
The AS/ATK records don't currently return whether a given record is a resource or a resource_component. This could be useful for other systems to render, and is also useful in order to allow the appraisal tab to determine whether a record is an AS/ATK record or a different type of record.


Hwesta:
:+1:

PR #335, opened 2015-12-01
Delete unneeded headers

sevein:
Also, removed shebang from clientConfig.conf and serverConfig.conf.


sevein:
Instead, we're adding the license where it's missing.

PR #336, opened 2015-12-01
Take more connection parameters in mysql_dev.sh

sevein:
It's backward compatible.

Before:

```
$ mysql_dev.sh <dbname> <root-password>
```

After:

```
$ mysql_dev.sh <dbname> <password> <dbhost> <dbuser> <dbseed>
```

Where `dbseed` is the location of the [mysql seed file](https://github.com/artefactual/archivematica/blob/115df7f4ddb0e9371f8084adfcd5751d996c159e/src/MCPServer/share/mysql). If given, the database is also re-created before the dump is loaded.

Also, updated the original mysql dump seed with `--extended-insert=FALSE` and  `--complete-insert=TRUE` to avoid too long lines.


src/MCPServer/share/mysql_dev.sh
```diff
@@ -1,37 +1,77 @@
 #!/bin/bash
 
-databaseName=$1
-dbpassword=$2  # should be "" or "-p<password>", like in localDevSetup/recreateDB.sh
```
Hwesta, line 4: Old behavior was `$dbpassword` was `-p <password>` or `""`, so that it handled an empty password without requiring more input. This isn't preserved - `-p` always prompts, even when the password is empty, most notably when running the mysql_dev scripts.

sevein, line 4: Good catch.

On Tue, Dec 1, 2015 at 5:10 PM, Holly Becker notifications@github.com
wrote:

> In src/MCPServer/share/mysql_dev.sh
> https://github.com/artefactual/archivematica/pull/336#discussion_r46365429
> :
> 
> > @@ -1,37 +1,77 @@
> >  #!/bin/bash
> > 
> > -databaseName=$1
> > -dbpassword=$2  # should be "" or "-p<password>", like in localDevSetup/recreateDB.sh
> 
> Old behavior was $dbpassword was -p <password> or "", so that it handled
> an empty password without requiring more input. This isn't preserved - -p
> always prompts, even when the password is empty, most notably when running
> the mysql_dev scripts.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/artefactual/archivematica/pull/336/files#r46365429.

## 

Jesús García Crespo,
Software Engineer, Artefactual Systems Inc.
http://www.artefactual.com | +1.604.527.2056

src/MCPServer/share/mysql_dev.sh
```diff
@@ -1,37 +1,80 @@
...
+  mysql_dev_hierarchical_dip_upload.sql
+  mysql_dev_8609_automate_matching_gui.sql
+  mysql_dev_8896_checksum_algorithms.sql
+  mysql_dev_delete_links.sql
+)
+
+failed=()
+
+for item in "${migrations[@]}"; do
+    echo "Loading migration ${item}... "
```
Hwesta, line 64: I like the feedback on what's being run.

src/MCPServer/share/mysql_dev.sh
```diff
@@ -1,37 +1,80 @@
...
+  mysql_dev_8415_checksum.sql
+  mysql_dev_8423_extract_logging.sql
+  mysql_dev_4757_sounds.sql
+  mysql_dev_8252_fpr_update.sql
+  mysql_dev_7595_archivesspace_upload.sql
+  mysql_dev_6488_aip_reingest2.sql
+  mysql_dev_hierarchical_dip_upload.sql
+  mysql_dev_8609_automate_matching_gui.sql
+  mysql_dev_8896_checksum_algorithms.sql
+  mysql_dev_delete_links.sql
```
Hwesta, line 58: Doublecheck this list is still correct when you rebase.

Hwesta:
:+1:  on the mysql_dev changes.  @jhsimpson, do you care about the changes to the seed mysql file?

jhsimpson:
:+1:   on both commits here.

PR #337, opened 2015-12-01
Dashboard: add gunicorn to requirements/local.txt

sevein:
It is going to be used in development environments (`gunicorn --reload`), it
may be added later to other environments.


Hwesta:
I don't think this should be in `base.txt`, because that's for production dependencies that are actually used.  Maybe it could go in `local.txt`, since that's for development things which can include experiments.  Or, it can remain a local customization for you until the experiments produce something?

Hwesta:
Commit message should be amended to reflect the new location, but :+1:

PR #339, opened 2015-12-01
MCPClient: Accept KeyboardInterrupt in main thread

sevein:
I know that we are not cleaning up our daemon threads but IMO that is a different
battle. What I want to achieve now is to listen to keyboard interrupt signals,
which is necessary when you are not using a supervisor and you are attached to
the process, e.g. when you are running the interpreter directly in a container.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -252,8 +252,10 @@ if __name__ == '__main__':
     logging.config.dictConfig(LOGGING_CONFIG)
     logger = logging.getLogger("archivematica.mcp.client")
 
-    loadSupportedModules(config.get('MCPClient', "archivematicaClientModules"))
-    startThreads(config.getint('MCPClient', "numberOfTasks"))
-    tl = threading.Lock()
-    tl.acquire()
-    tl.acquire()
+    try:
+        loadSupportedModules(config.get('MCPClient', "archivematicaClientModules"))
+        startThreads(config.getint('MCPClient', "numberOfTasks"))
+        while True:
+            time.sleep(100)
+    except (KeyboardInterrupt, SystemExit):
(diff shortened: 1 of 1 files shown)


Hwesta:
Does this have ramifications for how long an interrupt takes to process (up to 100 seconds)? What happens to Exceptions that aren't SystemExit or KeboardInterrupt in the old vs new way?  I assume SystemExit & KeyboardInterrupt both terminate the process in the new way - what happened before?

sevein:
> Does this have ramifications for how long an interrupt takes to process (up to 100 seconds)?

I don't think so. Apparently the CPython interpreter installs a signal handler in a separate thread. `SIGINT` is translated into a `KeyboardInterrupt` exception that will terminate the suspension of time.sleep() for us.

> What happens to Exceptions that aren't SystemExit or KeyboardInterrupt in the old vs new way? I assume SystemExit & KeyboardInterrupt both terminate the process in the new way - what happened before?

The old way created a lock object and tried to acquire it twice. The second try locked forever because the lock was already busy. It's kind of a hack. And it was blocking us from detecting `KeyboardInterrupt` because the interpreter wouldn't release our own lock for us obviously. Locks are a great tool to synchronize multiple threads, but in this particular case we were not even sharing the lock with others so nobody was going to release() for us.
I don't think that we have to worry about other exceptions because they are happening in different threads. Once our main thread is suspended really all we can do is to wait for the interpreter to raise `KeyboardInterrupt` or `SystemExit`.

This also introduces an opportunity to exit our threads cleanly. We need to make them non-daemonic and signal them with [https://docs.python.org/2/library/threading.html#threading.Event](Event). But that can be done some other day! Or maybe we use an event loop with `asyncio` :smile:.

Hwesta:
Ok, sounds like you looked into this. :) :+1:

PR #340, opened 2015-12-02
ArchivesSpaceClient: add_digital_object fixes

mistydemeo:
- The `title` and `identifier` fields are mandatory.
- Remove default `identifier` handling based around `uri`.
- Don't try to add a file_version if there is no `uri`.


sevein:
LGTM

PR #35, opened 2014-05-05
change hardcoded archivematica uid (333) in server to query os for uid

klindt:
Hi,

we had to change the MCPServer everytime after merges or checking out branches because our archivematica hasn't got the uid 333.

Cheers Marco (from ZIB)

PR #341, opened 2015-12-02
MCPServer clean-ups

sevein:
Getting rid of some stuff that is not needed anymore.


Hwesta:
Are there ramifications to changing the user ID? A lot of permissions expect the user & group to be archivematica.

sevein:
In practice, the process credentials are not really affected at this point because [we already determined](https://github.com/artefactual/archivematica/blob/24f5f6263d6fea28211cdc87c66dff2fc2b659b5/src/MCPServer/init/archivematica-mcp-server.conf#L27-L28) the `uid` and `gid` in upstart where the process is spawned.

The old init script always used `sudo -u archivematica` so that doesn't explain why we were using `setuid`. But after a bit of research I figured that this was due to the developer owning/running the process and scaling it down to `archivematica` at runtime, see https://github.com/artefactual/archivematica-history/commit/b3f36ec60ec8f2bd86f0259d409500b38315dbc8.

Hwesta:
Nice bit of spelunking.  

I was worried this would break #35 but it looks like it actually helps with their problem, since the UID will be set elsewhere. 

`RPCServer.py` also uses `databaseInterface.sqlLock`.  Can this be removed for the same reason the stuff in MCPServer could be?

sevein:
I believe so! The worker loop runs in the main thread (`gm_worker.work()`) and when a job comes it is processed within the loop, not in a new thread. So we can process only one job at a time, within the main thread. I can't imagine how that's going to produce race conditions.

`archivematicaMCP.py` [**main**] => `RPCServer.py` [**startRPCServer**]:

``` python
while True:
    try:
        gm_worker.work()
    except gearman.errors.ServerUnavailable:
        time.sleep(failSleep)
        if failSleep < failMaxSleep:
            failSleep += failSleepIncrementor
```

---

Unrelated thought: It's a bit confusing that the MCPServer is both a Gearman client and a worker simultaneously. And even more confusing that we are using Gearman for message passing (aka RPCServer).

Hwesta:
This looks good to me but I'd like a second opinion. 0.5:+1:  @jhsimpson @mistydemeo ?

mistydemeo:
Haven't tested, but looks good :+1:

PR #342, opened 2015-12-02
maindb Django app

sevein:
### It breaks Archivematica! WIP!

Feel free to suggest other name or a different location.

I'm not sure what should I do with `components.administration.models`. `ReplacementDict`, `ArchivistsToolkitConfig` and `ArchivesSpaceConfig` don't seem to be used anywhere else other than the dashboard.

I've also taken the chance to get rid of all the `sys.path` and `DJANGO_SETTINGS_MODULE` mangling that we're doing in the code. My idea is that environment-aware details should be handled outside the application. In the particular case or the dashboard, the best place for that IMHO is the Apache `VirtualHost`, but `SetEnv` and `mod_wsgi` don't seem to work together so I ended up setting them up in `apache/django.wsgi`. Make sure that your upstart configuration files are populated again:
- `/etc/init/archivematica-mcp-server.conf`
- `/etc/init/archivematica-mcp-client.conf`
### TODO:
- [ ] MCPClient depends on `fpr.models`: add as a git submodule?
- [ ] `maindb` requires `django_extensions`: needs to be installed wherever `maindb` is used.
- [ ] Migrations.


mistydemeo:
I wonder if we might want to break this out into a separate application/repo from archivematicaCommon entirely? Probably not that important, but might be nice to be able to install it separately.

We've been breaking out some more stuff lately, but most of the other ones are less coupled to Archivematica.

mistydemeo:
> I've also taken the chance to get rid of all the sys.path and DJANGO_SETTINGS_MODULE mangling that we're doing in the code. My idea is that environment-aware details should be handled outside the application.

:+1: I'm _really_ happy to get this kind of change made!

mistydemeo:
> MCPClient depends on fpr.models: add as a git submodule?

Submodule makes sense to me, since that's how we do it now.

> maindb requires django_extensions: needs to be installed wherever maindb is used.

If maindb becomes its own project, we can give it a `requirements.txt` that specifies that; otherwise add it to archivematicaCommon?

sevein:
My original intention was only to separate it from the dashboard so I can run the MCP server on a different host without having to install the dashboard sources as a dependency when I only need a small part of it.
Building a separate service would make a lot of sense, and we might as well create more than one service (some of the data may fit better in different types of stores) or push responsibilities to existing services. That's something I thought about it but I also want to keep this small and realistic. It's just one first step of possibly many others. Does that make sense?

sevein:
@mistydemeo @jhsimpson @Hwesta - still I'm fine starting with a new repo, up to you! Please let me know what is your preference. Only thing is that I'm not planning to work on building a service out of it (yet - I think that's a longer discussion), just breaking up the sources.

PR #343, opened 2015-12-02
Extract ArchivesSpaceClient and ArchivistsToolkitClient into new module

mistydemeo:
The new Python module currently lives at [artefactual-labs/agentarchives](https://github.com/artefactual-labs/agentarchives).

(The fact that I keep mixing up "agentarchives" vs "archivesagent" maybe isn't a good sign, hopefully I default to the correct one soon!)


src/archivematicaCommon/requirements/base.txt
```diff
@@ -2,4 +2,5 @@ bagit==1.5.2
 Django>=1.7,<1.8
 elasticsearch>=1.0.0,<2.0.0
 requests==2.7.0
-django-mysqlpool
\ No newline at end of file
+django-mysqlpool
+agentarchives
```
Hwesta: Is this in PyPI yet? Does it need to be for this to work?

mistydemeo: It'll be going into PyPI for the 1.6 release.

Hwesta:
Remove the tests and fixtures too?

mistydemeo:
Good catch, I'll remove those too.

Hwesta:
:+1:

mistydemeo:
Updated import paths.

Do we want pip to point at a git repo until this is published on pypi, or can we cope with it not being installable via pip in the interim?

Hwesta:
Probably the git repo, so installs of qa/1.x still work.  Once they get published, we can update the requirements.txt

mistydemeo:
Updated to use the git repo.

PR #344, opened 2015-12-03
upload-qubit.py : call get_script_logger() after django.setup()

hakamine:
For some reason get_script_logger() should be called after
django.setup() or the imports to work


Hwesta:
:+1:

PR #346, opened 2015-12-04
Events and Agents have a many-to-many relationship

Hwesta:
Previously, in createMETS2, Agents were added to an amdSec by adding everything in the Agents table, then finding all linkingAgentIdentifiers, assume they're User FKs, and creating Agents from that.   Event's linkingAgentIdentifier were added by checking for a linking agent, and if it existed, assume it's a User and create an agent Element, then add all Agents.

With external Agents, not all Agents are associated with every Event, and not all files (amdSecs) have Events associated with every Agent.

This updates the data model so Events and Agents have a many-to-many relationship (instead of only an implicit relationship), and Users have an associated Agent (this is done by adding a UserProfile, which includes a link to the related Agent. Other additional User information could go here too.).

Changes:
- SQL can be run before Django's syncdb/migrate is
  - This was needed to fix column types to what Django expected so Django could create foreign keys for new tables (UserProfile)
- New UserProfile model that links a User to an Agent
- Update insertIntoEvents to take a list of Agents (instead of a single User.id)
- Update digiprovMD creation to handle new Event/Agent relationship
- Create/update an Agent when a User is created/modified
- Update activeAgent (which tracks which user took an action) to refer to the Agent not the User

This still needs:
- [x] <s>Update ansible role to do pre-sync SQL</s>
- [x] Data migrations for things in `Events.linkingAgentIdentifier`
- [x] Create UserProfile for existing users on upgrade


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -443,24 +432,30 @@ def createDigiprovMD(fileUUID):
     """
     Create digiprovMD for PREMIS Events and linking Agents.
     """
+    global globalDigiprovMDCounter
```
mistydemeo, line 435: :sob:

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -479,60 +474,25 @@ def createEvent(digiprovMD, event_record):
...
-        etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = str(event_record.linking_agent)
-
     # linkingAgentIdentifier
-    for agent in Agent.objects.all():
+    for agent in event_record.agents.all():
         linkingAgentIdentifier = etree.SubElement(event, ns.premisBNS + "linkingAgentIdentifier")
         etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierType").text = agent.identifiertype
         etree.SubElement(linkingAgentIdentifier, ns.premisBNS + "linkingAgentIdentifierValue").text = agent.identifiervalue
 
+def createAgent(digiprovMD, agent_record):
```
mistydemeo, line 483: Maybe keep these as separate parameters so we can call `values` above and avoid the need to instantiate an `Agent` object? We're not using the stuff that a model object would give us.

src/MCPServer/lib/linkTaskManagerChoice.py
```diff
@@ -138,9 +138,12 @@ def xmlify(self):
 
     @log_exceptions
     @auto_close_db
-    def proceedWithChoice(self, chain, agent, delayTimerStart=False):
-        if agent:
-            self.unit.setVariable("activeAgent", agent, None)
+    def proceedWithChoice(self, chain, user_id, delayTimerStart=False):
+        if user_id is not None:
+            agent_id = UserProfile.objects.get(user_id=int(user_id)).agent_id
```
mistydemeo, line 143: Maybe avoid instantiating an `Agent` object here too?

mistydemeo, line 143: If we're doing this same thing in a few files, maybe make a helper function to return the `agent_id` as a string?

mistydemeo:
> Data migrations for things in Events.linkingAgentIdentifier

Would this handle creating the new `many-to-many` relationship for things created before the `create_user_agent` post-save signal existed?

src/MCPServer/lib/linkTaskManagerChoice.py
Hwesta, line 143: I don't think it is instantiating an Agent object, just a UserProfile one.  agent_id fetches the value in the column from UserProfile.

I'll look at moving this to utils.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
Hwesta, line 483: Possibly - I was modelling this on createEvent above which accepts a record.  Is instantiating an Agent that expensive?

Hwesta:
> > Data migrations for things in Events.linkingAgentIdentifier
> 
> Would this handle creating the new many-to-many relationship for things created before the create_user_agent post-save signal existed?

It would have to. This is part of why we're waiting - I'm hoping that Django migrations (including data migrations) will come in first.

jrwdunham:
This looks good to me.

PR #347, opened 2015-12-04
Dataverse-specific support

Hwesta:
This includes #346 

Add support for parsing an incoming METS file generated from Dataverse, and updating the AIP METS based on this.
- Display dataset name as the display string
- Parse and validate Dataverse METS fileGrp checksums, add Event for Dataverse fixity
- Create derivations for Dataverse-generated .tab files
- Add Dataverse as an Agents based on information in the newly-created `agents.json`
- Set fileGrpUse's from the METS
- Copy dmdSecs to the AIP METS

See also: https://wiki.archivematica.org/Dataverse

Todo:
- [ ] Only try to parse the Dataverse METS if it's a Dataverse transfer type (or equivalent solution)


mistydemeo:
Looks like there are merge conflicts, rebase this?

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -807,6 +766,13 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
                     else:
                         dspaceMetsDMDID = ids
 
+            # If it's a .tab file, check if there's a Dataverse METS with additional metadata
+            if f.originallocation.endswith('.tab'):
+                dv_metadata = create_dataverse_tabfile_dmdsec(baseDirectoryPath, os.path.basename(f.originallocation))
+                dmdSecs.extend(dv_metadata)
+                ids = ' '.join([ds.get('ID') for ds in dv_metadata])
```
mistydemeo, line 772: Nitpick, but you don't need the brackets when the comprehension is the only argument to a method call.

src/MCPClient/lib/clientScripts/parse_dataverse.py
```diff
@@ -0,0 +1,184 @@
...
+    Add external agent(s).
+
+    :return: ID of the first agent, assuming that's the Dataverse agent.
+    """
+    agents_jsonfile = os.path.join(unit_path, 'metadata', 'agents.json')
+    try:
+        with open(agents_jsonfile, 'r') as f:
+            agents_json = json.load(f)
+    except (OSError, IOError):
+        return None
```
mistydemeo, line 73: Nitpick, but could be `return`

Hwesta, line 73: Being explicit that it's not returning an ID but None instead.

Hwesta:
Context: This PR was part of some experimental work for starting transfers from Dataverse.  The associated storage service PR artefactual/archivematica-storage-service#99 was merged because it was fairly standalone.  This also relies on files created by artefactual/automation-tools#15 to work properly.

This adds a microservice that parses the METS file generated by automation tools containing the Dataverse metadata.  However, if the Dataverse METS file does not exist it throws an error for all Standard transfers.  Before this can be merged, it needs to be updated to not require an external METS file.  This could be done through a new Dataverse transfer type, or by adding optional parsing of external metadata files.

A Dataverse transfer type is not preferred, since we're trying to reduce the number of transfer types overall, as they are error prone.  However, if we could autodetect a transfer type instead of requiring the user to provide it, that would make this option much more feasible.

Adding a generic parsing of external metadata files would also benefit DSpace, and make it easier to support other systems that want to pass in metadata with a transfer. 

Work to make Events and Agents have an explicit many-to-many relationship also came out of this work, and was merged as #346.  It is included in this PR, but on rebase only commits from "API: display verbose name in browse display string" onwards are needed. 

Remaining todos:
- [ ] Optionally parse the Dataverse METS file in `metadata/METS.xml`
- [ ] Convert SQL migration to Django migration
- [ ] Confirm the original RData files are being returned from Dataverse and remove "Dataverse: Temporary: Ignore RData files" commit
- [ ] Ask on the dataverse mailing list about the JSON structure, and what fields are required/optional

sevein:
Replaced by #1242.

The branch won't be maintained for future reference.

PR #348, opened 2015-12-06
Better configuration

sevein:
# WIP, please ignore for now!
- **_I'm breaking this into smaller pull requests.**_
- Still figuring it out how to do this trying to keep backward compatibility.

---

The main goal of this pull request is to clean up the configuration system in `dashboard`, `mcp-sever` and `mcp-client` - making sure that they can run separately and completely abandon the assumption that they are running in the same machine. This is currently true when it's about configuration or source code availability. This pull request fixes the configuration part of it.

It is a big pull request, please read the commits individually.
#### Action items
- [x] MCPClient:
  - Abandon dbsettings
  - Use `EnvConfigParser`
  - **Move get_script_logger to MCPClient, add CLIENT_SCRIPTS_LOGGING_CONFIG?**. Not done.
- [ ] MCPServer:
  - Abandon dbsettings
  - Use `EnvConfigParser`...
- [ ] dashboard:
  - Abandon dbsettings
  - Don't read clientConf (currently: `kioskMode`,  `disableElasticsearchIndexing`, `sharedDirectoryMounted`... see also `get_client_config_value` function)
  - Make sure envs are used
- [ ] `elasticSearchFunctions` (archivematicaCommons):
  - Should not check on local installation of Elasticsearch
  - Should not depend on `clientConfig`, receive connection parameters instead!
- [ ] `databaseInterface` (archivematicaCommons):
  - It is going to be deleted. See #341 and #354.

### DIFF
--- a/src/MCPClient/etc/clientConfig.conf
+++ b/src/MCPClient/etc/clientConfig.conf
@@ -1,39 +1,28 @@
-#!/bin/bash
-
-# This file is part of Archivematica.
-#
-# Copyright 2010-2012 Artefactual Systems Inc. <http://artefactual.com>
-# 
-# Archivematica is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# Archivematica is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
-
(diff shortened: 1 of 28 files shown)


Hwesta:
`databaseInterface` is on it's way out, since it was mostly replaced by Django's ORM in functionality.  `setMaildirFileGrpUseAndFileIDs.py` is part of an already-broken codepath, so we don't care about that, and `archivematicaFITS.py` has been mostly (if not entirely) replaced by the FPR-based scripts for File ID/characterization etc. `emailFailReport.py` uses a SQL view that isn't in Django.  `upload-archivistsToolkit.py` needs to interface with an external DB that we don't control, so having it as a Django model doesn't make sense.  Discussions for that included a library to auto-detect DB structure, or just using `MySQLdb`/`pymysql`.  Removing `databaseInterface` could easily be its own PR.

sevein:
@Hwesta that makes a lot of sense, I'll do that. Thanks!

sevein:
@jhsimpson or @Hwesta

PR #349, opened 2015-12-07
MCPClient: leverage upstart's service logger

sevein:
Upstart logs every service (job) in a log file by the same name under
`/var/log/upstart`. Both `stdout` and `stderr` are captured. Furthermore,
Ubuntu rotates the log files as described in `/etc/logrotate.d/upstart`.
This commit takes away this responsibility from our upstart service, making
easier for Ubuntu administrators to inspect `archivematica-mcp-client` in the
standard manner:

```
$ sudo less +FG /var/log/upstart/archivematica-mcp-client.log
```

In addition, the `pre-start` stanza that waits for Gearman to become available
is being deleted because: (1) Gearman server may not be running in the same
host, (2) MCPClient is already dealing with the situation were the Gearman
server cannot be reached (see `gearman.errors.ServerUnavailable` in
`archivematicaClient.py`) using a retry pattern with incremental delays.

### DIFF
--- a/src/MCPClient/init/archivematica-mcp-client.conf
+++ b/src/MCPClient/init/archivematica-mcp-client.conf
@@ -15,8 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
-description     "Archivematica MCP Client"
-author          "Austin Trask <austin@artefactual.com>, Joseph Perry <joseph@artefactual.com>"
+description "Archivematica MCP Client"
 
 start on (net-device-up
           and local-filesystems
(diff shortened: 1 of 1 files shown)


Hwesta:
This looks good to me, but I'm unclear on potential ramifications. @jhsimpson @mistydemeo ?

sevein:
FYI: planning to do similarly in MCPSever at some point. I haven't changed that one yet because that service is configured to start after `mysql` is started (`/etc/init/archivematica-mcp-server.conf`):

```
start on (started mysql)
```

I can't change that just yet because I haven't verified that MCPServer will work if MySQL is not immediately available and can wait until it is.

jhsimpson:
:+1 from me, I like this change, my only quibble is with removing the GPL text, I am uncertain if that is a good idea or a bad idea.  This is a conf file, but it is in the repo and part of the application, so I think it probably should have the GPL text.  Other than that, I like it.

sevein:
Thank you!

PR #350, opened 2015-12-07
MCPClient: use thread-safe logger

sevein:
One of the many benefits of `logging` is its inherent thread safety. Using it
saves us from having to deal with thread synchronization in order to provide our
worker threads with safe access to `stdout` and `stderr`. This commits deletes
`printOutputLock` and replaces the old print statements with calls to the
logger.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -30,7 +30,7 @@
 #
 #For archivematica 0.9 release. Added integration with the transcoder.
 #The server will send the transcoder association pk, and file uuid to run.
-#The client is responsible for running the correct command on the file. 
+#The client is responsible for running the correct command on the file.
 
 import ConfigParser
 import cPickle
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/archivematicaClient.py
```diff
@@ -75,7 +115,7 @@ def loadSupportedModulesSupport(key, value):
     for key2, value2 in replacementDic.iteritems():
         value = value.replace(key2, value2)
     if not os.path.isfile(value):
-        print >>sys.stderr, "Warning - Module can't find file, or relies on system path:{%s}%s" % (key.__str__(), value.__str__())
+        logger.error('Warning! Module can\'t find file, or relies on system path: {%s} %s', key, value)
```
Hwesta: Tip: You don't have to escape `'` if they're inside a string defined by `"`

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -43,6 +43,48 @@
...
+        },
+    },
+    'handlers': {
+        'logfile': {
+            'level': 'INFO',
+            'class': 'custom_handlers.GroupWriteRotatingFileHandler',
+            'filename': '/var/log/archivematica/MCPClient/MCPClient.log',
+            'formatter': 'detailed',
+            'backupCount': 5,
+            'maxBytes': 4 * 1024 * 1024,  # 20 MiB
```
Hwesta, line 63: Fix comment to match actual value?

Hwesta:
Nice work!

Possible out of scope, but update/add more logging statements where they're useful? The current ones are pretty sparse and not always helpful, and now that we have loglevels we can make use of `.debug`.  This could also be another PR/done by someone else.

:+1:

sevein:
I'll totally do that, @Hwesta! Thank you so much.

PR #351, opened 2015-12-07
Add EnvConfigParser subclass with environment string support

sevein:
Planning to use `EnvConfigParser` in the future. Let me know what you think!

---

Main two differences comparing it to its ancestor `SafeConfigParser` are:
- The settings are looked up first in the string environment, trying to match
  both `section` and `option` first and only `option` as a fallback, e.g.
  `ARCHIVEMATICA_SECTION_OPTION` and `ARCHIVEMATICA_OPTION`.
- All the getters (`get`, `getint`, `getfloat`, `getbool`) learn a new keyword
  argument `fallback` so its value is returned when the configuration option
  is undefined, instead of throwing exceptions like `NoSectionError` or
  `NoOptionError`, as with a Python dictionary. This mimics the behavior
  introduced in the `configparser` module in Python 3. For example, you could
  write:
  
  ```
  tls_enabled = config.getboolean('network', 'tls', fallback=False)
  ```
  
  instead of:
  
  ```
  try:
      tls_enabled = config.getboolean('network', 'tls')
  except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
      tls_enabled = False
  ```

**Attention! This class is py2-only compatible!**

### DIFF
--- a/src/archivematicaCommon/lib/env_configparser.py
+++ b/src/archivematicaCommon/lib/env_configparser.py
@@ -1,0 +1,72 @@
+import os
+import ConfigParser
+import functools
+
+
+def fallback_option(fn):
+    def wrapper(*args, **kwargs):
+        fallback = kwargs.pop('fallback', None)
+        try:
+            return fn(*args, **kwargs)
+        except (ConfigParser.NoSectionError, ConfigParser.NoOptionError):
+            if fallback:
+                return fallback
+            raise
+    return functools.wraps(fn)(wrapper)
+
+
+class EnvConfigParser(ConfigParser.SafeConfigParser):
+    """
+    EnvConfigParser enables the user to provide configuration defaults using
+    the string environment, e.g. given:
+
+      - String environment prefix (prefix) = "ARCHIVEMATICA_MCPSERVER"
+      - Configuration section: "network"
+      - Configuration option: "tls"
+
+    This parser will first try to find the configuration value in the string
+    environment matching one of the two following keys:
+
+      - ARCHIVEMATICA_MCPSERVER_NETWORK_TLS
+      - ARCHIVEMATICA_MCPSERVER_TLS
+
+    If the variable is not set in the string environment the reader falls back
+    on the main configuration backend.
+
+    Additionally, the getters (get(), getint(), etc...) accept a new parameter
+    (fallback) that returns the value given to it instead of an exception when
+    the section or option trying to be match are undefined.
+    """
+    ENVVAR_SEPARATOR = '_'
+
+    def __init__(self, defaults=None, env=None, prefix=''):
+        self._environ = env or os.environ
+        self._prefix = prefix.rstrip('_')
+        ConfigParser.SafeConfigParser.__init__(self, defaults)
+
+    def _get_envvar(self, section, option):
+        for key in (
+            self.ENVVAR_SEPARATOR.join([self._prefix, section, option]).upper(),
+            self.ENVVAR_SEPARATOR.join([self._prefix, option]).upper(),
+        ):
+            if key in self._environ:
+                return self._environ[key]
+
+    @fallback_option
+    def get(self, section, option, **kwargs):
+        ret = self._get_envvar(section, option)
+        if ret:
+            return ret
+        return ConfigParser.SafeConfigParser.get(self, section, option, **kwargs)
+
+    @fallback_option
+    def getint(self, *args, **kwargs):
+        return ConfigParser.SafeConfigParser.getint(self, *args, **kwargs)
+
+    @fallback_option
+    def getfloat(self, *args, **kwargs):
+        return ConfigParser.SafeConfigParser.getfloat(self, *args, **kwargs)
+
+    @fallback_option
+    def getboolean(self, *args, **kwargs):
+        return ConfigParser.SafeConfigParser.getboolean(self, *args, **kwargs)
--- a/src/archivematicaCommon/tests/test_env_configparser.py
+++ b/src/archivematicaCommon/tests/test_env_configparser.py
@@ -1,0 +1,95 @@
+from __future__ import absolute_import
+import os
+import StringIO
+import ConfigParser
+
+from django.test import TestCase
+import pytest
+
+from env_configparser import EnvConfigParser
+
+
+class TestConfigReader(TestCase):
+    def setUp(self):
+        """
+        Make sure that we are not mutating the global environment. `os.environ`
+        is an instance of `os._Environ` which implements a `copy` method.
+        """
+        self.environ = os.environ.copy()
+
+    def tearDown(self):
+        self.environ = None
+
+    def read_test_config(self, test_config, prefix=''):
+        buf = StringIO.StringIO(test_config)
+        config = EnvConfigParser(env=self.environ, prefix=prefix)
+        config.readfp(buf)
+        return config
+
+    def test_env_lookup_int(self):
+        """
+        Note that the environment precedes the configuration.
+        """
+        self.environ['ARCHIVEMATICA_NICESERVICE_QUEUE_MAX_SIZE'] = '100'
+        config = self.read_test_config(
+            prefix='ARCHIVEMATICA_NICESERVICE',
+            test_config=
+"""
+[queue]
+max_size = 500
+""")
+        assert config.getint('queue', 'max_size') == 100
+
+    def test_env_lookup_nosection_bool(self):
+        """
+        The environment string matches the option even though the corresponding
+        section was not included.
+        """
+        self.environ['ARCHIVEMATICA_NICESERVICE_TLS'] = 'off'
+        config = self.read_test_config(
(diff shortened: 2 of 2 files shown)


src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,73 @@
...
+    on the main configuration backend.
+
+    Additionally, the getters (get(), getint(), etc...) accept a new parameter
+    (fallback) that returns the value given to it instead of an exception when
+    the section or option trying to be match are undefined.
+    """
+    ENVVAR_SEPARATOR = '_'
+
+    def __init__(self, defaults=None, env=None, prefix=''):
+        self._environ = os.environ if env is None else env
```
MikeFE: could possibly shorten to `self._environ = env or os.environ`

sevein: Thanks, I'll do that. I need to learn all the syntactic sugar in Python!

On Mon, Dec 7, 2015 at 11:23 AM, MikeFE notifications@github.com wrote:

> In src/archivematicaCommon/lib/env_configparser.py
> https://github.com/artefactual/archivematica/pull/351#discussion_r46863854
> :
> 
> > -    environment matching one of the two following keys:
> >   +
> > -      - ARCHIVEMATICA_MCPSERVER_NETWORK_TLS
> > -      - ARCHIVEMATICA_MCPSERVER_TLS
> >   +
> > -    It the variable is not set in the string environment the reader falls back
> > -    on the main configuration backend.
> >   +
> > -    Additionally, the getters (get(), getint(), etc...) accept a new parameter
> > -    (fallback) that returns the value given to it instead of an exception when
> > -    the section or option trying to be match are undefined.
> > -    """
> > -    ENVVAR_SEPARATOR = '_'
> >   +
> > -    def **init**(self, defaults=None, env=None, prefix=''):
> > -        self._environ = os.environ if env is None else env
> 
> could possibly shorten to self._environ = env or os.environ?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/artefactual/archivematica/pull/351/files#r46863854.

## 

Jesús García Crespo,
Software Engineer, Artefactual Systems Inc.
http://www.artefactual.com | +1.604.527.2056

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,73 @@
...
+        self._prefix = prefix.rstrip('_')
+        ConfigParser.SafeConfigParser.__init__(self, defaults)
```
Hwesta, line 45: Preferential to use `super` instead of naming the parent class directly when inheriting.

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,73 @@
+import os
+import ConfigParser
+import functools
+
+
+def fallback_option(fn):
```
Hwesta, line 6: I like this! Maybe this is too forward thinking, but how does it interact with the Python3 configparser's fallback?

src/archivematicaCommon/lib/env_configparser.py
sevein, line 45: `ConfigParser` (py2) does not subclass `object`. `configparser` (py3) does. I realized it wasn't easy to extend `SafeConfigParser` for both py2 and py3 at the same time. I'm not sure if it's worth the effort. I think it'll be easier to take care of it once we switch to py3 - unless the police is going to be to support both?

src/archivematicaCommon/lib/env_configparser.py
sevein, line 6: It doesn't. See my other comment above. I initially tried to support both py2 and py3 but the differences are important. Not sure if it's worth the time. Making this subclass py3-only would be very easy, but supporting both looked kind of a nightmare for a newcomer like me.

src/archivematicaCommon/lib/env_configparser.py
Hwesta, line 45: Ah, I didn't realize py2 configparser didn't subclass `object`. :(  We don't have a policy now, but I suspect we'll be supporting both, at least initially.  Not worth fixing now though if it's significant effort.

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,73 @@
...
+      - Configuration section: "network"
+      - Configuration option: "tls"
+
+    This parser will first try to find the configuration value in the string
+    environment matching one of the two following keys:
+
+      - ARCHIVEMATICA_MCPSERVER_NETWORK_TLS
+      - ARCHIVEMATICA_MCPSERVER_TLS
+
+    It the variable is not set in the string environment the reader falls back
```
Hwesta: Typo: s/It/If/

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,73 @@
...
+        for key in (
+            self.ENVVAR_SEPARATOR.join([self._prefix, section, option]).upper(),
+            self.ENVVAR_SEPARATOR.join([self._prefix, option]).upper(),
+        ):
+            if key in self._environ:
+                return self._environ[key]
+
+    @fallback_option
+    def get(self, *args, **kwargs):
+        section, option = args
```
Hwesta: If `args` is fixed length, why not just put `section, option` in the function definition?

src/archivematicaCommon/tests/test_config.py
```diff
@@ -0,0 +1,94 @@
...
+
+from env_configparser import EnvConfigParser
+
+
+class TestConfigReader(TestCase):
+    def setUp(self):
+        """
+        Make sure that we are not mutating the global environment.
+        """
+        self.environ = os.environ.copy()
```
Hwesta: You probably want `copy.deepcopy()` since `os.environ` is probably compound.  https://docs.python.org/2/library/copy.html

Hwesta:
:sparkling_heart:  for tests

Is this PR going to have anything using `EnvConfigParser` or will that be in later PRs?

src/archivematicaCommon/tests/test_config.py
sevein: Apparently `_Environ` implements a `copy` method! See http://git.io/vRH1J. I will add a note.

sevein:
Yes! Planning to use it very soon but in a different PR.

src/archivematicaCommon/lib/env_configparser.py
sevein: Very elegant, yes!

src/archivematicaCommon/lib/env_configparser.py
sevein: Sorry, where is that? `defaults`:question:

Hwesta: **It/If** the variable is not set in the string environment...

sevein: aaaah haha thanks

On Tue, Dec 8, 2015 at 4:04 PM, Holly Becker notifications@github.com
wrote:

> In src/archivematicaCommon/lib/env_configparser.py
> https://github.com/artefactual/archivematica/pull/351#discussion_r47034539
> :
> 
> > +class EnvConfigParser(ConfigParser.SafeConfigParser):
> > -    """
> > -    EnvConfigParser enables the user to provide configuration defaults using
> > -    the string environment, e.g. given:
> >   +
> > -      - String environment prefix (prefix) = "ARCHIVEMATICA_MCPSERVER"
> > -      - Configuration section: "network"
> > -      - Configuration option: "tls"
> >   +
> > -    This parser will first try to find the configuration value in the string
> > -    environment matching one of the two following keys:
> >   +
> > -      - ARCHIVEMATICA_MCPSERVER_NETWORK_TLS
> > -      - ARCHIVEMATICA_MCPSERVER_TLS
> >   +
> > -    It the variable is not set in the string environment the reader falls back
> 
> _It/If_ the variable is not set in the string environment...
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/artefactual/archivematica/pull/351/files#r47034539.

## 

Jesús García Crespo,
Software Engineer, Artefactual Systems Inc.
http://www.artefactual.com | +1.604.527.2056

sevein:
Should I merge?

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -0,0 +1,72 @@
...
+        ):
+            if key in self._environ:
+                return self._environ[key]
+
+    @fallback_option
+    def get(self, section, option, **kwargs):
+        ret = self._get_envvar(section, option)
+        if ret:
+            return ret
+        return ConfigParser.SafeConfigParser.get(self, *args, **kwargs)
```
Hwesta: *args is no longer defined

Hwesta:
:+1: Looking forward to seeing this used.

PR #352, opened 2015-12-07
MCPClient: README - clientScripts logging convention

sevein:
Please feel free to reword. I'm not very good at this, but I wanted to write down somewhere our clientScripts convention about logging vs printing to stdout/stderr.


Hwesta:
Our logging module isn't set up to direct to stdout/stderr in client scripts yet, but we want to do that, so :+1:

PR #353, opened 2015-12-08
upload-archivistsToolkit: don't use atk module

mistydemeo:
The only thing this module was used for was obtaining a db connection object; the function being called was a very thin wrapper around MySQLdb.connect, so just use that instead.

refs #8850


Hwesta:
:+1:

PR #354, opened 2015-12-08
Delete views and databaseInterface

sevein:
Uses 32b86a8 from #355 (CC @mistydemeo).

I couldn't test `_find_uuid_of_transfer_in_originals_directory_using_path` because it doesn't seem to be used at all, so maybe we can just delete it.


src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -20,166 +20,205 @@
...
+    'Completed successfully': '#dff0d8',
+    'Failed': '#f2dede',
+    'default': 'yellow',
+}
 
-def getEmailsFromDashboardUsers():
-    return [u[0] for u in User.objects.filter(is_active=True).values_list('email').exclude(email__in=['demo@example.com', ''])]
+def get_emails_from_dashboard_users():
+    users = User.objects.filter(is_active=True).values_list('email').exclude(email__in=['demo@example.com', ''])
+    return [u[0] for u in users]
```
Hwesta: Tip: This can be done with `flat=True`, eg `return User.objects.filter(is_active=True).values_list('email', flat=True).exclude(email__in=['demo@example.com', ''])`

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -20,166 +20,205 @@
...
 import django
 django.setup()
 # dashboard
+from django.db import connection
 from django.contrib.auth.models import User
+from main.models import Job, Report
 
 # archivematicaCommon
 from custom_handlers import get_script_logger
 import databaseInterface
```
Hwesta: Remove this, since it's unused now.

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -20,166 +20,205 @@
...
+    to2 = ", ".join(to)
     msg = MIMEMultipart('alternative')
     msg['Subject'] = subject
-    msg['From'] = from_
+    msg['From'] = efrom
     msg['To'] = to2
-    
+
     part1 = MIMEText(content, 'html')
     part2 = MIMEText("This email requires rendering of html to view", 'plain')
```
Hwesta: `part2` should be deleted, or re-added to the message below.

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -20,166 +20,205 @@
...
+
+
+def get_unit_job_log_html(sip_uuid):
+    """
+    Generate a string containing a HTML table with the jobs of a given unit
+    identifier.
+    """
+    parser = etree.HTMLParser(remove_blank_text=True)
+    jobs = Job.objects.filter(sipuuid=sip_uuid, subjobof='').exclude(jobtype='Email fail report').order_by('-createdtime', '-createdtimedec')
+    rows = [(item.jobtype, item.currentstep, item.createdtime) for item in jobs]
```
Hwesta: Tip: These two lines can be `rows = Job.objects.filter(sipuuid=sip_uuid, subjobof='').exclude(jobtype='Email fail report').order_by('-createdtime', '-createdtimedec').values_list('jobtype', 'currentstep', 'createdtime')`

src/MCPClient/lib/clientScripts/archivematicaFITS.py
```diff
@@ -33,93 +32,91 @@
...
-        exit(0)
+    if not FPCommandOutput.objects.filter(file=file_uuid).count():
+        logger.error('Warning: Fits has already run on this file. Not running again.')
+        return 0
 
-    tempFile="/tmp/" + uuid.uuid4().__str__()
-
-    command = "fits.sh -i \"" + escapeForCommand(target) + "\" -o \"" + tempFile + "\""
+    _, temp_file = tempfile.mkstemp()
+    args = 'fits.sh -i {} -o {}'.format(target, temp_file)
```
Hwesta: If this is built as an array, we won't have to worry about shlex.split, and executeOrRun knows how to handle it.  Suggest `args = ['fits.sh', '-i', target, '-o', temp_file]`

src/MCPClient/lib/clientScripts/archivematicaFITS.py
```diff
@@ -33,93 +32,91 @@
...
+    """
+    if file_grpuse in ('DSPACEMETS', 'maildirFile'):
+        logger.error("File's fileGrpUse in exclusion list, skipping")
+        return 0
 
-    sql = """SELECT fileUUID FROM FPCommandOutput WHERE fileUUID = %s;"""
-    if len(databaseInterface.queryAllSQL(sql, (fileUUID,))):
-        print >>sys.stderr, "Warning: Fits has already run on this file. Not running again."
-        exit(0)
+    if not FPCommandOutput.objects.filter(file=file_uuid).count():
```
Hwesta: `.exists()` is probably what you're looking for. https://docs.djangoproject.com/en/dev/ref/models/querysets/#exists

Hwesta:
> I couldn't test _find_uuid_of_transfer_in_originals_directory_using_path because it doesn't seem to be used at all, so maybe we can just delete it.

That seems pretty safe for deletion.

src/MCPClient/lib/clientScripts/upload-archivistsToolkit.py
```diff
@@ -27,6 +27,8 @@
 import MySQLdb
 import databaseInterface
```
Hwesta: This should also be removed.

Hwesta:
Nice cleanups!  Mostly just tips, but the last imports of databaseInterface should be removed before  :shipit:

Hwesta:
Please remove from `recreateDB.sh` since mysql2Views doesn't exist anymore.

```
echo "Creating MCP Views"
mysql -u root "${dbpassword}" --execute="USE ${databaseName}; SOURCE $currentDir/../src/MCPServer/share/mysql2Views;
```

sevein:
Right, good catch. Thanks, it's done!

Hwesta:
:+1: Glad to see it go!

PR #355, opened 2015-12-09
upload-archivistsToolkit: use agentarchives

mistydemeo:
Relies on artefactual-labs/agentarchives#13.


mistydemeo:
I'd like to test this but I don't have a copy of Archivist's Toolkit with data!

Hwesta:
:+1:

PR #356, opened 2015-12-09
MCPClient: create logger after django.setup()

sevein:
The attribute `config` in the `logging` module is not available until
`django.setup()` is invoked. This commit pospones the creation of our logger
`archivematica.mpc.client`.


Hwesta:
:+1:

PR #357, opened 2015-12-09
Remove archivematicaFITS

mistydemeo:
This is connected to https://github.com/archivematica/Issues/issues/77.


sevein:
`src/MCPClient/etc/archivematicaClientModules` still has a reference:

```
FITS_v0.0 = %clientScriptsDirectory%archivematicaFITS.py
```

Should that be deleted?

mistydemeo:
Removed.

The only workflow that seems to still use this is maildir, which isn't working either way.

Hwesta:
Remove DB entries that reference FITS (and associated)?

sevein:
I think that already happened in `share/1.1.0` and `share/1.2.0`?

mistydemeo:
> Remove DB entries that reference FITS (and associated)?

Some of them are in the middle of large chains that we're not removing yet, so I'll leave those there until we remove all of maildir at once.

Updated to switch the manualNormalization chain to use the new characterization code; haven't had the chance to test this yet, though. Our manual normalization test transfers aren't working for me.

sevein:
I've filed an issue: https://github.com/archivematica/Issues/issues/77.
Closing this PR until we have time to resume work.

PR #358, opened 2015-12-14
Use string environment to configure dashboard

sevein:
Stop touching `sys.path` from the application and update `django.wsgi` to set process environment variables coming from the WSGI environment (in our case, from `mod_wsgi`). See the commit messages for more details.

It also fixes the integration with the `monitor` module, which monitors changes in the source code every second and touches the `django.wsgi` file when it happens, forcing `mod_wsgi` to allocate new processes with the new code. Only used if `settings.DEBUG` equals `True`.

It requires changes in the deployment process. Please see https://github.com/artefactual-labs/ansible-role-archivematica-src/pull/30 for more details.


Hwesta:
Heads up: this and #375 will conflict, but not severely.

This breaks dashboard tests if the PYTHONPATH isn't set to include archivematicaCommon, but also makes them more flexible.  Previously if /usr/lib/archivematica/archivematicaCommon existed and pointed at archivematicaCommon, the tests just worked.  Now, archivematicaCommon can be anywhere (not hardcoded), but the PYTHONPATH has to be set. I think this is an improvement.

:+1: from me, but I'd like a second opinion. @jhsimpson @mistydemeo ?

sevein:
@jhsimpson gave me the ok IRL. Merged.

PR #359, opened 2015-12-14
elasticSearchFunctions: global client as module attribute

sevein:
Elasticsearch's client is thread-safe. We use it in two different contents:
our `dashboard` application and `clientScripts`.

This commit stores the Elasticserach instance as an attribute of its module.
Access to the instance if provided via `get_client()`, but it requires previous
initialization. In the `dashboard` this is achieved in `apache/django.wsgi` while
clientScripts have to initialize it manually for now. An exception is raised if
the user tries to access to the client before initialization.

Usage example:

```
import elasticSearchFunctions
elasticSearchFunctions.setup(hosts=['127.0.0.1:9200'])
client = elasticSearchFunctions.get_client()
results = client.search(**{})
```

It's been provided an extra setup method that reads the configuration param
from `clientConf.conf` to ease the transaction. This will be used by both
`dashboard` and `clientScripts` until they enable other ways to provide that
configuration, which is work in progress.

In addition, the dashboard receives a new middleware that replaces the old
`elasticsearch_required` decorator. See `middleware.common.ElasticsearchMiddleware`.
It redirects the user to a custom Elasticsearch error page when one of the two
following exceptions have been raised in the view: `ElasticsearchException` and
`ImproperlyConfigured`. The main advantage here is to avoid an extra hit to the
Elasticsearch server (as the previous `check_server_status` function used to
do).

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -21,12 +21,9 @@
 # @subpackage archivematicaCommon
 # @author Mike Cantelon <mike@artefactual.com>
 
-import base64
 import ConfigParser
-import cPickle
 import datetime
 import logging
-import MySQLdb
 import os
 import re
 import sys
@@ -45,11 +42,11 @@ import version
 sys.path.append("/usr/lib/archivematica/archivematicaCommon/externals")
 import xmltodict
 
-LOGGER = logging.getLogger("archivematica.common")
+from elasticsearch import Elasticsearch, ImproperlyConfigured
 
-from elasticsearch import Elasticsearch, ConnectionError, TransportError
 
-pathToElasticSearchServerConfigFile='/etc/elasticsearch/elasticsearch.yml'
+logger = logging.getLogger('archivematica.common')
+
 MAX_QUERY_SIZE = 50000  # TODO Check that this is a reasonable number
 MATCH_ALL_QUERY = {
     "query": {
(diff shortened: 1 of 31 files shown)


sevein:
I just rebased this so it merges cleanly.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -78,15 +78,74 @@
...
+
+def get_host():
+    """
+    Return one of the Elasticsearch hosts configured in our client. In the
+    future this function could look it up in the Elasticsearch client instead
+    of using the module attribute _es_hosts, because in an Elasticsearch
+    cluster, nodes can be added or removed dynamically.
+    """
+    if not _es_hosts:
+        raise ImproperlyConfigured('The Elasticsearch client has not been set up yet.')
```
Hwesta: Update this to mention the correct setup function? Eg `raise ImproperlyConfigured('The Elasticsearch client has not been set up yet. Please call setup() first')`

src/archivematicaCommon/tests/test_elasticsearch_functions.py
```diff
@@ -14,13 +14,14 @@
 class TestElasticSearchFunctions(unittest.TestCase):
 
     def setUp(self):
-        self.conn = Elasticsearch('127.0.0.1:9200')
+        hosts = os.environ.get('ELASTICSEARCH_SERVER', '127.0.0.1:9200')
+        self.client = elasticSearchFunctions.connect(hosts)
```
Hwesta: elasticSearchFunctions has no function `connect`. Should this be get_client?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -78,15 +78,74 @@
...
+
+def setup(hosts):
+    """
+    Initialize Elasticsearch client and share it as the attribute _es_client in
+    the current module. An additional attribute _es_hosts is defined containing
+    the Elasticsearch hosts (expected types are: string, list or tuple).
+    """
+    module = sys.modules[__name__]
+    setattr(module, '_es_hosts', hosts)
+    setattr(module, '_es_client', Elasticsearch(**{'hosts': _es_hosts}))
```
Hwesta: How about this?

```
_es_hosts = None
_es_client = None

def setup(hosts):
    global _es_hosts
    global _es_client
    # ...
    _es_hosts = hosts
    _es_client = Elasticsearch(**{'hosts': _es_hosts})
```

I think you only check for truthiness on _es_hosts & _es_client, and I think it's clearer to know that they're module attributes (and it keeps my linter from yowling).

src/dashboard/src/components/decorators.py
```diff
@@ -17,16 +17,13 @@
 
 import sys
 
-from django.shortcuts import render_to_response, render
+from django.shortcuts import render_to_response
 from django.http import Http404
 from django.utils.functional import wraps
 from django.template import RequestContext
 
-from contrib import utils
```
Hwesta, line 25: This is still used.

Hwesta:
Tests fail because cassettes mix `localhost:9200` and `127.0.0.1:9200` in URLs, and it fails to match.  Search-and-replace to all be `127.0.0.1:9200` should fix that.

sevein:
Heh, thanks for that cause' I was completely stuck on it.

```
src/archivematicaCommon/tests/test_elasticsearch_functions.py ......
==== 6 passed in 2.34 seconds
```

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -655,7 +676,7 @@ def index_transfer(conn, uuid, file_count, status=''):
     """
     Indexes transfer with UUID `uuid`
 
-    :param conn: The ElasticSearch connection.
+    :param client: The ElasticSearch client
```
Hwesta, line 683: The docstring was updated, but not the actual parameter name.

sevein:
I think that all the client scripts are now working.
Backlog and archival storage tabs are also working.

Hwesta:
localDevSetup/eraseElasticSearchIndexes.py also needs to be updated, thanks.

sevein:
`eraseElasticSearchIndexes.py` updated and tested.

Hwesta:
This will need to be rebased since #402 made it in first, but looks good! :shipit:

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -78,15 +75,83 @@
...
+
+
+def get_client():
+    """
+    Obtain the current Elasticsearch client. If undefined, an exception will be
+    raised. This function also checks the integrity of the indices expected by
+    this application and populate them when they cannot be found.
+    """
+    if not _es_client:
+        raise ImproperlyConfigured('The Elasticsearch client has not been set up yet. Please call setup() first.')
```
mistydemeo, line 150: Maybe just check for this and call `setup()` if needed, so the caller doesn't need to track this?

``` python
if _es_client is None:
    setup()
# ...
```

sevein, line 150: I'm not sure, because `setup()` excepts a list of hosts as a parameter, or you can call `setup_reading_from_client_conf()`. How would that look like?

mistydemeo, line 150: Hmmm, right. I wasn't taking into account the fact you'd added a new way to do this. Is calling `setup()` with an explicit host going to be the preferred way to do this going forward? I don't see a place that uses it right now.

mistydemeo, line 150: I think given the multiple ways to configure this, it probably doesn't make sense to automatically create the client, after thinking about it. Sorry for the noise!

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -246,7 +277,8 @@ def _sortable_string_field_specification(field_name):
         }
     }
 
-def set_up_mapping_aip_index(conn):
+
+def set_up_mapping_aip_index(client):
```
mistydemeo, line 281: This function changed in a recent qa/1.x commit, so there'll be a minor merge conflict on rebase.

src/archivematicaCommon/lib/elasticSearchFunctions.py
sevein, line 150: No problem!

src/archivematicaCommon/lib/elasticSearchFunctions.py
sevein, line 281: Yes, I'll rebase soon. Thank you!

mistydemeo:
:+1: 

After the discussion the other day, I have a better idea the purpose this serves - looks good to me! I'd be glad to include after a rebase.

PR #360, opened 2015-12-14
Move getDashboardUUID into archivematicaFunctions

mistydemeo:
This is one of only a few functions in `elasticSearchFunctions` that uses the database, which I'm slowly moving out of this module.


Hwesta:
:+1:

sevein:
@mistydemeo I think that you have to merge this again because your first try wasn't pushed to the internal git server.

PR #362, opened 2015-12-15
ES: specify all fields for AIP mapping

mistydemeo:
Archival storage now performs some faceting on archival storage data, which requires those fields to exist in the mapping. Depending on automatic mapping structure meant that these queries would throw exceptions until the first time data was indexed containing these fields. This was manifested in archival storage as an exception when trying to search when the AIP store is empty.


sevein:
:+1:

Hwesta:
:+1:

PR #364, opened 2015-12-16
WIP: customize tokenization for path fields

mistydemeo:
The default tokenization used by Elasticsearch doesn't deal well with the ways that path strings are constructed; substring searches often fail to bring up files and SIPs which would be expected by users. For example, trying to search for "events" will fail to bring up files in a SIP named "Events_2005-d1c6c1a2-f613-497f-a527-64eb6861f03c".

This adds several new custom analyzers to perform tokenization, which split strings on path delimeters and on logical word separators within strings (such as `-` and `_`). Fields which need these analyzers are now created as multifields, allowing searches for those fields to attempt to match all of the defined subfields. This WIP PR tests this by defining `transferfile`'s `relative_path` field as a multifield with unanalyzed, path-analyzed, and word-analyzed copies of the field.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -104,6 +104,21 @@ PATH_STRING_MULTIFIELD = {
     },
 }
 
+MULTI_FIELDS = {
+    'transfers': {
+        'transferfile': {
+            'relative_path': ['raw', 'path', 'word'],
+        },
+        'transfer': {},
+    },
+    'aips': {
+        'aipfile': {
+            'filePath': ['raw', 'path', 'word'],
+        },
+        'aip': {},
+    },
+}
+
 class ElasticsearchError(Exception):
     """ Not operational errors. """
     pass
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -283,14 +283,12 @@ def aip_mapping_is_correct(client):
     return mapping['aipfile']['properties']['AIPUUID']['index'] == 'not_analyzed'
 
 
-def create_index(client, index, attempt=1):
-    if attempt > 3:
-        return
-    response = client.indices.create(index, ignore=400)
-    if 'error' in response and 'IndexAlreadyExistsException' in response['error']:
-        return
+def create_index(client, index):
+    client.indices.create(index, ignore=400)
+    # Always put the mapping, even if the index already exists, because
+    # there may have been changes to the mapping structure; rerunning this
+    # will allow additional fields to be added to the index
     set_up_mapping(client, index)
-    create_index(client, index, attempt + 1)
 
 
 def _sortable_string_field_specification(field_name):
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -338,7 +338,7 @@ def set_up_mapping_aip_index(client):
         'AICID': MACHINE_READABLE_FIELD_SPEC,
         'sipName': {'type': 'string'},
         'indexedAt': {'type': 'double'},
-        'filePath': {'type': 'string'},
+        'filePath': PATH_STRING_MULTIFIELD,
         'fileExtension': {'type': 'string'},
         'origin': {'type': 'string'},
         'identifiers': {'type': 'string'},
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -76,6 +76,33 @@ MACHINE_READABLE_FIELD_SPEC = {
     'index': 'not_analyzed'
 }
 
+CUSTOM_ANALYZERS = {
+    'path': {
+        'type': 'pattern',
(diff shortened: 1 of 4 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -61,6 +61,19 @@
...
+CUSTOM_ANALYZERS = {
+    'path': {
+        'type': 'pattern',
+        # splits on path delimiters (/ and \)
+        'pattern': '(\\/|\\\)',
+    },
+    'word': {
+        'type': 'pattern',
+        # splits on path delimiters, underscores, and dashes
+        'pattern': '(\\/|\\\|_|-)'
```
Hwesta, line 88: Also split on `.` maybe?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -206,20 +219,13 @@ def aip_mapping_is_correct(conn):
...
-        response = conn.indices.create(index, ignore=400)
-        if 'error' in response and 'IndexAlreadyExistsException' in response['error']:
-            return conn
-
-        set_up_mapping(conn, index)
-        conn = connect_and_create_index(index, attempt + 1)
-    else:
-        conn = False
-
+def connect_and_create_index(index):
```
Hwesta: Why remove the retry?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -206,20 +219,13 @@ def aip_mapping_is_correct(conn):
...
+    conn = Elasticsearch(hosts=getElasticsearchServerHostAndPort())
+    conn.indices.create(index, ignore=400)
+    # Always put the mapping, even if the index already exists, because
+    # there may have been changes to the mapping structure; rerunning this
+    # will allow additional fields to be added to the index
```
Hwesta: Update put_mapping to `ignore=[400,404]` as well?

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: As written, the original version could never have worked as expected. It appeared to rely on `put_mapping` silently failing, which may or may not have been possible with pyes but which is definitely not possible in the current configuration.

src/archivematicaCommon/lib/elasticSearchFunctions.py
mistydemeo: What's the circumstances in which `404` might happen when creating the index?

src/archivematicaCommon/lib/elasticSearchFunctions.py
Hwesta: Ahh, that makes sense.

src/archivematicaCommon/lib/elasticSearchFunctions.py
Hwesta: Heh, not sure.  I was looking at http://elasticsearch-py.readthedocs.org/en/master/api.html?highlight=ignore#ignore which suggested 404 might be useful, but possibly not in this context.

Hwesta:
I like this idea, and I'd like to see the better analyzers on other fields.

mistydemeo:
Added several new commits to update how queries are constructed.

After quite a bit of testing, I determined that `match` queries weren't behaving well on these multifields; they were generally matching against only one of the subfields, which meant that queries that should have been valid weren't producing expected results. For example, matching a term against `relative_path` or `relative_path.*` returned no results, whereas matching it against `relative_path.word` returned the desired result.

We already have some support in our query generation code to expand queries on object fields into [multi_match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html) queries on the string subfields of those objects. I've updated this to support path fields like `relative_path`, and added a constant to `elasticSearchFunctions` which explicitly marks the subfields so we don't have to query the index for the structure in this case. This means queries on path fields go from looking like this:

``` json
{
  "filter": {
    "term": {
      "status": "backlog"
    }
  },
  "query": {
    "bool": {
      "must": [],
      "must_not": [],
      "should": [
        {
          "multi_match": {
            "query": "multimedia",
            "fields": [
              "relative_path.*"
            ]
          }
        }
      ]
    }
  }
}
```

To looking like this:

``` json
{
  "filter": {
    "term": {
      "status": "backlog"
    }
  },
  "query": {
    "bool": {
      "must": [],
      "must_not": [],
      "should": [
        {
          "multi_match": {
            "query": "multimedia",
            "fields": [
              "relative_path.raw",
              "relative_path.path",
              "relative_path.word"
            ]
          }
        }
      ]
    }
  }
}
```

With the latter query returning results which the former query will not. I haven't yet encountered any cases where this returns undesirable results, or otherwise has negative impact on existing queries.

I've also updated our search form JS to permit "path" as a type, which queries `relative_path` directly. We previously only had filename searches, which isn't sufficient for what we're doing here. Based on discussions with @jhsimpson, I think we should promote that to the default search field instead of "any".

mistydemeo:
Deployed to missy for QA.

mistydemeo:
Rebased on the new ES changes.

sevein:
I'm closing this PR as Misty is not actively working on this anymore - the branch will be maintained. I've filed a new issue to track the problem that she described - https://github.com/archivematica/Issues/issues/144. Thanks!

PR #368, opened 2015-12-17
Fixed issue with passing of MySQL password to mysql_dev.

mcantelon:
The password input by the user was getting modified into a MySQL argument before being passed to mysql_dev.


mistydemeo:
:+1:

Hwesta:
Good catch, this must have come in with #336 and everybody's using ansible so nobody noticed. :)

mcantelon:
Merged.

PR #369, opened 2015-12-21
elasticSearchFunctions: properly handle document_id_from_field_query failure

mistydemeo:
Fixes #9256.


Hwesta:
:+1: Better, but doesn't solve underlying problem.

mistydemeo:
Yeah, this'll just provide a usable error message. :( I'm going to adjust the chainlinks in another PR.

PR #370, opened 2016-01-01
Transfer: add API to retrieve transfer source locations

mistydemeo:


src/dashboard/src/components/transfer/views.py
```diff
@@ -64,6 +64,16 @@ def grid(request):
     hide_features = helpers.hidden_features()
     return render(request, 'transfer/grid.html', locals())
 
+def transfer_source_locations(request):
+    try:
+        return helpers.json_response(storage_service.get_location(purpose="TS"))
+    except:
```
Hwesta, line 69: Log something about why it failed? logger.exception?

Hwesta:
What's the use case for this? Since I don't see anything that calls it yet (did you forget to add?).

mistydemeo:
It's being used by the new transfer browser. The current version interpolates the transfer source directories into JavaScript at rendertime using Django templates; the new transfer browser instead requests them clientside after the page loads. Aside from being cleaner, it'll also allow the user to refresh the source directories after the page has been loaded without reloading the entire page (for example, if the storage service was broken when the page was loaded).

Hwesta:
Ah, nice!

Possibly out of scope for this, but it seems like this endpoint and `children/location/<location_uuid>/` from filesystem_ajax/urls.py should be together.

:+1:

src/dashboard/src/components/transfer/views.py
mistydemeo, line 69: Added a `logger.exception` call.

mistydemeo:
> Possibly out of scope for this, but it seems like this endpoint and children/location/<location_uuid>/ from filesystem_ajax/urls.py should be together.

Yeah, the current placement is a bit strange. I think the `filesystem_ajax` one is there because it was placed with the original version which actually hit the filesystem. Those APIs didn't really get reorganized when this got moved to the storage service.

Hwesta:
I think another PR for the API cleanup stuff, then. Probably related to the transfer browser changes.

PR #371, opened 2016-01-01
filesystem_ajax: return non-200 response on start_transfer error

mistydemeo:
refs #4757

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -16,6 +16,9 @@ class ResourceNotFound(Exception):
 class BadRequest(Exception):
     pass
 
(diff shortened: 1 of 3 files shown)


Hwesta:
Tests would be nice for this, but I don't think we have any existing to tweak.

:+1:

mistydemeo:
Yeah, unfortunately we don't have tests for this, and they're not straightforward to run right now.

mistydemeo:
Test problems I ran into:
- It's very hard to run the dashboard in a virtualenv, due to many requirements not specified in the requirements.txt.
- The dashboard tests can't run without several archivematicaCommon requirements, and also needs direct MySQLdb access and gearman in order to run `contrib.mcp.client`.

PR #372, opened 2016-01-05
Index transfer contents once

mistydemeo:
The current indexing process indexes files near the end of the transfer process, before a storage decision has been made. Subsequently, if the user decides to send the files to backlog, the records must be retrieved from the index and updated with a new status.

This instead defers indexing until after a storage decision has been made, and indexes them once with the final status.

Fixes #9265.


src/MCPClient/lib/clientScripts/backlogUpdatingTransferFileIndex.py
```diff
@@ -1,41 +0,0 @@
...
-
-if __name__ == '__main__':
-    logger = get_script_logger("archivematica.mcp.client.backlogUpdatingTransferFileIndex")
-
-    #"%SIPUUID%" "%SIPName%" "%SIPDirectory%"
-    transferUUID = sys.argv[1]
-    transferName = sys.argv[2]
-    transferDirectory = sys.argv[3]
-    print 'Processing ' + transferUUID + '...'
-    found = elasticSearchFunctions.connect_and_change_transfer_file_status(transferUUID, 'backlog')
```
Hwesta, line 40: I don't think `connect_and_change_transfer_file_status` is used elsewhere - can it be removed?

src/MCPServer/share/mysql_dev_9265_index_after_processing_decision.sql
```diff
@@ -0,0 +1,54 @@
...
+    SET nextMicroServiceChainLink=@sendToBacklog
+    WHERE microServiceChainLink=@updateTransferIndex;
+
+-- Finally, mark nextMicroServiceChainLink as null for the new last link
+UPDATE MicroServiceChainLinksExitCodes
+    SET nextMicroServiceChainLink=NULL
+    WHERE microServiceChainLink='561bbb52-d95c-4004-b0d3-739c0a65f406';
+
+-- Insert a new chainlink at the end of the "create one SIP" chain
+-- to index files with no specified status
```
Hwesta: Do we need this? I don't think non-backlogged files are ever queried from the transfer index.

src/MCPClient/lib/clientScripts/elasticSearchIndexProcessTransfer.py
```diff
@@ -52,12 +52,17 @@
...
+        except IndexError:
+            status = ''
 
         exitCode = elasticSearchFunctions.connect_and_index_files(
             'transfers',
             'transferfile',
             transferUUID,
-            pathToTransfer
+            pathToTransfer,
+            status=status
```
Hwesta: Nitpick: add trailing comma?

src/MCPClient/lib/clientScripts/backlogUpdatingTransferFileIndex.py
mistydemeo, line 40: Removed.

src/MCPClient/lib/clientScripts/elasticSearchIndexProcessTransfer.py
mistydemeo: Trailing comma added.

src/MCPServer/share/mysql_dev_9265_index_after_processing_decision.sql
mistydemeo: I don't think we use them, but we've been indexing files there for some time; there might be an expectation by some of our users they should be there?

mistydemeo:
Updated to stop indexing transfer files in the "create one SIP" workflow.

src/MCPServer/share/mysql_dev_9265_index_after_processing_decision.sql
```diff
@@ -0,0 +1,46 @@
...
+    SET defaultMicroServiceChainLink=@create_transfer_metadata_xml
+    WHERE defaultMicroServiceChainLink=@index_transfer_contents;
+
+-- "Identify DSpace METS file" should point at the chainlink after indexing
+UPDATE MicroServiceChainLinks
+    SET defaultNextChainLink=@create_transfer_metadata_xml
+    WHERE pk='8ec0b0c1-79ad-4d22-abcd-8e95fcceabbc';
+UPDATE MicroServiceChainLinksExitCodes
+    SET nextMicroServiceChainLink=@create_transfer_metadata_xml
+    WHERE nextMicroServiceChainLink=@index_transfer_contents;
```
Hwesta, line 20: Need to also delete the skipped index_transfer_contents MSCL (and TasksConfig/StandardTasksConfig if they're not used)

mistydemeo, line 20: Updated; deleted the STC and MSCLExitCodes rows in this file, and the MSCL and TC in the deletion file (since they may not be deletable).

Hwesta:
:+1:

PR #374, opened 2016-01-11
Upgrade to modern Bootstrap and jQuery

mistydemeo:
This branch updates several of our vendored libraries, and the places that use them. Major changes:
## Upgrades the vendored bootstrap from 1.x to 3.3.6

This also updates many of the templates in order to ensure compatibility. A few bits of the app now look subtly different, but for the most part things work as they did before.
## Upgrades the vendored jQuery to the latest release, along with the other vendored libraries we use.
# Known issues
- [ ] Modals are completely broken. @mcantelon and I both looked at this and couldn't figure out what was wrong. Suggest using the new transfer browser to replace the transfer's modal, and eliminating ingest's "enter AtoM slug" modal.
- [x] Installer not centred
- [x] Select fields not styled
- [x] Preservation planning unstyled
- [x] Admin/failures appear below sidebar
- [x] Jobs width not properly fluid
- [x] Messages don't display
- [ ] SIP arrange drag & drop broken
- [ ] Headers in the SIP tables (transfer/ingest tabs) are positioned strangely
- [ ] "Connected" icon in the corner isn't vertically centred properly
- [ ] Inline forms not styled properly
- [ ] Navbar active tab over effect doesn't look right
- [ ] User dropdown background color is changed and it shows a blue line when active
- [ ] Some margin needed after forms and buttons in the ingest and backlog tabs
- [ ] Review AIP page not centered


Hwesta:
Good start! From playing around with it, I found the following oddities:
- Installer not correctly centered - it move around from page to page
- Dropdown menus (eg selecting transfer source location, in processing config choices) aren't styled, which looks odd since everything else is
- Preservation planning is unstyled
- admin/failures appears below the sidebar - this may happen elsewhere too
- Ingest tab is completely greyed out, as if it should have a modal but doesn't. Cant interact with anything
- Jobs widths not properly fluid, the actions column is just extra wide
- Messages (warning, success, error) don't display (eg when there's an error in processing config)

mistydemeo:
> Preservation planning is unstyled

This is a separate repo, so I can't update that as a part of this PR.

> Dropdown menus (eg selecting transfer source location, in processing config choices) aren't styled, which looks odd since everything else is

Fixed in processing config choices. Not touching transfer source location, since that's being replaced.

This requires a CSS class in Bootstrap 3, it looks like; I'm guessing Bootstrap 1 automagically styled every `<select>` element.

> Ingest tab is completely greyed out, as if it should have a modal but doesn't. Cant interact with anything

This is one of the broken modals, yep. Not sure what's going on, but it's simply going to be removed and replaced with something else.

mistydemeo:
> admin/failures appears below the sidebar - this may happen elsewhere too

How wide is your screen? This looks like it's the responsiveness in action; it displays side by side if there's room, underneath otherwise.

Maybe I'm allocating too much width to the right side of the admin page, though... maybe I can make it less aggressive about reflowing.

Similarly, the navbar can overflow to two lines or to display vertically depending on the amount of room available.

mistydemeo:
Eliminated the AtoM modal, which makes the ingest tab viewable again.

mistydemeo:
> Installer not correctly centered - it move around from page to page

I can't repro this; can you show me what you mean?

Hwesta:
> This is a separate repo, so I can't update that as a part of this PR.

Update in that repo, and update submodule commit?

> Maybe I'm allocating too much width to the right side of the admin page, though... maybe I can make it less aggressive about reflowing.

Failures is styled with col-md-12, while other pages are col-md-9.

It looks like all our forms need styling TLC - metadata entry, most of the admin tab pages, archival storage search are all unstyled by default, and I didn't do an exhaustive search.  Should bootstrap be styling everything inside a `<form>`, or should we update all of them?

Mouseover active tab in navbar makes it unreadable - should probably make text dark, or not respond to mouseover (old behaviour). Same for user dropdown.

Hwesta:
> I can't repro this; can you show me what you mean?

It looks like this to me.  Notice that the second screen is further left than the other two.

installer1
installer2
installer3

mistydemeo:
> It looks like this to me. Notice that the second screen is further left than the other two.

Huh, interesting. I can't get that to repro at all. What browser is this? What is your window size? Does it still happen if you resize your browser smaller? I wonder if I can't see it because my laptop screen is too small to repro...

mistydemeo:
> Failures is styled with col-md-12, while other pages are col-md-9.

Updated.

> It looks like all our forms need styling TLC - metadata entry, most of the admin tab pages, archival storage search are all unstyled by default, and I didn't do an exhaustive search. Should bootstrap be styling everything inside a `<form>`, or should we update all of them?

I think Bootstrap 1.x automagically styled these, but Bootstrap 3.x doesn't. So yeah, we should be applying some new classes to the elements.

Hwesta:
> Huh, interesting. I can't get that to repro at all. What browser is this? What is your window size? Does it still happen if you resize your browser smaller? I wonder if I can't see it because my laptop screen is too small to repro...

Firefox 43.0 on Ubuntu, resolution is 1920x1080, actual browser width ~1457.  Still happens when browser width is ~1000px.  Also happens in Chromium 47.0.2526.73 Ubuntu 14.04 (64-bit) with browser width ~1400 and ~1000

mistydemeo:
Pushed a few more commits:
- Fixed alerts
- Fixed wrapping in the job view

mistydemeo:
> Admin/failures appear below sidebar

After some testing and the tweaks that have already been done, this appears to only happen if the screen is insufficiently wide, so I'm willing to take this as working okay now.

mistydemeo:
At this point, I think the major things remaining are:
- Preservation planning, which is a repo-wide change in fpr-admin

mistydemeo:
Preservation planning is addressed in artefactual/archivematica-fpr-admin#15.

Hwesta:
This has been replaced by #488

jraddaoui:
Merged in https://github.com/artefactual/archivematica/commit/89023d98da2cda8110c0cb1bc7a7b7f0401e6a27

PR #375, opened 2016-01-13
Workaround to troubleshoot mod_wsgi issues

hakamine:
Useful to troubleshoot "RuntimeError: populate() isn't reentrant"
occurring in Django 1.7.

See http://stackoverflow.com/questions/30954398/django-populate-isnt-reentrant

### DIFF
--- a/src/dashboard/src/apache/django.wsgi
+++ b/src/dashboard/src/apache/django.wsgi
@@ -17,7 +17,9 @@
 
 import os, sys
 from django.core.wsgi import get_wsgi_application
-
+import time
+import traceback
+import signal
 # Ensure that the path does not get added multiple times
 path = '/usr/share/archivematica/dashboard'
 if path not in sys.path:
@@ -25,7 +27,20 @@ if path not in sys.path:
 
 os.environ['DJANGO_SETTINGS_MODULE'] = 'settings.common'
 
-application = get_wsgi_application()
+# workaround to troubleshoot mod_wsgi issues
+# See http://stackoverflow.com/questions/30954398/django-populate-isnt-reentrant 
+
+try:
+    application = get_wsgi_application()
+    print 'WSGI without exception'
+except Exception:
+    print 'handling WSGI exception'
+    # Error loading applications
+    if 'mod_wsgi' in sys.modules:
+        traceback.print_exc()
+        os.kill(os.getpid(), signal.SIGINT)
+        time.sleep(2.5)
+
 
 # See http://blog.dscpl.com.au/2008/12/using-modwsgi-when-developing-django.html
 from django.conf import settings


src/dashboard/src/apache/django.wsgi
```diff
@@ -25,7 +27,20 @@ if path not in sys.path:
...
 
-application = get_wsgi_application()
+# workaround to troubleshoot mod_wsgi issues
+# See http://stackoverflow.com/questions/30954398/django-populate-isnt-reentrant 
+
+try:
+    application = get_wsgi_application()
+    print 'WSGI without exception'
+except Exception:
+    print 'handling WSGI exception'
```
Hwesta, line 37: Please update print statements to be compatible with Python3 eg `print('handling WSGI exception')`

src/dashboard/src/apache/django.wsgi
```diff
@@ -25,7 +27,20 @@ if path not in sys.path:
...
+    # Error loading applications
+    if 'mod_wsgi' in sys.modules:
+        traceback.print_exc()
+        os.kill(os.getpid(), signal.SIGINT)
+        time.sleep(2.5)
+
```
Hwesta, line 43: Should this re-raise the exception?  Some of the examples in https://groups.google.com/forum/#!topic/modwsgi/X3Upsk9Wfsw (linked from the SO question) have

```
except Exception:
    print('handling WSGI exception')
    # Error loading applications
    if 'mod_wsgi' in sys.modules:
        traceback.print_exc()
        os.kill(os.getpid(), signal.SIGINT)
        time.sleep(2.5)
   raise
```

The raise will probably kill that process, but I'm not sure of other ramifications.

sevein:
Please rebase when you have a chance. Your changes will conflict with baeff8d7f29ad41b45a67a61fcc54c18d2f3ed50.

sevein:
@hakamine, do you know a way to reproduce the `RuntimeError: populate() isn't reentrant` error? I want to integrate your solution with `baeff8d` but I'd like to reproduce first.

hakamine:
It happened when a debian package was missing in the VM installation, but I can't recall which one. You can try uninstalling: libapache2-mod-wsgi, python-gearman. 
(What caused the problem: I had removed the archivematica debian packages and then installed archivematica from source. Sometime later that I did apt-get autoremove, and this removed some archivematica package dependencies, breaking archivematica. Using the code in the PR I could track which packages were missing )

sevein:
This is not working for me though when combined with `baeff8d7f29ad41b45a67a61fcc54c18d2f3ed50`.
I can see the traceback (missing gearman module) but only once after starting Apache, with or without this patch.

We should probably find a better solution like giving a try to the `mod_wsgi` package in pip (which is not stuck on Ubuntu's old package, see https://github.com/GrahamDumpleton/mod_wsgi - but requires `apache-dev`) or make the final move to Nginx when we have a chance.

sevein:
@hakamine, feel free to keep trying (after rebasing) or close this PR.

PR #376, opened 2016-01-14
Migrations: SQL changes to make adding migrations easier

Hwesta:
Migrations are likely to come in with the 1.6 release, and these fixes will make it easier.
- Boolean field StandardTasksConfigs.requiresOutputLock doesn't have NULL values, only True/False
- MSCLs with no TasksConfig are deleted
- Add a synthetic key to FPCommandOutput.  MySQL seems to handle the AUTO_INCREMENT correctly if the table already has data.
- Add missing lastmodified to IDCommand from #120


mistydemeo:
:+1:

Hwesta:
Also cherry-picked to qa/1.x as deb55f182bc89dff0ff01426cbba7f4c4b53e3f0

PR #377, opened 2016-01-15
Use Django migrations

Hwesta:
Use Django's migrations, instead of custom-written SQL for database changes.  This assumes #376 will be in 1.5 - fixture data & schema will be re-generated/squashed once that is true.
- Add initial schema migration
- Load initial data from a fixture using the loaddata command in the data migration
- Add migration for changes since 1.5.
- Remove all mysql_dev migrations.

TODO:
- [x] Add FPR migrations & initial data artefactual/archivematica-fpr-admin#14
- [ ] Investigate why initial schema migration is slow
- [ ] Check for better fixture loading so tests don't have load all production data (or make it faster)
- [ ] Test with upgrades
- [x] Test with ansible artefactual-labs/ansible-role-archivematica-src#34


mistydemeo:
All this destroyed code... makes me happy.

mistydemeo:
:+1: Looks good to me.

re 14ef79d18e3a5c792350171516789c99b12ea1d5, maybe squash that into the preceding commit, since there isn't an associated migration?

Hwesta:
This was done as part of #409, and merged to qa/1.x as part of #462

PR #378, opened 2016-01-15
Experiment: try forcing browser to send data as UTF-8

mistydemeo:


mistydemeo:
This is, regrettably, not a real fix for the issue.

PR #379, opened 2016-01-18
Replace vendored base64 helpers with submodule

mistydemeo:


src/dashboard/src/media/js/vendor/base64.js
```diff
@@ -1,43 +0,0 @@
-// Adapted from https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64.btoa
```
Hwesta, line 1: Symlink rather than fixing the places that import it?

mistydemeo, line 1: This is a symlink!

Hwesta, line 1: Sorry, that was unclear. You're symlinking instead of fixing the places that import it? Wouldn't it be easier long term to modify the imports instead of having the indirection?

Hwesta:
:+1:

PR #381, opened 2016-01-19
unitTransfer: tempPath is already a unicode string

mistydemeo:
Fixes #9030.

### DIFF
--- a/src/MCPServer/lib/unit.py
+++ b/src/MCPServer/lib/unit.py
@@ -44,8 +44,7 @@ class unit:
         self.fileList = {}
         # currentPath must be a string to return all filenames as bytestrings,
         # and to safely concatenate with other bytestrings
-        currentPath = str(self.currentPath.replace("%sharedPath%", \
-                                                       archivematicaMCP.config.get('MCPServer', "sharedDirectory"), 1) + "/")
+        currentPath = os.path.join(self.currentPath.replace("%sharedPath%", archivematicaMCP.config.get('MCPServer', "sharedDirectory"), 1), "").encode('utf-8')
         try:
             for directory, subDirectories, files in os.walk(currentPath):
                 directory = directory.replace(currentPath, self.pathString, 1)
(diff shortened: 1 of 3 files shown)


Hwesta:
:+1:

src/MCPServer/lib/unit.py
```diff
@@ -44,8 +44,8 @@ def reloadFileList(self):
         self.fileList = {}
         # currentPath must be a string to return all filenames as bytestrings,
         # and to safely concatenate with other bytestrings
-        currentPath = str(self.currentPath.replace("%sharedPath%", \
-                                                       archivematicaMCP.config.get('MCPServer', "sharedDirectory"), 1) + "/")
+        currentPath = (self.currentPath.replace("%sharedPath%", \
+                                                       archivematicaMCP.config.get('MCPServer', "sharedDirectory"), 1) + "/").encode('utf-8')
```
Hwesta: Nitpick: Having both () and \ to handle the newline is overkill, I'd probably just put it all on one line and let the editor handle wrapping (especially since it's only 9 characters shorter split)

mistydemeo: Sounds good, I'll do that. I used `()` not for the newline though, but because this ends in a `+ "/"`; I'll just switch it to `os.path.join`.

Hwesta: Ohh, I'd missed the +'/' reason.

Hwesta:
:+1:

src/MCPServer/lib/unit.py
mistydemeo: Another good reason to change it - it wasn't very readable before!

Hwesta:
Also cherry-picked to 1.5 as 43705615061fdc9d226c5da58e51cb60a7fc104c 54e83c6d1307919139ccb1ff1cfe8af543af3c46 3a69144a16f5d69215b0a5b1a67aa4b6c6ef5fb3

PR #382, opened 2016-01-19
RCPServer: never return empty strings

mistydemeo:
- Several of these conditions were unreachable, e.g. the return value could never have been falsy.
- An empty string is not a valid cPickle serialization, so this results in confusing, hard-to-debug exceptions in places consuming these values.

Refs #9030.


Hwesta:
:+1:

mistydemeo:
Added a second commit which checks for failed jobs and raises an exception.

Hwesta:
:+1:

PR #384, opened 2016-01-20
Full AIP re-ingest

sevein:
See also: artefactual/archivematica-storage-service#100.

Five client scripts have received updates:
- archivematicaAssignFileUUID.py
- archivematicaSetTransferType.py
- archivematicaUpdateSizeAndChecksum.py
- restructureForCompliance.py
- verifyBAG.py


src/archivematicaCommon/lib/option_parsers.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import uuid
+
+
+def valid_uuid(uuid_str):
+    valid = False
```
mistydemeo: Instead of setting a `valid` variable, maybe just return the value directly?

``` python
def valid_uuid(uuid_str):
    if not uuid_str:
        return False
    try:
        uuid.UUID(uuid_str)
    except ValueError:
        return False

    return True
```

src/dashboard/src/components/api/views.py
```diff
@@ -392,49 +395,81 @@ def approve_transfer_via_mcp(directory, transfer_type, user_id):
...
+    source = os.path.join(shared_directory_path, 'tmp', sip_name)
+
+    if target == 'transfer':
+        dest = os.path.join(shared_directory_path, 'watchedDirectories', 'activeTransfers', 'standardTransfer')
+        name_has_uuid = len(sip_name) > 36 and re.match(UUID_REGEX, sip_name[-36:]) is not None
+        if name_has_uuid:
+            dest = os.path.join(dest, os.path.basename(source[:-37]))
+            if os.path.isdir(dest):
+                response = {'error': True, 'message': 'There is already a transfer in standardTransfer with the same name.'}
+                return helpers.json_response(response, status_code=500)
```
mistydemeo: This feels more like a client error than a server error; maybe use 400?

src/MCPClient/lib/clientScripts/archivematicaSetTransferType.py
```diff
@@ -35,4 +35,9 @@
     transferUUID = sys.argv[1]
     transferType = sys.argv[2]
 
-    Transfer.objects.filter(uuid=transferUUID).update(type=transferType)
+    transfer = Transfer.objects.get(uuid=transferUUID)
+    if transfer.type is not None:
+        sys.exit(0)
+
+    transfer.type = transferType
+    transfer.save()
```
Hwesta, line 44: Could this be `Transfer.objects.filter(uuid=transferUUID, type__isnull=False).update(type=transferType)` ?

src/archivematicaCommon/lib/option_parsers.py
```diff
@@ -0,0 +1,39 @@
...
+        return False
+
+    return True
+
+
+class UUIDArgAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        if not valid_uuid(values):
+            return
+        setattr(namespace, self.dest, values)
```
Hwesta: I think this can be done with builtins.  `parser.add_argument(..., action=uuid.UUID)` retuns a UUID object, which can be converted to a string with `str`.  It sacrifices a good error message, but `parser.add_argument('-u', type=lambda x: str(uuid.UUID(x)))` will return a string.

sevein: Readability vs conciseness :)

Hwesta: I found it more confusing to have a UUIDArgAction, since I'm familiar with Python's builtin stuff, but not with custom things, so I think it's more readable and more concise. :)

sevein: Ok

src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
```diff
@@ -17,26 +17,127 @@
...
+    size = checksum = checksum_type = None
+    r = pobject.xpath('premis:objectCharacteristics/premis:objectCharacteristicsExtension/Mediainfo/File/track[@type="General"]/File_size', namespaces=metsrw.utils.NAMESPACES)
+    if r:
+        size = r[0].text
+    r = pobject.xpath('premis:objectCharacteristics/premis:fixity/premis:messageDigest', namespaces=metsrw.utils.NAMESPACES)
+    if r:
+        checksum = r[0].text
+    r = pobject.xpath('premis:objectCharacteristics/premis:fixity/premis:messageDigestAlgorithm', namespaces=metsrw.utils.NAMESPACES)
+    if r:
+        checksum_type = r[0].text
```
Hwesta, line 97: The techMD will need to be replaced if the checksum_type here is different from what's configured in the dashboard.  We can discuss how to detect this, but I was thinking this script (and parse_mets_to_db) could check if the checksum_type is different from the configured one and regenerate it if so.

src/MCPClient/lib/clientScripts/archivematicaSetTransferType.py
sevein, line 44: Hmm I'm not sure, should we consider a scenario where `type` is an empty string?

src/archivematicaCommon/lib/option_parsers.py
sevein: It's working, thanks, very clever!

src/MCPClient/lib/clientScripts/verifyBAG.py
```diff
@@ -21,12 +21,19 @@
...
 
-logger = get_script_logger("archivematica.mcp.client.verifyBAG")
+logger = get_script_logger('archivematica.mcp.client.verifyBAG')
+
+PRINT_SUBPROCESS_OUTPUT = True
+BAG = '/usr/share/bagit/bin/bag'
+BAG_INFO = 'bag-info.txt'
 
+
+<<<<<<< 2c887e8c534aba024ae44c355a3a2388f9916ac1
```
Hwesta: Bad merge. :(

src/MCPClient/lib/clientScripts/restructureForCompliance.py
```diff
@@ -21,34 +21,113 @@
...
 import os
+import re
 import sys
 import shutil
 
+import django
+django.setup()
+from main.models import Transfer
+
+from verifyBAG import verifyBAG
```
Hwesta: Looks like this one also didn't rebase cleanly on #393 since I also did some bag cleanup there.

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -17,62 +18,106 @@
...
+    # TODO: is it ok to assume that the file structure is flat?
+    for entry in mets.all_files():
+        if entry.path == file_path_relative_to_sip:
+            logger.info('Archivematica AIP: file UUID of has been found in the METS document (%s).', entry.path)
+            return entry.file_uuid
+    logger.info('Archivematica AIP: file UUID has not been found in the METS document: %s', file_path_relative_to_sip)
+
+
+def main(file_uuid=None, file_path='', date='', event_uuid=None, sip_directory='', sip_uuid=None, transfer_uuid=None, use='original', update_use=True):
+    if file_uuid:
```
Hwesta, line 79: This needs to check `and file_uuid != "None"` because that's what the MCPServer passes when the file UUID is unknown.

src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
Hwesta, line 97: After later discussion, we decided not to generate checksums, but just to use the ones from the METS file, so this is good.

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -17,62 +18,106 @@
...
+    file_path_relative_to_sip = file_path_relative_to_sip.replace('%SIPDirectory%', '', 1)
+
+    # Warning! This is not the fastest way to achieve this. But we will focus
+    # on optimizations later.
+    # TODO: is it ok to assume that the file structure is flat?
+    for entry in mets.all_files():
+        if entry.path == file_path_relative_to_sip:
+            logger.info('Archivematica AIP: file UUID of has been found in the METS document (%s).', entry.path)
+            return entry.file_uuid
+    logger.info('Archivematica AIP: file UUID has not been found in the METS document: %s', file_path_relative_to_sip)
```
Hwesta, line 75: I added a `mets.get_file(path='foo')` (which does this internally, but will later be optimized)

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -17,62 +18,106 @@
...
+        logger.error('SIP exclusive-or Transfer UUID must be defined')
+        return 2
+
+    # Transfer
+    if transfer_uuid:
+        file_path_relative_to_sip = file_path.replace(sip_directory, '%transferDirectory%', 1)
+        transfer = Transfer.objects.get(uuid=transfer_uuid)
+        if transfer.type == 'Archivematica AIP':
+            file_uuid = get_file_uuid_from_mets(sip_directory, file_path_relative_to_sip)
+        if not file_uuid:
```
Hwesta, line 96: "None" also counts as no file_uuid.

src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
```diff
@@ -17,26 +17,126 @@
...
 
-    parser = OptionParser()
-    parser.add_option("-i", "--fileUUID", action="store", dest="fileUUID", default="")
-    parser.add_option("-p", "--filePath", action="store", dest="filePath", default="")
-    parser.add_option("-d", "--date", action="store", dest="date", default="")
-    parser.add_option("-u", "--eventIdentifierUUID", action="store", dest="eventIdentifierUUID", default="")
-    (opts, args) = parser.parse_args()
+    parser = argparse.ArgumentParser()
+    parser.add_argument('sharedPath')
+    parser.add_argument('-i', '--fileUUID', type=lambda x: str(uuid.UUID(x)), dest='file_uuid')
```
Hwesta, line 131: Need to import `uuid`

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
Hwesta, line 79: This and below could also be fixed by starting with `if file_uuid == "None": file_uuid = None`

Hwesta:
This will be merged as part of #391

PR #386, opened 2016-01-21
Endpoint to hide Transfers/SIPs

Hwesta:
Add an external API to allow units to be hidden.

Also, create a new 'units' app in the dashboard to store functions common to SIPs and Transfers.  Currently, details, microservices and delete/hide are there.


mistydemeo:
:+1:

PR #388, opened 2016-01-21
Move location of MCPClient archivematicaClientModules file

hakamine:
Would like to deprecate archivematica /etc/ config file info in the archivematica repository, moving those config files to the ansible role instead (and deleting the etc/ directories in the AM repo sometime in the future)
However it seems to be better for the MCPClients's archivematicaClientModules file to be kept in the AM repo. For this reason would like to move its location from etc to lib


sevein:
This is something that you probably want to port to all the branches that you currently deploy from Ansible, right? My preferred solution would be to move its contents to our codebase as a simple dictionary in Python. @mistydemeo @Hwesta thoughts?

hakamine:
The reason of wanting to remove archivematica /etc/ config file info from the archivematica repository and to put into the ansible role instead was because this seems more as deployment related data (that could vary with each deployment, as opposed to code in the AM repository that should be  mostly independent of the deployment). However there is the problem of how we do if using tools other than ansible for deployment ( deb/rpm packages, containers ). Thoughts ?

Hwesta:
I think the reason it's in /etc/ instead of /lib/ is it was considered a configuration file. However, it's not something that a user should ever be editing, and is very closely tied to the codebase, so I agree that it probably belongs in /lib/

I disagree that it should be moved to the ansible role.  What's defined in `archivematicaClientModules` is very closely related to the code, since it defines which scripts are available to gearman.  Unlike clientConfig.conf, this should be updated and overwritten on an upgrade, and will cause significant problems if it's out of sync.  The problems referenced in artefactual-labs/ansible-role-archivematica-src#45 are an example of this.  This seems like it's mostly independent of the deployment method, and closer tied to the installed code.

hakamine:
I agree, and this is the purpose of this PR. The reason of wishing to move archivematicaClientModules to lib/ is to keep it in the archivematica repository. In contrast, the clientConfig.conf file that is in etc/ contains config information such as the elasticsearch server location/MCP server location which I think is deployment dependent and can/should be configured by the ansible deployment role.

Hwesta:
Ah, that makes sense, both about archivematicaClientModules and clientConfig.conf.  Thanks for the clarification.

:+1:

PR #389, opened 2016-01-22
Advanced search: registar onclick to "add new", not entire div

mistydemeo:


mistydemeo:
The div is the width of the row, not the width of the text it contains, so attaching the onclick event to the div means that clicking on blank space has an effect.

Hwesta:
:+1:

PR #392, opened 2016-01-25
Reingest parse: Add checksumtype support

Hwesta:
Parse the checksumtype from the METS. Update tests.

Fixes reingest for the changes in  #317


mistydemeo:
:+1:

PR #393, opened 2016-01-25
Checksums: Use checksumtype setting everywhere

Hwesta:
Use configured checksumtype when creating pointer file and AIP bag. Remove from SQL. Also clean up the verify bag code and reduce duplicates.

The `-w` flag (`?w=1` on github) may be useful.


mistydemeo:
:+1:

PR #394, opened 2016-01-26
Fix dev-helper install for better sys.path handling

Hwesta:
Fix a dev-helper (not ansible) install for the changes in #358 (done in ansible in  artefactual-labs/ansible-role-archivematica-src#30 )


sevein:
:+1:

PR #380, opened 2016-01-19
Packaging & mysql seed for 1.5

Hwesta:
- Cherry pick 1.4.x SQL & packaging changes
- Generated mysql seed & update scripts for 1.5

PR #299, opened 2015-09-24
Dev/issue 8995 store after upload

mcantelon:
- add new chain/links for post-upload DIP storage
- modify AtoM and ATK upload links to continue to new chain
- add ability to select default post-upload DIP storage destination location

PR #395, opened 2016-01-26
Store DIP after upload. refs #8995

mcantelon:
Added, to the microservice chain, logic to ask the user if they'd like to store the DIP after upload.


Hwesta:
Should the `defaultNextChainLink` (where it goes in case of error) still point to the old (not start of a chain) "Move to the UploadedDIPs directory"?  Or do we want it to keep processing in case of error?

Otherwise, :+1: 

Does this need to go into qa/1.5.x and qa/1.x as well?  It will affect #380 if so.

Hwesta:
Is this a duplicate of #299 ?

mcantelon:
Closing at I re-did it for qa/1.x and made a separate PR.

PR #396, opened 2016-01-27
emailFailReport: restore --date argument

sevein:
It's not really being used but deleting it broke the report because the corresponding entry in StandardTasksConfigs use that argument. Not sure if it'd be best instead to update the task argument string?


Hwesta:
I'd probably remove the StandardTasksConfigs entry, if it's not being used.

sevein:
Ok, done. Thanks.

Hwesta:
:+1:

PR #399, opened 2016-01-27
dashboard: gearman config clean-ups

sevein:


Hwesta:
> We can't afford blocking a process for too long in the web context. Adding the
> `poll_timeout=2` parameter the client will block only for two seconds max. and
> return.

What happens if we hit the timeout and it hasn't responded yet?

sevein:
Good question, I was rushing this for another reason and didn't think about that! :sweat_smile:
`completed_job_request.complete` would equal `False` and `completed_job_request.timed_out` would equal `True`.

What do you think? Should I create a new `MCPClientTimedOut` subclass of `Exception` and raise it when it times out? Or maybe reuse `RPCError` for now?

Hwesta:
Reusing RPCError seems fine.

sevein:
I've deleted the commit that deals with the timeout issue because I can't reproduce at the moment. I'll address that if it's back in a different branch.

PR #400, opened 2016-01-28
pymysql (now mysqlclient)

sevein:
I'm all in with https://github.com/PyMySQL/PyMySQL, it's fast and also works with PyPy. However, it requires using `pymysql.install_as_MySQLdb` so Django won't break when it's trying to import `MySQdb`. That's fine, but as you'll see there is some monkey-patching work going on.

Django [prefers](https://docs.djangoproject.com/en/1.7/ref/databases/#mysql-db-api-drivers) to use https://github.com/PyMySQL/mysqlclient-python, which is a fork of the original MySQL-python connector with Python 3 support. It would not require changes in our code though other than the corresponding updates of requirements.txt files.

Thoughts? This pull request implements the pure-python version, it seems to ork but it's not complete (missing client scripts) and it has not been fully tested yet!

### DIFF
--- a/src/MCPClient/requirements/base.txt
+++ b/src/MCPClient/requirements/base.txt
@@ -4,7 +4,7 @@ django-annoying==0.7.7
 django-autoslug==1.7.1
 django-mysqlpool==0.1-9
 django-extensions==1.1.1
-MySQL-python==1.2.5
+mysqlclient==1.3.7
 gearman==2.0.2
 lxml==3.5.0
 requests==2.7.0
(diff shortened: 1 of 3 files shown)


Hwesta:
Turns out `pymysql.install_as_MySQLdb()` is the same as import `pymysql as MySQLdb` as discovered in artefactual-labs/agentarchives#7

However, if Django recommends mysqlclient-python, then I suspect they did more research than I did. As long as it supports Python3, I like it better than the original MySQLdb. Especially if it's less work to change to!

sevein:
I've been testing this and it works great, I couldn't tell the difference. Benchmarks out there say that it's slower than the original py2-only version, but I don't think it's a big deal for us.

Hwesta:
So, if we use mysqlclient, then no code changes are required and it works with Python3?

sevein:
Exactly.

Hwesta:
Then :+1:

PR #401, opened 2016-01-28
URLs: Fix Unit regex

Hwesta:
Fix transfer and ingest related URLs to prevent accidental greedy matching. All sub-url files required to match at the start of the string.

Notably this fixes `(transfer|ingest)/UUID` accidentally grabbing URLs intended for `transfer/component/UUID`


mistydemeo:
:+1:

Hwesta:
This is a fix for the changes introduced in #386

PR #402, opened 2016-01-28
Don’t autodetect METS mappings in ElasticSearch

mistydemeo:
In the past we used autodetection for most METS mapping fields, but this led to significant issues with type autodetection.

We tried to handle this in the past by special-casing certain fields which might get autodetected as dates; however, that game of whack-a-mole is going to be pretty hard to keep up. Instead, I’ve generated METS mappings using as complete a METS file as I could and included those in the archivematicaCommon tree as JSON files; when creating an index, we now use those to specify the entire format at once.

### DIFF
--- a/src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
+++ b/src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
@@ -1,0 +1,742 @@
+{
+  "properties": {
+    "ns0:mets_dict_list": {
+      "properties": {
+        "ns0:metsHdr_dict_list": {
+          "properties": {
+            "@CREATEDATE": {
+              "type": "date",
+              "format": "dateOptionalTime"
+            }
+          }
+        },
+        "ns0:amdSec_dict_list": {
+          "properties": {
+            "ns0:techMD_dict_list": {
+              "properties": {
+                "ns0:mdWrap_dict_list": {
+                  "properties": {
+                    "@MDTYPE": {
+                      "type": "string"
+                    },
+                    "ns0:xmlData_dict_list": {
+                      "properties": {
+                        "ns4:object_dict_list": {
+                          "properties": {
+                            "ns4:originalName": {
+                              "type": "string"
+                            },
+                            "@xsi:type": {
+                              "type": "string"
+                            },
+                            "ns4:objectCharacteristics_dict_list": {
+                              "properties": {
+                                "ns4:size": {
+                                  "type": "string"
+                                },
+                                "ns4:fixity_dict_list": {
+                                  "properties": {
+                                    "ns4:messageDigestAlgorithm": {
+                                      "type": "string"
+                                    },
+                                    "ns4:messageDigest": {
+                                      "type": "string"
+                                    }
+                                  }
+                                },
+                                "ns4:format_dict_list": {
+                                  "properties": {
+                                    "ns4:formatRegistry_dict_list": {
+                                      "properties": {
+                                        "ns4:formatRegistryName": {
+                                          "type": "string"
+                                        },
+                                        "ns4:formatRegistryKey": {
+                                          "type": "string"
+                                        }
+                                      }
+                                    },
+                                    "ns4:formatDesignation_dict_list": {
+                                      "properties": {
+                                        "ns4:formatVersion": {
+                                          "type": "string"
+                                        },
+                                        "ns4:formatName": {
+                                          "type": "string"
+                                        }
+                                      }
+                                    }
+                                  }
+                                },
+                                "ns4:compositionLevel": {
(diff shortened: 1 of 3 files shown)


src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
```diff
@@ -0,0 +1,739 @@
...
+                            "ns4:originalName": {
+                              "type": "string"
+                            },
+                            "@xsi:type": {
+                              "type": "string"
+                            },
+                            "ns4:objectCharacteristics_dict_list": {
+                              "properties": {
+                                "ns4:size": {
+                                  "type": "string"
```
Hwesta, line 35: Should this be a numeric type?

mistydemeo, line 35: This was autodetected by ES as a string, which suggests that non-obviously numberical fields can end up here.

src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
```diff
@@ -0,0 +1,739 @@
...
+                        },
+                        "dc:title": {
+                          "type": "string"
+                        },
+                        "dc:description": {
+                          "type": "string"
+                        },
+                        "dc:rights": {
+                          "type": "string"
+                        }
```
Hwesta, line 571: dcterms:isPartOf should also be here somewhere. AICs use it. Also, how does this handle dmdSecs that have non-DC metadata? We just don't index it?

src/archivematicaCommon/lib/elasticsearch/aipfile_mets_mapping.json
```diff
@@ -0,0 +1,474 @@
...
+                },
+                "@xsi:schemaLocation": {
+                  "type": "string"
+                },
+                "dc:title": {
+                  "type": "string"
+                },
+                "dc:description": {
+                  "type": "string"
+                }
```
Hwesta, line 32: We're only indexing a subset of DC for files?  Or should this match the aip/aips above?

src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
Hwesta, line 35: Fair, if odd.

Hwesta:
What happens if something is indexed that doesn't match this schema?  Does it raise an error, add it to the schema, omit that, something else?

mistydemeo:
> What happens if something is indexed that doesn't match this schema?

It adds it to the schema, using the type detected at the time the first data using it is indexed. That's the source of the bug that was the motivation for this: _all_ data in the METS was being autodetected this way, which led to some fields being typed as dates because the first encountered data in that field could be parsed as a date.

src/archivematicaCommon/lib/elasticsearch/aip_mets_mapping.json
mistydemeo, line 571: Curiously I did mark this as part of an AIC yet `isPartOf` didn't get indexed. I can manually add it though.

> Also, how does this handle dmdSecs that have non-DC metadata? We just don't index it?

Correct; when indexing objects you can always omit fields.

src/archivematicaCommon/lib/elasticsearch/aipfile_mets_mapping.json
mistydemeo, line 32: That appears to be the case; these were both generated after processing the same AIP!

Hwesta, line 32: Maybe the SIP-level DC from the form had all the fields, but the metadata.csv that created the file-level DC didn't have all the columns?  Do we index both SIP & file-level DC in aip/aipfile?  (Do we want to index both?)

mistydemeo, line 32: ¯_(ツ)_/¯

mistydemeo, line 32: Do we have an enumerated list of accepted fields, or do we accept anything?

Hwesta:
:+1:

src/archivematicaCommon/lib/elasticsearch/aipfile_mets_mapping.json
Hwesta, line 32: I think we accept anything, so may just specifying dc:date is enough.

mistydemeo:
Updated to add `dcterms:isPartOf`; omitted other fields for aipfile since we support anything.

Hwesta:
:+1:

PR #403, opened 2016-01-28
Don’t autodetect METS mappings (1.5 backport)

mistydemeo:
This is #402 backported to qa/1.5.x.


Hwesta:
:+1:  when #402 is :+1: :)

PR #407, opened 2016-02-04
Add API to create stub SIP objects; allow SIPs to be started from stubs

mistydemeo:
- Adds a new API to create an unstarted SIP object. A SIP is understood to be "unstarted" if its `currentpath` is null.
- Allows the SIP arrange "create SIP" API to accept a SIP UUID. If that UUID points to an unstarted SIP, uses that as the SIP object and updates it with the correct path. If it points to a started SIP, the API returns 400. If it doesn't point to a SIP that exists, then a SIP object is created with that UUID.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -331,8 +331,19 @@ def create_arranged_sip(staging_sip_path, files, sip_uuid):
     # Create SIP object
     sip_name = staging_sip_path.split('/')[1]
     sip_path = os.path.join(shared_dir, 'watchedDirectories', 'SIPCreation', 'SIPsUnderConstruction', sip_name)
+    currentpath = sip_path.replace(shared_dir, '%sharedPath%', 1) + '/'
     sip_path = helpers.pad_destination_filepath_if_it_already_exists(sip_path)
-    sip_uuid = databaseFunctions.createSIP(sip_path.replace(shared_dir, '%sharedPath%', 1)+'/', sip_uuid)
+    try:
+        sip = models.SIP.objects.get(uuid=sip_uuid)
+    except models.SIP.DoesNotExist:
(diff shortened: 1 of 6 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -331,8 +331,19 @@ def create_arranged_sip(staging_sip_path, files, sip_uuid):
...
+    # Update the already-created SIP with its path
+    if sips.count():
+        sip = sips[0]
+        if sip.currentpath is not None:
+            return "Provided SIP UUID ({}) belongs to an already-started SIP!".format(sip_uuid)
+        sip.currentpath = currentpath
+        sip.save()
+    # Create a SIP object if none exists
+    else:
+        databaseFunctions.createSIP(currentpath, sip_uuid)
```
Hwesta: This seems like it would be easier with a `.get` and try/catch? eg

```
try:
    sip = models.SIP.objects.get(uuid=sip_uuid)
    if sip.currentpath is not None:
        return "Provided SIP UUID ({}) belongs to an already-started SIP!".format(sip_uuid)
    sip.currentpath = currentpath
    sip.save()
except models.SIP.DNE:
    databaseFunctions.createSIP(currentpath, sip_uuid)
```

Hwesta:
:+1:

src/dashboard/src/components/filesystem_ajax/views.py
mistydemeo: Updated, though I moved most of the block into an `else`.

Hwesta:
:+1:

mistydemeo:
Made a few tweaks and added some new commits:
- Added access URLs which redirect to metadata and rights, creating a stub SIP if necessary.
- Access's `start_sip` will look up a proto-SIP if one exists, and pass that to the underlying view.

Hwesta:
:+1:

PR #408, opened 2016-02-10
Access: accept notes array unmodified

mistydemeo:
This allows multiple notes to be specified by callers. This depends on artefactual-labs/agentarchives#23.

Refs #9428.


Hwesta:
:+1:

PR #409, opened 2016-02-11
Use Django migrations (1.5 release)

Hwesta:
Use Django's migrations, instead of custom-written SQL for database changes. 
- Add schema migrations for all changes since 1.4.1
- Add data migrations for changes since 1.4.1
- Use dumpdata fixture for initial data
- Update code to reflect data - eg. set auto_now on lastModified columns, update Event creation to actually have unique IDs
- Includes #395/#417 as a migration
- Removes old SQL migrations
- Includes artefactual/archivematica-fpr-admin#14
- Edit: Update version & adds version update migration

Todo:
- [ ] Use data migration instead of loaddata for initial data migration
- [ ] Investigate why initial schema migration is slow
- [x] Test with upgrades
- [x] Test with artefactual-labs/ansible-role-archivematica-src#34


src/dashboard/src/requirements/base.txt
```diff
@@ -11,6 +11,7 @@ elasticsearch>=1.0.0,<2.0.0
 lazy-paged-sequence
 # Required by storage-service component
 slumber==0.6.0
+sqlparse
```
mistydemeo: What's this used by?

mistydemeo:
:raised_hands:

src/dashboard/src/requirements/base.txt
Hwesta: I think I added this when I was using RunSQL instead of RunPython for the data migrations. It can probably be removed.

src/dashboard/src/main/migrations/0010_dip_upload_store.py
```diff
@@ -0,0 +1,111 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
Hwesta, line 2: This should be checked to make sure it matches #417, which is what was actually merged.

src/dashboard/src/main/migrations/0008_fpcommandoutput.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('main', '0007_django_upgrade_tweaks'),
+    ]
+
+    operations = [
+        migrations.AlterModelTable(
+            name='fpcommandoutput',
+            table='FPCommandOutput',
+        ),
```
Hwesta: This needs to be updated to handle legacy databases.

Context: We originally had a table named FPCommandOutput.  When we added models, a FPCommandOutput model was created, but no legacy table name was created, so Django created main_fpcommandoutput.  This was later fixed (ab97a22203031bfca8e2bfa8906b18e9a62c2857) and this migration is intended to move main_fpcommandoutput to FPCommandOutput, but it fails if the legacy FPCommandOutput already exists.  We need to establish which table actually contains user data, and preserve that when we merged the two tables.

Hwesta: Updated this to revert the change pointing FPCommandOutput model at FPCommandOutput table. Modified this migration to move all data from FPCommandOutput table into main_fpcommandoutput table.

src/dashboard/src/main/migrations/0010_dip_upload_store.py
Hwesta, line 2: Checked and updated.

sevein:
:shipit:

src/dashboard/src/create_datamigration.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
```
Hwesta: This should be moved to devtools.

localDevSetup/recreateDB.sh
```diff
@@ -57,19 +57,13 @@ echo "Creating MCP database"
 mysql -u root "${dbpassword}" --execute="CREATE DATABASE ${databaseName} CHARACTER SET utf8 COLLATE utf8_unicode_ci;"
 echo "Creating ${username} user"
 mysql -u root "${dbpassword}" --execute="CREATE USER '${username}'@'localhost' IDENTIFIED BY '${password}';"
-mysql -u root "${dbpassword}" --execute="GRANT SELECT, UPDATE, INSERT, DELETE, CREATE, ALTER, INDEX ON ${databaseName}.* TO '${username}'@'localhost';"
+mysql -u root "${dbpassword}" --execute="GRANT SELECT, UPDATE, INSERT, DELETE, CREATE, ALTER, INDEX, DROP ON ${databaseName}.* TO '${username}'@'localhost';"
```
Hwesta: This is no longer necessary.

PR #414, opened 2016-02-17
Allow notes to be copied without a label

jhsimpson:
refs #9446

When inheriting notes from an Archival Object to a Digital Object, if the note does not have a label, this change allows the note to still be copied, instead of thorwing an error.

### DIFF
--- a/src/archivematicaCommon/lib/archivesspace/client.py
+++ b/src/archivematicaCommon/lib/archivesspace/client.py
@@ -506,11 +506,11 @@ class ArchivesSpaceClient(object):
                 content = [subnote["content"] for subnote in pnote["subnotes"]]
             else:
                 content = pnote["content"]
-
+        
             new_object["notes"].append({
                 "jsonmodel_type": "note_digital_object",
                 "type": dnote,
-                "label": pnote["label"],
+                "label": pnote.get("label", ""),
                 "content": content,
                 "publish": pnote["publish"],
             })


src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -506,11 +506,16 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
...
+            if "label" in pnote:
+                dlabel = pnote["label"]
+            else:
+                dlabel = ''
 
             new_object["notes"].append({
                 "jsonmodel_type": "note_digital_object",
                 "type": dnote,
-                "label": pnote["label"],
+                "label": dlabel,
```
Hwesta: This could be `"label": pnote.get("label", ""),` and omit the other chunk.

src/archivematicaCommon/lib/archivesspace/client.py
```diff
@@ -506,11 +506,11 @@ def add_digital_object(self, parent_archival_object, dashboard_uuid, title="", i
                 content = [subnote["content"] for subnote in pnote["subnotes"]]
             else:
                 content = pnote["content"]
-
+
```
Hwesta, line 509: Nitpick: remove this?

Hwesta:
:+1:

Hwesta:
This change was merged to qa/1.5 as part of some other work.

PR #416, opened 2016-02-17
MCPServer: prioritize currentpath over UUID in path

mistydemeo:
This rewrites the SIP lookup by path and alters the prioritizatation of options.

There are three possible outcomes of this route:
1. If the path name ends in a UUID, looks that up in the database, creating
   a SIP if nothing is found.
2. Looks up any SIPs with `currentpath` and uses the first located.
3. If nothing was found from the first two options, creates a new SIP.

The prioritization of options here leads to a possible bug in the case that a directory ends in a UUID, or any string which can be read as a UUID, where that UUID is not the UUID of the SIP in question. If this happens, Archivematica tries to look up a SIP using that UUID and creates one if necessary - even if the SIP already exists in the database under a different UUID.

This rewrites the lookup, prioritizing path lookups over UUID. It also brings the two paths a bit closer to each other, and merges the database query so that the path and UUID lookups happen in a simultaneous OR clause.

Fixes #9417.

### DIFF
--- a/src/MCPServer/lib/archivematicaMCP.py
+++ b/src/MCPServer/lib/archivematicaMCP.py
@@ -50,6 +50,8 @@ import django
 sys.path.append("/usr/share/archivematica/dashboard")
 django.setup()
 
+from django.db.models import Q
+
 # This project, alphabetical by import source
 import watchDirectory
 import RPCServer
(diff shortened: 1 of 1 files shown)


Hwesta:
:+1:

PR #419, opened 2016-03-09
Matching GUI: tweaks and fixes

Hwesta:
When filtering in the matching GUI, don't select checkboxes that aren't visible or are disabled. Also, fix viewing resource component and be explicit about what's sent to the template.

Update: Sorting is now case insensitive.
Update: Add sorting by date. This works best with artefactual-labs/agentarchives#29

This may also be merged into 1.5

### DIFF
--- a/src/dashboard/src/components/ingest/pair_matcher.py
+++ b/src/dashboard/src/components/ingest/pair_matcher.py
@@ -102,7 +102,18 @@ def render_resource_component(client, request, resource_component_id, query, pag
             reverse(match_redirect_target, args=[uuid, resource_component_id])
         )
     else:
-        return render(request, resource_detail_template, locals())
+        return render(request, resource_detail_template, {
+            'match_redirect_target': match_redirect_target,
+            'page': page,
+            'query': query,
+            'reset_url': reset_url,
+            'resource_component_data': resource_component_data,
+            'resource_component_id': resource_component_id,
+            'search_params': search_params,
+            'sort_by': sort_by,
+            'sort_direction': sort_direction,
+            'uuid': uuid,
+        })
 
 
 def match_dip_objects_to_resource_levels(client, request, resource_id, match_template, parent_id, parent_url, reset_url, uuid, matches=[]):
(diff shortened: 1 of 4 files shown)


src/dashboard/src/components/ingest/pair_matcher.py
```diff
@@ -102,7 +102,18 @@ def render_resource_component(client, request, resource_component_id, query, pag
...
+            'page': page,
+            'query': query,
+            'reset_url': reset_url,
+            'resource_component_data': resource_component_data,
+            'resource_component_id': resource_component_id,
+            'search_params': search_params,
+            'sort_by': sort_by,
+            'sort_direction': sort_direction,
+            'uuid': uuid,
+        })
```
sevein, line 116: Yes to this.

src/dashboard/src/templates/ingest/as/match.html
```diff
@@ -16,7 +16,12 @@
   function toggle_objects(source) {
     checkboxes = document.querySelectorAll('.atk-matcher-object-path > input');
     for (var i = 0, n = checkboxes.length; i < n; i++) {
-      checkboxes[i].checked = source.checked;
+      // Only select if the checkbox is visible and not disabled
+      if ( $(checkboxes[i]).is(":visible") && ! $(checkboxes[i]).is(":disabled") ) {
```
sevein: Instantiating jQuery objects is slow (I think that's the reason why `this` in `jQuery.each` is a DOM object and not a jQuery object). Given that this is happening inside a loop it may be worth to optimize a bit by storing a reference in a variable and use it from your two conditions of the `if`.

sevein:
```
var $cb = $(checkboxes[i]);
if ($cb.is(':visible') && !$cb.is('disabled')) {
  $cb.prop('checked', false);
}
```

Hwesta: Updated.

sevein:
:+1:

Hwesta:
Also cherry-picked into qa/1.5.x as 0a424fce09325faa3ae31cef38cdfcca93cc5a68 c1993de244420d89fe529051bf0f5f456c821ba2 8f6663ce4c08e67b417c96853dc753e14f12e60a a455c5246de3927187cd7d17c5d1e04f2ed09f0f

PR #420, opened 2016-03-15
Usage: Fix assumed block size

Hwesta:
df gives block size of 1024 by default (512 if POSIXLY_CORRECT env var set).  Explicitly set block size when getting disk size.

Django's filesizeformat filter displays this using SI prefixes, even though it calculates with 1024.

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -355,12 +355,12 @@ def _usage_check_directory_volume_size(path):
     """
     Check the size of the volume containing a given path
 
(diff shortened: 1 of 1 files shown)


sevein:
:+1:

Hwesta:
Cherry-picked to qa/1.5.x as 894896988c5b2b3f1430bebca0e5cca24d254e94 plus parent commits a5d2c8d30103ca7f398d818e38aec09c17461f01 and bcb42c988d5c22703130793b30f602fce09bf6a1

PR #423, opened 2016-03-16
dashboard: handle null sipName ES hit

sevein:


Hwesta:
Is there any valid case where sipName shouldn't be indexed? It looks like the only time `aips/aipfile` indexes sipName is from `indexAIP.py`, which should always provide the sipName.

sevein:
I saw it happening in that case in production but I couldn't figure out why.

Hwesta:
I'd rather know what's happening, but I'm okay merging it as-is, given that ES isn't a 100% reliable source of data.

:+1:

sevein:
Thank you! e6f83f7e52a2609e0dc9459b48875382c0f21ad0

PR #424, opened 2016-03-20
Add support for gunicorn in Dashboard

sevein:


src/dashboard/src/wsgi.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+from django.core.wsgi import get_wsgi_application
+
+
+application = get_wsgi_application()
```
Hwesta, line 6: This is intentionally not reusing `src/dashboard/src/apache/django.wsgi`?

src/dashboard/src/requirements/base.txt
```diff
@@ -9,6 +9,7 @@ django-autoslug==1.7.1
 django-annoying==0.7.7
 elasticsearch>=1.0.0,<2.0.0
 gearman==2.0.2
+gunicorn==19.4.5
```
Hwesta, line 12: If this is in base, remove from local.txt please.

sevein, line 12: Thank you :) - done now.

src/dashboard/src/wsgi.py
sevein, line 6: Yes, that's being ignored purposely. There might be more changes after testing (like finding a place for ES initialization when using gunicorn), but this is our first barrier.

Hwesta:
:+1:

PR #427, opened 2016-03-29
Normalize: fix preservation file UUID to match path on disk

Hwesta:
Preservation files have the taskUUID appended to the output name. Instead of generating a new UUID, use the existing task UUID as the file UUID in the database, to match the path on disk.

Access & thumbnail copies use the original file UUID.


sevein:
👍

Hwesta:
Also cherry-picked to 1.5 as 56007df671017db4785f787181866ec15d8ff3ff

PR #430, opened 2016-04-05
ArchivesSpace: Location of originals is DIP UUID (1.5)

Hwesta:
Instead of dashboard UUID, location of originals should be the AIP/DIP UUID. 

This also includes #428 and a fix from agentarchives.

PR #431, opened 2016-04-05
ArchivesSpace: location of originals is DIP UUID (1.6)

Hwesta:
Instead of dashboard UUID, location of originals should be the AIP/DIP UUID.

refs #9656  Duplicate of #430 for 1.6


sevein:
👍

PR #433, opened 2016-04-13
ArchivesSpace: optional inheritance of notes

sevein:


sevein:
Ready!

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -190,20 +191,21 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
                                   size=size,
                                   format_name=format_name,
                                   format_version=format_version,
-        )
+                                  inherit_notes=inherit_notes,)
```
Hwesta: This is more style, but I prefer to have the closing ) on its own line so that next time we add a parameter we're only touching one line.

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -217,9 +219,10 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
     parser.add_argument('--uri_prefix', metavar="uri prefix")
     parser.add_argument('--access_conditions', metavar="conditions governing access", default="")
     parser.add_argument('--use_conditions', metavar="conditions governing use", default="")
+    parser.add_argument('--inherit_notes', metavar="Inherit digital object notes from the parent component", default=False, type=lambda x: x.lower() in ('yes', 'y', 'true', '1'))
```
Hwesta: A better option here would be `action='store_true'` https://docs.python.org/2/library/argparse.html#action.  However, the dashboard config makes that difficult.  The `type` isn't for enforcing correctness, though.  `choices` is probably what you're looking for.  Also, the metavar is the stand in value for paramaters that take arguments (eg in `--foo BAR` BAR is the metavar).  The description should be in `help`

src/MCPServer/share/mysql_dev_9647_archivesspace_inherit_notes.sql
```diff
@@ -0,0 +1,5 @@
+-- Add column inherit_notes column to administration_archivesspaceconfig
+ALTER TABLE `administration_archivesspaceconfig` ADD COLUMN `inherit_notes` bool NOT NULL AFTER `repository`;
```
Hwesta: Does this need a default value?  It's bool but NOT NULL.

src/dashboard/src/components/administration/models.py
```diff
@@ -50,3 +51,4 @@ class ArchivesSpaceConfig(models.Model):
     access_conditions = models.CharField(max_length=50, blank=True, verbose_name='Conditions governing access')
     use_conditions = models.CharField(max_length=50, blank=True, verbose_name='Conditions governing use')
     repository = models.IntegerField(default=2, verbose_name='ArchivesSpace repository number')
+    inherit_notes = models.BooleanField(default=True, verbose_name='Inherit digital object notes from the parent component')
```
Hwesta: Since the inherit_notes is particular for a client, do we want to set the default to True?

sevein: Default behaviour used to be to **inherit** notes so I think we're good?

sevein: I've noticed that `syncdb` populates the column as ``inherit_notes` tinyint(1) NOT NULL`, so I've updated my migration script to update the existing row so `inherit_notes` equals `True`. I guess that under normal circumstances the according Django migration would do that for me?

src/MCPServer/share/mysql_dev_9647_archivesspace_inherit_notes.sql
sevein: I've noticed that syncdb populates the column as inherit_notes tinyint(1) NOT NULL, so I've updated my migration script to update the existing row so inherit_notes equals True. I guess that under normal circumstances the according Django migration would do that for me?

src/MCPClient/lib/clientScripts/upload-archivesspace.py
sevein: I tried what you're saying without much luck because what I get is one of the choices, which are strings, but I want a boolean right?

src/dashboard/src/components/administration/models.py
Hwesta: But those changes were for a particular client, and I think we wanted to move away from the somewhat unusual behavior by having it default to not inheriting, and the client can set it to True for their usecase.

src/MCPServer/share/mysql_dev_9647_archivesspace_inherit_notes.sql
Hwesta: syncdb (not migrate) will only create tables, not modify them.  On a fresh install, this will work correctly, but on an upgrade the new column won't be created.  SQL changes or a real migration will fix this. Since we expected migrations to make it into 1.5, and this is scheduled for 1.6, I think we can handle this when Django migrations come in.

sevein: Ok so if I understood correctly, my migration is going to add the column and we deal with upgrades for existing users later? Out of curiosity, how have you dealt with similar situations before?

sevein:
Ok, I hope I got it right this time :sweat_smile:

src/MCPClient/lib/clientScripts/upload-archivesspace.py
```diff
@@ -190,36 +191,40 @@ def upload_to_archivesspace(files, client, xlink_show, xlink_actuate, object_typ
...
+    parser.add_argument('--dip_uuid', help='DIP UUID')
+    parser.add_argument('--restrictions', help='Restrictions apply', default='premis', choices=RESTRICTIONS_CHOICES)
+    parser.add_argument('--object_type', help='object type', default='')
+    parser.add_argument('--xlink_actuate', help='XLink actuate', default='onRequest', choices=EAD_ACTUATE_CHOICES)
+    parser.add_argument('--xlink_show', help='XLink show', default='new', choices=EAD_SHOW_CHOICES)
+    parser.add_argument('--use_statement', help='USE statement')
+    parser.add_argument('--uri_prefix', help='URI prefix')
+    parser.add_argument('--access_conditions', help='Conditions governing access', default='')
+    parser.add_argument('--use_conditions', help='Conditions governing use', default='')
+    parser.add_argument('--inherit_notes', help='Inherit digital object notes from the parent component', default=False, type=str.lower, choices=INHERIT_NOTES_CHOICES)
```
Hwesta: I like this solution much better than the previous one!

Nitpick: `default=False`, but `type=str` - the `default` should be the same type as the `type`.  I'd also move the `str.lower()` to below, eg `args.inherit_notes = args.inherit_notes.lower() in INHERIT_NOTES_CHOICES` instead of having it in the type.

src/MCPServer/share/mysql_dev_9647_archivesspace_inherit_notes.sql
Hwesta: We haven't really. :sweat_smile: Previously all tables were created by SQL. We recently started adding tables created by Django, but I think this might be the first case where we had to modify one (so syncdb created something different than an upgrade produced).

Hwesta:
:+1:  Thanks for being so responsive to feedback!

PR #435, opened 2016-04-18
Rights: fix validation issue, refs #9703

sevein:
Redmine: https://projects.artefactual.com/issues/9703

Return if the data bound to the form does not validate.

This is a quick fix. I think that the whole rights app could use some clean-ups. This is trying to solve a small issue that I believe is the result of the way that things are being done in that view, but this commit addresses the issue in the fastest way possible.

### DIFF
--- a/src/dashboard/src/components/rights/views.py
+++ b/src/dashboard/src/components/rights/views.py
@@ -182,6 +182,11 @@ def rights_edit(request, uuid, id=None, section='ingest'):
         form=forms.RightsOtherRightsForm
     )
 
+    copyrightFormset = CopyrightFormSet()
+    statuteFormset = StatuteFormSet()
+    licenseFormset = LicenseFormSet()
+    otherFormset = OtherFormSet()
+
     # handle form creation/saving
     if request.method == 'POST':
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/rights/views.py
```diff
@@ -194,6 +199,9 @@ def rights_edit(request, uuid, id=None, section='ingest'):
             createdRights = form.save()
 
         copyrightFormset = CopyrightFormSet(request.POST, instance=createdRights)
+        if not copyrightFormset.is_valid():
+            return render(request, 'rights/rights_edit.html', locals())
+
```
Hwesta, line 204: Add this for the other formsets? eg statute_formset, license_formset, other_formset

sevein:
Done, please review!

sevein:
I know locals() is not a good thing, but I think that's also out of scope!

Hwesta:
:+1:

Hwesta:
Also cherry-picked to qa/1.5.x as 1f1876797f7d7320afbcbd4384da73f73aca3eb6

PR #436, opened 2016-04-19
MCPClient: fix dip_generation_helper.py logging

hakamine:


hakamine:
Found errors. Will fix and re-push

src/MCPClient/lib/clientScripts/dip_generation_helper.py
```diff
@@ -36,9 +39,11 @@ def create_archivesspace_client():
             repository=config['%repository%']
         )
     except archivesspace.AuthenticationError:
+        logging.error('Unable to authenticate to ArchivesSpace server using the default user! Check administrative settings.')
```
Hwesta: These should be `logger` not `logging`.  `logging` logs to the root logger, while `logger` logs to the define logger (below), and it's very useful for determining where a log originated and getting it filtered to the correct log files to use defined loggers.

sevein (approved):
@hakamine I've fixed the issue where the logger was defined but the you were still using the default one. I think this is ready to merge but please take a look!

hakamine:
Thank you for fixing! Looks good to me :+1:

PR #439, opened 2016-05-13
dbsettings: use database instead of db

sevein:
Note: planning to merge into `qa/1.5.x` and `qa/1.x`.

When using `read_default_file` (as in `databaseInterface.py`), the `db`
parameter seems to be ignored. It's an alternative name for `database` but its
use is not recommended (and apparently abandoned?). See
https://dev.mysql.com/doc/connector-python/en/connector-python-connectargs.html.

Related: https://github.com/artefactual-labs/ansible-role-archivematica-src/pull/53

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -25,7 +25,7 @@ config.read('/etc/archivematica/archivematicaCommon/dbsettings')
 DATABASES = {
     'default': {
         'ENGINE': 'django_mysqlpool.backends.mysqlpool',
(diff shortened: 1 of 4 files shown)


Hwesta:
:+1:

sevein:
Merged in `qa/1.5.x` and `qa/1.x`. `stable/1.4.x` has the `database` parameter [hardcoded](https://github.com/artefactual/archivematica/blob/stable/1.4.x/src/archivematicaCommon/lib/databaseInterface.py#L35).

PR #440, opened 2016-05-16
METS: clean up empty metadata dir, refs #8427

sevein:
Redmine: https://projects.artefactual.com/issues/8427


Hwesta:
:+1: 

Another option that would delete all empty directories (not just the `metadata` one) is:

``` python
for root, dirs, files in os.walk('temp', topdown=False):
     if not dirs and not files:  # Directory is empty
         os.rmdir(root)
```

`topdown=False` walks from the leaves back up, so this will delete empty subdirectories before the (possibly now empty) parent directory.

sevein:
I like your solution way better, I'll give it a try to confirm that it works.

Hwesta:
:+1:

PR #445, opened 2016-05-20
Use Format description in premis:formatName

Hwesta:
Use the Format's description in `<premis:formatName>` instead of FormatVersion's description.  The Format description is populated from PRONOM's name field, while the FormatVersion is from the version name.

Also, remove insertIntoFilesIDs because it was trivial with Django's ORM.

Redmine 8319


sevein:
👍

PR #438, opened 2016-05-12
Storage Service: Add API key support

Hwesta:
Add an API key auth to the slumber storage service config. Add SS user & api key to the SS config form.

This makes all storage service related fields in the form required - previously the URL was optional.

Adds Archivematica support for artefactual/archivematica-storage-service#124

PR #446, opened 2016-05-20
Storage Service: Add auth to redirect URLs

Hwesta:
Add new auth required by the storage service to URLs generated for redirection.  Fixes for #438

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -6,6 +6,7 @@ import requests
 from requests.auth import AuthBase
 import slumber
 import sys
+import urllib
 
 sys.path.append("/usr/share/archivematica/dashboard")
 from main.models import DashboardSetting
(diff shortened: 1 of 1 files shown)


sevein:
Your PR is 👍 but are we ok exposing the credentials to dashboard users? One solution would be to proxy the response but that would require significant changes, maybe in 1.6?

Hwesta:
The API key is already visible to dashboard users in the Administration > General > Storage Service options, so I don't think having it in the URL here is a huge problem. We could move it to the headers, though that would be somewhat harder to implement and only slightly more obscure.

sevein:
You're right, I was thinking on the dashboard as a publicly available site. Employing HTTPS should be enough?

Hwesta:
Cherry-picked to qa/1.x as 8e49b8bdaac911dac0dbfab9d814ff89342f4c1f

PR #443, opened 2016-05-19
ArchivesSpace: optional inheritance of notes

sevein:
In qa/1.x (without migration): ecb385b51fa53585d09570f50df0f3144d72cccc
It includes two migrations:
- `administration/migrations/0004_archivesspaceconfig_inherit_notes.py`:<br />This migration adds a new field (`inherit_notes`) to the `ArchivesSpaceConfig` model.
- `main/migrations/0013_upload_archivesspace_inherit_notes.py`:<br />This migration updates the task configuration.

PR #447, opened 2016-05-24
Add inherit_notes opt to ArchivesSpace client, refs #9872

sevein:
qa/1.x is using agentarchives (already updated) but #443 did not take into
account that in qa/1.5.x the ArchivesSpaceClient is part of
archivematicaCommon.

Related commit in agentarchives: https://github.com/artefactual-labs/agentarchives/pull/31/commits/42f471278b08c5e972c381270e7065055c7e1815


Hwesta:
:+1:

PR #448, opened 2016-05-25
FITS: Use models not databaseInterface

Hwesta:
FPCommandOutput doesn't exist anymore - as part of migrations (#98 and #409 ) it became `main_fpcommandoutput`. Update this to use the models instead of the outdated SQL.  This is already fixed in qa/1.x with #354


sevein:
👍

PR #450, opened 2016-05-26
AIC fixes

Hwesta:
Fixes for bugs relating to #440 and deleting empty directories.
- Add DMDID to the top-level structMap div, instead of the objects directory, since it no longer exists
- Deleting bagged files continues processing if it has an error, which happens for AICs since there's no objects directory
- Add the metadata directory explicitly to the structMap when it's not in the objects dir
- Delete empty directories anywhere in the AIP, not just the objects directory.

And other minor fixes
- Only try to decode `str` not `unicode` strings. Fixes displaying unicode in the task detail view
- Set default project settings, instead of the boilerplate

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -144,7 +144,7 @@ def escapeForCommand(string):
 # and is primarily used for arbitrary strings (e.g. filenames, paths)
 # that might not be valid unicode to begin with.
 def escape(string):
-    if isinstance(string, basestring):
+    if isinstance(string, str):
         string = string.decode('utf-8', errors='replace')
     return string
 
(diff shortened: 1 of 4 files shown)


src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -144,7 +144,7 @@ def escapeForCommand(string):
 # and is primarily used for arbitrary strings (e.g. filenames, paths)
 # that might not be valid unicode to begin with.
 def escape(string):
-    if isinstance(string, basestring):
+    if isinstance(string, str):
```
sevein, line 147: Should we use `django.utils.six.binary_type` instead?

sevein:
👍

src/archivematicaCommon/lib/archivematicaFunctions.py
Hwesta, line 147: Neat, I didn't realize Django shipped with a copy of `six`!  We've tried to keep from requiring Django in archivematicaCommon however, and when we do the Py3 conversion we'll probably just use `six` as a dependency instead of using Django's.

PR #452, opened 2016-05-30
Reingest: Do not use formatversion.description to try and match a premis:formatRegistryKey entry

jhsimpson:
refs #9926

Do not use formatversion.description to try and match a premis:formatRegistryKey entry.

See the notes on the redmine ticket.  https://projects.artefactual.com/issues/9926

### DIFF
--- a/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
+++ b/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
@@ -62,11 +62,6 @@ def parse_files(root):
                 key = amdsec.findtext('.//premis:formatRegistryKey', namespaces=ns.NSMAP)
                 print('FPR key', key)
                 format_version = fpr_models.IDRule.active.get(command_output=key).format
-            # If not, look for formatName
-            if not format_version:
-                name = amdsec.findtext('.//premis:formatName', namespaces=ns.NSMAP)
-                print('Format name', name)
-                format_version = fpr_models.FormatVersion.active.get(description=name)
         except fpr_models.FormatVersion.DoesNotExist:
             pass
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/parse_mets_to_db.py
```diff
@@ -66,7 +66,10 @@ def parse_files(root):
             if not format_version:
                 name = amdsec.findtext('.//premis:formatName', namespaces=ns.NSMAP)
                 print('Format name', name)
-                format_version = fpr_models.FormatVersion.active.get(description=name)
+                try:
+                    format_version = fpr_models.FormatVersion.active.get(description=name)
+                except:
```
sevein: I think that it'd be best to explicitly catch`fpr_models.FormatVersion.MultipleObjectsReturned` to avoid ignoring other possible exceptions like `fpr_models.FormatVersion.DoestNotExist` or others?

jhsimpson: I thought about doing that, but I don't think we want to ignore other possible exceptions.  In fact, I am starting to think that it would be better to remove the entire if statement:

https://github.com/artefactual/archivematica/blob/52c1bbdd538e92fda8879cfc5a8a4cb0a10f7a77/src/MCPClient/lib/clientScripts/parse_mets_to_db.py#L66-L72

The purpose of this code is to make sure that the premis object for each file , in the original METS file, gets copied into the database.  I think trying to guess the match from the premis object in the METS, to the fpr_formatversion, based on matching the premis name to the format version description, is not guaranteed to be accurate, and so we probably should not even try.

sevein:
:+1:

PR #453, opened 2016-05-30
Dashboard: strip SS API key field, refs #9838

sevein:
Not entirely sure if this is the most idiomatic way to achieve it. One
limitation I've found is that the HTML document that Django returns after the
form is posted by the browser does not contain the stripped value but the
original, however the value stored is the cleaned one. I'm not sure if the
problem is in the way that the view is built or somewhere else.

### DIFF
--- a/src/dashboard/src/components/administration/forms.py
+++ b/src/dashboard/src/components/administration/forms.py
@@ -60,6 +60,17 @@ class SettingsForm(forms.Form):
 
 
 class StorageSettingsForm(SettingsForm):
+
+    class StripCharField(forms.CharField):
+        """
+        Strip the value of leading and trailing whitespace.
+
+        This won't be needed in Django 1.9, see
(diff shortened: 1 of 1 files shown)


jhsimpson:
deploying for testing now

jhsimpson:
this is working, with the caveat you mention, where the form does not contain the cleaned version after saving.  Reloading the page shows the correct value.

I think this is sufficient to solve the bug, for the 1.5.0 release.

:+1:

src/dashboard/src/components/administration/forms.py
```diff
@@ -60,6 +60,17 @@ def save(self, *args, **kwargs):
 
 
 class StorageSettingsForm(SettingsForm):
+
+    class StripCharField(forms.CharField):
```
Hwesta, line 64: I think using the `clean_storage_service_apikey` would be better here. https://docs.djangoproject.com/en/1.8/ref/forms/validation/#cleaning-a-specific-field-attribute

PR #455, opened 2016-06-02
Archival storage: fix AIP pagination

sevein:
Redmine: https://projects.artefactual.com/issues/9030

**NOTE**: this is a back-port of #322 (qa/1.5.x 01b7e65, qa/1.x 38dad3d) for `stable/1.4.x`. There was a small code conflict but it was solved and after testing this seems to work properly.

---

This reworks archival storage's search view in order to fix paging of
AIP searches.

AIP searches are very broken in the current iteration. For historical
reasons, searches are performed against the aips/aipfile index, which
is limited to returning only AIP-related fields. This is done because
aips/aipfile contains relevant information which is not present in the
aips/aip index, which contains actual AIP data.

The issue arises when paging is applied. Because these are file results,
there are many files to one AIP; these are incorrectly paged as file
results. This means that each page, instead of containing ten AIPs as
intended, contains ten files' worth of AIPs - which could be a single
AIP, if the first returned AIP contains more than one file.

In addition to fetching metadata from aips/aipfile, the view later uses
the fetched UUIDs to fetch records from aips/aip - in order to have
enough information to seed the AIC form, and some other additional
metadata that aips/aipfile is missing. It currently does this by
performing one query for every fetched UUID, which is fairly wasteful.

This patch simplifies the current workflow to the following:
- Run the query against aips/aipfile, with and an aggregation which
  returns all unique UUIDs in the requested set and the count of files
  within that UUID. This query is set to return no results - only the
  aggregation is used.
- Uses that set of UUIDs to fetch pages of AIP data from aips/aip in a
  single query, which feeds the template.

Because the paging is now based on AIPs, not files within an AIP, a
page in archival storage can display the appropriate number of AIPs.
In addition, because AIPs are fetched in pages and the total number of
queries performed has been reduced substantially, performance is
substantially better.

Fixes #9030.
Fixes #9085.

### DIFF
--- a/src/dashboard/src/components/archival_storage/views.py
+++ b/src/dashboard/src/components/archival_storage/views.py
@@ -106,73 +106,78 @@ def search(request):
         # pulling only one field (we don't need field data as we augment
         # the results using separate queries)
         if not file_mode:
+            # Fetch all unique AIP UUIDs in the returned set of files
+            query['aggs'] = {'aip_uuids': {'terms': {'field': 'AIPUUID', 'size': 0}}}
+            # Don't return results, just the aggregation
+            query['size'] = 0
             # Searching for AIPs still actually searches type 'aipfile', and
             # returns the UUID of the AIP the files are a part of.  To search
             # for an attribute of an AIP, the aipfile must index that
             # information about their AIP in
             # elasticSearchFunctions.index_mets_file_metadata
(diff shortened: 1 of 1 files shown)


jhsimpson:
This change has been tested in qa/1.5.x and qa/1.x branches already.

:+1:  to merging in stable/1.4.x

PR #459, opened 2016-06-03
Use UTC in MCPClient and MCPServer

sevein:
Also, use standard `Date.prototype.toLocaleString()` in the dashboard.


jhsimpson:
:+1:

PR #366, opened 2015-12-16
MoMA specific changes

Hwesta:
Changes from the MoMA private repo, squashed into a handful of commits. This is likely to be broken into several PRs as each feature gets merged. 

Changes:
- Add DRMC README
- copyToNetX client script, for after uploading to AtoM/DRMC
- New config page for NetX related config
- Normalization job displays red if any normalization fails
- Overload the accession number as the TMS object ID and use it as the AtoM/DRMC slug
- Add LDAP support

TODO:
- [x] Check if FPR changes are already in QA - yes, commit dropped
- [x] Test DRMC upload & copy to NetX script
- [ ] Make displaying the NetX config conditional (like DSpace transfer type) - not needed with #517 ?
- [x] Allow passing in a DRMC slug properly, instead of overloading the accession number field - uses new access system identifier field
- [ ] Allow the accession # field to have a custom label - not needed with access system ID?
- [x] Strip private information from LDAP implementation and merge to qa
- [x] Make DIP upload to AtoM/Binder/NetX more generic - new chain for upload to DRMC that includes NetX
- [x] Convert SQL migration to Django migration (from #463)

PR #463, opened 2016-06-13
MoMA specific changes (1.5)

Hwesta:
Copy of #366 for 1.5 release branch.

Changes from the MoMA private repo, squashed into a handful of commits. This is likely to be broken into several PRs as each feature gets merged. Note this does not (yet) include the LDAP auth changes.

Changes:
- Add DRMC README
- copyToNetX client script, for after uploading to AtoM/DRMC
- New config page for NetX related config
- Normalization job displays red if any normalization fails
- Overload the accession number as the TMS object ID and use it as the AtoM/DRMC slug

TODO:
- [x] Convert SQL migration to Django migration
- [ ] Test DRMC upload & copy to NetX script
- [ ] Make displaying the NetX config conditional (like DSpace transfer type)
- [ ] Allow passing in a DRMC slug properly, instead of overloading the accession numebr field
- [ ] Strip private information from LDAP implementation and merge to mainline


Hwesta:
See #366 for current work on this feature. This branch is now available as qa/1.5.x-moma

PR #462, opened 2016-06-10
Bring 1.5 changes into qa

Hwesta:
This will merge `stable/1.5.x` into `qa/1.x`, and bring all the changes that only went into 1.5 into mainline. Notably, Django migrations instead of raw SQL migrations (#409). 

Includes #460 #459 #456 #454 #453 #452 #451 #450 #448 #447 #446 #445 #443 #442 #441 #439 #430 #412 #403 #373 #308 

For code review, only the two most recent commits should need it: merge conflict resolution, and adding Django migrations for 1.6-only SQL migrations.

PR #464, opened 2016-06-21
Added agentType select field to admin settings.

jrwdunham:
Bug #9523 in RedMine.


Hwesta:
Looks good for now. I think this should wait until #462 brings in Django migrations, and then update this with a migration.

jrwdunham:
@Hwesta how's this?

Hwesta (approved):
Looks good.

PR #465, opened 2016-06-27
Made copyright status a controlled vocabulary.

jrwdunham:
Given 4.1.3.1 copyrightStatus of PREMIS v. 3.0, I used "copyrighted", "publicdomain", and "unknown" as the underlying values with their human-readable counterparts displayed in the <select>.


Hwesta:
:+1: Should wait on #462 though, and will need a DB migration afterwards.

jrwdunham:
Added migration.

Hwesta (approved):
Looks good. :+1:

PR #467, opened 2016-06-30
MCPServer: create reingestAIP directory as part of package install

scollazo:
This fixes the problem with AIP reingest

### DIFF
--- a/src/MCPServer/debian/postinst
+++ b/src/MCPServer/debian/postinst
@@ -13,6 +13,8 @@ else
   rsync -a /usr/share/archivematica/sharedDirectoryStructure/* /var/archivematica/sharedDirectory/.
 fi
 
+mkdir -p /var/archivematica/sharedDirectory/watchedDirectories/system/reingestAIP/
+
 chown -R archivematica:archivematica /var/archivematica/
 
 /usr/share/archivematica/postinstSharedWithDev
--- a/src/MCPServer/debian/postinst
+++ b/src/MCPServer/debian/postinst
@@ -5,15 +5,9 @@
 
 dbc_go archivematica-mcp-server $@
 
-if [ -d /var/archivematica/sharedDirectory ]; then
-  echo "/var/archivematica/sharedDirectory exists"
-else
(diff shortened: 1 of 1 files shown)


Hwesta:
:+1: 

Since this problem should only happen if the shared directory structure already existed, that could probably be moved to inside the if block above.

hakamine:
Do you see any problem if we just do the rsync of the sharedDirectory structure each time there is an upgrade (i.e., remove the "if/then/else" conditionals in the block above ) ? Currently the rsync is not done when the /var/archivematica/sharedDirectory exists, so a similar problem is going to happen when the sharedDirectory structure changes again in the future.

scollazo:
Yes, that won't hurt. But we need to be sure that the sharedDirectory/system/reingest folder is created before the rsync happens, I wasn't able to find it in the git repo.

hakamine:
the reingestAIP folder is inside the structure here  ( watchedDirectories/system/reingestAIP/ ): https://github.com/artefactual/archivematica/tree/stable/1.5.x/src/MCPServer/share/sharedDirectoryStructure

jhsimpson:
:+1:  this looks good to me, and I think it can be merged.

We have analogous code for building rpm's in the am-packbuild repo here:
https://github.com/artefactual-labs/am-packbuild/blob/master/rpm/archivematica/package.spec

I think there needs to be a code change in am-packbuild, to make sure the behaviour when building debs is the same when building rpms, i.e., make sure the sharedDirectory structure is copied over on upgrades not just on installs. 

In a future release of Archivematica, we can move these debian directories from the Archivematica git repo and put them in am-packbuild.

scollazo:
Merged in qa/1.x

scollazo:
Merged into stable/1.5.x too, closing.

Hwesta:
Cherry-picked to qa/1.x as 5ee579c330c601fa88b3927ad078b0411e4c1ea3 and stable/1.5.x as 80c5f2b3c7cd64e608654075a68705f101893563

PR #391, opened 2016-01-25
Reingest: Add full reingest, use metsrw

Hwesta:
Add full reingest (from transfer) and update to use metsrw instead of parsing/updating METS manually.
- Use metsrw for all reingest METS changes
- Delete SIP-level DublinCore from the METS if the form was deleted
- Add ability to start full reingest, which runs from Transfer
- In full reingest, parse METS for file information (file UUID, size, checksum, format)
- Add all Events, not just reingestion Events
- Update & version the techMD if the object information has changed
- Create new & remove deprecated preservation derivatives
- Parse multiple rightsGranted in PREMIS Rights
- Parse an external METS file (which may be from an AIP) for additional metadata
- Cleanup in createMETS2 - split several functions into smaller, more modular functions.
- Delete a RightsStatement from the DB if there's no associated RightsGranted (and it doesn't appear in the GUI)

Depends on storage service branch dev/issue-9211-full-aip-reingest artefactual/archivematica-storage-service#100 and metsrw branch dev/reingest artefactual-labs/mets-reader-writer#22

PR #468, opened 2016-07-05
Metadata-only dip upload, refs #9752

sevein:
Redmine: https://projects.artefactual.com/issues/9752


src/archivematicaCommon/requirements/base.txt
```diff
@@ -4,4 +4,4 @@ elasticsearch>=1.0.0,<2.0.0
 requests==2.7.0
 python-dateutil==2.4.2
 django-mysqlpool
-git+git://github.com/artefactual-labs/agentarchives.git
+-e git+git://github.com/sevein/agentarchives.git
```
Hwesta: I assume this will be removed once artefactual-labs/agentarchives#32 is merged?

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -16,16 +16,50 @@
...
+from components.archival_storage.atom import get_atom_client
 
 
 class CreateAICForm(forms.Form):
     results = forms.CharField(label=None, required=True, widget=forms.widgets.HiddenInput())
 
+
+class UploadMetadataOnlyAtomForm(forms.Form):
+    slug = forms.CharField(label=_('Insert slug'), required=True)
+    slug.widget.attrs['class'] = 'span8'
```
Hwesta: This should be done with `widget=forms.TextInput(attrs={'class': 'span8'}))`

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -16,16 +16,50 @@
...
+        (OBJECTS, _('Metadata and objects'))
     )
-    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES,  widget=forms.RadioSelect, required=True)
+    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES, widget=forms.RadioSelect, required=True)
+
+
+class DeleteAIPForm(forms.Form):
+    uuid = forms.CharField(label=_('Please type in the UUID of the repository to confirm'), required=True)
+    uuid.widget.attrs['class'] = 'xxlarge'
+    uuid.widget.attrs['placeholder'] = 'UUID'
```
Hwesta: As above, these should be set in the field constructor using the widgets attribute, or in the `__init__`of the form.

src/dashboard/src/templates/archival_storage/list.html
```diff
@@ -91,15 +103,9 @@
   {% else %}
 
     <div class="alert-message">
-      <p>Archival storage is empty.</p>
+      <p>{% trans "Archival storage is empty." }</p>
```
Hwesta: Missing closing %}

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -306,53 +302,12 @@ def create_aic(request, *args, **kwargs):
...
-        reason_for_deletion = request.POST.get('reason_for_deletion', '')
 
-        response = storage_service.request_file_deletion(
-           uuid,
-           request.user.id,
-           request.user.email,
-           reason_for_deletion
-        )
+def delete_aip(request, uuid):
+    raise Http404
```
Hwesta: What is the tombstone on `delete_aip` for? The view function was renamed and the URL is changed, so the only places I'd expect to call it are from new code.  Same question for `reingest_aip` though the URL & view are unchanged.

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -16,16 +16,50 @@
...
-        (OBJECTS, 'Re-ingest metadata and objects')
+        (METADATA_ONLY, _('Metadata only')),
+        (OBJECTS, _('Metadata and objects'))
     )
-    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES,  widget=forms.RadioSelect, required=True)
+    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES, widget=forms.RadioSelect, required=True)
+
+
+class DeleteAIPForm(forms.Form):
+    uuid = forms.CharField(label=_('Please type in the UUID of the repository to confirm'), required=True)
```
Hwesta: What is this field looking for? It doesn't validate the input against anything, and appears to discard it unused.

src/archivematicaCommon/requirements/base.txt
sevein: Yes!

sevein:
Thanks, Holly. I've added a new commit with the fixes (is that easier?), but I'll squash it later.

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -16,16 +16,57 @@
...
-        (OBJECTS, 'Re-ingest metadata and objects')
+        (METADATA_ONLY, _('Metadata only')),
+        (OBJECTS, _('Metadata and objects'))
     )
-    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES,  widget=forms.RadioSelect, required=True)
+    reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES, widget=forms.RadioSelect, required=True)
+
+
+class DeleteAIPForm(forms.Form):
+    uuid = forms.CharField(label=_('Please type in the UUID to confirm'), required=True, widget=forms.TextInput(attrs={'class': 'xxlarge', 'placeholder': _('UUID')}))
```
Hwesta: Clarify what UUID you're looking for? Eg "Please enter the AIP's UUID to confirm"  Also, what's the motivation for having this? Were people accidentally deleting AIPs?

sevein: There was a confirmation page before. This may not be a good idea after all, I can get rid of this change.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -613,8 +574,78 @@ def document_json_response(document_id_modified, type):
...
+                messages.success(request, message)
+            return redirect('archival_storage_index')
+
+    # Process delete form
+    if request.POST and 'submit-delete-form' in request.POST:
+        form_delete = forms.DeleteAIPForm(request.POST, prefix='delete', uuid=uuid)
+        active_tab = 'delete'
+        if form_delete.is_valid():
+            response = storage_service.request_file_deletion(uuid, request.user.id, request.user.email, form_delete.cleaned_data['reason'])
+            messages.info(request, response['message'])
```
Hwesta, line 634: Elasticsearch deletion seems to have gotten lost.  Original aip_delete included `es_client = elasticSearchFunctions.get_client(); elasticSearchFunctions.mark_aip_deletion_requested(es_client, uuid)` (see above)

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -613,8 +574,78 @@ def document_json_response(document_id_modified, type):
...
+            return redirect('archival_storage_index')
+
+    context = {
```
Hwesta, line 639: :+1:  on being explicit about context passed to template.

src/dashboard/src/templates/archival_storage/list.html
```diff
@@ -1,5 +1,6 @@
 {% extends "layout_fluid.html" %}
 {% load breadcrumb %}
+{% load i18n %}
```
Hwesta: Can the i18n changes be in their own commit please? They're conceptually unrelated to the metadata DIP upload.

src/dashboard/src/templates/archival_storage/view.html
```diff
@@ -0,0 +1,182 @@
...
+        <form action="" method="POST" class="form-stacked">
+          {% csrf_token %}
+          <fieldset>
+            <legend>{% trans "Re-ingest" %}</legend>
+            <div class="clearfix">
+              <label>{% trans "Choose type of re-ingest:" %}</label>
+              <div class="input">
+                {{ forms.reingest.reingest_type.errors }}
+                <ul class="inputs-list">
+                  {% for choice in forms.reingest.reingest_type %}
```
Hwesta: Any reason the forms aren't done with Django's built-in form display, or Archivematica's form display (`{% include "_form.html" with form=forms.reingest %}`) ? It would be more flexible with new changes, such as full reingest's extra fields.

src/dashboard/src/templates/archival_storage/search.html
```diff
@@ -41,68 +48,51 @@
...
+          <th>{% trans "Date stored" %}</th>
+          <th>{% trans "Status" %}</th>
+          <th>{% trans "Actions" %}</th>
+        </thead>
+        <tbody>
+        {% for term_usage in results %}
+          <tr>
+            <td>
+              <a href="{% url 'view_aip' term_usage.uuid %}">{{ term_usage.name }}</a>
+              <a href='/archival-storage/search/json/aip/{{ term_usage.document_id_no_hyphens }}/' target='_blank'>({% trans "view raw" %})</a>
```
Hwesta: I like this - can you change the link to the AIP in files view to go to view_aip instead of download?

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,104 @@
...
+
+from main.models import DashboardSetting
+
+
+logger = logging.getLogger('archivematica.dashboard')
+
+
+def get_atom_client():
+    url = DashboardSetting.objects.get(name='dip_upload_atom_url').value
+    key = DashboardSetting.objects.get(name='dip_upload_atom_key').value
```
Hwesta: Are these always populated? If not, `helpers.get_setting` might be better.

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,104 @@
...
+
+def upload_dip_metadata_to_atom(aip_name, aip_uuid, parent_slug):
+    """
+    Write to a AtoM's resource (parent_slug) the metadata of the objects of a
+    AIP given its name and UUID. Return the slug of the new container resource
+    created to hold the metadata objects.
+    """
+    try:
+        # Download METS file
+        _, temp_path = tempfile.mkstemp()
```
Hwesta: [TemporaryFile](https://docs.python.org/2/library/tempfile.html#tempfile.TemporaryFile) can be used with a `with` statement, which will automatically close itself so you don't have to do the try/except/finally.  This also means the exception handling can be around the smallest part of the code likely to generate the exception.

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,104 @@
...
+            if res is not None and res.text:
+                dict_[name] = res.text
+                logger.debug('Extracted property %s from METS: %s', name, res.text)
+            logger.debug('Failed to extract property {} from METS: not found', name)
+
+        # Add objects
+        for item in mw.all_files():
+            if item.type == 'Directory' or item.use != 'original':
+                continue
+            attrs = {
```
Hwesta: Can the items in `attrs` have a key but a value of None? If so, `add_prop_from_xml` can be refactored to something like `attrs['key'] = amdsec.findtext(elem, 'query')`

src/dashboard/src/components/archival_storage/views.py
sevein, line 634: Fixed.

src/dashboard/src/templates/archival_storage/list.html
sevein: Hahaha sure

src/dashboard/src/templates/archival_storage/view.html
sevein: Nice one, thanks.

src/dashboard/src/templates/archival_storage/search.html
sevein: Ok, done.

sevein:
Thanks Holly, please take a look again!

src/dashboard/src/components/archival_storage/forms.py
```diff
@@ -17,15 +17,55 @@
...
+
+
 class ReingestAIPForm(forms.Form):
     METADATA_ONLY = 'metadata'
     OBJECTS = 'objects'
     REINGEST_CHOICES = (
-        (METADATA_ONLY, 'Re-ingest metadata only'),
-        (OBJECTS, 'Re-ingest metadata and objects')
+        (METADATA_ONLY, 'Metadata only'),
+        (OBJECTS, 'Metadata and objects')
```
sromkey, line 52: Since both this and full AIP re-ingest are targetted for 1.6 this should have all three re-ingest options available.

src/dashboard/src/components/archival_storage/forms.py
Hwesta: Ah, ok.  Maybe run it by the archivists to see what they think?

src/dashboard/src/components/archival_storage/forms.py
Hwesta, line 52: I assume the full reingest option would come in with #384 / #391 with the rest of the feature.  If those get merged first, then this will have to be updated.

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,104 @@
...
+                add_prop_from_xml(attrs, 'format_registry_name', amdsec, './/{info:lc/xmlns/premis-v2}objectCharacteristics/{info:lc/xmlns/premis-v2}format/{info:lc/xmlns/premis-v2}formatRegistry/{info:lc/xmlns/premis-v2}formatRegistryName')
+                add_prop_from_xml(attrs, 'format_registry_key', amdsec, './/{info:lc/xmlns/premis-v2}objectCharacteristics/{info:lc/xmlns/premis-v2}format/{info:lc/xmlns/premis-v2}formatRegistry/{info:lc/xmlns/premis-v2}formatRegistryKey')
+                title = os.path.basename(item.path)
+                slug = client.add_child(parent_slug=file_slug, title=title, level='Item')
+                logger.info('Creating child with slug %s and title %s', slug, title)
+                client.add_digital_object(slug, **attrs)
+
+            return file_slug
+
+    except (AtomError, MetsError, SlumberHttpBaseException):
```
Hwesta: Can you move the exception handling to be around the smallest piece likely to generate it? It makes it clearer what the expected problem is, and means that unexpected errors don't get accidentally caught.

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,104 @@
...
+                if res is not None and res.text:
+                    dict_[name] = res.text
+                    logger.debug('Extracted property %s from METS: %s', name, res.text)
+                logger.debug('Failed to extract property %s from METS: not found', name)
+
+            # Add objects
+            for item in mw.all_files():
+                if item.type == 'Directory' or item.use != 'original':
+                    continue
+                attrs = {
```
Hwesta: Can the items in `attrs` have a key with a value of None? If so, `add_prop_from_xml` can be refactored to something like `attrs['key'] = amdsec.findtext(elem, 'query')` and probably doesn't need to be its own function.

Hwesta:
Looks better! Just a couple of small comments.

src/dashboard/src/components/archival_storage/atom.py
sevein: I can do that, but I want to omit the property in JSON, not to give it the null value.  I can get rid of `add_prop_from_xml` but I'll need a custom subclass of `json.JSONEncoder` to omit items with`None` values, what do you prefer?

Hwesta: The way it is is probably better. I was wondering if the properties could be given a default value of None, but it sounds like that's not workable.

sevein:
Please review again!

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import tempfile
+
+from agentarchives.atom.client import AtomClient, AtomError, CommunicationError
+from metsrw import METSDocument
+from storageService import extract_file
+from slumber.exceptions import SlumberHttpBaseException
+
+from components import helpers
+from main.models import DashboardSetting
```
Hwesta: Unused

Hwesta:
:+1: But please wait to merge until after #462 comes in

PR #470, opened 2016-07-13
Endpoint to hide Transfers/SIPs (1.5)

Hwesta:
Backport of #386 

Add an external API to allow units to be hidden.

Also, create a new 'units' app in the dashboard to store functions common to SIPs and Transfers. Currently, details, microservices and delete/hide are there.

Works with artefactual/automation-tools#10


src/dashboard/src/templates/transfer/detail.html
```diff
@@ -18,14 +18,14 @@
         <div class="alert-message block-message">
           <p>There are micro-services awaiting user decisions.</p>
           <div class="alert-actions">
-            <a class="btn small" href="{% url 'components.ingest.views.ingest_microservices' uuid %}">View micro-services</a>
```
jrwdunham, line 21: It looks like this also fixes a bug where a transfer detail used to incorrectly link back to an ingest's list of micro-services. Is that right?

jrwdunham:
Looks good to me.

src/dashboard/src/templates/transfer/detail.html
Hwesta, line 21: Yup, good catch.  Though, the views are almost identical except for the breadcrumbs.

PR #471, opened 2016-07-13
API: Encode unicode paths

Hwesta:
When dealing with filesystem paths coming from the API, ensure they're strs.  Otherwise, it generates Unicode exceptions.

There are likely other cases with this problem; these are the ones I found.

refs #9941

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -31,6 +31,7 @@ from annoying.functions import get_object_or_None
 from tastypie.authentication import ApiKeyAuthentication
 
 # This project, alphabetical
+import archivematicaFunctions
 from contrib.mcp.client import MCPClient
 from components.filesystem_ajax import views as filesystem_ajax_views
 from components.unit import views as unit_views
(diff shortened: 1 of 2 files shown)


sevein:
Sorry I missed this! 👍

jrwdunham:
👍

See https://github.com/artefactual/archivematica/pull/481 for a related PR.

src/dashboard/src/media/js/transfer/component_form.js
```diff
@@ -264,6 +272,7 @@ var TransferComponentFormView = Backbone.View.extend({
             'accessionNumber': $('#transfer-accession-number').val(),
             'sourcePaths':     paths
           };
+          console.log(transferData);
```
Hwesta: Looks like debugging output that shouldn't be committed. :)

Hwesta:
@jrwdunham These are definitely good fixes!  Encoding support is something that's a constant work in progress.  However, I think these should go into qa/1.x , not a point release.  Can you make a new branch and PR with just your commits against qa/1.x, and we'll move the conversation over there?

Hwesta:
So it doesn't get forgotten, https://github.com/artefactual/archivematica/compare/qa/1.x...dev/issue-9234-unicode-transfers may be useful

Hwesta:
Cherry-picked to stable/1.5.x as b8f4a550d649ad7bfc2fc78710f8b98af851d6b9

PR #473, opened 2016-07-14
Reingest: Clear files from ES index

Hwesta:
Delete AIP files from ES index when storing. Since ES is indexed from the METS file, all files will be re-indexed even if they aren't on disk (eg a metadata-only reingest)

Cherry-picked from #391 for 1.5


sevein:
👍

PR #476, opened 2016-07-21
Normalization report, refs #10146

sevein:
Redmine: https://projects.artefactual.com/issues/10146

---

Normalization report sent by email to active users.

Screenshot

### DIFF
--- a/src/MCPClient/lib/clientScripts/normalizeReport.py
+++ b/src/MCPClient/lib/clientScripts/normalizeReport.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python2
+
+from __future__ import print_function
+
+from argparse import ArgumentParser
+import logging
+import os.path
+from smtplib import SMTPException
+import sys
+
+import django
+from django.contrib.auth.models import User
+from django.core.mail import send_mail
+from django.template import Context, Template
+
+from main.models import File, Job, Report, SIP, Task
+import components.helpers as helpers
+
+from custom_handlers import get_script_logger
+
+
+django.setup()
+
+logger = get_script_logger('archivematica.mcp.client.normalizeReport')
+
+# Based on http://leemunroe.github.io/responsive-html-email-template/email.html
+EMAIL_TEMPLATE = """
+    <!doctype html>
+    <html>
+        <head>
+            <meta name="viewport" content="width=device-width">
+            <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+            <title>Normalization failure report - failures detected!</title>
+        </head>
+        <body>
+            <table cellpadding="10" width="100%">
+                <tr>
+                    <td></td>
+                    <td>
+                        <h1>Archivematica</h1>
+                        <p>
+                            Pipeline: {{ pipeline_uuid }}<br />
+                            SIP: <strong>{{ name }}</strong> ({{ uuid }})
+                        </p>
+                        <hr />
+                        <h3>Normalization failure report - failures detected!</h3>
+                        <table cellpadding="10" border="1" width="100%">
+                            <thead>
+                                <tr>
+                                    <th>File name</th>
+                                    <th>File UUID</th>
+                                    <th>Exit code</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                {% for type, tasks in failed_tasks.items %}
+                                    <tr>
+                                        <td colspan="3" style="background-color: #aaa; font-weight: bold;">
+                                            {{ type }} (total = {{ tasks.count }})
+                                        </td>
+                                    </tr>
+                                    {% for item in tasks %}
+                                        <tr>
+                                            <td>{{ item.location }}</td>
+                                            <td>{{ item.fileuuid }}</td>
+                                            <td>{{ item.exitcode }}</td>
+                                        </tr>
+                                    {% endfor %}
+                                {% endfor %}
+                            </tbody>
+                        </table>
+                    </td>
+                    <td></td>
+                </tr>
+            </table>
+            <table class="footer-wrap">
+                <tr>
+                    <td></td>
+                    <td class="container">
+                        <div class="content">
+                            <table>
+                                <tr>
+                                    <td align="center">
+                                        <p>Archivematica</p>
+                                    </td>
+                                </tr>
+                            </table>
+                        </div>
+                    </td>
+                    <td></td>
+                </tr>
+            </table>
+        </body>
+    </html>
+"""
+
+
+def report(uuid):
+    """
+    Generate normalization report using Django's template module and send it to
+    every active user.
+    """
+    recipient_list = User.objects.filter(is_active=True).values_list('email', flat=True).exclude(email__in=['demo@example.com', ''])
+    if not recipient_list:
+        logger.info('Normalization report is not being sent because the recipient list is empty.')
+        return 0
+
+    logger.info('Sending report to %s.', ', '.join(recipient_list))
+
+    try:
+        sip = SIP.objects.get(uuid=uuid)
+    except SIP.DoesNotExist:
+        logger.error('SIP with UUID %s not found.', uuid)
+        return 1
+
+    failed_tasks = {}
+    for jobtype in ('Normalize for preservation', 'Normalize for access'):
+        try:
+            job = Job.objects.filter(sipuuid=uuid, jobtype=jobtype).order_by('-createdtime')[0]
+        except IndexError:
+            logger.info('No normalization failures have been detected in type "%s"', jobtype)
+            continue
+        tasks = Task.objects.filter(job=job).exclude(exitcode=0)
+        if not tasks.exists():
+            logger.info('No normalization failures have been detected in type "%s"', jobtype)
+            continue
+        failed_tasks[jobtype] = tasks.values('filename', 'fileuuid', 'exitcode')
+        for item in failed_tasks[jobtype]:
+            try:
+                item['location'] = File.objects.get(uuid=item['fileuuid']).currentlocation.replace('%SIPDirectory%', '')
+            except File.DoesNotExist:
+                pass
+
+    if not len(failed_tasks):
+        logger.info('Normalization report is not being sent because no failures have been detected.')
+        return 0
+
+    ctxdict = {
+        'uuid': uuid,
+        'name': os.path.basename(sip.currentpath.rstrip('/')).replace('-' + sip.uuid, ''),
+        'pipeline_uuid': helpers.get_setting('dashboard_uuid'),
+        'failed_tasks': failed_tasks
+    }
+
+    logger.info('Building HTML message')
+    ctx = Context(ctxdict)
+    tmpl = Template(EMAIL_TEMPLATE)
+    html_message = tmpl.render(ctx)
+
+    logger.info('Storing report in database')
+    Report.objects.create(content=html_message, unittype='SIP', unitname=ctxdict['name'], unitidentifier=ctxdict['uuid'])
+
+    try:
+        logger.info('Sending email...')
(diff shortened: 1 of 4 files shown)


sevein:
@Hwesta, what do you think? Am I on the right path?

src/dashboard/src/main/migrations/0022_normalization_report.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Introduce new normalization report microservice
+    StandardTaskConfig.objects.create(id=normalization_report_stc, requires_output_lock=False, execute='normalizeReport_v0.0', arguments='''--uuid "%SIPUUID%"''')
+    TaskConfig.objects.create(id=normalization_report_tc, tasktype_id=one_instance_task_type, tasktypepkreference=normalization_report_stc, description='Normalization report')
+    MicroServiceChainLink.objects.create(id=normalization_report_mscl, currenttask_id=normalization_report_tc, defaultnextchainlink_id=move_to_approve_normalization_mscl, microservicegroup='Normalize', defaultexitmessage='Failed')
+    MicroServiceChainLinkExitCode.objects.create(id='cbf1b16e-5260-4548-ba80-eeeaf5e54cc2', microservicechainlink_id=normalization_report_mscl, exitcode=0, nextmicroservicechainlink_id=move_to_approve_normalization_mscl, exitmessage='Completed successfully')
+
+    # Introduce between "Remove files without linking information" and "Move to approve normalization"
+    MicroServiceChainLinkExitCode.objects.filter(microservicechainlink_id='dba3028d-2029-4a87-9992-f6335d890528').update(nextmicroservicechainlink_id=normalization_report_mscl)
```
Hwesta: Don't forget to update the `MicroServiceChainLink.defaultnextchainlink` if it points to something other than then failed SIP chain.

Hwesta:
Looks good! Sending the report as an email attachment is an interesting choice - why not as the body of an HTML email?  Did the client want the report as a table or as paragraphs?

sevein:
It's not an attachment but I put the contents in a file to take the screenshot, I have to send a real email and see if it works.
I'll convert it into a table, I was rushing the layout, trying to get a proof of concept working - wasn't sure if using Django's template was going to be possible!

Hwesta:
Ah, makes sense. Using Django's templating is a good idea. Depending on the size, it might be cleaner to store the template as a separate file instead of inside the script itself, but either could work.

src/dashboard/src/main/migrations/0022_normalization_report.py
sevein: I've checked, thanks.

sevein:
@Hwesta, I think it's ready!

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+                </tr>
+            </table>
+        </body>
+    </html>
+"""
+
+
+def query_normalization_results(cursor, uuid):
+    """
+    Execute SQL query on given cursor. Based on https://git.io/vK1Km.
```
Hwesta: A path to the source (eg `src/dashboard/src/components/ingest/views_NormalizationReport.py`) or even a github link would be clearer an less likely to suffer from bitrot.

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+                OR (CASE WHEN b.exitCode <= 255 AND a.fileID IS NOT NULL THEN TRUE ELSE FALSE END AND CASE WHEN a.fileID IS NOT NULL AND b.exitCode > 0 THEN TRUE ELSE FALSE END)
+            );
+    """
+
+    logger.info('Extracting normalization results from the database.')
+    logger.debug('Raw SQL query: %s.', sql)
+
+    start = time.time()
+    cursor.execute(sql, (uuid, uuid))
+    logger.info('Time spent in SQL query: %0.2f seconds.', (time.time() - start) * 1000.0)
```
Hwesta: Nitpick: is this info or debug level?

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Generate normalization report using Django's template module and send it to
+    every active user.
+    """
+    recipient_list = User.objects.filter(is_active=True).values_list('email', flat=True).exclude(email__in=['demo@example.com', ''])
+    if not recipient_list:
+        logger.error('Normalization report is not being sent because the recipient list is empty.')
+        return 0
+
+    logger.debug('Sending report to %s.', ', '.join(recipient_list))
```
Hwesta: Nitpick: Seems like info level to know who's getting the email?

src/MCPClient/lib/clientScripts/normalizeReport.py
sevein: git.io is owned by GitHub.

src/MCPClient/lib/clientScripts/normalizeReport.py
sevein: It's a very slow query so I'm voluntarily using info to increase its visibility.

src/MCPClient/lib/clientScripts/normalizeReport.py
sevein: Ok! I'll fix it.

src/MCPClient/lib/clientScripts/normalizeReport.py
Hwesta: Ah, that makes sense.

src/MCPClient/lib/clientScripts/normalizeReport.py
Hwesta: I didn't realize that! Still, I think a path to the source would be clearer what you're referring to

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+        logger.info('Total count of normalization failures: %d.', failed_count)
+
+        ctxdict = {
+            'failed_count': failed_count,
+            'failures': []
+        }
+
+        # Expected columns (order matters!)
+        sql_cols = [
+             'sip_name',
```
Hwesta: Nitpick, but these aren't indented to multiple of 4.

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+                                <tr>
+                                    <th>File</th>
+                                    <th>Type</th>
+                                    <th>Exit code</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                {% for item in failures %}
+                                    <tr>
+                                        <td><abbr title="{{ item.file_uuid }}">{{ item.location }}</abbr></td>
```
Hwesta: Since it's often convenient to copy-paste the UUID, maybe move this out of the title into another column?

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,248 @@
...
+                        <h3>
+                            {{ name }}<br />
+                            Normalization failure report: {{ failed_count }} failures found.
+                        </h3>
+                        <table cellpadding="10" border="1" width="100%">
+                            <thead>
+                                <tr>
+                                    <th>File</th>
+                                    <th>Type</th>
+                                    <th>Exit code</th>
```
Hwesta: When there's both an access and preservation error, this only displays the header for one

Hwesta:
A couple of issues: when there's both an access and preservation derivative, the header row is only displayed for one.  Also, when normalization is re-run, it emails about failures from the previous iterations, even if those don't appear in the current run.

sevein:
> Also, when normalization is re-run, it emails about failures from the previous iterations, even if those don't appear in the current run.

That's tricky. I've tried but without much luck, I need to ask @jhsimpson for some help!

sevein:
Ready!

sevein:
I've moved `django.setup()` to make flake8-E402 happy.

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python2
+
+from __future__ import print_function
+
+from argparse import ArgumentParser
+import logging
+from smtplib import SMTPException
+from os.path import basename
```
Hwesta: Two nitpicks: I try to keep the import list sorted (ie os before smtplib), and I usually `import os.path` so when I use `os.path.basename` it's clear where it comes from, and not a locally-defined basename.

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def report(uuid):
+    """
+    Generate normalization report using Django's template module and send it to
+    every active user.
+    """
+    recipient_list = User.objects.filter(is_active=True).values_list('email', flat=True).exclude(email__in=['demo@example.com', ''])
+    if not recipient_list:
+        logger.error('Normalization report is not being sent because the recipient list is empty.')
```
Hwesta: Is this an error case, or an expected do-nothing case? I lean towards the latter.

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,174 @@
...
+        return 1
+
+    failed_tasks = {}
+    for jobtype in ('Normalize for preservation', 'Normalize for access'):
+        try:
+            job = Job.objects.filter(sipuuid=uuid, jobtype=jobtype).order_by('-createdtime')[0]
+        except IndexError:
+            logger.info('No normalization failures have been detected in type "%s"', jobtype)
+            continue
+        tasks = Task.objects.filter(Q(job=job) & ~Q(exitcode=0))
```
Hwesta: I think this is the same as `Task.objects.filter(job=job).exclude(exitcode=0)` since `exclude` is filter-not

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,174 @@
...
+        if not tasks.count():
```
Hwesta: I think you want `tasks.exists()`

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python2
+
+from __future__ import print_function
```
Hwesta, line 3: :+1: on python3 compatibility (even if you don't use print in this function)

Hwesta:
:+1: looks good!

src/MCPClient/lib/clientScripts/normalizeReport.py
```diff
@@ -119,19 +120,24 @@ def report(uuid):
...
+        failed_tasks[jobtype] = tasks.values('filename', 'fileuuid', 'exitcode')
+        for item in failed_tasks[jobtype]:
+            try:
+                item['location'] = File.objects.get(uuid=item['fileuuid']).currentlocation.replace('%SIPDirectory%', '')
+            except File.DoesNotExist:
+                pass
 
     if not len(failed_tasks):
         logger.info('Normalization report is not being sent because no failures have been detected.')
-        return 0
```
Hwesta: Did you mean to delete this? If there's no failed tasks we probably want an early abort.

sevein:
Rebased, migration updated and tested again.

PR #477, opened 2016-07-21
Metadata-only DIP upload: add help text, refs #9752

sevein:


sevein:
👍  consensus after long technical discussion in the chatroom.

PR #478, opened 2016-07-22
Index Islandora identifier

Hwesta:
Add the Islandora identifier to the list of identifiers specifically indexed.

This usually happens when using the FEDORA Space in the storage service, which accepts a METS POSTed to it (usually from Islandora).  The METS is parsed and the files listed are fetched, then sent to Archivematica.  The METS originally posted has an OBJID that is the Islandora identifier, that needs to be indexed.

Also, added tests for the identifiers section of code.  Fixtures can be trimmed to a more minimal example if requested.

This will also need to be committed to qa/1.x

### DIFF
--- a/src/MCPClient/lib/clientScripts/indexAIP.py
+++ b/src/MCPClient/lib/clientScripts/indexAIP.py
@@ -6,8 +6,6 @@ from glob import glob
 import os
 import sys
 
-import django
-django.setup()
 # dashboard
 from main.models import UnitVariable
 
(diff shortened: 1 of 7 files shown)


sevein:
:+1:

Hwesta:
Merged to qa/1.x as 77b48dfca9c135d3b1e567953163ea040231601a c47dfc45f3b0f9f9cba564373d7f55a3454fa59f 7facea4bf83639d4710cc3c1069170616e124bcf

PR #482, opened 2016-07-27
Fix backlog search tab for transfers

Hwesta:
When querying for transfers, we usually actually want the transfer containing a file that matches, because most of the metadata is stored in the transferfile index. To do this, query transferfile but only return the transfer UUIDs, then query transfer.

Also, only return fields used by the GUI, and minor cleanup.

refs redmine #9174

### DIFF
--- a/src/dashboard/src/components/backlog/views.py
+++ b/src/dashboard/src/components/backlog/views.py
@@ -145,6 +145,22 @@ def search(request):
             doc_type = 'transferfile'
             source = 'filename,sipuuid,relative_path'
         else:  # Transfer mode
+            # Query to transfers/transferfile, but only fetch & aggregrate transfer UUIDs
+            # Based on transfer UUIDs, query to transfers/transfer
+            query['aggs'] = {'transfer_uuid': {'terms': {'field': 'sipuuid'}}}
+            hits = es_client.search(
+                index='transfers',
+                doc_type='transferfile',
+                body=query,
+                size=0,  # Don't return results, only aggregation
+            )
+            uuids = [x['key'] for x in hits['aggregations']['transfer_uuid']['buckets']]
+
+            query['query'] = {
+                'terms': {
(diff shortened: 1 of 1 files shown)


sevein:
Wouldn't make more sense to extend the metadata stored in `transferfile` documents? I think one principle in Elasticsearch is: a single document should contain all of the information that is required to decide whether it matches a search request. I think that denormalization at index time is the way to go. You also have nested objects and parent/child relationships.

Hwesta:
Search is something that needs more attention generally, but that's beyond the scope of this fix.

Adding more information to `transfer` would make more sense than more information to `transferfile` - the problem was `transfer` had different columns & less information that `transferfile`.

We want to be able to search for transfers in backlog, or files in backlog.  Searches for files in backlog worked, because we were querying the columns matching `transferfile`. Searches for transfers failed because `transfers` columns had different names than what was being provided (set up to search `transferfile`) or the metadata was only indexed in `transferfile`.  This fix mimics the behavior in archival storage, by querying `transferfile`, returning a list of matching transfer UUIDs, and sending the transfer information to the GUI.

Can you clarify what you mean by "I think that denormalization at index time is the way to go. You also have nested objects and parent/child relationships."?

sevein:
> Can you clarify what you mean by "I think that denormalization at index time is the way to go. You also have nested objects and parent/child relationships."?

I meant that we should optimize read performance by adding all that redundant data to our documents when we are indexing. It's ok if the process becomes slower, it could be done asynchronously later on - if it's indexed from Django. I mentioned nested objects or parent/child relationships (one-to-many) as alternatives offered by Elasticsearch to create relationships between entities that perform well and don't require application-side joins.

Hwesta:
Ah, okay.  Those are good ideas, and you should bring them up again when we take a closer look at search generally.

Otherwise, does this fix look OK?

sevein:
Yes 👍

PR #483, opened 2016-07-29
Dev/rpm packaging

scollazo:
Update requirements for all packages, based on the ones from qa/1.x

### DIFF
--- a/src/MCPClient/requirements/base.txt
+++ b/src/MCPClient/requirements/base.txt
@@ -1,0 +1,11 @@
+bagit==1.5.2
+Django>=1.7,<1.8
+django-annoying==0.7.7
+django-autoslug==1.7.1
+django-mysqlpool==0.1-9
+django-extensions==1.1.1
+mysqlclient==1.3.7
+gearman==2.0.2
+lxml==3.5.0
+requests==2.7.0
+unidecode==0.04.19
--- a/src/MCPServer/requirements/base.txt
+++ b/src/MCPServer/requirements/base.txt
@@ -4,3 +4,4 @@ django-extensions==1.1.1
 mysqlclient==1.3.7
 gearman==2.0.2
 lxml==3.5.0
+pyinotify
--- a/src/MCPServer/requirements/base.txt
+++ b/src/MCPServer/requirements/base.txt
@@ -1,0 +1,6 @@
+Django>=1.7,<1.8
+django-mysqlpool==0.1-9
+django-extensions==1.1.1
+mysqlclient==1.3.7
+gearman==2.0.2
+lxml==3.5.0
--- a/src/dashboard/src/requirements/base.txt
+++ b/src/dashboard/src/requirements/base.txt
@@ -8,12 +8,17 @@ django-extensions==1.1.1
 django-autoslug==1.7.1
 django-annoying==0.7.7
 elasticsearch>=1.0.0,<2.0.0
+gearman==2.0.2
 gunicorn==19.4.5
 lazy-paged-sequence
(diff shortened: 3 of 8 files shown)


src/MCPServer/requirements/base.txt
```diff
@@ -0,0 +1,7 @@
+Django>=1.7,<1.8
+django-mysqlpool==0.1-9
+django-extensions==1.1.1
+mysqlclient==1.3.7
+gearman==2.0.2
+lxml==3.5.0
+pyinotify
```
sevein, line 7: pyinotify is not used anymore I believe!

src/dashboard/src/requirements/base.txt
```diff
@@ -8,12 +8,17 @@ django-extensions==1.1.1
 django-autoslug==1.7.1
 django-annoying==0.7.7
 elasticsearch>=1.0.0,<2.0.0
+gearman==2.0.2
 gunicorn==19.4.5
 lazy-paged-sequence
+git+git://github.com/artefactual-labs/agentarchives.git
```
jhsimpson: A question about this , I don't think this dependency is required for Archivematica 1.5.x.  It will be required for 1.6.0, but for this PR, merging into stable/1.5.x I think this line should be removed.

When doing the 1.6.0 release, I think we will want to pull a specific version of agentarchives from pypi, rather than link to a branch in github.

src/MCPClient/requirements/base.txt
```diff
@@ -0,0 +1,11 @@
+bagit==1.5.2
+Django>=1.7,<1.8
+django-annoying==0.7.7
+django-autoslug==1.7.1
+django-mysqlpool==0.1-9
+django-extensions==1.1.1
+mysqlclient==1.3.7
```
jhsimpson, line 7: When adding this requirement to the MCPClient , shouldn't mysql-client also be added to the debian/control file?  If MCPClient and the dashboard are both installed on the same machine, then it won't be a problem, but I think it would be good to add the os dependency in the same PR as the pip requirement.

src/MCPServer/requirements/base.txt
jhsimpson, line 7: I don't think you are correct, @sevein .  In stable/1.5.x pyinotify is still used:

https://github.com/artefactual/archivematica/blob/stable/1.5.x/src/MCPServer/lib/archivematicaMCP.py#L40

It is not used in qa/1.x, but this PR is for stable/1.5.x, so pyinotify is still needed.

jhsimpson:
:+1:  this looks good to me now, for stable/1.5.x

scollazo:
Rebased into one commit and merged into stable/1.5.x as 1df87a0

Hwesta:
Squashed to stable/1.5.x as 1df87a0f33bf5a3586f44edaf78e7ad1a67261c5

src/MCPClient/requirements/base.txt
sevein, line 7: `mysqlclient` is a fork of the `MySQL-python` library, not the actual MySQL command-line tool, if that's what you meant. I don't really think that the Ubuntu package `mysql-client` is a dependency whereas `rsync` is a dependency because it's used by one of the client scripts. Thoughts? It won't hurt to have the MySQL command-line tool installed, but I don't think it's strictly necessary.

src/MCPServer/requirements/base.txt
sevein, line 7: You are right! I made a change in qa/1.x to stop using it. It's funny though that we are declaring a dependency even though that specific code path is never executed. Should we fix that or would that not be a backward-compatible fix?

PR #484, opened 2016-07-30
clientScripts: use python provided by environment

scollazo:
In the rpm packages the reingest aip feature doesn't work, due to the use of a different virtualenv for each service. This script was using the python provided by the system, instead of the one from the virtualenv


jhsimpson:
:+1: 

This looks good to me.

Hwesta:
:+1:

PR #485, opened 2016-08-02
Models: File.currentlocation & File.originallocation are longblobs

Hwesta:
Pre-migrations, File paths were stored as blobs so they correctly matched paths on disk in the MCPServer. However, TextFields are represented as longtext, which broke that behaviour. Reverting to longblob in the database.

This may not be the approach we take in 1.6.

### DIFF
--- a/src/dashboard/src/main/migrations/0018_blob_fields.py
+++ b/src/dashboard/src/main/migrations/0018_blob_fields.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


jrwdunham:
👍 

I tested this with a set of client files (that Justin emailed me) with UTF-8 encoded filenames and there were no issues. Confirmed that the "Original name" values in the resulting METS file match the original file names (when the character entities were decoded to Unicode).

Hwesta:
Also committed to qa/1.x as f363d6880e95128ab5e95ae72a40593367dcb496

PR #486, opened 2016-08-03
moved removeUnneededFiles config to clientConfig.conf and added .DS_Store

axfelix:
This should work, but I can't actually find where in the Archivematica codebase this script would be called from normally, and the file_uuid references make it seem like directories will never actually get passed to this script, only regular files.

### DIFF
--- a/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
+++ b/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import shutil
 
 # databaseFunctions requires Django to be set up
 import django
@@ -9,14 +10,24 @@ django.setup()
 # archivematicaCommon
 from custom_handlers import get_script_logger
 from databaseFunctions import fileWasRemoved
+import ConfigParser
 
-REMOVEABLE_FILES = ["Thumbs.db", "Icon", u"Icon\u000D"]
+clientConfigFilePath = '/etc/archivematica/MCPClient/clientConfig.conf'
+config = ConfigParser.SafeConfigParser()
+config.read(clientConfigFilePath)
+try:
+    removableFiles = [e.strip() for e in config.get('MCPClient', 'removableFiles').split(',')]
+except:
+    removableFiles = ["Thumbs.db", "Icon", u"Icon\u000D", ".DS_Store"]
 
 def remove_file(target_file, file_uuid):
     basename = os.path.basename(target_file)
-    if basename in REMOVEABLE_FILES:
+    if basename in removableFiles:
         print "Removing {filename} (UUID: {uuid})".format(uuid=file_uuid, filename=basename)
-        os.remove(target_file)
+        try:
+            os.remove(target_file)
+        except OSError:
+            shutil.rmtree(target_file)
         # Gearman passes parameters as strings, so None (NoneType) becomes
         # "None" (string)
         if file_uuid and file_uuid != "None":
--- a/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
+++ b/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
 
+import ConfigParser
 import os
 import sys
 import shutil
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,21 +2,32 @@
...
 # databaseFunctions requires Django to be set up
 import django
 django.setup()
 # archivematicaCommon
 from custom_handlers import get_script_logger
 from databaseFunctions import fileWasRemoved
+import ConfigParser
 
-REMOVEABLE_FILES = ["Thumbs.db", "Icon", u"Icon\u000D"]
+clientConfigFilePath = '/etc/archivematica/MCPClient/clientConfig.conf'
```
Hwesta: Since this is being done through reading a config file, instead of as a constant, can you move it inside the remove_file function?

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,21 +2,32 @@
...
+config = ConfigParser.SafeConfigParser()
+config.read(clientConfigFilePath)
+try:
+    removableFiles = [e.strip() for e in config.get('MCPClient', 'removableFiles').split(',')]
+except:
```
Hwesta: What exceptions are you expecting here?  `ConfigParser.NoOptionError` ? Please catch only the exceptions you're expecting.

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,21 +2,32 @@
...
 import sys
+import shutil
 
 # databaseFunctions requires Django to be set up
 import django
 django.setup()
 # archivematicaCommon
 from custom_handlers import get_script_logger
 from databaseFunctions import fileWasRemoved
+import ConfigParser
```
Hwesta: Nitpick: Since this is part of python's standard library, can you move it to the line above `import os`?

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
axfelix: Sure -- I actually copied the syntax from https://github.com/artefactual/archivematica/blob/ae0c0f1c9a55cd19f32105179d7908cb02939cfd/src/dashboard/src/components/helpers.py and https://github.com/artefactual/archivematica/blob/ae0c0f1c9a55cd19f32105179d7908cb02939cfd/src/archivematicaCommon/lib/elasticSearchFunctions.py since that seemed to be how you're handling this config file elsewhere :)

Hwesta:
Thanks for the contribution!  I checked it locally, and `.DS_Store` files were deleted.

This script, like other clientScripts, is run through gearman, which is controlled by our workflow engine.  You're correct - this script is run once per file in the Transfer, and the file is deleted if it matches the removable files list.

If you want more details about the MCPServer and our workflow engine work, it's documented here https://wiki.archivematica.org/MCP though it's pretty dense.  Information on scripts running once per file is here https://wiki.archivematica.org/MCP/TaskTypes#Run_for_each_file though it will make more sense if you read the MCP page first.

You can make the fixes I suggested and push your update to this branch, or I can make them as a second commit when I merge this. Let me know which you prefer.

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
Hwesta: That's reasonable! We're incrementally trying to improve the codebase, and this is part of it.  https://github.com/artefactual/archivematica/blob/stable/1.5.x/src/MCPClient/lib/clientScripts/indexAIP.py#L37-L41 has a better model for catching the exception.

axfelix:
last commit should handle the fixes. Thanks for getting to this so quickly!

Hwesta:
The tweaks all look good!

Can you also fill out a [Contributor's Agreement](https://wiki.archivematica.org/images/2/25/Contributor_agreement.txt) and send it to agreement@artefactual.com?  More information on why we require this is [in our (in progress) contribution guidelines](https://github.com/artefactual/archivematica/blob/55a61616a92242e6db0e56ba57d4fcb600fab57b/CONTRIBUTING.md#contributors-agreement) or [on the wiki](https://wiki.archivematica.org/Contributor_Agreement)

We're also discussing if we should change the defaults to include `.DS_Store`, and are planning on soliciting community feedback on the discussion list.  I'll hold off merging this until that's decided.

sromkey:
Based on feedback in the user forum, people seem pretty keen to have DS_Store files in the list of files to remove, and since this makes it more easily configurable anyway, I suggest we merge the PR as is, with removing DS_Store by default. Thanks again Alex!

Hwesta:
Cherry-picked and squashed as a3d8af8915e217eebeecc20a02462a7843e71aa4 to qa/1.x

Thanks for the improvement!

PR #487, opened 2016-08-03
Python3: modernize python2 code (futurize --stage1)

Hwesta:
Run `futurize --stage1 --nobackups --write` on the code. This performs only changes that don't require any external dependencies, mostly converting print to a function.

Notably, all the scripts in archivematicaCommon are treated as independent modules, and cannot do relative imports from each other.  The same is true for MCPClient scripts, but less common. See http://stackoverflow.com/a/14664354/2475775 for details

To support this, some additional cleanup:
- Removed unused modules (eg template tags)
- Removed unneeded print statements
- Moved get_file_checksum to archivematicaFunctions instead of its own module

We could add `from __future__ import absolute_import, division, print_function` to all files.  That would prevent future problems with imports & print, but I thought it might be too much noise.

The scope of this PR is `futurize --stage1` and removing code that would otherwise generate a change from futurize.


src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -800,8 +800,8 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
                 label = "mets.xml-%s" % (GROUPID)
                 dspace_dmdsecs = createDSpaceDMDSec(label, itemdirectoryPath, directoryPathSTR)
                 if dspace_dmdsecs:
-                    dmdSecs.extend(dspace_dmdsecs.values())
-                    ids = ' '.join(dspace_dmdsecs.keys())
+                    dmdSecs.extend(list(dspace_dmdsecs.values()))
+                    ids = ' '.join(list(dspace_dmdsecs.keys()))
```
jrwdunham: Is it really necessary to convert these dict views to lists? In Python 3.4 you can extend a list with the output of .values() and also join that dict view on a string.

src/MCPClient/lib/clientScripts/jsonMetadataToCSV.py
```diff
@@ -88,7 +88,7 @@ def object_to_row(row, headers):
     """
     def sort_row(keyvalue):
         return headers.index(keyvalue[0])
-    row = sorted(row.items(), key=sort_row)
+    row = sorted(list(row.items()), key=sort_row)
```
jrwdunham: Is it necessary to call `list` here? `sorted` can handle any iterable as its first argument, can't it?

src/MCPServer/lib/linkTaskManagerDirectories.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, jobChainLink, pk, unit):
         # Apply unit (SIP/Transfer) replacement values
         commandReplacementDic = unit.getReplacementDic(directory)
         # Escape all values for shell
-        for key, value in commandReplacementDic.items():
+        for key, value in list(commandReplacementDic.items()):
```
jrwdunham: Why the call to `list` here? And also in the next file? If we're just looping through a dict.items(), it doesn't seem to be necessary to call `list` in either Py2 or Py3. I see that `commandReplacementDic` is a `ReplacementDict` instance, but I don't think that changes how `.items` behaves.

jrwdunham:
I've looked through all of the files changed in this PR and nothing looks surprising to me. (Just the minor questions about `list` above.) I'll pull these changes into one of my dev branches and play around with it.

src/MCPServer/lib/linkTaskManagerDirectories.py
Hwesta: The fixer automatically converts from `d.items()` to `list(d.items())`, since in py2 it's explicitly a list and the fixer is trying to preserve that behaviour. I reverted the ones I caught that didn't need that fix, thanks for catching the ones I missed.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
Hwesta: I thought they'd fail on an iterable, but it looks like that's fine.

jrwdunham:
:+1:

PR #488, opened 2016-08-04
Upgrade to modern Bootstrap and jQuery (UPDATED)

jraddaoui:
This branch updates several of our vendored libraries, and the places that use them. Major changes:
## Upgrades the vendored bootstrap from 1.x to 3.3.7

This also updates many of the templates in order to ensure compatibility. A few bits of the app now look subtly different, but for the most part things work as they did before.
## Upgrades the vendored jQuery to the latest release, along with the other vendored libraries we use.
# Known issues
- [x] File browsers do not render inside modals. Moved them outside keeping a similar behaviour until the new transfer browser is included.
- [x] Arrage edit metadata and upload DIP to AtoM modals broken
- [x] Installer not centred
- [x] Select fields not styled
- [x] Preservation planning unstyled
- [x] Admin/failures appear below sidebar
- [x] Jobs width not properly fluid
- [x] Messages don't display
- [x] SIP arrange drag & drop broken
- [x] Headers in the SIP tables (transfer/ingest tabs) are positioned strangely
- [x] "Connected" icon in the corner isn't vertically centred properly
- [x] Inline forms not styled properly
- [x] Navbar active tab over effect doesn't look right
- [x] User dropdown background color is changed and it shows a blue line when active
- [x] Some margin needed after forms and buttons in the ingest and backlog tabs
- [x] Review AIP and download DIP pages not centered
- [x] Breadcrumb slashs duplicated
- [x] AIP view page not properly styled
- [x] No margins in SIP/Transfer add metadata form
- [x] SIP/Transfer add rights form not working and not properly styled
- [x] User form actions missing margins


Hwesta:
Changes look great! :+1: If you're okay with the few commits I made, then this is ready to merge.

jraddaoui:
Merged in https://github.com/artefactual/archivematica/commit/89023d98da2cda8110c0cb1bc7a7b7f0401e6a27

PR #489, opened 2016-08-08
Normalization report, refs #10146 (1.5)

sevein:
Back-port of #476 for `stable/1.5.x`.


Hwesta:
:+1:  Migrations look good, everything else looks the same.

PR #490, opened 2016-08-10
requirements.txt: Make names & contents more consistent

Hwesta:
For every project, have a base.txt, test.txt and dev.txt. base.txt contains dependencies required to run the project, test.txt has the testing dependencies, and dev.txt has anything that makes development easier. dev.txt includes test.txt which includes base.txt.

This should also make it easier to update the ansible role to install dev.txt or test.txt when not in production.

I'm not sure if it makes sense to separate put the test.txt dependencies in base.txt (since you should always be able to run tests), or perhaps dev.txt in test.txt (since there's few dev-only non-testing dependencies).

### DIFF
--- a/src/MCPClient/requirements/dev.txt
+++ b/src/MCPClient/requirements/dev.txt
@@ -1,0 +1,1 @@
+-r test.txt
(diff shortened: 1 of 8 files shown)


jhsimpson:
good idea, looks good to me.

:+1:

PR #494, opened 2016-08-19
Modify removeUnneededFiles and removeCacheFiles to only run once per …

dnarc:
…submission rather than once per file.

### DIFF
--- a/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
+++ b/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
@@ -2,7 +2,6 @@
 
 import os
 import sys
-import ntpath
 
 # databaseFunctions requires Django to be set up
 import django
@@ -18,16 +17,15 @@ def remove_file(sip_obj_dir, sip_UUID):
 
     ### 1. Remove unneeded files from package.
     for dir_path, dir_name, filenames in os.walk(sip_obj_dir):
-        for remove_name in REMOVEABLE_FILES:
-            for filename in filenames:
-                if remove_name == filename:
-                    del_file = dir_path + '/' + filename
-                    os.remove(del_file)
-                    logger.debug('removeUnneededFiles file deleted: %s', del_file)
+        for filename in filenames:
+            if filename in REMOVEABLE_FILES:
+                del_file = os.path.join(dir_path, filename)
+                os.remove(del_file)
+                logger.debug('removeUnneededFiles file deleted: %s', del_file)
 
     ### 2. Insert removal notification into events database
     sip_files = File.objects.filter(sip=sip_UUID).values_list('uuid', 'currentlocation')
-    remove_files = [file_UUID for file_UUID, current_location in sip_files if (ntpath.basename(current_location) in REMOVEABLE_FILES and file_UUID != 'None')]
+    remove_files = [file_UUID for file_UUID, current_location in sip_files if (os.path.basename(current_location) in REMOVEABLE_FILES and file_UUID != 'None')]
     logger.debug('removeUnneededFiles db results :  %s', remove_files)
     for file_UUID in remove_files:
         fileWasRemoved(file_UUID)
--- a/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
+++ b/src/MCPClient/lib/clientScripts/removeUnneededFiles.py
@@ -2,6 +2,7 @@
 
 import os
 import sys
+import ntpath
 
 # databaseFunctions requires Django to be set up
 import django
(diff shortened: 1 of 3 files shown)


dnarc:
Testing on submissions with large number of files (>500) I see speed improvements of 2% (multimedia files) to 8% (text files) on AIP/DIP generation runs with this change to run the removeUnneededFiles micro-service only once rather than on every single file. 

removeUnneededFiles is itself modified to work correctly when only called once per submission.

There are at least two more micro-services (assignFileUUID and updateSizeAndChecksum) that I believe can be changed in the same way and that will yield similar speed improvements.

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,30 +2,41 @@
 
 import os
 import sys
+import ntpath
```
Hwesta: We try to avoid using platform-specific libraries. This can be done using [`os.path`](https://docs.python.org/2/library/os.path.html#os.path.basename)

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,30 +2,41 @@
...
-        # "None" (string)
-        if file_uuid and file_uuid != "None":
-            fileWasRemoved(file_uuid)
+def remove_file(sip_obj_dir, sip_UUID):
+
+    ### 1. Remove unneeded files from package.
+    for dir_path, dir_name, filenames in os.walk(sip_obj_dir):
+        for remove_name in REMOVEABLE_FILES:
+            for filename in filenames:
+                if remove_name == filename:
```
Hwesta: The previous 3 lines can be done more efficiently and elegantly as:

```
for filename in filenames:
    if filename in REMOVEABLE_FILES:
```

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,30 +2,41 @@
...
+                    del_file = dir_path + '/' + filename
```
Hwesta: Please use `os.path.join` here, eg `del_file = os.path.join(dir_path, filename)`

src/MCPClient/lib/clientScripts/removeUnneededFiles.py
```diff
@@ -2,30 +2,41 @@
...
+                    os.remove(del_file)
+                    logger.debug('removeUnneededFiles file deleted: %s', del_file)
+
+    ### 2. Insert removal notification into events database
+    sip_files = File.objects.filter(sip=sip_UUID).values_list('uuid', 'currentlocation')
```
Hwesta, line 27: While I think it's slightly less efficient, I think it makes for clearer and less error-prone code if the `fileWasRemoved` call happens in the same loop as the file being deleted, so the database is kept as close to in sync as the disk as possible. I'm open to discussion on this, especially if it is much less efficient, so please tell me if you disagree.

Hwesta:
Thanks for the contribution, especially one improving performance!

Can you move the database changes into a migration? Your changes work on a fresh install, but will not appear in an upgraded installation.  Django provides [documentation](https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations) on how to do that, but a summary:
- Go to the directory with manage.py (`cd src/dashboard/src/`)
- Create an empty migration (`python manage.py makemigrations --empty main`)
- Rename the migration to something meaningful. Please keep the same number
- Add the data migration content. For an example, https://github.com/artefactual/archivematica/blob/stable/1.5.x/src/dashboard/src/main/migrations/0019_normalization_report.py

I think this fix would work well for assignFileUUID, but probably not for updateSizeAndChecksum. Part of the motivation using microservices is that they can be run in parallel - by moving everything to once-per-SIP we save on startup costs, but lose parallelism. For something small like checking if a file needs to be removed or assigning checksums, startup costs can easily dominate. However, for assigning checksums, I think the parallelism gains are greater, especally with large files. Please test it out and let me know if I'm wrong though!

dnarc:
@Hwesta thanks for the review:
- I'd rather keep the SQL and filesystem operations as they are. I realise this means some temporary state inconsistency but there would be a lot of context switching and (likely) cache issues to interpolate them.
- Point taken on the parallelism, although with the constant need to sync to the filesystem and database I wonder what level of gains are actually realised for many micro-services. I did a lot of benchmarking to see which micro-services were taking the most time and updateSizeAndCheksum was one that stood out. I'll make the changes and see if they make a difference.

Hwesta:
Thanks for making the changes and adding the migration.  The migration happens instead of needing changes to initial_data.json, so the initial_data changes should be removed. Sorry that was unclear initially.

Previously, this ran on every file in the Transfer, including non-objects (metadata, service files etc).  I think we should preserve that by passing `%SIPDirectory%` instead of `%SIPObjectsDirectory%` 

Otherwise, looking good!

Hwesta:
This merges fine into stable/1.5.x but has conflicts in qa/1.x because of #486 That PR also adds deleting directories in addition to files which doesn't easily map to the query to find deleted files.  However, this runs before files have been assigned UUIDs, so it might be a moot point.  I can resolve the conflicts when merging to qa/1.x, unless you want to make a second PR. :)

dnarc:
Hey Holly, thanks for the review. If you don't mind fixing the conflicts I would appreciate it; it appears I have some work to do on the other PR you reviewed.

dnarc:
Superseded by merge into QA branch in PR#518.

PR #497, opened 2016-09-07
Decode Task stderr to Unicode prior to db save.

jrwdunham:
Fixes bug (cf. #9967) where `logTaskCompletedSQL` in databaseFunctions.py can
receive Latin-1-encoded stderr output (e.g., from GhostScript normalization),
triggering an exception from MySQLdb/Django's ORM. Note: this issue doesn't
arise in newer versions (e.g., qa/1.x), probably because of newer versions of
Django or MySQL-python.

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -77,9 +77,17 @@ def unicodeToStr(string):
         string = string.encode("utf-8")
     return string
 
-def strToUnicode(string):
+def strToUnicode(string, obstinate=False):
     if isinstance(string, str):
-        string = string.decode("utf-8")
+        try:
+            string = string.decode('utf8')
+        except UnicodeDecodeError:
+            if obstinate:
+                # Obstinately get a Unicode instance by replacing
+                # indecipherable bytes.
+                string = string.decode('utf8', 'replace')
+            else:
+                raise
     return string
 
 
--- a/src/archivematicaCommon/lib/databaseFunctions.py
+++ b/src/archivematicaCommon/lib/databaseFunctions.py
@@ -28,6 +28,8 @@ import string
 import sys
 import uuid
 
+from archivematicaFunctions import strToUnicode
+
 sys.path.append("/usr/share/archivematica/dashboard")
 from django.db.models import Q
 from django.utils import timezone
@@ -233,9 +235,13 @@ def logTaskCompletedSQL(task):
     task = Task.objects.get(taskuuid=taskUUID)
     task.endtime = getUTCDate()
     task.exitcode = exitCode
-    task.stdout = stdOut
-    task.stderror = stdError
+    # ``strToUnicode`` here prevents the MCP server from crashing when, e.g.,
+    # stderr contains Latin-1-encoded chars such as \xa9, i.e., the copyright
+    # symbol, cf. #9967.
+    task.stdout = strToUnicode(stdOut, obstinate=True)
+    task.stderror = strToUnicode(stdError, obstinate=True)
     task.save()
+
 
 def logJobCreatedSQL(job):
     """


Hwesta:
`to_unicode` looks promising, but we already have 2 str-to-unicode functions (in [databaseFunctions](https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/databaseFunctions.py#L317-L326) and [archivematicaFunctions](https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/archivematicaFunctions.py#L73-L81)).  I'd rather see them consolidated, probably to the more-used strToUnicode, than a new one added.  There was some discussion about this in https://projects.artefactual.com/issues/10226#note-9 and on Slack about how to handle this, including when to use replacement characters etc.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -219,10 +219,32 @@ def logTaskCompletedSQL(task):
...
 
+
+def to_unicode(output):
+    """
+    Attempt to decode ``output`` to Unicode using the 'ascii', 'utf8', and
+    finally 'latin-1' codecs. If all fail, decode using 'ascii', ignoring
+    indecipherable bytes.
+    """
+    try:
+        return unicode(output)
```
sevein: Should we explicitly pass the `ascii` encoding here to be more clear?

Hwesta: Possibly - I think I'd rather try `utf8` as our first choice, since `ascii` is a subset of it.

Hwesta:
Summarizing external discussion: we'll modify `strToUnicode` with a second param (defaulting to False) that, when True, causes the multiple-decode-attempt behavior of `to_unicode`

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -219,10 +221,14 @@ def logTaskCompletedSQL(task):
     task = Task.objects.get(taskuuid=taskUUID)
     task.endtime = getUTCDate()
     task.exitcode = exitCode
-    task.stdout = stdOut
-    task.stderror = stdError
+    # ``strToUnicode`` here prevents the MCP server from crashing when, e.g.,
+    # stderr contains Latin-1-encoded chars such as \xa9, i.e., the copyright
+    # symbol, cf. #9967.
+    task.stdout = strToUnicode(stdOut, True)
+    task.stderror = strToUnicode(stdError, True)
```
Hwesta: Nitpick, but for clarity `strToUnicode(stdError, obstinate=True)` would be better.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -75,9 +75,24 @@ def unicodeToStr(string):
...
-def strToUnicode(string):
+def strToUnicode(string, obstinate=False):
     if isinstance(string, str):
-        string = string.decode("utf-8")
+        if obstinate:
+            # Obstinately try to get a Unicode instance: attempt to decode to
+            # using the 'ascii', 'utf8', and finally 'latin-1' codecs. If all
+            # fail, decode using 'ascii', ignoring indecipherable bytes.
+            try:
+                string = string.decode('ascii')
```
Hwesta: Given that ASCII is a subset of UTF8, I think we can probably skip this attempt.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -75,9 +75,24 @@ def unicodeToStr(string):
...
+            except UnicodeDecodeError:
+                try:
+                    string = string.decode('utf8')
+                except UnicodeDecodeError:
+                    try:
+                        string = string.decode('latin-1')
+                    except UnicodeDecodeError:
+                        string = string.decode('ascii', 'ignore')
```
Hwesta: If we skip ASCII (as suggested above), this should probably be `string.decode('utf8', 'ignore')`

Hwesta (commented):
A couple of minor feedback suggestions, otherwise looks good!

src/archivematicaCommon/lib/archivematicaFunctions.py
jrwdunham: Good points all. How's it look now?

Hwesta: Looks good!

I was thinking about this more, and I think it might make more sense to immediately fall back to `string.decode('utf8', 'replace')`

Eg

```
>>> sj = u'元気だ'.encode('shiftjis')
>>> sj
'\x8c\xb3\x8bC\x82\xbe'
>>> print sj.decode('latin1')
³C¾
>>> print sj.decode('utf8', 'ignore') 
C
>>> print sj.decode('utf8', 'replace')
���C��
```

They're all nonsense, but utf8 with ignore makes it clear what happened.

jrwdunham: Yes, that makes sense. And also—why give special preference to latin1?

Hwesta: The reason to potentially keep latin1 is because we've had some problematic input that was latin1 (eg the copyright symbol noted in the comment) and not user data (produced by our tools). I'm ambivalent on that, though.

Hwesta:
:+1: Looks good!

Hwesta:
Also cherry-picked to stable/1.5.x as 0fc5f34ec58152d95581cbb09233cb188ce3bf9c for https://projects.artefactual.com/issues/10226

PR #498, opened 2016-09-19
MediaConch tool and FPR command.

jrwdunham:
Created migration that adds FPR tool and command for using MediaConch for validation.


sevein:
Replaced by #557.

kellyannewithane:
@jrwdunham Just some feedback - the STDOUT message is maybe not as useful as it could be.  For future development we could look at refining what appears there.

PR #499, opened 2016-09-21
Fixed message when editing taxonomy, refs #10336

jrwdunham:
Previous to this, editing a Taxonomy would result in a dict literal `{'text': 'Saved.'}` being displayed in the HTML.

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -396,7 +396,7 @@ def term_detail(request, term_uuid):
         form = TaxonomyTermForm(request.POST, instance=term)
         if form.is_valid():
             form.save()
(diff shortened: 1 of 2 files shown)


Hwesta:
Since you're fixing things in taxonomy, can you add `class='btn btn-primary'` to the save button so it's styled correctly?

Hwesta (approved):
:+1:

Hwesta:
Still :+1:

sevein:
Accidentally pushed this to the GH remote instead of the internal one and the commits were lost. Here are again: 3bd6b2455a0b404e8550797d5bb5d89039caeba7 and 35c89b1e689a0e7ca44fe781022bb9d7d3ac98d9. Sorry for not squashing it, I was doing too many things and didn't pay enough attention.

PR #500, opened 2016-09-21
Fixes for unicode in transfer names

Hwesta:
- Upgrade slumber so it doesn't use deprecated simplejson library (Edit: This won't be needed with #550)
- Encode filesystem paths as bytestrings
- In JS, assume strings are unicode and can be utf8 encoded.  Note that this probably removes support for non-unicode transfer names and paths, but needs further testing

refs #9234

Edit:
- Add MCPServer bugfix for unicode (from database) vs str (from disk) paths (Edit: Merged as part of #545)

### DIFF
--- a/src/dashboard/src/requirements/base.txt
+++ b/src/dashboard/src/requirements/base.txt
@@ -15,7 +15,7 @@ git+https://github.com/artefactual-labs/agentarchives.git#egg=agentarchives
 git+https://github.com/artefactual-labs/mets-reader-writer#egg=metsrw
 mysqlclient==1.3.7
 # Required by storage-service component
-slumber==0.6.0
+slumber
 pytz
 pyopenssl
 python-dateutil==2.4.2
--- a/src/MCPServer/lib/unit.py
+++ b/src/MCPServer/lib/unit.py
@@ -25,6 +25,8 @@ import logging
 import os
 import sys
 
+import archivematicaFunctions
+
 import archivematicaMCP
 from unitFile import unitFile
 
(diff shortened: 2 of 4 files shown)


src/dashboard/src/requirements/base.txt
```diff
@@ -15,7 +15,7 @@ git+https://github.com/artefactual-labs/agentarchives.git#egg=agentarchives
 git+https://github.com/artefactual-labs/mets-reader-writer#egg=metsrw
 mysqlclient==1.3.7
 # Required by storage-service component
-slumber==0.6.0
+slumber
```
sevein, line 18: Can you pin the dependency to the specific version you were testing?

sevein:
Holly, what's the difference between this `base64.js` and the base64-helpers submodule?

Hwesta:
Added some related unicode work. This fixes having filenames with unicode in the name (something I thought worked already?).  It was failing to match the path-on-disk with the path-in-the-db because of a unicode/str mismatch.  Encode the unicode as utf8 for the match.

This will probably not work with non-unicode filenames.

Hwesta:
We haven't run into this issue recently, even though this fix hasn't been merged. This may not be required.

jhsimpson:
I am closing this PR, we are not going to merge it now.  There is an issue that has been opened on the acceptance-tests repo to define user stories for unicode transfer names, so we know what needs to be supported.  https://github.com/artefactual-labs/archivematica-acceptance-tests/issues/11

PR #501, opened 2016-09-23
Merge updateSizeAndChecksum microservice into assignFileUUID.

dnarc:
- assignFileUUID is always followed by updateSizeAndChecksum, and is called with the same arguments.
- updateSizeAndChecksum is merely a wrapper around the updateSizeAndChecksum() database call.
- Merging the two micro-services speeds up ingests by 2-5% by avoiding the overhead of an extra micro-service and simplifies the microservice call chain.
- Other microservices such as transcribeFile already and extractContents already use the updateSizeAndChecksum database call in-line so this change actually conforms with how the rest of the code base works.
- The Django migration depends on PR#494 being merged.


src/dashboard/src/main/migrations/0021_remove_update_checksum.py
```diff
@@ -0,0 +1,69 @@
...
+        for next_msclec in next_msclecs:
+            MicroServiceChainLinkExitCode.objects.filter(microservicechainlink_id=prev_msclec).update(nextmicroservicechainlink_id=next_msclec)
+
+    # Delete all MSCLECs corresponding to updateSizeAndChecksum
+    msclecs_delete = ['c4c7ff3c-3eef-42e3-a68d-f11ca3e7c6dd', 'ea5b8e14-519e-473b-818c-e62879559816', \
+                      'd5690bcf-1c0f-44a1-846e-e63cea2b9087', '5cd0e0ee-75a1-419f-817c-4edd6adce857', \
+                      'ad09d973-5c5d-44e9-84a2-a7dbb27dd23d', 'fe188831-76b3-4487-8712-5d727a50e8ce', \
+                      'a0f33c59-081b-4427-b430-43b811cf0594', 'e0d9e83b-89e1-4711-89e4-14dbe15bea4c']
+    for msclec in msclecs_delete:
+        MicroServiceChainLinkExitCode.objects.filter(pk=msclec).delete()
```
Hwesta, line 35: These can also be done using [pk__in](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#in) eg `MicroServiceChainLinkExitCode.objects.filter(pk_in=msclecs_delete).delete()`

src/dashboard/src/main/migrations/initial_data.json
```diff
@@ -1550,19 +1550,6 @@
   }
 },
 {
-  "pk": "2714cd07-b99f-40e3-9ae8-c97281d0d429",
```
Hwesta, line 1553: Similar to #494 , the data migration happens instead of modifications to initial_json, so the should be removed.

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -32,8 +33,15 @@
 from custom_handlers import get_script_logger
 from fileOperations import addFileToTransfer
 from fileOperations import addFileToSIP
+from fileOperations import updateSizeAndChecksum
 
 
+# Update the file size and checksum in the database.
+def updateFileData(fileUUID, filePath, date, uuid):
+    updateSizeAndChecksum(fileUUID, \
```
Hwesta, line 41: All on one line is fine, since Python doesn't need `\` to continue lines inside `()`. I know where you copied it from had them, but we're trying to remove them when we touch a section of code. Thanks!

src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py
```diff
@@ -74,5 +77,7 @@
         addFileToSIP(filePathRelativeToSIP, fileUUID, opts.sipUUID, opts.eventIdentifierUUID, opts.date, use=opts.use)
 
     else:
-        print >>sys.stderr, "SIP exclusive-or Transfer uuid must be defined"
+        logger.error("SIP exclusive-or Transfer uuid must be defined")
         exit(2)
+
+    updateFileData(fileUUID, opts.filePath, opts.date, uuid.uuid4().__str__())
```
Hwesta, line 83: This doesn't happen if any of the error cases above occur, unlike when it was a separate microservice. This should be checked to see if there are any case where this is a problem.

src/dashboard/src/main/migrations/0021_remove_update_checksum.py
```diff
@@ -0,0 +1,69 @@
...
+                    'b21018df-f67d-469a-9ceb-ac92ac68654e', '8c8bac29-4102-4fd2-9d0a-a3bd2e607566']
+
+    previous_msclecs = ['88a36d00-af47-4845-8bcf-a72529ae78f8', '800259c7-f6d2-4ac6-af69-07985f23efec', \
+                        '1462359e-72c8-467f-b08a-02e4e3dfede1', 'e090fea3-2e44-4dd9-b17d-73c4d2088e0c', \
+                        '8bec6c39-8b98-4e2c-9a91-f7a9c8130f2e', '18db3a8a-ee5f-47c2-b5a7-7d223c3023c0', \
+                        '4de0e894-8eda-49eb-a915-124b4f6c3608', '5e481248-543b-463d-b0a4-eee87c79e71f']
+
+    for prev_msclec in previous_msclecs:
+        for next_msclec in next_msclecs:
+            MicroServiceChainLinkExitCode.objects.filter(microservicechainlink_id=prev_msclec).update(nextmicroservicechainlink_id=next_msclec)
```
Hwesta, line 27: I think it should be pk__in=previous_msclecs.  Also, how does this not set everything to the last next_msclecs?

Hwesta (commented):
Thanks for the contribution! Looks very promising, with a few minor comments.

dnarc:
Superseded by PR#522 merge into QA branch.

PR #503, opened 2016-09-26
Look for non-zero exit code, not just > 0

jhsimpson:
Fixes #10354

As suggested by Holly Becker.
This has been tested in a case where some tasks return -1.


Hwesta:
:+1: Though since I suggested the fix, so I'm not sure I should be the only reviewer.

PR #505, opened 2016-09-26
Switch to Django persistent connections from django-mysqlpool

jhsimpson:
As described in https://projects.artefactual.com/issues/10355

This PR removes django-mysqlpool, in favour of the built in connection pooling functionality available in Django >= 1.6.

The MCP Client changes in here have been tested, but the MCP Server changes have not.

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -51,7 +51,6 @@ CONFIG_MAPPING = {
     'db_password': {'section': 'client', 'option': 'password', 'type': 'string'},
     'db_host': {'section': 'client', 'option': 'host', 'type': 'string'},
     'db_port': {'section': 'client', 'option': 'port', 'type': 'string'},
-    'db_pool_max_overflow': {'section': 'client', 'option': 'max_overflow', 'type': 'string'},
 }
 
 CONFIG_DEFAULTS = """[MCPClient]
@@ -81,9 +80,8 @@ user = archivematica
 password = demo
 host = localhost
 database = MCP
-max_overflow = 40
 port = 3306
-engine = django_mysqlpool.backends.mysqlpool
+engine = django.db.backends.mysql
 """
 
 config = Config(env_prefix='ARCHIVEMATICA_MCPCLIENT', attrs=CONFIG_MAPPING)
@@ -102,16 +100,14 @@ DATABASES = {
         'PASSWORD': config.get('db_password'),
         'HOST': config.get('db_host'),
         'PORT': config.get('db_port'),
+
+        # Recycling connections in MCPClient is not an option because this is
+        # a threaded application. We need a connection pool but we don't have
+        # one we can rely on at the moment - django_mysqlpool does not support
+        # Py3 and seems abandoned.
+        'CONN_MAX_AGE': 0,
     }
 }
-
-MYSQLPOOL_BACKEND = 'QueuePool'
-MYSQLPOOL_ARGUMENTS = {
-    'use_threadlocal': False,
-    'max_overflow': config.get('db_pool_max_overflow'),
-}
-
-CONN_MAX_AGE = 14400
 
 # Make this unique, and don't share it with anybody.
 SECRET_KEY = config.get('secret_key', default='e7b-$#-3fgu)j1k01)3tp@^e0=yv1hlcc4k-b6*ap^zezv2$48')
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -50,8 +50,7 @@ from django.conf import settings as django_settings
 
 from main.models import Task
 
-from django_mysqlpool import auto_close_db
-import databaseFunctions
+from databaseFunctions import auto_close_db, getUTCDate
 from executeOrRunSubProcess import executeOrRun
 
 
@@ -90,7 +89,7 @@ def executeCommand(gearman_worker, gearman_job):
         execute = gearman_job.task
         logger.info('Executing %s (%s)', execute, gearman_job.unique)
         data = cPickle.loads(gearman_job.data)
-        utcDate = databaseFunctions.getUTCDate()
+        utcDate = getUTCDate()
         arguments = data["arguments"]  # .encode("utf-8")
         if isinstance(arguments, unicode):
             arguments = arguments.encode("utf-8")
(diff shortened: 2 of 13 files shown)


src/MCPClient/lib/settings/common.py
```diff
@@ -16,20 +16,20 @@
...
 DATABASES = {
     'default': {
-        'ENGINE': 'django_mysqlpool.backends.mysqlpool',
+        'ENGINE': 'django.db.backends.mysql',
         'NAME': config.get('client', 'database'),          # Or path to database file if using sqlite3.
         'USER': config.get('client', 'user'),              # Not used with sqlite3.
         'PASSWORD': config.get('client', 'password'),      # Not used with sqlite3.
         'HOST': config.get('client', 'host'),              # Set to empty string for localhost. Not used with sqlite3.
         'PORT': '',                                        # Set to empty string for default. Not used with sqlite3.
+        'CONN_MAX_AGE': 500,
```
Hwesta: Do we want to set this to `None`, which allows unlimited persistent connections? https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-CONN_MAX_AGE

jhsimpson: My initial assumption was that setting an actual limit like 500 seconds prevents a runaway query from chewing up one connection for ever.  If there is a query that takes longer than 500 seconds, it will time out and die, causing an exception, with this setting.  

I don't know enough about how Django handles edge cases like connections that die due to bugs in the driver, or network issues etc.  Probably it works just fine.

I tested with a value of 500, it would be good to test with other values. I suggest we are just looking for a default value that is known to work and the user should/could change as needed.

Perhaps I should change 500 to 'config.get('client', 'conn_max_age') and set it in the config file?

sevein: Something to remember is that Django achieves thread-safety by not sharing connections between threads.

https://docs.djangoproject.com/en/dev/ref/databases/#caveats

> Since each thread maintains its own connection, your database must support at least as many simultaneous connections as you have worker threads.

With this caveat in mind, my two concerns related to scalability issues are:
- We are deploying gunicorn using the `sync` [worker class](http://docs.gunicorn.org/en/stable/settings.html#worker-class) (synchronous workers). Under this concurrency model, persistent connections created within a worker will be successfully reused on new requests processed by the same worker. However, if we ever switch to greenlets (`eventlet` or `gevent` worker class) - which we should seriously consider for performance and reliability reasons, persistent connections will become an issue because Django's thread-safety model will force these green threads to make new connections on each request, while a connection pool allows these connections to be shared. I also think that there are big chances (not confirmed) that these connections created by the green threads won't ever be closed by the client and will accumulate.
- In MCPServer, we create a significant amount of POSIX threads. By adding a pool we allowed these threads to share and reuse existing connections and keep the total number of connections to the database as low as the size of the pool. To make sure threads return connections to the pool as soon as possible we used the `@auto_close_db` decorator. It also helped to reduce thread lock contention. Not having a pool I believe puts us back to a concurrency model where connections are not shared and created on a per-thread basis.

If we stop using our pool, we should at least make sure that MCPServer's threaded functions close the database connections as soon as they return. Simplest solution would be to implement our own `@auto_close_db` decorator. The fundamental reason why the MCPServer needs extra care is because Django has its own mechanism to close these connections when a web request starts and/or finishes - this mechanism just never happens when you are using Django's ORM outside the web context. From [/django/db/**init**.py](https://github.com/django/django/blob/f2ff1b2fabbe26d5e61d690c4c5a47f9582f9300/django/db/__init__.py#L57-L63), see the callbacks triggered after signals `request_started` and `request_finished` are emitted:

``` python
# Register an event to reset transaction state and close connections past
# their lifetime.
def close_old_connections(**kwargs):
    for conn in connections.all():
        conn.close_if_unusable_or_obsolete()
signals.request_started.connect(close_old_connections)
signals.request_finished.connect(close_old_connections)
```

Hwesta: Based on this and our conversation, I re-implemented `auto_close_db` using `close_old_connections` and added it to the same functions.

We should definitely document that this will probably behave poorly with async gunicorn workers, but I'm not sure the best spot.

I like the idea of making the timeout configurable in the settings file.

sevein:
Looking good!

sevein:
@Hwesta, I've just realized that this PR is not removing from our settings the directives specific to `mysqlpool`, e.g. `MYSQLPOOL_BACKEND` or `MYSQLPOOL_ARGUMENTS`.

Hwesta:
> I've just realized that this PR is not removing from our settings the directives specific to mysqlpool, e.g. MYSQLPOOL_BACKEND or MYSQLPOOL_ARGUMENTS.

Removed, good catch. I also set the `CONN_MAX_AGE` in the databases dict in all settings files.  Previously it was set to 14400 (4 hours) in some places and 500 (~8 mins) in others - updated all to use 500 seconds.

sevein:
I've found this comment in StackOverflow very useful to understand the differences between persistent connections (introduced by Django 1.6) and connection pools (added by django_mysqlpool and similar extensions): http://stackoverflow.com/a/9736414.

I think that we need to live with the fact that Django's ORM wasn't designed to run in multithreaded environments. We use the ORM from different scenarios with different execution models. Below is my attempt to summarize the best approach for each scenario.

| Service | v1.6 | v1.7 | Notes
|---|---|---|---|
| **MCPServer** | Pooling | Pooling | MCPServer benefited from having a pool. Problem: MCPServer creates too many threads (see #553), e.g. user can run 100 transfers simultaneously - there should be contention. It needs to be revised.
| **MCPClient** | Pooling | None | MCPClient did not benefit from having a pool. MCPClient starts a small number of long-running Gearman workers as threads (as many as CPUs you have) hence any kind of connection sharing here is not needed.
| **Client scripts** | Pooling | None | Client scripts are not multithreaded.  Use `CONN_MAX_AGE = 0` and close connections when it exits.
| **Dashboard (sync)** | None | Persistent connections | Use `CONN_MAX_AGE > 0`.
| **Dashboard (async)** | N/A | Pooling | *We're not using async workers yet.* E.g. [django-db-geventpoo](https://github.com/jneight/django-db-geventpool), read [this](https://github.com/benoitc/gunicorn/issues/996) for more details.

sevein:
I've deleted a comment accidentally! Sorry...

sevein:
#762 replaces this PR.

PR #506, opened 2016-09-28
Dashboard: i18n, refs #8161 (1/2)

sevein:
**i18n**-*ize* all the things,

See also:
- Redmine: https://projects.artefactual.com/issues/8161
- Notes: https://gist.github.com/sevein/75732d85e129348dc32e6c4b15982bf8

Tasks:
- [x] Dashboard: templates, views, etc...
- [x] Dashboard: JavaScript code
- [x] Dashboard: appraisal tab (https://github.com/artefactual-labs/appraisal-tab/pull/151)
- [x] Dashboard: transfer browser (https://github.com/artefactual-labs/transfer_browser/pull/12)
- [x] Archivematica FPR (https://github.com/artefactual/archivematica-fpr-admin/pull/32)
- [x] Field `Job.currentstep`
- [x] Dashboard: use new client, DashboardSetting dicts, etc...


sevein:
Actually, not sure if this is meant to be done for 1.6 o 1.7... feel free to update!

sevein:
Updated models.

src/dashboard/src/requirements/base.txt
```diff
@@ -9,6 +9,7 @@ django-extensions==1.1.1
 django-autoslug==1.7.1
 django-annoying==0.7.7
 elasticsearch>=1.0.0,<2.0.0
+git+https://github.com/artefactual/archivematica-fpr-admin.git@dev/issue-8161-dashboard-i18n#egg=fpr
```
Hwesta: This seems to generate a warning (error?) `Running setup.py (path:/tmp/pip-build-iBYv5A/fpr/setup.py) egg_info for package fpr produced metadata for project name archivematica-fpr-admin. Fix your #egg=fpr fragments`

sevein: I've fixed this, also now pointing to qa/1.x.

src/dashboard/src/components/administration/urls.py
```diff
@@ -29,6 +30,8 @@
     url(r'dips/atk/$', views.administration_atk_dips),
     url(r'dips/atom/$', views.atom_dips),
     url(r'dips/atom/edit_levels/$', views.atom_levels_of_description),
+    url(r'^i18n/', include('django.conf.urls.i18n')),
```
Hwesta: What's the `i18n` url doing? I can't get anything at `administration/i18n/` and `administration/i18n/setlang/` just redirects back to transfers.

sevein: It exposes the named urlpattern `set_language`. I'm POST'ing to it from `language.html`, see the form tag:

``` html
<form action="{% url 'set_language' %}" method="post">
   <!-- ... -->
</form>
```

Hwesta: Ah, okay.

src/dashboard/src/components/administration/forms.py
```diff
@@ -149,30 +156,14 @@ class Meta:
...
-    dip_upload_atom_debug = forms.ChoiceField(required=False,
-        label="Debug mode",
-        help_text="Show additional details.",
-        choices=((False, 'No'), (True, 'Yes')))
-    dip_upload_atom_key = forms.CharField(required=False,
-        label="REST API key",
-        help_text="Used in metadata-only DIP upload.")
+    dip_upload_atom_url = forms.CharField(required=True, label=_("Upload URL"), help_text=_("URL where the Qubit index.php frontend lives, SWORD services path will be appended."))
+    dip_upload_atom_email = forms.CharField(required=True, label=_("Login email"), help_text=_("E-mail account used to log into Qubit."))
+    dip_upload_atom_password = forms.CharField(required=True, label=_("Login password"), help_text=_("E-mail account used to log into Qubit."))
```
Hwesta: This is the original text, but 'E-mail account used to log into Qubit.' doesn't make sense as help text on the password.

src/dashboard/src/components/administration/forms.py
```diff
@@ -210,114 +201,114 @@ class ProcessingConfigurationForm(forms.Form):
...
+        'label': _('Send transfer to quarantine'),
         'yes_option': '97ea7702-e4d5-48bc-b4b5-d15d897806ab',
         'no_option': 'd4404ab1-dc7f-4e9e-b1f8-aa861e766b8e',
     }
     processing_fields['19adb668-b19a-4fcb-8938-f49d7485eaf3'] = {
         'type': 'days',
         'name': 'quarantine_expiry_days',
-        'label': 'Remove from quarantine after (days)',
+        'label': _('Remove from quarantine after (days)'),
         'placeholder': 'days',
```
Hwesta: Should placeholder be translated too?

src/dashboard/src/components/administration/models.py
```diff
@@ -18,37 +21,38 @@ class Meta:
 EAD_ACTUATE_CHOICES = [('none', 'None'), ('onLoad', 'onLoad'), ('other', 'other'), ('onRequest', 'onRequest')]
 EAD_SHOW_CHOICES = [('embed', 'Embed'), ('new', 'New'), ('none', 'None'), ('other', 'Other'), ('replace', 'Replace')]
```
Hwesta: Should these be translated too?

src/dashboard/src/main/views.py
```diff
@@ -22,6 +22,12 @@
 from django.core.urlresolvers import reverse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.http import Http404, HttpResponse
+from django.utils import timezone
+from django.utils.translation import get_language
+from django.utils.translation import ugettext as _
+from django.views.decorators.cache import cache_page
+from django.views.decorators.http import last_modified
+from django.views.i18n import javascript_catalog
```
Hwesta: Are all these imports used? I only see `_`

src/dashboard/src/templates/administration/term_detail.html
```diff
@@ -35,7 +31,8 @@
       {% include "_form.html" %}
 
       <div style='float: right'>
-        <input type='submit' value='Save' /> <a href='/administration/taxonomy/term/{{ term.pk }}/delete/' class='btn btn-danger'>Delete</a>
+        <input type='submit' value='{% trans "Save" %}' />
+        <a href='/administration/taxonomy/term/{{ term.pk }}/delete/' class='btn btn-danger'>{% trans "Delete" %}</a>
```
Hwesta, line 35: Out of scope for this, but a PR updating all URLs to use `{% url ... %}` should happen at some point.

src/dashboard/src/templates/ingest/backlog/_search_form.html
```diff
@@ -1,10 +1,12 @@
+{% load i18n %}
+
 <form id='search_form' class='form-inline'>
   <div class='clearfix'>
     <div id='search_form_container'></div>
     <div id='search_form_submit_container'>
-      <input type='button' class='form-control btn-primary' id='search_submit' value='Search transfer backlog'>
       {% if show_files %}
-      <span style='margin-left: 10px'>Show files?</span>
+      <input type='button' class='form-control btn-primary' id='search_submit' value='{% trans "Search transfer backlog" %}'>
```
Hwesta: The search transfer backlog button should not be inside `{% if show files %}`

src/dashboard/src/templates/ingest/grid.html
```diff
@@ -180,32 +181,33 @@
...
-      <span class="microservice-group-name">Micro-service: <%= name %></span>
+      <span class="microservice-group-name">{% trans "Microservice" %}: <%= name %></span>
     </div>
   </script>
 
   <script type="text/template" id="job-template">
     <div class="job-detail-microservice">
-      <span class="job-type-label">Job:&nbsp;</span><span title="<%= uuid %>"><%= type %> <% if(obj.filename) { %>(file: <i><%= filename %></i>)<% } %></span>
+      <span class="job-type-label">{% trans "Job" %}</span>
+      <span title="<%= uuid %>"><%= type %> </span>
```
Hwesta: Is `<% if(obj.filename) { %>(file: <i><%= filename %></i>)<% } %>` unnecessary now?

src/dashboard/src/templates/main/tasks.html
```diff
@@ -11,21 +12,55 @@
...
+
             <span>&nbsp;</span> {{ item.exec_field }} (exit code: {{ item.exitcode }})
           </td>
           <td>
-            <span>Start time:</span> <span class='datetime'>{{ item.starttime.isoformat }}</span><br />
-            <span>End time:</span> <span class='datetime'>{{ item.endtime.isoformat }}</span><br />
-            <span>Created time:</span> <span class='datetime'>{{ item.createdtime.isoformat }}</span><br />
-            <span>Duration:</span> {{ item.duration }} second(s)
+            {% blocktrans with datetime=item.starttime.isoformat %}
+              <span>Start time:</span> <span class="datetime">{{ datetime }}</span>
```
Hwesta: Does the HTML & value need to be included in the translation or just the label?  I suppose it might be necessary for RTL reordering.  Somewhat concerned about putting styling in translations.

src/dashboard/src/templates/rights/rights_edit.html
```diff
@@ -13,43 +14,56 @@
...
 
       {% if new_content_type_created %}
           <div class="alert alert-info">
             {% if new_content_type_created == 'Other' %}
-              <p>New content added.</p>
+              <p>{% trans "New content added." %}</p>
             {% else %}
-              <p>New {{ new_content_type_created|lower }} content added.</p>
+              <p>
+                {% blocktrans with type=new_content_type_created|lower %}
```
Hwesta: Is `new_content_type_created` already translated? If so, should it be `|lower`, since that may not apply to the language (or in German where I think all nouns have caps)?

src/dashboard/src/templates/rights/rights_grants_edit.html
```diff
@@ -13,37 +14,50 @@
...
 
       {% if new_content_type_created %}
           <div class="alert alert-info">
             {% if new_content_type_created == 'Other' %}
-              <p>You may now add additional information such as document identifers and notes.</p>
+              <p>{% trans "You may now add additional information such as document identifers and notes." %}</p>
             {% else %}
-              <p>You may now add additional {{ new_content_type_created|lower }} information such as document identifers and notes.</p>
+              <p>
+                {% blocktrans with type=new_content_type_created|lower %}
```
Hwesta: Same comment about `|lower`

src/dashboard/src/templates/transfer/grid.html
```diff
@@ -99,13 +100,13 @@
...
         <%= directory %>
-        <abbr title="<%= uuid %>">UUID</span>
+        <abbr title="<%= uuid %>">{% trans "UUID" %}</span>
       </div>
       <div class="sip-detail-uuid"><%= uuid %></div>
       <div class="sip-detail-timestamp"><%= this.getIngestStartTime() %></div>
       <div class="sip-detail-actions">
-        <a class="btn_show_metadata" href="#" title="Metadata"><span>Micro-Services</span></a>
-        <a class="btn_remove_sip" href="#" title="Remove"><span>Remove</span></a>
+        <a class="btn_show_metadata" href="#" title="{% trans 'Metadatos' %}"><span>{% trans "Metadatos" %}</span></a>
```
Hwesta: This should be left as English.

src/dashboard/src/settings/common.py
```diff
@@ -47,10 +50,31 @@
...
 # Language code for this installation. All choices can be found here:
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-us'
 
+# https://docs.djangoproject.com/en/dev/ref/settings/#languages
+LANGUAGES = [
+    ('fr', _('French')),
+    ('en', _('English')),
+    ('es', _('Spanish')),
+]
```
Hwesta, line 72: Is this the final set of languages, or just the ones for testing? If final, why these three, and not others?

src/dashboard/src/components/administration/migrations/0007_auto_20161116_1728.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('administration', '0006_use_statement_optional'),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
Hwesta: This migration should be squashed with the commit that introduced the change before merging.

src/MCPClient/lib/clientScripts/emailFailReport.py
```diff
@@ -29,9 +29,7 @@
 import argparse
 from lxml import etree
 
-import django
-django.setup()
-# dashboard
+import django; django.setup()
```
Hwesta: Why oneline these? Python generally prefers statements on different lines.

src/dashboard/src/main/models.py
```diff
@@ -403,7 +403,19 @@ class Job(models.Model):
...
+    STATUS_EXECUTING_COMMANDS = 3
+    STATUS_FAILED = 4
+    STATUS = (
+        (STATUS_UNKNOWN, _('Unknown')),
+        (STATUS_AWAITING_DECISION, _('Awaiting decision')),
+        (STATUS_COMPLETED_SUCCESSFULLY, _('Completed successfully')),
+        (STATUS_EXECUTING_COMMANDS, _('Executing command(s)')),
+        (STATUS_FAILED, _('Failed'))
+    )
+    currentstep = models.IntegerField(db_column='currentStep', choices=STATUS, default=0, blank=False)
```
Hwesta: Don't forget to update `src/dashboard/tests/test_api.py` & related fixtures for these changes.

Hwesta (commented):
Wow we have a lot of user-facing strings! Code reviewed up to commit "Dashboard: i18njs, import transfer-browser"

src/dashboard/src/main/views.py
sevein: Right, that's going to be later used by the i18njs catalogue view. I'll fix it.

src/dashboard/src/templates/administration/term_detail.html
sevein: Right! I didn't even notice!

src/dashboard/src/templates/ingest/grid.html
sevein: That was a mistake!

src/dashboard/src/templates/main/tasks.html
sevein: You're right. I'm fixing it. I'm making some rearrangements so the different properties are described using `<dl></dl>`.

src/dashboard/src/templates/transfer/grid.html
sevein: Opssss

src/dashboard/src/templates/rights/rights_grants_edit.html
sevein: This whole `rights/` component is a pain.

src/dashboard/src/settings/common.py
sevein: French is a must and I added Spain so I could make my own translations for testing purposes. We can add more later but I wanted to start small to keep it manageable.

src/dashboard/src/main/migrations/0030_job_currentstep_choices.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def data_migration(apps, schema_editor):
+    Job = apps.get_model('main', 'Job')
+    MicroServiceChainLink = apps.get_model('main', 'MicroServiceChainLink')
+    MicroServiceChainLinkExitCode = apps.get_model('main', 'MicroServiceChainLinkExitCode')
+
+    # I can't rely on the model registry neither the current model.
+    replacements = {
+        'Awaiting decision': 1,
```
Hwesta: Should this also include unknown: 0?

src/dashboard/src/main/migrations/0031_dashboardsetting_scope.py
```diff
@@ -0,0 +1,238 @@
...
+    old_prefix = 'dip_upload_atom_'
+    fields = ('url', 'email', 'password', 'rsync_target', 'rsync_command', 'version', 'debug', 'key')
+
+    qs = apps.get_model('main', 'DashboardSetting').objects.filter(name__in=['{}{}'.format(old_prefix, field) for field in fields])
+    old_dict = dict(qs.values_list('name', 'value'))
+    new_dict = {field: old_dict.get('{}{}'.format(old_prefix, field)) for field in fields}
+
+    # Set new dict and delete previous tuples. Reminder: migrations run inside
+    # a transaction.
+    DashboardSetting.objects.set_dict('upload-qubit_v0.0', new_dict)
```
Hwesta: Will this migration still work if set_dict changes in the future?

src/protos/workflow.proto
```diff
@@ -0,0 +1,107 @@
...
+        uint32 code = 1;
+        Job.Status jobStatus = 2;
+        string linkId = 3;
+    }
+    message LinkConfig {
+        message Standard {
+            string execute = 1;
+            string arguments = 2;
+            string filterSubdir = 3;
+            string filterFileStart = 4;
```
Hwesta: Looks like this field is actually never used (the value is always null) - may we can just remove it?  Possibly out of scope for this change.

src/archivematicaCommon/lib/mcpserver_cli.py
```diff
@@ -0,0 +1,60 @@
...
+    ]
+
+    if len(sys.argv) == 1:
+        usage(commands)
+
+    cmd = sys.argv[1]
+    client = mcpserver.get_client()
+
+    if cmd not in commands:
+        usage(commands)
```
Hwesta: Why not use `argparse` for this sort of checking? It automatically generates usage help, can check that a parameter is in a list, and use subparsers if different commands need different parameters.

src/archivematicaCommon/lib/mcpserver_cli.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if cmd == 'ApproveJob':
+        try:
+            job_uuid = sys.argv[2]
+            choice_uuid = sys.argv[3]
+        except IndexError:
+            sys.exit('Missing parameters (job_uuid, choice_uuid)')
+        approved = client.approve_job(job_uuid, choice_uuid)
+        print("Has the job been approved? {}".format("Yes!" if approved else "No, :("))
```
Hwesta: Clearly this should be `print("Has the job been approved? {}".format("Yes!" if approved else "No, 😞"))` ;)

src/MCPServer/lib/gRPCServer.py
```diff
@@ -0,0 +1,181 @@
...
+                        description = item[1]['en']
+                    except KeyError:
+                        continue
+                    if description != 'Approve transfer':
+                        continue
+                    match = task_manager
+                    break
+            try:
+                match
+            except NameError:
```
Hwesta: I think it's more idiomatic to set `match = None` at the top, and use `if match is None`.

src/dashboard/src/media/js/jobs.js
```diff
@@ -433,132 +410,56 @@ var MicroserviceGroupView = Backbone.View.extend({
...
+      var previewUrl = function(linkId, uuid) {
+        // Store AIP
+        if (linkId == '2d32235c-02d4-4686-88a6-96f4d6c7b1c3') {
+          return '/ingest/preview/aip/' + uuid;
+        // Approve normalization
+        } else if (linkId == 'de909a42-c5b5-46e1-9985-c031b50e9d30') {
+          return '/ingest/preview/normalization/' + uuid;
+        // Move to the uploadedDIPs directory
+        } else if (linkId == '2e31580d-1678-474b-83e5-a53d97d150f6' || linkId == 'e3efab02-1860-42dd-a46c-25601251b930') {
+          return '/ingest/preview/dip/' + uuid;
```
Hwesta: Is it possible to use Django's `{% url %}` here?

src/archivematicaCommon/lib/workflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Client(object):
+    DEFAULT_ADDRESS = 'localhost:50050'
+
+    def __init__(self, addr=None):
+        self.addr = addr if addr is not None else self.DEFAULT_ADDRESS
+
+        self._init_metadata()
+
+        self.channel = grpc.insecure_channel(self.addr)
```
Hwesta: Are we planning on making this a `secure_channel` later?

Hwesta (commented):
Looked at 'Models: add i18n-choices to Job.currentstep' to half of 'gRPC: MCPServer and Workflow API'  

A few small comments, but overall I like the direction this is going. Removing the DB as the method of communication and replacing it with something that's actually intended to be spread across different architecture parts makes a lot of sense, and will hopefully make database management easier down the line.

I'm interested in what possibilities other than gRPC exist that might also serve the same purpose. Did you look at any other libraries when considering this?

I'll look at the rest of this PR and artefactual-labs/archivematica-workflow after I get back from vacation.

sevein:
> Did you look at any other libraries when considering this?

I briefly looked into other solutions. Apache Thrift is very similar but gRPC is gaining some momentum. The list of open source projects using it is really impressive: etcd, docker, mesos, kubernetes... gRPC is interesting in that it was designed to support multiple data formats, protobuf is the default.

src/archivematicaCommon/lib/workflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Error(Exception):
+    def __init__(self, message):
+        self.message = 'Workflow client error: ' + message
+
+
+class ServerUnavailableError(Error):
+    def __init__(self, code):
+        self.code = code
+        self.message = 'server unavailable, code ' + self.code
```
Hwesta: Can't concatenate a str & int - `self.message = 'server unavailable, code %s' % (self.code,)` will work.  Should probably do the same precaution above.

src/dashboard/src/main/migrations/0031_dashboardsetting_scope.py
sevein: I'm glad you've asked this question because I hadn't thought about it!

I've found two relevant sections in the docs:
- [Historical models](https://docs.djangoproject.com/en/1.8/topics/migrations/#historical-models)
- [Model managers](https://docs.djangoproject.com/en/1.8/topics/migrations/#model-managers)

If I understood properly I can add the attribute `use_in_migrations` (`True`) to `DashboardSettingManager` so the manager becomes part of the migration. However, the manager has to stay around as long as the migration references to the manager. The method `set_dict` should be compatible with all past migrations using it.

More here: https://code.djangoproject.com/ticket/23822.

I could take the two following actions:
- Use `use_in_migrations` (will probably require to regenerate the migration).
- Add a warning to the docs of `set_dict` mention this issue.

Thoughts?

src/protos/workflow.proto
sevein: I think it makes sense to get rid of it. I've done the same with other unused fields IIRC. I will double check before but sure!

src/archivematicaCommon/lib/mcpserver_cli.py
sevein: You're right. I will do it. They were conceived for testing purposes but I think they can be handy. If `archivematica-common` becomes its own pypi package, these two could be defined as extra entry points. Eventually I think we should have just one public-facing CLI tool but we're still far from that.

src/archivematicaCommon/lib/workflow.py
sevein: Yes to this but I'm hoping I can simplify as much as possible. If we adopt the sidecar model (i.e. we run something like the linkerd proxy in every host) we could just use the insecure channel to speak to the sidecar and let it take care of the encryption for us. See this article: [transparent TLS with linkerd](https://blog.buoyant.io/2016/03/24/transparent-tls-with-linkerd/). That would make things easier from the perspective of our microservices but also operationally speaking. E.g. we would only have to set up the certificates in the sidecar during deployement (private key, public key, authority cert...), we wouldn't have add extra config parameters to our apps, etc...

PR #508, opened 2016-09-29
Replace pbzip2 with lbzip2 for better speed with similar compression ratio.

dnarc:


src/dashboard/src/main/migrations/initial_data.json
```diff
@@ -9977,7 +9977,7 @@
     "replaces": null, 
     "choiceavailableatlink": "01d64f58-8295-4b7b-9cab-8f1b153a504f", 
     "description": "Parallel bzip2", 
-    "replacementdic": "{\"%AIPCompressionAlgorithm%\":\"pbzip2-\"}"
+    "replacementdic": "{\"%AIPCompressionAlgorithm%\":\"lbzip2-\"}"
```
Hwesta, line 9980: The initial data shouldn't be modified - the migration is enough.

src/MCPClient/lib/clientScripts/createPointerFile.py
```diff
@@ -15,6 +15,9 @@
 # dashboard
 from django.utils import timezone
 from main.models import DublinCore
+import os
+import subprocess
```
Hwesta, line 19: Nitpick: Please move these up to the section for standard library imports. You could also modify `import os.path` to be just `import os` above.

src/dashboard/src/main/migrations/0022_use_pbzip_aip.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # Modify user choice to indicate use of lbzip2 instead of pbzip2.
+    newAipCompressionAlgorithmString = '{"%AIPCompressionAlgorithm%":"lbzip2-"}'
+    MicroServiceChoiceReplacementDic.objects.filter(pk='f61b00a1-ef2e-4dc4-9391-111c6f42b9a7').update(replacementdic=newAipCompressionAlgorithmString)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0021_remove_update_checksum'),
```
Hwesta, line 18: This will have to be updated for migrations that have happened since, but I can do on merge.

Hwesta (commented):
Sorry for taking so long to get back to you!

The code changes for this look fine, but I'm wondering about the motivation. As far as I can tell, pbizp2 and lbzip2 have similar speed and compression characteristics, especially given the wide range of formats Archivematica might be compressing.  Can you point me to some sources for why it's worth switching?

Thank you!

dnarc:
Closing.

PR #509, opened 2016-10-04
Models: Index File.currentlocation

Hwesta:
We often query for a file by it's location, notably in indexAIP. Add an index to improve performance.

Suggested by Aaron Elkiss on the archivematica-tech list  https://groups.google.com/forum/#!msg/archivematica-tech/oYFHiJZ2eis/PZqbF4TLBQAJ

This is connected to https://github.com/archivematica/Issues/issues/75.

### DIFF
--- a/src/dashboard/src/main/migrations/0027_index_currentlocation.py
+++ b/src/dashboard/src/main/migrations/0027_index_currentlocation.py
@@ -12,9 +12,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AlterField(
-            model_name='file',
-            name='currentlocation',
-            field=main.models.BlobTextField(null=True, db_column=b'currentLocation', db_index=True),
-        ),
+        migrations.RunSQL('CREATE INDEX Files_currentLocation ON Files(currentLocation(255))')
     ]
--- a/src/dashboard/src/main/migrations/0027_index_currentlocation.py
+++ b/src/dashboard/src/main/migrations/0027_index_currentlocation.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import main.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('main', '0026_agent_m2m_event'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='file',
+            name='currentlocation',
+            field=main.models.BlobTextField(null=True, db_column=b'currentLocation', db_index=True),
+        ),
+    ]
--- a/src/dashboard/src/main/models.py
+++ b/src/dashboard/src/main/models.py
@@ -345,7 +345,7 @@ class File(models.Model):
     transfer = models.ForeignKey(Transfer, db_column='transferUUID', to_field='uuid', null=True, blank=True)
     # both actually `longblob` in the database
     originallocation = BlobTextField(db_column='originalLocation')
-    currentlocation = BlobTextField(db_column='currentLocation', null=True, db_index=True)
+    currentlocation = BlobTextField(db_column='currentLocation', null=True)
     filegrpuse = models.CharField(max_length=50, db_column='fileGrpUse', default='Original')
     filegrpuuid = models.CharField(max_length=36, db_column='fileGrpUUID', blank=True)
     checksum = models.CharField(max_length=128, db_column='checksum', blank=True)
--- a/src/dashboard/src/main/models.py
+++ b/src/dashboard/src/main/models.py
@@ -345,7 +345,7 @@ class File(models.Model):
     transfer = models.ForeignKey(Transfer, db_column='transferUUID', to_field='uuid', null=True, blank=True)
     # both actually `longblob` in the database
     originallocation = BlobTextField(db_column='originalLocation')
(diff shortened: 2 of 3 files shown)


jhsimpson:
:+1: 

nice enhancement, thanks Aaron

aelkiss:
This produces the following error for me when attempting to run the migration:

`django.db.utils.OperationalError: (1170, \"BLOB/TEXT column 'currentLocation' used in key specification without a key length\")`

(I found some suggestion that adding a `max_length` parameter in the migration might help, but that doesn't seem to work either.)

I'm not familiar with Django, but it looks like you can specify custom SQL to run in a migration, for example:

`migrations.RunSQL('create index Files_currentLocation on Files(currentLocation(767))'`

https://docs.djangoproject.com/en/1.10/ref/migration-operations/#runsql

I tried that (see https://github.com/mlibrary/archivematica/tree/dev/issue-xxxx-index-currentlocation) but that produces the error:

```
  Applying main.0027_index_currentlocation...
:stderr: Traceback (most recent call last):
  File \"./manage.py\", line 10, in <module>
    execute_from_command_line(sys.argv)
  File \"/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py\", line 354, in execute_from_command_line
    utility.execute()
  File \"/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py\", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File \"/usr/local/lib/python2.7/dist-packages/django/core/management/base.py\", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File \"/usr/local/lib/python2.7/dist-packages/django/core/management/base.py\", line 445, in execute
    output = self.handle(*args, **options)
  File \"/usr/local/lib/python2.7/dist-packages/django/core/management/commands/migrate.py\", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py\", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/migrations/executor.py\", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/migrations/migration.py\", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/special.py\", line 102, in database_forwards
    self._run_sql(schema_editor, self.sql)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/migrations/operations/special.py\", line 125, in _run_sql
    statements = schema_editor.connection.ops.prepare_sql_script(sqls)
  File \"/usr/local/lib/python2.7/dist-packages/django/db/backends/base/operations.py\", line 270, in prepare_sql_script
    import sqlparse
ImportError: No module named sqlparse
", "path": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "state": "absent", "syspath": ["/tmp/ansible_pOPHxf", "/tmp/ansible_pOPHxf/ansible_modlib.zip", "/tmp/ansible_pOPHxf/ansible_modlib.zip", "/usr/lib/python2.7", "/usr/lib/python2.7/plat-x86_64-linux-gnu", "/usr/lib/python2.7/lib-tk", "/usr/lib/python2.7/lib-old", "/usr/lib/python2.7/lib-dynload", "/usr/local/lib/python2.7/dist-packages", "/usr/lib/python2.7/dist-packages"]}
```

Not sure what the best way to fix that is or even if this is a good approach...

Hwesta:
@aelkiss Is `create index Files_currentLocation on Files(currentLocation(767))` what you used to troubleshoot this issue initially?  It looks like [mysql doesn't like indexing a `blob`/`text`](http://stackoverflow.com/questions/1827063/mysql-error-key-specification-without-a-key-length)  because of it's variable length.

Thanks for catching the bug - I ran the tests but against sqlite so it didn't catch this.

aelkiss:
Mysql can index it, just not the entire field. You just need to specify the prefix length (that's what the '767' is for - that's by default the maximum prefix length (see http://dev.mysql.com/doc/refman/5.7/en/innodb-restrictions.html). I suspect indexing 767 characters would be more than enough for most filenames. But django doesn't have a way of adding that index prefix length to the CREATE INDEX option, hence `migrations.RunSQL` suggestion.

Hwesta:
To run `migrations.RunSQL` the new dependency `sqlparse` has to be added.  I added a commit that replaces Django's add-index migration with your RunSQL migration. Tests pass when run against MySQL.  Does this still fix your issue?

aelkiss:
Works for me!

Hwesta:
Uh-oh, my niggling concern that this would work differently between different databases appears to be coming true.

MySQL (what we use in production) requires the length of the index to be specified, while sqlite (what we use for tests) doesn't accept that.  Given that we already found that Django doesn't create the index easily, I think this needs more consideration.

sevein:
I'm closing this issue as Holly is not actively working on this.
I've filed a new issue and this PR is connected to it - see https://github.com/archivematica/Issues/issues/75.

PR #510, opened 2016-10-05
API: unit status handles bad sub-second ordering

Hwesta:
For fast microservices & small transfers it can take less than a second to run and the createdtimedec can be set wrong. To handle this, check for existence of a matching Job, instead of assuming it's the most recent.

Add tests and fixtures for jobs.


Hwesta:
With #513 this may not need the fix, but the tests are still useful.

Hwesta:
Removed code that handled misordered jobs - this is now just tests for get_status.

sevein:
:+1:

PR #513, opened 2016-10-12
MCPServer: Fix microsecond job ordering

Hwesta:
Microseconds with a leading 0 were having the leading 0 dropped.  This meant that if several microservices ran within a second, sometimes the first ones were reordered to the end.  Use zfill to preserve leading 0.
#510 may not be needed now.


jrwdunham:
👍

Hwesta:
Also cherry-picked to stable/1.5.x as 1fbb92c5a839630fa586d827c74080bbbf46625a

PR #514, opened 2016-10-13
Requirements: Update for agentarchives & metsrw on PyPI

Hwesta:
agentarchives and metsrw are published on PyPI, so update requirements files to use that instead of Github.


sevein:
:+1:

jhsimpson:
:+1:

PR #515, opened 2016-10-13
Add PRONOM88 FPR-admin commit, fix dependencies

Hwesta:
Update django-autoslug to work with Django 1.8.

Adds artefactual/archivematica-fpr-admin#25


Hwesta:
Pushed fix for name of migration data file. Branch dev/issue-9213-pronom-88-1.6 may also be use for reviewing.

jrwdunham:
I deployed and reviewed this and it looks good to me: AM tests pass and confirmed that the new Pronom formats show up in the preservation planning tab. I have one small concern: Does it matter that django-autoslug is still 1.7.1 in setup.py of the fpr admin?

Hwesta:
> Does it matter that django-autoslug is still 1.7.1 in setup.py of the fpr admin?

Good catch - since I didn't end up using any features from the update django-autoslug, I think I'll remove that so we don't have version confusion.

jrwdunham:
👍

Hwesta:
Also cherry-picked to qa/1.6.x as 8f87b8bab3e929faf64c1ca4edd6c2f0f9ebe07f

PR #517, opened 2016-10-19
Administration: Remove hiding certain UI elements

Hwesta:
We had the ability to hide certain parts of the UI if they weren't used. However, this was implemented inconsistently, both in terms of what features could be hidden and hiding the features properly. This removes the feature, since it is straightforward to ignore UI elements unrelated to your current
task.

This also removes SettingsForm, since the reverse_checkboxes it provided were only used in hiding UI elements.

Redmine https://projects.artefactual.com/issues/10447

### DIFF
--- a/src/dashboard/src/components/accounts/views.py
+++ b/src/dashboard/src/components/accounts/views.py
@@ -28,13 +28,11 @@ from django.template import RequestContext
 from tastypie.models import ApiKey
 from components.accounts.forms import UserCreationForm
 from components.accounts.forms import UserChangeForm
-from components.helpers import hidden_features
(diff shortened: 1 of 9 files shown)


sevein:
LGTM

Hwesta:
Merged as 909bac26d77e58e32bc7888c945a5fbcfc22c19f (forgot to push this branch before merging)

sallain:
Searching suggests that this doesn't appear in the docs; however, it might appear in READMEs or other dev-side docs? @sevein, do you know?

sevein:
I don't think so, we should be fine.

PR #519, opened 2016-10-19
Added API endpoint to get completed transfers.

jrwdunham:


src/dashboard/src/components/api/views.py
```diff
@@ -89,13 +89,17 @@ def get_unit_status(unit_uuid, unit_type):
...
+            .filter(jobtype='Create SIP from transfer objects').exists():
         ret['status'] = 'COMPLETE'
         # Get SIP UUID
         sips = models.File.objects.filter(transfer_id=unit_uuid, sip__isnull=False).values('sip').distinct()
         if sips:
             ret['sip_uuid'] = sips[0]['sip']
-    elif models.Job.objects.filter(sipuuid=unit_uuid).filter(jobtype='Move transfer to backlog').exists():
+    elif models.Job.objects\
+            .filter(sipuuid=unit_uuid)\
+            .filter(jobtype='Move transfer to backlog').exists():
```
Hwesta: Nitpick, but why split this into multiple lines?

src/dashboard/src/components/api/views.py
```diff
@@ -247,6 +251,34 @@ def start_transfer_api(request):
...
+    """Return all completed transfers::
+
+        GET /api/transfer/completed?username=<am-username>&api_key=<am-api-key>
+
+    """
+    if request.method == 'GET':
+        auth_error = authenticate_request(request)
+        response = {}
+        if auth_error is None:
+            error = None
```
Hwesta: Unused

Hwesta (approved):
Looks good!  A more efficient way to get the unit status would be nice, but I'm not sure that's easily possible since there's no Django-foreign key between Jobs & Transfers.

src/dashboard/src/components/api/views.py
jrwdunham: Because otherwise it would be over 79 characters long, of course!

src/dashboard/src/components/api/views.py
jrwdunham: Good point. Will delete.

Hwesta:
If #521 gets merged first this can use the new `_api_endpoint` decorator.

src/dashboard/src/components/api/views.py
Hwesta: Since there's no semantic changes, can you leave these as one line?

Hwesta (approved):
Looks good!

src/dashboard/src/components/api/views.py
jrwdunham: Ok. Changed back.

Hwesta:
Also cherry-picked to qa/1.x as 7556cf41f1c92a8cc8f62e88a54b4dcc6d72cc88

PR #521, opened 2016-10-27
API cleanup, clarify

Hwesta:
Cleanups and fixes to the Archivematica API

Added a `_api_endpoint` decorator to handle checking HTTP verbs & auth for all endpoints and updated endpoints to use it instead. May be useful to view diff without whitespace changes (?w=1 in github or -w at command line)

Documentation for the API is now at https://wiki.archivematica.org/Archivematica_API based on this branch

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -45,6 +45,37 @@ SHARED_DIRECTORY_ROOT = helpers.get_server_config_value('sharedDirectory')
 UUID_REGEX = re.compile(r'^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$', re.IGNORECASE)
 
 
+def _api_endpoint(expected_methods):
+    """
+    Decorator for authenticated API calls that handles boilerplate code.
+
+    Checks if method is allowed, and request is authenticated.
+    """
+    def decorator(func):
+        """The decorator applied to the endpoint."""
+        def wrapper(request, *args, **kwargs):
+            """Wrapper for custom endpoints with boilerplate code."""
+            # Check HTTP verb
+            if request.method not in expected_methods:
+                return django.http.HttpResponseNotAllowed(expected_methods)  # 405
+
+            # Auth
+            auth_error = authenticate_request(request)
+            if auth_error is not None:
+                response = {'message': auth_error, 'error': True}
+                return django.http.HttpResponseForbidden(  # 403
+                    json.dumps(response),
+                    content_type='application/json'
+                )
+
+            # Call the decorated method
+            result = func(request, *args, **kwargs)
+
+            return result
+        return wrapper
+    return decorator
+
+
 def authenticate_request(request):
     error = None
 
@@ -107,18 +138,10 @@ def get_unit_status(unit_uuid, unit_type):
     return ret
 
 
+@_api_endpoint(expected_methods=['GET'])
 def status(request, unit_uuid, unit_type):
     # Example: http://127.0.0.1/api/transfer/status/?username=mike&api_key=<API key>
-    if request.method not in ('GET',):
-        return django.http.HttpResponseNotAllowed(['GET'])
-    auth_error = authenticate_request(request)
     response = {}
-    if auth_error is not None:
-        response = {'message': auth_error, 'error': True}
-        return django.http.HttpResponseForbidden(
-            json.dumps(response),
-            content_type='application/json'
-        )
     error = None
 
     # Get info about unit
@@ -132,7 +155,7 @@ def status(request, unit_uuid, unit_type):
     if unit is None:
         response['message'] = 'Cannot fetch {} with UUID {}'.format(unit_type, unit_uuid)
         response['error'] = True
-        return django.http.HttpResponseBadRequest(
+        return django.http.HttpResponseBadRequest(  # 400
             json.dumps(response),
             content_type='application/json',
         )
@@ -149,7 +172,7 @@ def status(request, unit_uuid, unit_type):
     if error is not None:
         response['message'] = error
         response['error'] = True
-        return django.http.HttpResponseServerError(
+        return django.http.HttpResponseServerError(  # 500
             json.dumps(response),
             content_type='application/json'
         )
@@ -157,23 +180,13 @@ def status(request, unit_uuid, unit_type):
     return helpers.json_response(response)
 
 
+@_api_endpoint(expected_methods=['GET'])
 def waiting_for_user_input(request):
-    # Example: http://127.0.0.1/api/transfer/waiting?username=mike&api_key=<API key>
-    if request.method not in ('GET',):
-        return django.http.HttpResponseNotAllowed(['GET'])
-
-    auth_error = authenticate_request(request)
+    # Example: http://127.0.0.1/api/ingest/waiting?username=mike&api_key=<API key>
     response = {}
-    if auth_error is not None:
-        response = {'message': auth_error, 'error': True}
-        return django.http.HttpResponseForbidden(
-            json.dumps(response),
-            content_type='application/json'
-        )
-
-    error = None
     waiting_units = []
 
+    # TODO should this filter based on unit type into transfer vs SIP?
     jobs = models.Job.objects.filter(currentstep='Awaiting decision')
     for job in jobs:
         unit_uuid = job.sipuuid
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -157,23 +182,14 @@ def status(request, unit_uuid, unit_type):
     return helpers.json_response(response)
 
 
+@_api_endpoint(expected_methods=['GET'])
 def waiting_for_user_input(request):
```
Hwesta, line 184: Despite the url (`/api/ingest/waiting`) this returns all units, SIP or Transfer, waiting for user input. Should we add a transfer vs ingest endpoint and limit results?

src/dashboard/src/components/api/views.py
```diff
@@ -541,6 +522,8 @@ def fetch_levels_of_description_from_atom(request):
     else:
         return get_levels_of_description(request)
 
+# TODO should this have auth?
+@_api_endpoint(expected_methods=['GET', 'POST'])
 def path_metadata(request):
```
Hwesta, line 551: This had no auth or verb restriction previously - needs to be tested to see if it still works with external consumers.

Also, it should return 'message': status and possibly 'error': False to match the rest of the API.

src/dashboard/src/components/api/views.py
```diff
@@ -582,6 +565,8 @@ def path_metadata(request):
         return helpers.json_response(body, status_code=201)
 
 
+# TODO should this have auth?
+@_api_endpoint(expected_methods=['GET'])
 def processing_configuration(request, name):
```
Hwesta, line 594: This had no auth or verb restriction previously - needs to be tested to see if it still works with external consumers.

This returns an XML document as the body instead of JSON, which is different from the rest of the API. Should this be changed?

src/dashboard/src/components/api/views.py
```diff
@@ -519,6 +498,8 @@ def get_levels_of_description(request):
     response = [{l.id: l.name} for l in levels]
     return helpers.json_response(response)
 
+# TODO should this have auth?
+@_api_endpoint(expected_methods=['GET'])
 def fetch_levels_of_description_from_atom(request):
```
Hwesta, line 527: This had no auth or verb restriction previously - needs to be tested to see if it still works with external consumers.

src/dashboard/src/components/api/views.py
```diff
@@ -508,6 +485,8 @@ def copy_metadata_files_api(request):
     paths = request.POST.getlist('source_paths[]')
     return filesystem_ajax_views.copy_metadata_files(sip_uuid, paths)
 
+# TODO should this have auth?
+@_api_endpoint(expected_methods=['GET'])
 def get_levels_of_description(request):
```
Hwesta, line 514: This had no auth or verb restriction previously - needs to be tested to see if it still works with external consumers.

Also, this returns a list, not a json (with eg a 'results' list) - should it be changed for consistency?

src/dashboard/src/components/api/views.py
```diff
@@ -500,6 +475,8 @@ def reingest(request, target):
         response = {'message': 'Approval successful.', 'reingest_uuid': reingest_uuid}
         return helpers.json_response(response)
 
+# TODO should this have auth?  Does it support GET?
+@_api_endpoint(expected_methods=['POST'])
 def copy_metadata_files_api(request):
```
Hwesta, line 504: This had no auth or verb restriction previously - needs to be tested to see if it still works with external consumers.

src/dashboard/src/components/api/views.py
```diff
@@ -149,31 +174,22 @@ def status(request, unit_uuid, unit_type):
...
+    # Example: http://127.0.0.1/api/ingest/waiting?username=mike&api_key=<API key>
     response = {}
-    if auth_error is not None:
-        response = {'message': auth_error, 'error': True}
-        return django.http.HttpResponseForbidden(
-            json.dumps(response),
-            content_type='application/json'
-        )
-
     error = None
```
jrwdunham: This is unrelated to your specific changes, but I'm not seeing how ``error`` could ever be modified between lines 174-192. Therefore, isn't the entire `if error is not None` block at 193 unnecessary?

src/dashboard/src/components/api/views.py
jrwdunham: Yes. Though something might depend on this "get all waiting units" endpoint, no? Why not provide 3:
1. /api/unit/waiting
2. /api/ingest/waiting
3. /api/transfer/waiting

src/dashboard/src/components/api/views.py
```diff
@@ -309,103 +290,84 @@ def _completed_units(unit_type='transfer'):
...
-            response['error'] = True
-            return helpers.json_response(response, status_code=403)
+    if error is not None:
+        response['message'] = error
+        response['error'] = True
     else:
-        return django.http.HttpResponseNotAllowed(permitted_methods=['GET'])
+        response['message'] = 'Fetched unapproved transfers successfully.'
+
+        if error is not None:
```
jrwdunham: Again, how is it possible for `error` to ever be changed from `None`? Also, why are we checking `error` twice if there is no possibility of it changing after the first check? Lastly, if the `if error is not None` block stays, I think it needs to return the response, right?

src/dashboard/src/components/api/views.py
```diff
@@ -45,6 +45,39 @@
 UUID_REGEX = re.compile(r'^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$', re.IGNORECASE)
 
 
+def _api_endpoint(expected_methods, required_fields=[]):
```
jrwdunham: I don't think it's a problem in this case but lists as default values is generally a bad idea, right. Also, this `required_fields` argument is never used, so why have it?

src/dashboard/src/components/api/views.py
jrwdunham: Maybe the return value could be sensitive to the request's content-type. I see that storage_service/locations/models/pipeline.py expects the response body to be an XML string, so that would have to change if the default return value is JSON (containing the XML in an object attribute, I assume)

src/dashboard/src/components/api/views.py
jrwdunham: ... which would require changes to dashboard/src/media/js/ingest/directory-metadata-form-dialog.js

src/dashboard/src/components/api/views.py
jrwdunham: If so (for posterity), changes to dashboard/src/media/js/ingest/directory-metadata-form-dialog.js would be required.

jrwdunham (commented):
Overall looks good: I like the removal of repetitive code. A couple of questions and comments.

src/dashboard/src/components/api/views.py
```diff
@@ -250,48 +248,31 @@ def start_transfer_api(request):
...
-        else:
-            response['message'] = auth_error
-            response['error'] = True
-            return helpers.json_response(response, status_code=403)
-    else:
-        return django.http.HttpResponseNotAllowed(permitted_methods=['GET'])
+    response = {}
+    completed = _completed_units(unit_type='ingest')
+    response['results'] = completed
+    response['message'] = 'Fetched completed ingests successfully.'
```
jrwdunham, line 264: Doesn't this need a return value?

src/dashboard/src/components/api/views.py
Hwesta: I wonder how often callers want info about both transfers & SIPs? Does it make sense to have versions of many of these endpoints as unit-agnostic?  We don't actually have an `/api/unit/` or `/unit/` set of endpoints right now - the `unit` component is just functionality common between ingest & transfer that's offered at both `/transfer/` and `/ingest/`

Given time constraints right now though, I'm inclined to document the current behaviour and fix it later.

src/dashboard/src/components/api/views.py
Hwesta: Good catch - lists can be problematic if they're being modified instead of reassigned inside a function, though that isn't the case here.  Since we didn't end up using the required fields functionality, I'll remove the stub of it.

src/dashboard/src/components/api/views.py
Hwesta: You're right - I'll removed the unneeded code.

src/dashboard/src/components/api/views.py
Hwesta: Oops, yes.  Proof that this section needs tests too.

src/dashboard/src/components/api/views.py
Hwesta: FIxed.

src/dashboard/src/components/api/views.py
Hwesta: I think it makes sense to leave the odd return types as they are for now. As nice as a consistent set of return bodies is, this stage is more about making it code-consistent and documenting current behaviour. And for this endpoint in particular, it doesn't make sense to return JSON containing an XML body.

Hwesta (commented):
Updated based on feedback, I'll rebase the fixup commits before merging.

Overall, I think it makes more sense to accept the current inconsistencies and document them rather than fix them right now. Fixes make more sense if we have a versioned API as well.

PR #523, opened 2016-10-31
remove empty bulkextractor logs

aelkiss:
Submitting this PR at the recommendation of @eckardm -- this partially addresses https://github.com/mlibrary/archivematica/issues/2 by drastically cutting down on the number of files that need to be index by removing any empty BulkExtractor logs. The generated `report.xml` still lists any scanners that had no results, so there is still evidence that BulkExtractor ran the scanners.

### DIFF
--- a/src/MCPClient/lib/clientScripts/examineContents.py
+++ b/src/MCPClient/lib/clientScripts/examineContents.py
@@ -12,11 +12,11 @@ def main(target, output):
     try:
         os.makedirs(output)
         subprocess.call(args)
-	# remove empty BulkExtractor logs
-	for filename in os.listdir(output):
-	    filepath = os.path.join(path,filename)
-	    if os.path.getsize(filepath) == 0:       
-		os.remove(filepath)
+        # remove empty BulkExtractor logs
+        for filename in os.listdir(output):
+            filepath = os.path.join(output,filename)
+            if os.path.getsize(filepath) == 0:
+                os.remove(filepath)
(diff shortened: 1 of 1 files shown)


Hwesta (approved):
Looks good, thank you!

src/MCPClient/lib/clientScripts/examineContents.py
```diff
@@ -12,6 +12,11 @@ def main(target, output):
     try:
         os.makedirs(output)
         subprocess.call(args)
+	# remove empty BulkExtractor logs
+	for filename in os.listdir(output):
+	    filepath = os.path.join(path,filename)
```
Hwesta: Looks like this should be `filepath = os.path.join(output, filename)` since `path` is undefined.

Hwesta (changes requested):
Sorry, I took a closer look at this and it has whitespace problems, which are meaningful in Python.  Can you convert all tabs to spaces, and strip trailing whitespace?

aelkiss:
Sorry, I missed this ... fix is pushed now.

Hwesta (approved):

Hwesta:
Looks good now, thank you!

Hwesta:
Committed to qa/1.6.x as 7d02cbf5e449de0d03bf85da1aeb2d10890f746e and qa/1.x as 460ed562a64b7aacbd690a2e62f39b11bfcb092f

PR #524, opened 2016-11-02
Templates: Hide 'show files' in backlog search

Hwesta:
Show files was added for the backlog tab, but since the code is reused in the ingest tab it also appeared there, without being functional.  Update the template to hide the non-functional checkbox.

refs Redmine #10502


sevein:
👍

PR #527, opened 2016-11-09
Transfer browser: Fix display of disk image metadata icon 

Hwesta:
Remove CSS hiding the disk image metadata icon in the transfer browser and fix bug in transfer browser so it shows up for disk image transfers.

Includes artefactual-labs/transfer_browser#11  I'm not sure if the webpack is generated correctly - it seems to include different base64 encoding helper stuff.


sevein:
👍

PR #528, opened 2016-11-09
Models: Update ArchivesSpaceDOComponent to Digital Object 

Hwesta:
The appraisal tab workflow changed, and no longer has digital object components as children of digital objects as children of archival objects. SIPs are now created directly from digital objects. Remove extra structure used by digital object components and rename model.

This PR contains two possible solutions to this problem.

1. Add a migration that renames the model
2. Update the migration that originally introduced ArchivesSpaceDOComponent with the new name.

Normally option 1 is the correct solution. However, the archivematica user doesn't have DROP table permissions, so renaming the model fails (as does renaming the model but setting the db_table to the old name).  Since the first migration hasn't been released yet, it might be possible to just alter it, which will work for fresh installs but not upgrades.  I'm not sure if this is an acceptable solution - thoughts?


sevein:
We don't have to face this problem now but I suspect that we'll need `DROP` in the future. Maybe we should have an extra settings file that uses different database credentials. Maybe we can unset `SECRET_KEY` to make sure that it's not used in the web context, Django commands should be able to function properly without it.

> Since the first migration hasn't been released yet, it might be possible to just alter it, which will work for fresh installs but not upgrades. I'm not sure if this is an acceptable solution - thoughts?

We never do that in AtoM because we sometimes have users running of our development branches. I'm not sure if that's the case in Archivematica, probably not?

Hwesta:
>  Maybe we should have an extra settings file that uses different database credentials.

Use a different settings file that knows about a different DB user that has more permissions when doing upgrades?  It's a possibility. Django [won't start](https://docs.djangoproject.com/en/1.8/ref/settings/#secret-key) if `SECRET_KEY` isn't set though 

> We never do that in AtoM because we sometimes have users running of our development branches. I'm not sure if that's the case in Archivematica, probably not?

We try to keep production deploys on release branches (or sometimes client branches based off of release branches). I know there's deploys with the old migration, but I don't know if they're production, and how much we care about supporting non-production deploys. @jhsimpson might know.  On the other hand, qa/1.x is supposed to be latest-stable which implies not breaking things too much, which is why I'm hesitant.

Hwesta:
Discussion says that we don't have any significant deploys affected by this, and we're not concerned with upgrading from development to development (even if that's qa/1.x).  Open to discussion on this if someone feels strongly.

sevein:
:+1:

PR #529, opened 2016-11-09
Dashboard: remove dead code

sevein:
Stuff that I've found that is unused and we can get rid of safely:

1. Function `components.administration.views.dips_formset` and underlying form `main.forms.MicroServiceChoiceReplacementDicForm`
2. In `js/{job,ingest,transfer}.js` backbone app:
    - Unused `status` property of `Job` model
    - Buttons approve job, reject job, browse job
      - Its images when not used by others and CSS
    - The job browser (named as directory-browser) and its CSS 
      - Corresponding endpoint `/jobs/explore/` (entry in `urls.py` and view)
      - Corresponding endpoint `/jobs/list-objects/` (entry in `urls.py` and view)

I think **`2)`** is mostly related to the fact that we replaced the old job [approve/reject] buttons with the more sophisticated drop-down of choices, but some of the old stuff was left behind. This old screenshot may help to clarify what I am talking about:

screenshot


src/dashboard/src/templates/transfer/grid.html
```diff
@@ -131,80 +131,11 @@
...
-      <span class="transfer_path_icons">
-        <span class="transfer_path_edit_icon"><img src="{{STATIC_URL}}images/table_edit.png" /></span>
-        <span class="transfer_path_delete_icon"><img src="{{STATIC_URL}}images/delete.png" /></span>
-      </span>
-    </div>
-  </script>
-
-  <!-- File explorer templates -->
-
-  <script type="text/template" id='template-dir-level'>
```
sevein, line 196: Used in `js/transfer/component_directory_select.js`?

src/dashboard/src/templates/transfer/grid.html
```diff
@@ -131,80 +131,11 @@
...
-        </form>
-      </div>
-      <div id="transfer_paths_area">
-        <%= transfer_paths %>
-      </div>
-    </div>
-    <br clear='all'/>
-  </script>
-
-  <script type="text/template" id='transfer-component-path-item'>
```
sevein, line 184: Used in `js/transfer/component_form.js`?

src/dashboard/src/templates/transfer/grid.html
```diff
@@ -131,80 +131,11 @@
...
-    </div>
-  </script>
-
-  <!-- File explorer templates -->
-
-  <script type="text/template" id='template-dir-level'>
-    <div class="backbone-file-explorer-level"></div>
-  </script>
-
-  <script type="text/template" id='template-dir-entry'>
```
sevein, line 200: Used in `js/transfer/component_directory_select.js`?

Hwesta:
> Used in `js/transfer/component_directory_select.js`? 

`component_directory_select.js` is include two places: `transfer/grid.html` & `ingest/metadata_add_files.html`.  No template extends or includes `transfer/grid.html`, so it can't be used in other places. `component_directory_select.js` was only used in the transfer tab for the old transfer browser, which is now replaced, so I think all templates & JS related to that can be deleted. `metadata_add_files.html` defines its own templates for the `component_directory_select.js`.

> Used in `js/transfer/component_form.js`?

Looks like `component_form.js` was only used in the old transfer browser, and can be deleted completely.

Following the above logic, I removed a bunch of old transfer-browser code that was only still in use in the add-metadata-files template now.  Let me know if you find any issues.

Hwesta:
:+1: on your commits

sevein:
So can I merge? Just want to double check, not sure if `qa/1.6.x` is in practice being used.

Hwesta:
1.6 is code freeze, so this would go to qa/1.x (for 1.7)

If you think my commits are good & don't find anything they break, then it's good to merged.

PR #531, opened 2016-11-11
Requests: attach timeout to avoid hanging forever

sevein:
Redmine: https://projects.artefactual.com/issues/10575

See http://docs.python-requests.org/en/latest/user/advanced/#timeouts:
> Most requests to external servers should have a timeout attached, in case the
> server is not responding in a timely manner. By default, requests do not time
> out unless a timeout value is set explicitly. Without a timeout, your code
> may hang for minutes or more.

This is particularly important in the Dashboard when it's deployed using synchronous workers because these workers can process only one request at a time.

Try for example to fetch the level of descriptions in the AtoM settings. If the AtoM server is not available your request will hang for a long time (in the order of minutes). With this change, it will hang only for five seconds (probably too generous, but much better) and return an error if AtoM didn't make it in time. This example is particularly annoying because [the XHR call is not asynchronous](https://github.com/artefactual/archivematica/blob/stable/1.5.x/src/dashboard/src/media/js/administration/atom_levels_of_description.js#L11), completely blocking the whole browser tab, i.e. the user interface.

### DIFF
--- a/src/MCPClient/lib/clientScripts/post_store_aip_hook.py
+++ b/src/MCPClient/lib/clientScripts/post_store_aip_hook.py
@@ -53,7 +53,7 @@ def dspace_handle_to_archivesspace(sip_uuid):
     url = archivesspace_url + '/users/' + config.user + '/login'
     params = {'password': config.passwd}
     logger.debug('Log in to ArchivesSpace URL: %s', url)
-    response = requests.post(url, params=params)
+    response = requests.post(url, params=params, timeout=60)
     logger.debug('Response: %s %s', response, response.content)
     session_id = response.json()['session']
     headers = {'X-ArchivesSpace-Session': session_id}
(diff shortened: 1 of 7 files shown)


sevein:
@Hwesta, I've seen your attempt to get rid of slumber. This PR makes a change in our module using it in order to support timeouts just in case this gets merged first and slumber sticks for a while.

sevein:
IMO this is a great candidate for 1.6 but feel free to change.

Hwesta (approved):
Looks good! The session creation code should still be useful sans slumber.

I'm not sure if we're still accepting commits to AM 1.6 - I'll mention it at the release planning meeting.

Hwesta:
This should be discussed further since @jhsimpson mentioned that e.g. ArchivesSpace can take longer than 5 seconds to respond.

sevein:
I can give 10 seconds to that one, ok?

sevein:
After discussing it with @jhsimpson, I've increased the timeouts in AS and merged.

Hwesta:
For the storage service, a timeout of 5 isn't always long enough. I ran into issues starting a transfer that was ~600MB from the same disk. Increasing the timeout from 5 to 30 seconds form `storage_service.copy_files` fixed it.

PR #534, opened 2016-11-18
housekeeping

retokromer:
- "modern" shebang
- uniformed indenting to 4
- $(...) preferred over \`...\`
- = preferred over == inside [[...]]
- ${VAR} preferred over $VAR

### DIFF
--- a/dev-installer
+++ b/dev-installer
@@ -26,7 +26,7 @@ cd $(dirname "${0}")
 part="create the archivematica user"
 echo -n "\"Would you like to ${part}?\" (y/N) "
 read a
-if [[ $a = "Y" || $a = "y" ]]; then
+if [[ "${a}" = [Yy] ]]; then
     echo "Going to ${part} ..."
     sudo adduser --uid 333 --group --system --home /var/lib/archivematica/ archivematica
     sudo adduser --uid 334 --group --system --home /var/lib/archivematica-django/ archivematicadashboard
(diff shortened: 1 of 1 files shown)


dev-installer
```diff
@@ -21,126 +21,121 @@
 # @author Joseph Perry <joseph@artefactual.com>
 # @version svn: $Id$
 
-cd `dirname $0`
+cd $(dirname "${0}")
 
 part="create the archivematica user"
 echo -n "\"Would you like to ${part}?\" (y/N) "
 read a
-if [[ $a == "Y" || $a == "y" ]]; then
```
Hwesta, line 29: Why is `=` preferred over `==` inside `[[...]]`? I couldn't find an explanation via Google.

Hwesta (commented):
This looks good, thanks for the contribution! Can you sign a [contributor's agreement](https://wiki.archivematica.org/images/2/25/Contributor_agreement.txt) and email it to agreement@artefactual.com ? Thank you!

By the way, `dev-installer` and `dev-helper` are considered deprecated - we recommend using [Ansible and Vagrant](https://wiki.archivematica.org/Getting_started#Installation) for development or a [packge install](https://www.archivematica.org/docs/latest/admin-manual/installation/installation/). We're also interested if you have feedback on making the documentation clearer around this.

dev-installer
retokromer: I’ll send you the technical reference as soon as back home.

retokromer: In `[[ ... ]]` there is no reason to use `==`. However, if you prefer, I can undo.

Hwesta:
Cherry-picked to qa/1.x as 379ad89212ba0f7d5ea6968ec13c659c77e7fb84 and 2f9b602abd0ce4be042892ec084dc070c2aaeecf  Thanks for the cleanup!

PR #538, opened 2016-11-25
MCPClient: Fix fetching FPCommand in extract

Hwesta:
Filtering on a FPRule uses all Rules, not just the active ones, because the filter doesn't use the 'active' manager. Update to explicitly filter on enabled FPR rules.

Refs Redmine https://projects.artefactual.com/issues/10604


sevein:
Makes sense! :+1:

Hwesta:
Cherry-picked to qa/1.x as 63ea6343b1ee8f7ca93cab3773389475d119215f

PR #539, opened 2016-12-04
Add coverage requirement / config

helenst:
Thinking about Python 3 compatibility, I wanted to see what coverage was like so set up the pytest-cov plugin. (It's reporting 42%)

At the moment, it can just be run manually (`py.test --cov`) but could be part of a future CI.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+omit = **/migrations/*,**/settings/*,**/wsgi.py,**/tests/*,**/manage.py
+source = .
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,16 @@
 [run]
-omit = **/migrations/*,**/settings/*,**/wsgi.py,**/tests/*,**/manage.py
-source = .
+
+omit =
+	src/archivematicaCommon/lib/externals/
+	**/migrations/*
+	**/settings/*
+	**/wsgi.py
+	**/tests/*
+	**/manage.py
+
+source =
+	src/archivematicaCommon/lib/
+	src/dashboard/src/
+	src/MCPClient/lib/
+	src/MCPClient/lib/clientScripts/
+	src/MCPServer/lib/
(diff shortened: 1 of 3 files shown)


.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+omit = **/migrations/*,**/settings/*,**/wsgi.py,**/tests/*,**/manage.py
+source = .
```
Hwesta: Coverage only looks at files that are at the root of the specified directory or in a module (with `__init__.py`) and much of Archivematica is set up as a collection of scripts and isn't seen by coverage.  I think this will be more accurate:

```
source =
    src/dashboard/src/
    src/archivematicaCommon/lib/
    src/MCPClient/lib/
    src/MCPClient/lib/clientScripts/
    src/MCPServer/lib/
```

Sadly this brings test coverage down to 25%

.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+omit = **/migrations/*,**/settings/*,**/wsgi.py,**/tests/*,**/manage.py
```
Hwesta: We probably want to omit `src/archivematicaCommon/lib/externals/` (but not `src/dashboard/src/external/archivematica-fpr-admin/` though it may be irrelevant since the fpr-admin will be installed as a proper dependency once the translations work #506 comes in)

Hwesta (commented):
Looks good!  Unfortunately I think it's missing part of Archivematica because many sections aren't technically python modules and coverage doesn't see them.

Do you know of any disadvantages to setting `branch = True`?

.coveragerc
helenst: Ah, that's worth knowing, thank you. Branch is a good idea, unfortunately it reduces coverage further to 22% :disappointed:

Hwesta:
Squash-committed to qa/1.x as 471abef7802608f5cad46306821c2f7d5becf1a7  Thank you!

PR #540, opened 2016-12-05
Backlog removal event: removed objects/ filterSubDir

jrwdunham:
The "Create removal from backlog PREMIS events" micro-service was, prior to this PR, ignoring all files not in a root objects/ directory of the transfer. When you re-arrange a backlogged transfer, chances are you are removing that condition, i.e., you are creating a SIP without a root objects/ directory. The solution here implemented is simply to remove the condition that in order to receive a "removal from backlog" event you must be somewhere in objects/.

### DIFF
--- a/src/MCPClient/lib/clientScripts/createEvent.py
+++ b/src/MCPClient/lib/clientScripts/createEvent.py
@@ -46,10 +46,19 @@ if __name__ == '__main__':
 
     (opts, args) = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


Hwesta (approved):
Looks good

jrwdunham:
Cherry-picked to qa/1.6.x as 6afb0f30d4d888eb72e5e7615a7ec3aa03059c9f

PR #542, opened 2016-12-09
Updated transfer browser to sort directories before files

jrwdunham:


sevein:
😢  - 50,835 changes, I think we need to fix this sooner rather than later or our git repo is going to become an unsustainable beast.

Hwesta:
I find that if I make sure my npm dependencies are up to date (and presumably the appraisal tab & transfer browser webpacks were generated with similarly up to date dependencies) then the diff is quite small, and I can usually see what changes I actually made.  Of course, if the npm dependencies change, then it can still have a large diff...

sevein:
Well that's always going to be a problem in the front-end because deps are constantly changing, unless we start hard pinning dependencies. The other thing is that both appraisal tab and transfer browser should be able to share a bunch of code. bower is calling!

jrwdunham:
Oops. This is partially a bigger mistake on my end: appraisal_tab.js should have been untouched. I'll fix that and I'll update my npm stuff and re-create the PR.

jrwdunham:
Ok, now there are "only" 6,512 changes. I did the following:

    $ cd transfer_browser
    $ npm update
    $ npm install
    $ cp app/transfer_browser.js /path/to/archivematica/src/dashboard/src/media/js/

Should I have done something different? Do we have our JS dependency versions at last build time documented somewhere?

Hwesta:
Looking at the diff I see Angular 5.8 vs 6.0 - I suspect that's the source of a bunch of the noise.  I don't think we have the currently-installed version listed anywhere yet.

Was there a transfer-browser PR for this? I can't seem to find it...

jrwdunham:
@Hwesta @sevein Ok, the diff should be more manageable now. Cf. https://github.com/artefactual-labs/archivematica-browse-helpers/pull/2

sevein:
👍

jrwdunham:
Cherry-picked to qa/1.6.x as c435ce80fdc2f067cdd13b2d1385afaf1457cf38

PR #543, opened 2016-12-10
Rights: Handle cascade delete of RightsStatementRightsGranted

Hwesta:
When a RightsStatement is deleted, it cascade deletes anything with a foreign key to it. However, RightsStatementRightsGranted checks to see if its RightsStatement had any other RightsGranted, so it can delete Statements with no Rights.  When the RightsStatement had already been deleted, this raised an exception. Also, add tests.

Refs redmine 9938


sevein:
Clear! 👍

PR #545, opened 2016-12-13
UTF8-encode db paths when matching to fsystem paths

jrwdunham:
Allows transfers containing files with non-ASCII characters in their names.

### DIFF
--- a/src/MCPServer/lib/unit.py
+++ b/src/MCPServer/lib/unit.py
@@ -25,6 +25,8 @@ import logging
 import os
 import sys
 
+import archivematicaFunctions
+
 import archivematicaMCP
 from unitFile import unitFile
 
@@ -60,9 +62,10 @@ class unit:
             else:
                 files = File.objects.filter(sip_id=self.UUID)
             for f in files:
-                if f.currentlocation in self.fileList:
-                    self.fileList[f.currentlocation].UUID = f.uuid
-                    self.fileList[f.currentlocation].fileGrpUse = f.filegrpuse
+                currentlocation = archivematicaFunctions.unicodeToStr(f.currentlocation)
+                if currentlocation in self.fileList:
+                    self.fileList[currentlocation].UUID = f.uuid
+                    self.fileList[currentlocation].fileGrpUse = f.filegrpuse
                 else:
                     LOGGER.warning('%s %s has file (%s) %s in the database, but file does not exist in the file system',
                         self.unitType, self.UUID, f.uuid, f.currentlocation)


src/MCPServer/lib/unit.py
```diff
@@ -60,12 +60,20 @@ def reloadFileList(self):
...
                 files = File.objects.filter(sip_id=self.UUID)
             for f in files:
-                if f.currentlocation in self.fileList:
-                    self.fileList[f.currentlocation].UUID = f.uuid
-                    self.fileList[f.currentlocation].fileGrpUse = f.filegrpuse
+                try:
+                    file_unit = self.fileList.get(
+                        f.currentlocation,
+                        self.fileList.get(f.currentlocation.encode('utf8')))
+                except UnicodeError:
```
sevein: Is it worth also including `KeyError` here?

jrwdunham: I don't think it's possible for `KeyError` to be triggered here. `dict.get(key)` returns `None` if `key` is not in `dict` and `dict.get(key, dflt)` returns `dflt`.

sevein: Ahhh you're right! I keep making that mistake.

src/MCPServer/lib/unit.py
```diff
@@ -60,12 +60,20 @@ def reloadFileList(self):
             else:
                 files = File.objects.filter(sip_id=self.UUID)
             for f in files:
-                if f.currentlocation in self.fileList:
-                    self.fileList[f.currentlocation].UUID = f.uuid
-                    self.fileList[f.currentlocation].fileGrpUse = f.filegrpuse
+                try:
+                    file_unit = self.fileList.get(
+                        f.currentlocation,
+                        self.fileList.get(f.currentlocation.encode('utf8')))
```
Hwesta: Should this use the `archivematicaFunctions.unicodeToStr` ?

Hwesta (commented):
This is fairly similar to [this commit](https://github.com/artefactual/archivematica/pull/500/commits/47eeaafba5963bf50a3cc3efeab86b45024e0173). Does that seem like a simpler approach?

jrwdunham:
@Hwesta yes, PR 500 fixes this issue too. I think we want this fix on 1.6 but that PR is deferred to 1.7. What do you suggest? Replace this PR's commit with 47eeaafba5963bf50a3cc3efeab86b45024e0173?

Hwesta:
#500 was a collection of fixes that didn't have bugs, and I made the PR so the potentially useful fixes wouldn't get lost. Since both commits fix the issue, pick whichever one you think is better.  :+1: for this one.

jrwdunham:
Cherry-picked to qa/1.6.x as 358ecc84a2fb31f49b8e0db6c519317eae1785ca

PR #546, opened 2016-12-16
Reingest: Handle directory names

Hwesta:
When checking if a path has a UUID, only consider the basename, and handle directories by normalizing the path.

Requires artefactual/archivematica-storage-service#168

refs redmine 10080


jhsimpson:
I tested this and it is working.  

:+1:

jhsimpson:
:+1:  looks good.  tested with ss#168 as well.

Hwesta:
Cherry-picked to qa/1.x as 61a183ab81a60d54a254df53b45075425ceb7d61

PR #547, opened 2016-12-21
Code review: Initial draft of code review guidelines

Hwesta:
The previous contents of code_review.md now live in the CONTRTIBUTING.md.

Requesting feedback on what should or shouldn't be in the code review guidelines.

Do any of the existing questions need a longer description?

### DIFF
--- a/code_review.md
+++ b/code_review.md
@@ -4,109 +4,54 @@ Code review
 Every new feature and bugfix to a project is expected to go through code review before inclusion.
 This applies both to developers at Artefactual and to outside contributors.
 
-Here's an outline of the code review process:
+The contribution process in general is outlined in the [contributing document](CONTRIBUTING.md).
+This provides a checklist and set of guidelines for what to do when doing code review.
+It also helps set expectations for contributors about what to expect.
 
-1. Fork the Artefactual project on GitHub, and commit your changes to a branch.
-2. Open a pull request.
-3. Back and forth discussion with developers on the branch.
-4. Make any changes suggested by reviewers.
-5. Repeat 3 and 4 as necessary.
-6. Clean up commit history, if necessary.
-7. Your branch will be merged!
+Checklist
+---------
 
-Getting started
----------------
+### Meta
+- [ ] Have they submitted a signed contributor's agreement?
+- [ ] Does this meet the requirements? Does it fix the bug?
+- [ ] Is this feature useful to multiple users or potential users?
+- [ ] Have the changes been discussed with an analyst?
+- [ ] Is this one logical piece of work? Does it make sense to be split into multiple PRs for each sub-feature?
+- [ ] Is there documentation?  This includes docstrings, comments & potentially user-facing documentation.
+- [ ] Is the commit history clean?
 
-So you have something to contribute to an Artefactual project. Great!
+### Architecture
+- [ ] Does this duplicate functionality found elsewhere?
+- [ ] Does it make more sense for this to be implemented elsewhere? E.g. Should this happen in the storage service vs a client script, or should a helper function live in archivematicaCommon?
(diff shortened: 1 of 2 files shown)


code_review.md
```diff
@@ -4,109 +4,53 @@ Code review
...
+### Style
+- [ ] Does it follow [PEP8](https://www.python.org/dev/peps/pep-0008/) (code style) and [PEP257](https://www.python.org/dev/peps/pep-0257/) (docstrings)?
+- [ ] Do docstrings list the parameters and behaviour?
+- [ ] Are the names accurate and sensible?
+- [ ] Are there massive functions, classes or files? Can they be split?
+- [ ] Are errors handled?  Is the error handling consistent with similar code?
+
+### Syntax:
+- [ ] Is it Python 2.7 & Python 3 compatible?
+- [ ] Are all user-facing strings marked for translation?
```
sevein, line 42: This may lead to confusion because we haven't merged the i18n work yet but it should be happening very soon so I would just leave it! Thanks.

sevein (approved):
I like the changes. It can be a great help.

code_review.md
```diff
@@ -4,109 +4,53 @@ Code review
...
-While there's no need to rehash anything that code comments already say, if there's more detail that helps a reader understand *why* a change was made, be as verbose as you need to!
-Remember: future-you (or another developer) will read this when going through the commit history to understand why a change was made.
-Make their life easier.
+The contribution process in general is outlined in the [contributing document](CONTRIBUTING.md).
+This provides a checklist and set of guidelines for what to do when doing code review.
+It also helps set expectations for contributors about what to expect.
+
+Checklist
+---------
+
```
sromkey, line 13: Is the checklist intended to be prescriptive or a list of suggestions? It might be helpful to clarify.

code_review.md
```diff
@@ -4,109 +4,53 @@ Code review
...
+- [ ] Are there tests?
+- [ ] Do the tests cover all the new functionality or fixes?
+- [ ] Do the tests handle error cases?
+- [ ] Is unit testing or integration testing more appropriate?
+- [ ] Does this work with non-ASCII?
+
+### Doing
+- [ ] Do I understand what this is supposed to do?
+- [ ] Have I checked out and run this code to verify it does what it says & is supposed to?
+- [ ] Is someone else an expert in this area? Should I ask them to also review this?
```
sromkey, line 55: Could I suggest that we add to this, or in a new checkbox, "Have I discussed new functionality or changes to existing functionality with an analyst?" or similar

sromkey:
I made a couple of small suggestions, but overall I agree with @sevein , looks good and I think it will be a big help.

code_review.md
Hwesta: Hmm, I tried to cover a bunch of that in the Meta section (Does it meet the reqs? Is it useful to multiple users?) but an explicit statement is a good idea.

code_review.md
Hwesta: Suggestions, expectations management, and mental structure.  It can be hard to remember the big picture things, as well as project reqs unrelated to the feature (eg translation) and this is a reference. Does "This provides a checklist and set of guidelines for what to do when doing code review. It also helps set expectations for contributors about what to expect." not adequately explain that?

sromkey: You're right, describing this as a "guideline" is clear enough." thanks!

code_review.md
sromkey: Yeah, I wouldn't mind an explicit statement to actually communicate with an archivist/analyst. Either here or in the Meta section.

Hwesta: Added a line in the Meta section "Have the changes been discussed with an analyst?" so I think this is ready to go

PR #548, opened 2016-12-24
Reingest: keep deleted preservation derivatives in METS

Hwesta:
This reverts commit 94c2238f9b493f19c7181f3129f461fc28ce2b58 from #391 

Keep deleted preservation derivatives in the METS file.

refs redmine 10713


jrwdunham:
Confirmed that with this change a re-ingested AIP's METs has a <mets:fileGrp USE="deleted"> element while, without this change, there is no such element. 👍

Hwesta:
Also cherry-picked to qa/1.x as 192391c6ab42706cf26dcca957dfbeab0cfe2e02

PR #550, opened 2017-01-03
Remove 'slumber' dependency

Hwesta:
Remove `slumber` as a dependency.  Slumber is a wrapper around `requests`, but doesn't provide much additional use, is less standard in it's API than stock requests, and adds another dependency to keep up to date.  Removing to help make Python 3 support easier as well.

### DIFF
--- a/src/MCPClient/lib/clientScripts/failedSIPCleanup.py
+++ b/src/MCPClient/lib/clientScripts/failedSIPCleanup.py
@@ -22,9 +22,10 @@ def main(fail_type, sip_uuid):
     models.SIPArrange.objects.filter(sip_id=sip_uuid).delete()
 
     # Update storage service that reingest failed
-    api = storage_service._storage_api()
+    session = storage_service._storage_api_session()
+    url = storage_service._storage_service_url() + 'file/' + sip_uuid + '/'
     try:
-        api.file(sip_uuid).patch({'reingest': None})
+        session.patch(url, json={'reingest': None})
     except Exception:
         # Ignore errors, as this may not be reingest
         pass
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -50,11 +50,11 @@ def _storage_service_url():
     LOGGER.debug("Storage service URL: {}".format(storage_service_url))
     return storage_service_url
 
-def _storage_api_session(auth=None, timeout=None):
+def _storage_api_session(timeout=5):
     """ Returns a requests.Session with a customized adapter with timeout support. """
 
     class HTTPAdapterWithTimeout(requests.adapters.HTTPAdapter):
-        def __init__(self, timeout=5, *args, **kwargs):
+        def __init__(self, timeout=None, *args, **kwargs):
             self.timeout = timeout
             super(HTTPAdapterWithTimeout, self).__init__(*args, **kwargs)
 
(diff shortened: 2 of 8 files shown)


src/archivematicaCommon/lib/storageService.py
```diff
@@ -393,58 +354,60 @@ def request_reingest(package_uuid, reingest_type, processing_config):
...
 
 def index_backlogged_transfer_contents(transfer_uuid, file_set):
-    api = _storage_api()
-    try:
-        api.file(transfer_uuid).contents.put(file_set)
-    except slumber.exceptions.HttpClientError as e:
-        raise BadRequest("Unable to add files to transfer: {}".format(e))
+    url = _storage_service_url() + 'file/' + transfer_uuid + '/contents/'
+    response = _storage_api_session().put(url, json=file_set)
+    if 400 <= response.status_code < 500:
```
jrwdunham, line 412: Why `< 500` here. Should a 500 status code raise the BadRequest exception too?

src/archivematicaCommon/lib/storageService.py
```diff
@@ -296,20 +252,20 @@ def create_file(uuid, origin_location, origin_path, current_location,
         'related_package_uuid': related_package_uuid
     }
 
-    LOGGER.info("Creating file with {}".format(new_file))
+    LOGGER.info("Creating file with %s", new_file)
     try:
+        session = _storage_api_session(timeout=None)
```
sevein, line 257: Should we define a sensible timeout instead here?

src/MCPClient/lib/clientScripts/failedSIPCleanup.py
```diff
@@ -22,9 +22,10 @@ def main(fail_type, sip_uuid):
     models.SIPArrange.objects.filter(sip_id=sip_uuid).delete()
 
     # Update storage service that reingest failed
-    api = storage_service._storage_api()
+    session = storage_service._storage_api_session()
+    url = storage_service._storage_service_url() + 'file/' + sip_uuid + '/'
```
sevein, line 26: Do you know why we don't have a function in `storageService.py` for this case? It feels clumsy that we're using functions that are meant to be used only internally.

sevein (approved):
Thank you, Holly! I'm very happy to see slumber gone, I wasn't a big fan. `storageService.py` looks much nicer now! I think in the future we could look into a class-based approach but this is already a huge improvement.

src/archivematicaCommon/lib/storageService.py
Hwesta: Possibly - but what's a sensible timeout? This is the function that store AIP calls, so it could need a very long timeout (which is why it's disabled, unlike the other calls)

src/MCPClient/lib/clientScripts/failedSIPCleanup.py
Hwesta: Different authors, I think. I agree that a good future improvement would be to move this to `storageService.py`

jrwdunham (commented):
I have one outstanding question but in general this looks good to me. I ran the AM and SS tests as well as the relevant Gherkin acceptance tests and all pass.

src/archivematicaCommon/lib/storageService.py
Hwesta: I was matching previous behaviour - `slumber.exceptions.HttpClientError` happens on 4xx responses but not 5xx responses.

PR #551, opened 2017-01-03
URLs: fix trailing $ with no /

Hwesta:
Django's APPEND_SLASH setting can handle URLs that end in `/$`, `/` or nothing, but not just `$`.  This causes the version of the url with a trailing / to not exist.  Notably, this fixes `api/transfer/approve/` to not return a 404.

refs redmine 10707


Hwesta:
This should also be backported to stable/1.5.x, since it also appears on that branch.

hakamine:
Confirmed to be working in a client's testing server👍

jrwdunham:
👍

Hwesta:
Cherry-picked to qa/1.6.x as a2223b4ba0e426aae415450cd9c3efbb3953982f
Cherry-picked to stable/1.5.x as 413ab26b347c0f7e2c06974c705a006b8cbd3b71

PR #552, opened 2017-01-04
Sanitize Names: Handle unicode transfer names

Hwesta:
The transfer name coming from the database was unicode, which forced the (byte)strings to also be promoted to unicode, which failed with a `UnicodeDecodeError`.  Converting the transfer name to a bytestring by encoding as utf8 solves the problem.

Added tests to the sanitize object names code.

This should be cherry-picked to qa/1.x when merged.

### DIFF
--- a/src/MCPClient/lib/clientScripts/sanitizeNames.py
+++ b/src/MCPClient/lib/clientScripts/sanitizeNames.py
@@ -26,7 +26,9 @@ import string
 import os
 from shutil import move as rename
 import sys
+import unicodedata
 from unidecode import unidecode
+from archivematicaFunctions import unicodeToStr
 
 VERSION = "1.10." + "$Id$".split(" ")[1]
 valid = "-_.()" + string.ascii_letters + string.digits
@@ -74,7 +76,9 @@ def sanitizeRecursively(path):
 
     sanitizedName = sanitizePath(path)
     if sanitizedName != path:
-        sanitizations[path] = sanitizedName
+        path_key = unicodeToStr(
+            unicodedata.normalize('NFC', path.decode('utf8')))
+        sanitizations[path_key] = sanitizedName
     if os.path.isdir(sanitizedName):
         for f in os.listdir(sanitizedName):
             sanitizations.update(sanitizeRecursively(os.path.join(sanitizedName, f)))
--- a/src/MCPClient/lib/clientScripts/sanitizeObjectNames.py
+++ b/src/MCPClient/lib/clientScripts/sanitizeObjectNames.py
@@ -24,6 +24,7 @@ from __future__ import print_function
 import logging
 import sys
 import os
+import unicodedata
 
 import django
 django.setup()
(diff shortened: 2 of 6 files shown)


jrwdunham:
When I tested with this branch, I was still not able to initiate a transfer with non-ASCII characters in the transfer name. In order to make that work, I had to modify filesystem_ajax/views.py so that processing_location['path'] was a UTF8-encoded bytestring in the `copy_from_transfer_sources` func. Cf. the fix in dfa1e3e4ff196a4b5065c207ea2a11476886915f

Also, the tests were not passing for me, maybe because of Mac OS X's different behaviour when it comes to Unicode normalization. In order to make the tests pass, I had to perform Unicode NFC normalization on the sanitizations dict's keys as well as on dirpath when attempting to match. I made a disposable branch implementing these changes because I got fixated on trying to fix them; see https://github.com/artefactual/archivematica/commit/68a54250e47fda81d1f0ea03ebbd7c4172952716.

Hwesta:
That's very odd, since I was unable to reproduce those problems. I deployed a VM with vagrant & ansible using this branch, and was able to create transfers with the names `ユニコード` and `éve`, and the tests passed when run on the VM (also locally).

jrwdunham:
Ok, after further investigation, it turns out that the fixes in dfa1e3e4ff196a4b5065c207ea2a11476886915f were not required to solve the first issue listed above. I had been using a branch of ansible-archivematica-src in my dev env that lacked the correct encoding env vars (d6f99e484da7066dc9241f811080149886ee5dc0). Rebasing my ansible-archivematica-src with master solved this issue.

However, the tests were still failing for me. Further investigation revealed that sanitizeNames.py is getting NFD-normalized strings from the filesystem while sanitizeObjectNames.py is getting NFC-normalized strings from the database. When `sanitize_object_names` attempts to find `dirpath` in `sanitizations` (`if dirpath in sanitizations:`), it can't because the keys of sanitizations are NFD-normalized and it's trying to match them to a specific NFC-normalized key.

The source of these strings is the fixture src/MCPClient/tests/fixtures/files-transfer-unicode.json, which contains NFC-normalized "é" in "ev\u00e9lyn's photo". I think the issue arises because test_sanitize.py writes test files to disk and Mac OS stores filenames as NFD-normalized UTF8 strings by default. Because my dev setup has the source dirs synced between the mac host and linux guest, for some reason the test is writing Unicode filenames to disk in the Mac OS way, even though the tests are being run from the command line in Ubuntu.

In order for these tests to pass on a Mac-host/Linux-guest dev setup, fixes like those in 68a54250e47fda81d1f0ea03ebbd7c4172952716 need to be incorporated.  If comparing Unicode strings for equality is important throughout AM, then we should probably implement a normalization policy so that we are always sure that all strings are normalized in the same way.

Hwesta:
Good sleuthing to figure that out! I like the commit that adds NFC normalization, and I'll include that in this PR if that's okay with you. 

It wouldn't solve the underlying problem, but would changing the name of the file in the test json to be NFD normalized also fix the problem?

jrwdunham:
Yep, I'm okay with you including that commit in this PR. And yea, I would guess that NFD-normalizing the json fixture would also fix the immediate testing problem.

jrwdunham:
👍

Hwesta:
Cherry-picked to qa/1.x as bdaf1a1991417ea805fa97a73018bc65e152383c a51c114397206d30d4ad6cbf733644d4acff7bbc a98a7d2a735a268722cd3a769308dc24f5b70bcd d92cebebde4bdf61310d1624b18386e463707dfa

PR #553, opened 2017-01-06
django-mysqlpool: unlimited max_overflow #10622

hakamine:
Add parameter "max_overflow: -1" to MYSQLPOOL_ARGUMENTS in both MCPServer and MCPClient configuration, to avoid hitting a limit in the total number of mysql concurrent connections (default value is 10), that causes some scalability issues.

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -36,6 +36,7 @@ DATABASES = {
 MYSQLPOOL_BACKEND = 'QueuePool'
 MYSQLPOOL_ARGUMENTS = {
     'use_threadlocal': False,
+    'max_overflow': -1,
 }
 
 CONN_MAX_AGE = 14400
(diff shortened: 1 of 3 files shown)


jhsimpson:
after discussing with @hakamine irl, I made the parameter configurable, with a default of 40.  Edit the dbsettings file (at /etc/archivematica/archivematicaCommon/dbsettings) to change this, it will be the same setting for both MCPServer and MCPClient.

This is related to #505 - This PR is a temporary change, keeping the django-mysqlpool library, but making it possible to use more than 10 database connections at a time, which should help scale MCPClient when there are more than 12 or so cpu cores available.

#505 will get bumped to the next release, after it has had more testing.

jhsimpson:
:+1:

jhsimpson:
:+1:

jhsimpson:
These two commits where cherry-picked into qa/1.x as

d5d397f1d194e305f4b2b0cada1b670ba53d674d
and
715c95241f1f79b6c4ca5e947afde8a660f2a939

PR #554, opened 2017-01-09
copyToNetX: handle multiple Jobs when fetching DIP path

Hwesta:
Since we're uploading to multiple systems, there are multiple Upload DIP jobs, all with the same path. Take the first one instead of erroring.


hakamine:
This fix also committed to branch dev/1.6.x-moma

hakamine:
👍  Tested and working

PR #555, opened 2017-01-12
save target slug in dip upload better

jhsimpson:
Fixes issue #10759 https://projects.artefactual.com/issues/10759

For some reason (unknown to me), the value of input.filter(':text').val()
is empty when it is used to populate the form POST data sent to Archivematica.

This commit saves the value to a variable and uses the variable instead.

I'd love to know why this is necessary.  The code here did not change from
the 1.5.1 release, other than the change to a new bootstrap modal.

I assume that is the reason the behaviour has changed, but I am not sure
why it didn't show up until now.

### DIFF
--- a/src/dashboard/src/media/js/ingest.js
+++ b/src/dashboard/src/media/js/ingest.js
@@ -381,10 +381,11 @@ $(function()
 
                   if (input.filter(':text').val())
                   {
+                    var v_target = input.filter(':text').val();
                     $('#upload-dip-modal-spinner').show();
                     // get AtoM destination URL (so we can confirm it's up)
                     $.ajax({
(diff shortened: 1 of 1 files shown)


jraddaoui (commented):
I don't really know the reason but it looks like a good fix, as it avoids filtering the input two times ;)

We also upgraded jQuery alongside Bootstrap but I couldn't find anything about it.

jrwdunham:
👍 Looks good to me. One nitpicky question: why not just call `input.filter(':text').val()` once instead of twice and define the `v_target` var at line 381?

jhsimpson:
I wanted to change as little as possible, but you are probably right @jrwdunham .  However, this works, I suggest just leaving it as is.

Cherry-picked to qa/1.x as 787e89d63025a23b5dcc0695806a89a19670bb19

PR #557, opened 2017-01-18
PREFORMA: MediaConch integration

jrwdunham:
- Pre-ingest conformance: can check whether .mkv files are in
  conformance with the .mkv specification (implementation check).

- Post-normalization conformance: can check whether files normalized
  to .mkv are in conformance with the .mkv specification.

- Transfer policy check: check whether original files conform to
  a pre-defined MediaConch policy.

- Ingest policy check: can check that derivatives
  (normalized for access or preservation, manually or AM-created)
  conform to a pre-defined policy.

### DIFF
--- a/src/MCPClient/lib/clientScripts/policyCheck.py
+++ b/src/MCPClient/lib/clientScripts/policyCheck.py
@@ -1,0 +1,425 @@
+#!/usr/bin/env python2
+"""Runs zero or more FPR policy check against the provided file and returns an
+exit code. May also print to stdout, generate an Event models in the db,
+and/or write command-specific stdout to disk.
+
+If a format has no defined policy check commands, no command is run.
+
+Usage::
+
+    $ ./policyCheck.py [FILE_PATH] [FILE_UUID] [SIP_UUID] [SHARED_PATH] \
+                       [FILE_TYPE]
+"""
+from __future__ import print_function
+import json
+import os
+import sys
+
+from custom_handlers import get_script_logger
+
+import django
+django.setup()
+from fpr.models import FPRule, FormatVersion
+from main.models import Derivation, File, SIP, Transfer
+
+from executeOrRunSubProcess import executeOrRun
+import databaseFunctions
+from dicts import replace_string_values
+
+# Note that linkTaskManagerFiles.py will take the highest exit code it has seen
+# from all tasks and will use that as the exit code of the job as a whole.
+SUCCESS_CODE = 0
+NOT_APPLICABLE_CODE = 0
+FAIL_CODE = 1
+
+
+LOGGER = get_script_logger("archivematica.mcp.client.policyCheck")
+
+
+class PolicyChecker(object):
+    """Checks whether a given file conforms to all of the MediaConch policies
+    that the system is configured to run against that type of file, given the
+    file's format and its purpose, i.e., whether it is intended for access or
+    preservation. Usage involves initializing on a file and then calling the
+    ``check`` method.
+    """
+
+    def __init__(self, file_path, file_uuid, sip_uuid, shared_path, file_type):
+
+        self.file_path = file_path
+        self.file_uuid = file_uuid
+        self.sip_uuid = sip_uuid
+        self.shared_path = shared_path
+        self.file_type = file_type
+        self.policies_dir = self._get_policies_dir()
+        self.is_manually_normalized_access_derivative = (
+            self._get_is_manually_normalized_access_derivative())
+        self._sip_logs_dir = None
+        self._sip_subm_doc_dir = None
+        self._sip_policy_checks_dir = None
+
+    def check(self):
+        """Check the file identified by ``self.file_uuid`` against any
+        policy-check FPR commands that are applicable. If any fail, return a
+        non-zero exit code; otherwise return ``0``.
+        """
+        if not self.is_manually_normalized_access_derivative:
+            try:
+                self.file_model = File.objects.get(uuid=self.file_uuid)
+            except File.DoesNotExist:
+                print('Not performing a policy check because there is no file'
+                      ' with UUID {}.'.format(self.file_uuid))
+                return NOT_APPLICABLE_CODE
+        if not self._we_check_this_type_of_file():
+            return NOT_APPLICABLE_CODE
+        rules = self._get_rules()
+        if not rules:
+            print('Not performing a policy check because there are no relevant'
+                  ' FPR rules')
+            return NOT_APPLICABLE_CODE
+        rule_outputs = []
+        for rule in rules:
+            rule_outputs.append(self._execute_rule_command(rule))
+        if 'failed' in rule_outputs:
+            return FAIL_CODE
+        else:
+            return SUCCESS_CODE
+
+    purpose = 'policy_check'
+
+    def _we_check_this_type_of_file(self):
+        """Return ``True`` if this policy checker should perform a check on
+        this file; ``False`` otherwise. This will depend on ``self.file_type``
+        and on whether the file is an original or a preservation/access
+        derivative.
+        """
+        if self.file_type == 'original':
+            # During transfer we check all files. Is this correct or are there
+            # classes of file that we do not want to perform policy checks on?
+            return True
+        elif self.file_type == 'preservation':
+            if (not self._file_is_derivative()) or self._file_is_for_access():
+                print('File {uuid} is not a preservation derivative; not'
+                      ' performing a policy check.'.format(uuid=self.file_uuid))
+                return False
+            return True
+        elif self.file_type == 'access':
+            if (self._file_is_derivative(for_access=True) and
+                    self._file_is_for_access()):
+                return True
+            print('File {uuid} is not an access derivative; not performing'
+                  ' a policy check.'.format(uuid=self.file_uuid))
+            if not self._file_is_derivative(for_access=True):
+                print('File {uuid} is not a derivative.'.format(
+                    uuid=self.file_uuid))
+            if not self._file_is_for_access():
+                print('File {uuid} is not for access.'.format(
+                    uuid=self.file_uuid))
+            return False
+        else:
+            return True
+
+    def _file_is_derivative(self, for_access=False):
+        """Return ``True`` if the target file is a derivative; ``False``
+        otherwise.
+        """
+        if self.is_manually_normalized_access_derivative:
+            return True
+        # Access derivatives have Derivation rows with NULL event types (cf.
+        # normalize.py client script).
+        event_type = 'normalization'
+        if for_access:
+            event_type = None
+        try:
+            Derivation.objects.get(derived_file__uuid=self.file_uuid,
+                                   event__event_type=event_type)
+            return True
+        except Derivation.DoesNotExist:
+            return False
+
+    def _get_policies_dir(self):
+        return os.path.join(self.shared_path, 'sharedMicroServiceTasksConfigs',
+                            'policies')
+
+    def _get_is_manually_normalized_access_derivative(self):
+        """Manually normalized access derivatives are never given UUIDs.
+        Therefore, we need this heuristic for determining if that is what we
+        are dealing with. TODO/QUESTION: will this return false positives?
+        """
+        if (self.file_uuid == 'None' and
+                os.path.split(self.file_path)[0].endswith('/DIP/objects')):
+            return True
+        return False
+
+    def _file_is_for_access(self):
+        """Returns ``True`` if the file with UUID ``self.file_uuid`` is "for"
+        access.
+        """
+        if (self.is_manually_normalized_access_derivative or
+                self.file_model.filegrpuse == 'access'):
+            return True
+        return False
+
+    def _get_manually_normalized_access_derivative_file_uuid(self):
+        """If the file-to-be-policy-checked is a manually normalized access
+        derivative it will have no file UUID in the database. We therefore have
+        to retrieve the UUID of the original file that was format-identified,
+        i.e., the file that was in manualNormalization/access/, which we do by
+        querying the database based on the original location of the file. This
+        file UUID is needed so that we can get the format (PRONOM id) in order
+        to retrieve the appropriate policy-check FPR rule for this type of file
+        (see ``self._get_rules()``).
+        """
+        manually_normalized_file_name = os.path.basename(self.file_path)[37:]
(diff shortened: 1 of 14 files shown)


src/MCPServer/share/sharedDirectoryStructure/sharedMicroServiceTasksConfigs/policies/.keep
```diff
@@ -0,0 +1,2 @@
+This file exists just to ensure that git will add this (otherwise empty)
```
sevein: Our convention to track empty directories in git is to leave a `.gitignore` file like the following: https://goo.gl/RN90mS. I don't know if it's the best approach toady but perhaps we should be consistent. More here: http://stackoverflow.com/a/11579945.

src/MCPClient/lib/clientScripts/policyCheck.py
```diff
@@ -0,0 +1,331 @@
...
+from dicts import replace_string_values
+
+# Note that linkTaskManagerFiles.py will take the highest exit code it has seen
+# from all tasks and will use that as the exit code of the job as a whole.
+SUCCESS_CODE = 0
+NOT_APPLICABLE_CODE = 0
+FAIL_CODE = 1
+
+
+class PolicyChecker:
```
sevein: Should your new class subclass `object`? I think we always use the new style.

src/MCPClient/lib/clientScripts/validateDerivative.py
```diff
@@ -0,0 +1,179 @@
...
+import databaseFunctions
+from dicts import replace_string_values
+
+SUCCESS_CODE = 0
+FAIL_CODE = 1
+NOT_DERIVATIVE_CODE = 0
+NO_RULES_CODE = 0
+
+
+class DerivativeValidator:
```
sevein: Same here, I think we want to subclass `object`.

src/MCPClient/lib/clientScripts/validateDerivative.py
```diff
@@ -0,0 +1,179 @@
+from __future__ import print_function
```
sevein: It seems that `validateDerivative.py` is not an executable script but your commit includes the file with the permission bit on it. This is not the only case I think, e.g. `transcoder.py` has both a shebang and the execution bit - we should take care of those some other time, I'll file an issue.

I think I would prefer to put files that are not entry points somewhere else. The easiest way could be a `lib/` folder inside `clientScripts`. This is not strictly required but I think it would be nice. I like your class-based approach. Some of these classes like `PolicyChecker` could as well go to `lib`?

sevein: @jhsimpson @Hwesta, what do you think?

src/dashboard/src/components/administration/urls.py
```diff
@@ -39,6 +39,9 @@
     url(r'processing/delete/(?P<name>\w{1,16})/$', views_processing.delete),
     url(r'premis/agent/$', views.premis_agent),
     url(r'api/$', views.api),
+    url(r'policies/$', views.policies),
+    url(r'policies/view/(?P<policy>[^/]+)/$', views.view_policy),
```
sevein: Can we replace this one with `r'policies/(?P<policy>[^/]+)/$'`? It seems consistent with other URLs across the dashboard.

src/dashboard/src/components/administration/urls.py
```diff
@@ -39,6 +39,9 @@
...
+    url(r'policies/delete/(?P<policy>[^/]+)/$', views.delete_policy),
```
sevein: I realize we're not consistently naming our URLs, sometimes we use the `GET /foobar/action/$id/` format but in other cases we've been using `GET /foobar/$id/action/`. I guess we can discuss that some other day.

src/dashboard/src/templates/administration/policies.html
```diff
@@ -0,0 +1,79 @@
...
+      {% endif %}
+
+      <div class="actions">
+        <form action='{% url 'components.administration.views.policies' %}'
+              method='post' enctype='multipart/form-data'>
+
+          <div class='form-group'>
+            <label class='btn btn-default' for='file'>
+              <input id='file' type='file' name='file' style='display:none;'
+                onchange='$("#filename").html($(this).val().replace("C:\\fakepath\\", ""));'>
```
sevein: What is the routine in `onchange` for?

src/dashboard/src/templates/administration/policy.html
```diff
@@ -0,0 +1,60 @@
...
+
+  <style>
+    pre {
+      overflow-x: scroll;
+    }
+    code {
+      white-space: pre;
+      line-height: 2em;
+      background-color: transparent;
+    }
```
sevein: This `<style>...</style>` is the same you have above in the same file excepting this extra `background-color` here.

src/dashboard/src/templates/administration/policy.html
```diff
@@ -0,0 +1,60 @@
...
+      {% else %}
+        <p>There is no policy named {{ policy }}.</p>
+      {% endif %}
+
+    </div>
+
+  </div>
+
+{% endblock %}
+
```
sevein: This blank line is not needed.

src/dashboard/src/components/administration/forms.py
```diff
@@ -258,6 +265,20 @@ class ProcessingConfigurationForm(forms.Form):
...
+        'type': 'boolean',
+        'name': 'policy_checks_preservation_derivatives',
+        'label': 'Perform policy checks on preservation derivatives',
+        'yes_option': '3a55f688-eca3-4ebc-a012-4ce68290e7b0',
+        'no_option': 'b7ce05f0-9d94-4b3e-86cc-d4b2c6dba546',
+    }
+    processing_fields['8ce07e94-6130-4987-96f0-2399ad45c5c2'] = {
+        'type': 'boolean',
+        'name': 'policy_checks_access_derivatives',
+        'label': 'Perform policy checks on access derivatives',
```
sevein: Justin mentioned that this should probably be renamed to `Validate access derivatives`, similar with the other cases. We can talk about this later.

sevein (commented):
@jrwdunham I've got a few observations for you, see my comments inline.

Some of the stuff you're doing in this pull request is interacting with parts of the application that I don't really know very well. I noticed that some of the changes you do in `0030_mediaconch.py` belong in `fpr-admin` and the scripts under `migrations-data` should perhaps be moved to https://github.com/artefactual/archivematica-fpr-tools. Justin also mentioned a couple of things that need to be discussed, like the labels used in the processing configuration page may have to be reworded and some stuff about the validation tools I can't remember well! He's going to either participate in this pull request or schedule a meeting to discuss this IRL. I'm interested to participate so I can get a better understanding on how the FPR works!

src/MCPClient/lib/clientScripts/policyCheck.py
jrwdunham: Yes, you're right. I'm used to writing Python3-only code where subclassing `object` is redundant.

src/MCPServer/share/sharedDirectoryStructure/sharedMicroServiceTasksConfigs/policies/.keep
jrwdunham: Hey, who you callin' toady! Just kidding: change effected.

src/dashboard/src/components/administration/forms.py
jrwdunham: This proposed change seems confusing to me since this decision point governs whether "policyCheckAccessDerivative.py" gets called and there is a distinct client script named "validateAccessDerivative.py" that has nothing to do with this...

src/dashboard/src/components/administration/urls.py
jrwdunham: Ok, that makes sense.

src/dashboard/src/components/administration/urls.py
jrwdunham: I guess I was just being consistent with what's done in this particular file.

src/dashboard/src/templates/administration/policies.html
jrwdunham: When the user selects a file to upload, the `<span#filename>` displays the filename. Some browsers list this "fakepath" stuff and it looks ugly, so the `replace` just removes it. See http://stackoverflow.com/questions/6365858/use-jquery-to-get-the-file-inputs-selected-filename-without-the-path

src/dashboard/src/templates/administration/policy.html
jrwdunham: Yeah, that's odd I did that. I'll merge into a single <style> block.

src/MCPClient/lib/clientScripts/validateDerivative.py
jrwdunham: Ok.

src/MCPClient/lib/clientScripts/validateDerivative.py
jrwdunham: Yes, assuming the clientScripts/lib/ approach is what we go with, the entire policyCheck.py could be moved there too and made non-executable. It's not really used as a client script on its own; it's always imported and subclassed just like validateDerivative.py

jrwdunham:
@sevein I take your point about the 0030_mediaconch.py migration potentially being better placed in the archivematica-fpr-admin repo. However, I'd like some clarification on this and a better high-level understanding of the motivations for this (from you or @Hwesta / @jhsimpson) before I go ahead and make changes. 

Some of the models I'm adding are AM models with classes defined in the dashboard and others are FPR models. I could move the entire migration to archivematica-fpr-admin, but some quick greping suggests that the migrations there never touch the AM models proper.

So would the idea be to split this migration across 2 different files in 2 different repos? That would seem to needlessly complicate things, especially given that this migration is conceptually atomic.

Relatedly, I could put the two new commands in archivematica-fpr-tools, but the "deployment process" in the README talks about generating raw SQL migrations, which doesn't seem to jive with how we currently do migrations (Django).

src/MCPClient/lib/clientScripts/validateFile.py
```diff
@@ -1,91 +1,297 @@
...
+                try:
+                    os.makedirs(_sip_pres_val_dir)
+                except OSError:
+                    pass
+                else:
+                    self._sip_pres_val_dir = _sip_pres_val_dir
+        return self._sip_pres_val_dir
+
+
+def _get_logger_name(file_type):
```
sevein: Is this still relevant?

src/MCPClient/lib/clientScripts/validateFile.py
```diff
@@ -1,91 +1,297 @@
...
+
+
+if __name__ == '__main__':
     file_path = sys.argv[1]
     file_uuid = sys.argv[2]
     sip_uuid = sys.argv[3]
-    sys.exit(main(file_path, file_uuid, sip_uuid))
+    shared_path = _get_shared_path(sys.argv)
+    file_type = _get_file_type(sys.argv)
+    logger = get_script_logger(_get_logger_name(file_type))
```
sevein: I think you should use the logger instead of `sys.stderr`. The events are currently [written to disk](https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/archivematicaCommon/lib/custom_handlers.py#L20) but we're planning to change that soon.

sevein: See #606.

src/MCPClient/requirements/base.txt
```diff
@@ -1,3 +1,4 @@
+git+git://github.com/artefactual-labs/ammcpc@v0.1.0#egg=ammcpc
```
sevein: I tagged and updated this.

src/dashboard/src/requirements/base.txt
```diff
@@ -9,7 +9,7 @@ django-extensions==1.1.1
 django-autoslug==1.7.1
 django-annoying==0.7.7
 elasticsearch>=1.0.0,<2.0.0
-git+https://github.com/artefactual/archivematica-fpr-admin.git@v1.3.0-beta.1#egg=archivematica-fpr-admin
+git+https://github.com/artefactual/archivematica-fpr-admin.git@dev/issue-9478-preforma#egg=archivematica-fpr-admin
```
sevein: Now in qa/1.x we import archivematica-fpr-admin using pip. I've updated this using your branch.

src/dashboard/src/main/migrations/0034_mediaconch_policy_checks.py
```diff
@@ -0,0 +1,451 @@
...
+    ###########################################################################
+
+    # 'Validate formats' exits to 'Policy checks for originals'
+    # CHOICE cl.
+    validate_formats_cl = MicroServiceChainLink.objects.filter(
+        currenttask__description='Validate formats').first()
+    validate_formats_cl.defaultnextchainlink = rgnl_policy_check_choice_cl
+    MicroServiceChainLinkExitCode.objects\
+        .filter(microservicechainlink=validate_formats_cl)\
+        .update(nextmicroservicechainlink=rgnl_policy_check_choice_cl)
```
sevein: With this you're leaving behind the old task config, right? Should we delete it?

sevein (approved):
@jrwdunham - please take a look at my changes. There is a minor observation about a chain link config left behind. Other than that this is 👍. Good job, I've learned a lot! The commits prefixed with #601 can be deleted, I needed them to run the test suite.

src/MCPClient/lib/clientScripts/validateFile.py
jrwdunham: No, it's not. Thanks. Will remove.

jrwdunham:
This PR should be closed since it is replace by https://github.com/artefactual/archivematica/pull/634. Before closing, I would like to make sure that the issues reference above are addressed or are being addressed:

- The commits prefixed with #601 that sevein mentions: are these still in existence? still relevant?
- What is the "observation about a chain link config left behind" that sevein mentions?
- Requesting more details on the suggestion to use the logger instead of sys.stderr, and whether that will affect what is displayed in the /tasks/<UUID> pages

sevein:
Can we delete these other two branches?
- dev/issue-9478-preforma-test
- dev/issue-9478-preforma-test-qa-1-x

PR #558, opened 2017-01-25
Fixes #10785: AIC restructure for compliance bug

jrwdunham:


jhsimpson:
This PR restores the behaviour from 1.5.1.  I tried it and it works, I think this client script could be refactored, but that is not in scope for this bug fix, so looks good to me

:+1:

jrwdunham:
Merged to qa/1.6.x as dc205d871834b8537cbafeabde14dfda59128ad6 and cherry-picked to qa/1.x as 75bb89e28cd690ea66b7b247bc8d3f51d1b5bb6b

PR #559, opened 2017-01-25
Packaging updates for 1.6

scollazo:
Changes apache/wsgi to nginx + gunicorn.

I'm not sure about how to handle the upgrade from 1.5.1 using apache, should we stop it from the package, or update the archivematica docs to explain it?

### DIFF
--- a/src/dashboard/debian/archivematica-dashboard.install
+++ b/src/dashboard/debian/archivematica-dashboard.install
@@ -1,2 +1,4 @@
 src/* /usr/share/archivematica/dashboard/
-
+install/archivematica-dashboard.conf /etc/init/
+install/dashboard.conf /etc/nginx/sites-available/
+install/dashboard.gunicorn-config.py /etc/archivematica/
--- a/src/dashboard/debian/control
+++ b/src/dashboard/debian/control
@@ -11,15 +11,16 @@ Architecture: any
 Depends: 
 	${shlibs:Depends},
 	${misc:Depends},
-	apache2-mpm-prefork,
-	libapache2-mod-wsgi,
+	nginx,
 	python-pip,
 	python-dateutil,
 	python-gearman,
 	python-simplejson,
 	libffi-dev,
 	libssl-dev,
-	python-dev
+	python-dev,
+	libxml2-dev,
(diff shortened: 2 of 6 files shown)


jhsimpson:
I think we have to stop apache, in order to bind to port 80?  not sure about uninstalling it though, that we might want to do in documentation.

src/MCPClient/debian/control
```diff
@@ -45,5 +45,6 @@ Depends:
 	tree,
 	ufraw,
 	unrar-free,
+	python-unidecode,
```
jhsimpson: Why is unidecode needed?  If there is a python dependency, couldn't this get added via the archivematica-common requirements file?  That way it wouldn't need to get added in multiple places (spec file for rpm, ansible vars file).

scollazo: Without it, the sanitizeObjectNames microservice fails with

`
  File "/usr/lib/archivematica/MCPClient/clientScripts/sanitizeObjectNames.py", line 38, in <module>
    import sanitizeNames
  File "/usr/lib/archivematica/MCPClient/clientScripts/sanitizeNames.py", line 30, in <module>
    from unidecode import unidecode
ImportError: No module named unidecode`

I added it as a package dependency, to avoid messing with requirements.txt, but I'll do the opposite to make packaging simpler.

sevein:
Looks good to me but the branch seems to need a rebase. Also, I know nothing about Debian packages but @jhsimpson and @scollazo are working together on it.

👍

scollazo:
Rebased and removed the mcp client dependecy with python-unidecode, as it's now in the requirements.txt

jhsimpson:
:+1:  

My earlier comment about stopping apache can/should be dealt with in the upgrade instructions, so this looks good as is.

jhsimpson:
merged into qa/1.6.x as
2b73bb65523aa6c111d1c01a789e739b9fccd619
8d166f18f39a9661dce9de50f6533f779bfe7e7a

and cherry-picked into qa/1.x as
89ebee17cd78f9b7814eb1974571a227d8201524
d9130b673a8c1eb9a87e1f7859d03c8467c979e5

PR #560, opened 2017-01-28
metadata/ and logs/ dirs viewable in SIP arrange

jrwdunham:
UI has "Show metadata & logs directories?" checkbox to control whether
these directories will be displayed when browsing transfers in backlog.

### DIFF
--- a/src/dashboard/src/components/ingest/views.py
+++ b/src/dashboard/src/components/ingest/views.py
@@ -504,9 +504,16 @@ def transfer_backlog(request, ui):
     # Get search parameters from request
     results = None
 
+    # GET params in SIP arrange can control whether files in metadata/ and
+    # logs/ are returned. Appraisal tab always hides these dirs and their files
+    # (for now).
+    backlog_filter = elasticSearchFunctions.BACKLOG_FILTER
+    if ui == 'appraisal' or request.GET.get('hidemetadatalogs'):
+        backlog_filter = elasticSearchFunctions.BACKLOG_FILTER_NO_MD_LOGS
+
     if not 'query' in request.GET:
         query = elasticSearchFunctions.MATCH_ALL_QUERY.copy()
-        query['filter'] = elasticSearchFunctions.BACKLOG_FILTER
+        query['filter'] = backlog_filter
     else:
         queries, ops, fields, types = advanced_search.search_parameter_prep(request)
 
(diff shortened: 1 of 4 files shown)


src/dashboard/src/media/js/ingest/backlog.js
```diff
@@ -61,6 +61,9 @@ function renderBacklogSearchForm(search_uri, on_success, on_error) {
     function backlogSearchSubmit() {
       // Query Django, which queries ElasticSearch, to get the backlog file info
       var query_url = search_uri + '?' + search.toUrlParams();
+      if (!$('#id_show_metadata_logs').is(':checked')) {
+          query_url = query_url + '&hidemetadatalogs=1';
```
sevein: Should we use [`jQuery.param`](http://api.jquery.com/jquery.param/) to make sure that we're building the query string consistently? Source: https://goo.gl/RfVJkl. It may not be necessary.

src/dashboard/src/components/ingest/views.py
```diff
@@ -504,9 +504,16 @@ def transfer_backlog(request, ui):
     # Get search parameters from request
     results = None
 
+    # GET params in SIP arrange can control whether files in metadata/ and
+    # logs/ are returned. Appraisal tab always hides these dirs and their files
+    # (for now).
+    BACKLOG_FILTER = elasticSearchFunctions.BACKLOG_FILTER
+    if ui == 'appraisal' or request.GET.get('hidemetadatalogs'):
+        BACKLOG_FILTER = elasticSearchFunctions.BACKLOG_FILTER_NO_MD_LOGS
```
sevein: If `BACKLOG_FILTER` is not a constant perhaps we shouldn't use capitals. What do you think?

jrwdunham: Makes sense.

src/dashboard/src/media/js/ingest/backlog.js
jrwdunham: Ok.

sevein:
Cool! I think we think of safer ways to build URLs (e.g. we've used URI.js in other projects) - but I'm not sure if it's really needed. Thanks for the changes! 👍

jhsimpson:
:+1:  from me too.

jrwdunham:
Cherry-picked to qa/1.x as d22d8a711316ec9978db70dda9d36f6fc95dfb81

PR #561, opened 2017-01-28
Reorder migrations

jhsimpson:
refs #10830

This puts the dashboards django migrations in the same order, in qa/1.6.x as the order in stable/1.5.x
to avoid problems when upgrading.

This change will be needed in qa/1.x as well.

### DIFF
--- a/src/dashboard/src/main/migrations/0019_normalization_report.py
+++ b/src/dashboard/src/main/migrations/0019_normalization_report.py
@@ -30,7 +30,7 @@ def data_migration(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('main', '0022_blob_fields'),
+        ('main', '0018_archivesspace_sip_arrange'),
     ]
 
     operations = [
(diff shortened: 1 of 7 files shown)


jhsimpson:
I added another commit to this branch, to deal with a problem that came up in testing.  I assume it is bad data, but it is possible for the linkingAgentIdentifier column of the Events table to be null, which broke the migration.

I can't think of a situation where these rows are worth saving, any one have any thoughts on that?

sevein:
I've confirmed that this works on a fresh `qa/1.6.x` installation.

- `linkingAgentIdentifier` is not a field in the `Events` table,
- I can see in `main_userprofile` that my user `demo` is associated with a row in `Agents` (`agentIdentifierType='Archivematica user pk'`).

:+1:

sevein:
👍 - tested in two different VMs.

jhsimpson:
cherry-picked into qa/1.x as
7fd3a21f190d8374367cf238f6bdb6ec7e03f37f
0f1583ff2dc7a6384c5649ba90e9ec3f6dd2cabe

PR #563, opened 2017-01-30
Dev/packaging 1.6 virtualenvs

scollazo:
First draft of conversion to dh-virtualenv

 - Changed some paths in initscripts in order to reflect the current ones in the virtulenvs. This should be a configuration option in /etc/defaults/archivematica-(-dashboard,-mcp-client,-mcp-server), so we can add something like `PYTHON_BIN=${AM_DASHBOARD_PYTHON:/usr/bin/python}` to the init scripts
- I'm usure about archivematica-common, as all python depedencies should be intalled into each virtualenv, to avoid messing with  PYTHON_PATHs


jhsimpson:
We discussed this irl, and are leaving it out of the 1.6.0 release, to look at in more detail after the release.

Hwesta:
This was against qa/1.6.x, which has been replaced by stable/1.6.x Changing what branch this is merging to and reopening.

sevein:
This is going to be done in AM17. However, @scollazo is moving the debian files to [am-packbuild](https://github.com/artefactual-labs/am-packbuild).

PR #564, opened 2017-01-30
Full reingest: old AIP METS removed from fileGrp

jrwdunham:
Fixes #10829 where, during full reingest, the METS file of the original
AIP was correctly removed from the new AIP but it was still incorrectly
referenced in the <mets:fileGrp USE="submissionDocumentation"> of the
new METS.


sevein:
👍

Hwesta (approved):

jrwdunham:
Cherry-picked to qa/1.x as ce0d2863fc703e0a84e8d70b038c47870bfa5652

PR #565, opened 2017-02-08
Increase timeout to 120 seconds

sevein:
For the storage service, a timeout of 5 isn't always long enough. Since the
values are currently harcoded we're defining larger values for safety reasons.

This is a continuation of #531.


sevein:
I've confirmed it's working. I put a `time.sleep(N)` in SS's `LoginRequiredMiddleware` with different values. **Only** when N is greater than 120, the dashboard (`/filesystem/children/location/$uuid/?path=$path`) throws the following error after 120 seconds waiting:

```
ReadTimeout: HTTPConnectionPool(host='localhost', port=8000): Read timed out. (read timeout=120)
```

jhsimpson:
:+1:

PR #570, opened 2017-02-16
MCPClient: update version dependencies

scollazo:
Use current versions for fido and siegfried.

Also, add python-unidecode

### DIFF
--- a/src/MCPClient/debian/archivematica-mcp-client.install
+++ b/src/MCPClient/debian/archivematica-mcp-client.install
@@ -1,5 +1,7 @@
 etc/* /etc/archivematica/MCPClient/
 lib/* /usr/lib/archivematica/MCPClient/
 init/* /etc/init/
+requirements.txt /usr/share/archivematica/MCPClient/
+requirements/* /usr/share/archivematica/MCPClient/requirements/
 
 
--- a/src/MCPClient/debian/control
+++ b/src/MCPClient/debian/control
@@ -35,8 +35,6 @@ Depends:
 	p7zip-full,
 	pbzip2,
 	postfix,
-	python-unidecode,
-	python-fido (>=200.1.3.5.88),
 	readpst,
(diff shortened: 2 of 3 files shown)


hakamine (commented):
are we using fido from packages or from pypi (https://pypi.python.org/pypi/opf-fido) ?

scollazo:
Removed the dependency with python-pip, to avoid installing ubuntu pip package.

pip needs to be installed before archivematica, with easy_install pip

jhsimpson:
I just tested and python-unidecode is being added in the pip requirements file for the MCP Client, so it doesn't need to be in the controlfile, I don't think.

I think python-pip does need to be removed as a dependency.

opf-fido has been added to the pip requirements file, so for a fresh installation of Archivematica 1.6.0, it is not necessary to add python-fido 1.3.5 to the debian control file - it should be possible to use the version pip installs.  I tested this and on my test vm, I have fido 1.3.4 from packages.archivematica.org and 1.3.5 from pypi, and it is the 1.3.5 version that gets used:

vagrant@am-local:~$ which fido
/usr/local/bin/fido
vagrant@am-local:~$ fido -v
FIDO v1.3.5 (formats-v88.xml, container-signature-20160121.xml, format_extensions.xml)
vagrant@am-local:~$ apt-cache policy python-fido
python-fido:
  Installed: 200.1.3.4-84
  Candidate: 200.1.3.4-84
  Version table:
 *** 200.1.3.4-84 0
        500 http://packages.archivematica.org/1.5.x/ubuntu-externals/ trusty/main amd64 Packages
        100 /var/lib/dpkg/status


I think this PR could be modified just to include the last change (remove python-pip).

scollazo:
MCPCient requirements.txt are not installed in the debian package :(

scollazo:
Tested and :+1:

hakamine:
👍

PR #572, opened 2017-02-17
METS: Update METS namespace to 1.11

Hwesta:
This should also include artefactual-labs/mets-reader-writer#24

refs redmine 10881


jrwdunham:
Should this change also include an update of the version number in metsrw's setup.py (in https://github.com/artefactual-labs/mets-reader-writer/pull/24) and a corresponding update in MCPClient/requirements/base.txt and dashboard/src/requirements/base.txt?

Hwesta:
Yes. I don't think this change is significant enough to justify a new release of metsrw. However, since dev on that has slowed, it might make sense to do a point release (0.1.1) so that PR isn't hanging unreleased too long.

jrwdunham:
👍

Hwesta:
Decided to make a 0.1.1 release, added updated dependency.

PR #575, opened 2017-02-23
Dev/issue 10140 rights csv import 1 6 rebase

mcantelon:
Adds microservice to parse metadata/rights.csv, containing rights criteria for specific files,  and create rights entries (that later get added to METS)

### DIFF
--- a/src/MCPClient/lib/clientScripts/rightsFromCSV.py
+++ b/src/MCPClient/lib/clientScripts/rightsFromCSV.py
@@ -1,0 +1,373 @@
+#!/usr/bin/env python2
+
+# This file is part of Archivematica.
+#
+# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+# @package Archivematica
+# @subpackage archivematicaClientScript
+# @author Mike Cantelon <mike@artefactual.com>
+
+from __future__ import print_function
+import csv
+import os
+import sys
+
+import django
+django.setup()
+
+# dashboard
+from main import models
+
+
+class RightsRowException(Exception):
+    def __init__(self, message, reader):
+        message = '[Row {}] {}'.format(reader.rows_processed + 1, message)
+        super(RightsRowException, self).__init__(message)
+
+
+class RightCsvReader(object):
+
+    metadata_applies_to_type = None
+    current_row = None
+    object_basis_act_usage = {}
+    rows_processed = 0
+
+    required_column_names = [
+        'file',
+        'grant_act',
+    ]
+
+    optional_column_names = [
+        'basis',
+        'status',  # mandatory for copyright
+        'determination_date',
+        'start_date',
+        'end_date',
+        'jurisdiction',  # mandatory for copyright and statute basis
+        'terms',
+        'citation',  # mandatory for statute basis
+        'note',
+        'grant_restriction',
+        'grant_start_date',
+        'grant_end_date',
+        'grant_note',
+        'doc_id_type',
+        'doc_id_value',
+        'doc_id_role'
+    ]
+
+    allowed_column_names = optional_column_names + required_column_names
+
+    def __init__(self, transfer_uuid, rights_csv_filepath):
+        """Initialize parser."""
+        #self.allowed_column_names = self.optional_column_names + self.required_column_names
+        self.transfer_uuid = transfer_uuid
+        self.rights_csv_filepath = rights_csv_filepath
+
+    def parse(self):
+        """Read and parse rights CSV file."""
+        # Cache metadata applies to type
+        self.metadata_applies_to_type = models.MetadataAppliesToType.objects.filter(description='File').first()
+
+        # Use universal newline mode to support unusual newlines, like \r
+        with open(self.rights_csv_filepath, 'rbU') as f:
+            reader = csv.DictReader(f)
+            for row in reader:
+                self.parse_row(row)
+
+        return self.rows_processed
+
+    def parse_row(self, row):
+        """Parse a row of a CSV file and create rights records in database."""
+        if self.current_row is None:
+            self.validate_columns(row)
+
+        self.current_row = row
+
+        # If no file specified, fail
+        filepath = self.column_value('file')
+        if not filepath:
+            raise RightsRowException('No file specified', self)
+
+        # If restriction specified, ensure it has an allowed value
+        restriction = self.column_value('grant_restriction')
+        if restriction and restriction.lower() not in ["disallow", "conditional", "allow"]:
+            raise RightsRowException("The value of element restriction must be: 'Allow', 'Disallow', or 'Conditional'", self)
+
+        # Initialize hash to note which basis/act combinations for file have alredy been imported
+        if not filepath in self.object_basis_act_usage:
+            self.object_basis_act_usage[filepath] = {}
+
+        basis = self.column_value('basis').lower().capitalize()
+
+        if not basis in self.object_basis_act_usage[filepath]:
+            self.object_basis_act_usage[filepath][basis] = {}
+
+        # Check that act is set and normalize value
+        act = self.column_value('grant_act')
+        if not act:
+            raise RightsRowException('No act specified', self)
+        act = act.lower().capitalize()
+
+        # Process row if basis/act combination for file hasn't yet been imported
+        if not act in self.object_basis_act_usage[filepath][basis]:
+            self.store_row()
+            self.object_basis_act_usage[filepath][basis][act] = True
+        else:
+            print("Skipping duplicate basis/act combination at row", self.rows_processed + 1)
+
+        self.rows_processed += 1
+
+    def store_row(self):
+        """Create rights records in database using row data."""
+        rights_statement = self.generate_rights_statement()
+
+        if rights_statement.rightsbasis == 'Copyright':
+            self.store_copyright_info(rights_statement)
+
+        elif rights_statement.rightsbasis == 'License':
+            self.store_license_info(rights_statement)
+
+        elif rights_statement.rightsbasis == 'Statute':
+            self.store_statute_info(rights_statement)
+
+        elif rights_statement.rightsbasis in ['Other', 'Donor', 'Policy']:
+            self.store_other_info(rights_statement)
+
+        self.store_grant_info(rights_statement)
+
+    def generate_rights_statement(self):
+        """Generate rights statement."""
+        basis = self.column_value('basis').lower().capitalize()
+
+        if basis not in dict(models.RightsStatement.RIGHTS_BASIS_CHOICES):
+            raise RightsRowException('Invalid basis: {}'.format(basis), self)
+
+        # Get file data
+        filepath = self.column_value('file')
+        transfer_file = models.File.objects.get(originallocation='%transferDirectory%' + filepath, transfer_id=self.transfer_uuid)
+
+        # Create rights statement
+        rights_statement = models.RightsStatement()
+        rights_statement.metadataappliestotype = self.metadata_applies_to_type
+        rights_statement.metadataappliestoidentifier = transfer_file.uuid
+        rights_statement.rightsbasis = basis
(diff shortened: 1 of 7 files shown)


src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+        'grant_start_date',
+        'grant_end_date',
+        'grant_note',
+        'doc_id_type',
+        'doc_id_value',
+        'doc_id_role'
+    ]
+
+    def __init__(self, transfer_uuid, rights_csv_filepath):
+        """Initialize parseri."""
```
Hwesta: Typo? s/parseri/parser

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+import sys
+
+import django
+django.setup()
+
+# dashboard
+from main import models
+
+
+class RightCsvReader:
```
Hwesta: This should inherit from `object`, unless you meant to create an old-style class. [More details](http://stackoverflow.com/questions/13816849/old-style-and-new-style-classes-in-python-2-7)

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+        """Initialize parseri."""
+        self.allowed_column_names = self.optional_column_names + self.required_column_names
+        self.transfer_uuid = transfer_uuid
+        self.rights_csv_filepath = rights_csv_filepath
+
+    def parse(self):
+        """Read and parse rights CSV file."""
+        # Print notice if no rights CSV file found
+        if not os.path.isfile(self.rights_csv_filepath):
+            print "No such file:", self.rights_csv_filepath
```
Hwesta: Since we're trying to write python2/3 compatible code, please use the new print format.  `from __future__ import print_function` at the top of the file will enable the syntax.

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+        'grant_end_date',
+        'grant_note',
+        'doc_id_type',
+        'doc_id_value',
+        'doc_id_role'
+    ]
+
+    def __init__(self, transfer_uuid, rights_csv_filepath):
+        """Initialize parseri."""
+        self.allowed_column_names = self.optional_column_names + self.required_column_names
```
Hwesta: Is this modified anywhere? Can it be moved to the class definition?

src/dashboard/src/main/migrations/0030_rights_import.py
```diff
@@ -0,0 +1,39 @@
...
+
+def data_migration(apps, schema_editor):
+    StandardTaskConfig = apps.get_model('main', 'StandardTaskConfig')
+    TaskConfig = apps.get_model('main', 'TaskConfig')
+    MicroServiceChainLink = apps.get_model('main', 'MicroServiceChainLink')
+    MicroServiceChainLinkExitCode = apps.get_model('main', 'MicroServiceChainLinkExitCode')
+
+    one_instance_task_type = '36b2e239-4a57-4aa5-8ebc-7a29139baca6'
+    characterize_and_extract_md_mscl = '303a65f6-a16f-4a06-807b-cb3425a30201'
+    fail_mscl = '61c316a6-0a50-4f65-8767-1f44b1eeb6dd' # TODO
```
Hwesta: Is this still TODO?

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+        self.rows_processed += 1
+
+    def exit_with_error(self, message, value=None):
+        """Print message to standard error and exit with status 1."""
+        message = '[Row {}] {}'.format(self.rows_processed + 1, message)
+        if value:
+            print >>sys.stderr, message, value
+        else:
+            print >>sys.stderr, message
+        exit(1)
```
Hwesta: I think this could be replaced by just raising an exception where the problem occurred. An exception prints to stderr, shows the traceback of the problem, and if unhandled with exit the program with a non-0 exit code.

On the other hand, I like that this prints the row with the error, while tracebacks don't.

If you keep it, this should use `sys.exit`, not `exit`. [Explanation](http://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if (transfer_file is None):
+            self.exit_with_error('Record not found for file', filepath)
+
+        # Create rights statement
+        rights_statement = models.RightsStatement()
+        rights_statement.metadataappliestotype = self.metadata_applies_to_type
+        rights_statement.metadataappliestoidentifier = transfer_file.uuid
+        rights_statement.rightsbasis = basis
+        rights_statement.status = 'ORIGINAL'
```
Hwesta, line 169: Was this tested with reingest? What happens if the rights.csv is added on reingest?

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+
+        for column_name in self.required_column_names:
+            if not column_name in row:
+                self.exit_with_error('Missing required column:', column_name)
+
+    def column_value(self, column_name):
+        """Return value of a row column by it's name in the header, None if missing/blank."""
+        if column_name in self.current_row:
+            value = self.current_row[column_name]
+            return None if value.strip() == '' else value
```
Hwesta: `.get` could simplify this function to 
```python
value = self.current_row.get(column_name, '').strip()
return value if value else None
```

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+            self.validate_columns(row)
+
+        self.current_row = row
+
+        # If no file specified, fail
+        filepath = self.column_value('file')
+        if not filepath:
+            self.exit_with_error('No file specified')
+
+        # If restriction specified, insure it has an allowed value
```
Hwesta: Typo: s/insure/ensure

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+        restriction = self.column_value('grant_restriction')
+        if restriction and not restriction.lower() in ["disallow", "conditional", "allow"]:
```
Hwesta: Tip: More pythonic to be `restriction.lower() not in ["disallow", "conditional", "allow"]`  Same for other `not ... in`

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+    def generate_rights_statement(self):
+        """Generate rights statement."""
+        basis = self.column_value('basis').lower().capitalize()
+
+        if basis not in dict(models.RightsStatement.RIGHTS_BASIS_CHOICES):
+            self.exit_with_error('Invalid basis:', basis)
+
+        # Get file data
+        filepath = self.column_value('file')
+        transfer_file = models.File.objects.filter(originallocation='%transferDirectory%' + filepath, transfer_id=self.transfer_uuid).first()
```
Hwesta: Has this been tested with non-ASCII filenames?

src/MCPClient/lib/clientScripts/rightsFromCSV.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if (transfer_file is None):
+            self.exit_with_error('Record not found for file', filepath)
```
Hwesta: If the queryset's [`.get`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#get) is used instead of `.first` above, then this error checking isn't needed (and the exception provides more details)

src/dashboard/src/main/migrations/0030_rights_import.py
```diff
@@ -0,0 +1,39 @@
...
+
+    rights_import_stc = '82579891-1cb7-486e-a0a6-b4cdf99c282f'
+    rights_import_tc = '637cc23a-c5f5-4df5-a352-5d47014be63b'
+    rights_import_mscl = '1a136608-ae7b-42b4-bf2f-de0e514cfd47'
+
+    # Introduce new rights import microservice
+    StandardTaskConfig.objects.create(id=rights_import_stc, requires_output_lock=False, execute='rightsFromCSV_v0.0', arguments='''%SIPUUID%'' ''%SIPDirectory%metadata/rights.csv''')
+    TaskConfig.objects.create(id=rights_import_tc, tasktype_id=one_instance_task_type, tasktypepkreference=rights_import_stc, description='Load rights')
+    MicroServiceChainLink.objects.create(id=rights_import_mscl, currenttask_id=rights_import_tc, defaultnextchainlink_id=fail_mscl, microservicegroup='Characterize and extract metadata', defaultexitmessage='Failed')
```
Hwesta, line 24: Is there a reason it's in Characterize & Extract metadata? There's a group for 'Reformat metadata files' that might be a better fit, or maybe a new group?

Hwesta (changes requested):
Looks good overall. Some stylistic or cleanup feedback, especially around Python2/3 compatibility.

Since this is adding a new, relatively standalone, feature, can you add tests? I'd like to see some example rights.csv files, and especially tests for things like non-ASCII filenames, checking that duplicate basis results in multiple acts, duplicate basis with different information to confirm what gets dropped.

There should be examples in the MCPClient's test folder for setting up database fixtures, and I recommend reusing the fixtures that are already there as much as possible. If you're unsure how to do this, please ask.

src/dashboard/src/main/migrations/0030_rights_import.py
mcantelon: Nope... removed.

src/dashboard/src/main/migrations/0030_rights_import.py
mcantelon: I'll look at moving it.

src/MCPClient/lib/clientScripts/rightsFromCSV.py
mcantelon: I renamed a file to have a unicode snowman and adjusted the CSV similarly and it found the rights data, so seems to work.

src/dashboard/src/main/migrations/0030_rights_import.py
mcantelon: (I ran it by Sara and Sarah and they think Characterize/Extract is probably best fit. There's a "load labels" microservice in that group which does something similar.)

src/MCPClient/lib/clientScripts/rightsFromCSV.py
mcantelon: The originally imported rights metadata stays in METS, but new/changed rights.csv aren't supported during reingest (I checked with Sara and this is how it's supposed to be as per the spec).

src/MCPClient/lib/clientScripts/rightsFromCSV.py
mcantelon: Nice!

Hwesta:
This was against qa/1.6.x, which has been replaced by stable/1.6.x  Changing what branch this is merging to and reopening.

mcantelon:
I'd added tests. Let me know if anything seems weird or missing in them. Thanks!

src/MCPClient/lib/clientScripts/rightsFromCSV.py
Hwesta: Ok, thanks for checking that.

src/MCPClient/lib/clientScripts/rightsFromCSV.py
Hwesta: Sorry I was unclear before - you don't need the `if column_name in self.current_row:`.  `.get` is like `dict[index]`, except if the entry isn't there it doesn't raise an exception, but returns the second parameter instead.

src/dashboard/src/main/migrations/0030_rights_import.py
Hwesta: Ok, good to know.

src/MCPClient/lib/clientScripts/rightsFromCSV.py
Hwesta: Great, and I see the test for unicode as well.

Hwesta (approved):
This looks good! The test file you provided worked, and the tests look like they cover all the common use cases.

mcantelon:
Awesome... Thanks Holly!

mcantelon:
Merged into qa/1.x and stable/1.6.x.

nickwilkinson:
[Dev/issue 10140 rights csv import 1 6 rebase](https://trello.com/c/jpJuUIz0/7-dev-issue-10140-rights-csv-import-1-6-rebase)

PR #578, opened 2017-03-01
osdeps: put OS dependencies information in repository

hakamine:
Add a directory osdeps/ for each archivematica component
(archivematicaCommon, dashboard, MCPServer, MCPClient) with files
containing information of OS dependencies. To be used in deployment
(ansible, etc.)
( Currently, files in osdeps/ directory are used by [PR104 of the ansible role](https://github.com/artefactual-labs/ansible-archivematica-src/pull/104) to be able to deploy qa/1.6.x on CentOS/RH 7 in addition to Ubuntu trusty )


src/MCPClient/osdeps/Ubuntu-14.json
```diff
@@ -0,0 +1,49 @@
...
+    "deb [arch=amd64] http://packages.archivematica.org/1.6.x/ubuntu-externals trusty main",
+    "deb http://archive.ubuntu.com/ubuntu/ trusty multiverse",
+    "deb http://archive.ubuntu.com/ubuntu/ trusty-security universe",
+    "deb http://archive.ubuntu.com/ubuntu/ trusty-updates multiverse"
+  ],
+  "osdeps_packages": [
+  { "name": "atool", "state": "latest"},
+  { "name": "bagit", "state": "latest"},
+  { "name": "bulk-extractor", "state": "latest"},
+  { "name": "ffmpeg", "state": "latest"},
```
Hwesta, line 22: Is there a way to specify minimum version? The control file lists `>=7:30`

src/MCPClient/osdeps/RedHat-7.json
```diff
@@ -0,0 +1,65 @@
+{
```
Hwesta: The RedHat and CentOS json lists are identical - do we need both?

src/MCPServer/osdeps/Ubuntu-14.json
```diff
@@ -0,0 +1,15 @@
...
+    "Include here keys, repos, and packages required by this component not already in archivematicaCommon osdeps file"
+  ],
+  "osdeps_repokeys": [
+  ],
+  "osdeps_repos": [
+  ],
+  "osdeps_packages": [
+  { "name": "dbconfig-common", "state": "latest"},
+  { "name": "logapp", "state": "latest"},
+  { "name": "uuid", "state": "latest"}
```
Hwesta, line 13: Should this also install gearman & a database?

src/dashboard/osdeps/Ubuntu-14.json
```diff
@@ -0,0 +1,13 @@
...
+  "_comments": [
+    "Add any comments here",
+    "Include here keys, repos, and packages required by this component not already in archivematicaCommon osdeps file"
+  ],
+  "osdeps_repokeys": [
+  ],
+  "osdeps_repos": [
+  ],
+  "osdeps_packages": [
+  { "name": "gettext", "state": "latest"}
```
Hwesta, line 11: There are several dependencies listed in the control file that are not in here - are they handled elsewhere?

Hwesta (commented):
Are we planning to replace the debian control files with this, or is this intended to be just for Ansible?  How do we plan to keep the dependencies & versions in sync?

src/MCPServer/osdeps/Ubuntu-14.json
scollazo: I don't think so. Gearman, elasticsearch and mysql are services archivematica relies on, but they should be handled on their own. 

I'm thinking about environments with their own mysql/es servers, were installing the db servers as part of archivematica doesn't make sense.

src/MCPClient/osdeps/Ubuntu-14.json
scollazo: I also think that we should add a known working version as minimum, keeping things explicit helps in the long term.

scollazo:
@Hwesta , yes, we plan to populate the `Depends:` field in debian/control files with the content provided by those. This should help handle different dependency versions for ubuntu 14.04 and 16.04, without having different control files for each distribution.

src/MCPServer/osdeps/Ubuntu-14.json
Hwesta: That makes sense - they don't need to be installed on the same machine, just available.

hakamine:
>Are we planning to replace the debian control files with this, or is this intended to be just for Ansible? How do we plan to keep the dependencies & versions in sync?

@Hwesta The idea is to try to put the dependencies in one place and reuse as much as possible (not only by ansible). That said, the current format of the .json files are made so that they are easily usable by ansible (I think those could probably be readily usable if we try to make containers with them).

Hwesta:
Ah, great. Sound like you have that covered.

I noticed the RedHat & CentOS lists are identical - do they both need to exist?

src/MCPClient/osdeps/Ubuntu-14.json
hakamine: The current format for the .json file is based on what the ansible apt / yum modules takes as input (which in turn is based on the way apt and yum OS commands work), it is possible to specify a fixed version or to install the "latest" version available from the repository, but not possible to specify a "minimum" version. 
As a workaround I think that we could add some parameters to the .json file to be used by the package building scripts (example `{ "name": "ffmpeg", "state": "latest", "debcontrol":">= 7:3.0"}`)

src/MCPClient/osdeps/RedHat-7.json
hakamine: I expected there could be some small differences between RH and CentOS but so far this wasn't the case. We could probably use a symlink to one for now (and split into separate files if we find any difference in the future)

src/dashboard/osdeps/Ubuntu-14.json
hakamine: I think some of the dependencies listed in the current control files may not be required. I tested deploying archivematica from source with the dependencies in the osdeps files in this PR and so far haven't found issues.

hakamine:
> I noticed the RedHat & CentOS lists are identical - do they both need to exist?

Currently the ansible role uses the variable `ansible_distribution` ( which gets the OS the target system is using ) to pick the correct osdeps .json file to use for deployment.
I expected there could be some small differences between RH and CentOS but so far this wasn't the case. We could probably use a symlink to one of them instead of having two files with duplicate content (and remove the symlink and have two separate files if we find any difference in the future).

src/dashboard/osdeps/Ubuntu-14.json
Hwesta: Good to know. Maybe worthwhile to consider removing them from the control file after this release?

Hwesta:
Ahh, Ansible wants the two versions.  I like the symlink idea.

Hwesta (approved):

hakamine:
Merged to qa/1.6.x and also cherry-picked to stable/1.6.x

hakamine:
Also cherry picked to qa/1.x

PR #579, opened 2017-03-10
Migration: Empty checksums are '' not None

Hwesta:
When setting the default value for old rows, check against '' not None, since the column is blank=True but null=False.

refs redmine 10966

Will be cherry-picked to qa/1.x


sevein (approved):

jhsimpson:
:+1:

Hwesta:
Cherry-picked to qa/1.x as 6ccd2002061a047e8bdf2e69707ef5c6a3e737c8

PR #580, opened 2017-03-11
Truncate client task stdout/stderror logged to db

hakamine:
Truncate client task stdout/stderror logged to the MCP database.

Some client scripts generate huge stdout/stderr messages, sometimes
exceeding hundreds of MB, causing database server errors when
attempting to save them.

Refs. #10732

### DIFF
--- a/src/archivematicaCommon/lib/databaseFunctions.py
+++ b/src/archivematicaCommon/lib/databaseFunctions.py
@@ -229,17 +229,29 @@ def logTaskCompletedSQL(task):
     print("Logging task output to db", task.UUID)
     taskUUID = task.UUID.__str__()
     exitCode = task.results["exitCode"].__str__()
-    stdOut = task.results["stdOut"]
-    stdError = task.results["stdError"]
+    # ``strToUnicode`` here prevents the MCP server from crashing when, e.g.,
+    # stderr contains Latin-1-encoded chars such as \xa9, i.e., the copyright
+    # symbol, cf. #9967.
+    stdOut = strToUnicode(task.results["stdOut"], obstinate=True)
+    stdError = strToUnicode(task.results["stdError"], obstinate=True)
+
+    # limit task stdout/stderror string size saved to database to 1MB
+    # (to prevent mysql server errors, refs. #10732) 
+    # if size limit exceeded, keep only initial and last part
+    TASK_STDOUT_LIMIT = 1024 * 1024
+    TASK_STDERR_LIMIT = 1024 * 1024
+    stdOut = (stdOut[:(TASK_STDOUT_LIMIT / 2 - 1)] + '..' +
+              stdOut[-(TASK_STDOUT_LIMIT / 2 - 1):]) \
+        if len(stdOut) > TASK_STDOUT_LIMIT else stdOut
+    stdError = (stdError[:(TASK_STDERR_LIMIT / 2 - 1)] + '..' +
+                stdError[-(TASK_STDERR_LIMIT / 2 - 1):]) \
+        if len(stdError) > TASK_STDERR_LIMIT else stdError
 
     task = Task.objects.get(taskuuid=taskUUID)
     task.endtime = getUTCDate()
     task.exitcode = exitCode
-    # ``strToUnicode`` here prevents the MCP server from crashing when, e.g.,
-    # stderr contains Latin-1-encoded chars such as \xa9, i.e., the copyright
-    # symbol, cf. #9967.
-    task.stdout = strToUnicode(stdOut, obstinate=True)
-    task.stderror = strToUnicode(stdError, obstinate=True)
+    task.stdout = stdOut
+    task.stderror = stdError
     task.save()
 
 


src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -229,8 +229,12 @@ def logTaskCompletedSQL(task):
...
     taskUUID = task.UUID.__str__()
     exitCode = task.results["exitCode"].__str__()
-    stdOut = task.results["stdOut"]
-    stdError = task.results["stdError"]
+    # truncate task stdout/stderror logged to the database
+    # (to prevent mysql server errors, refs. #10732) 
+    TASK_STDOUT_LIMIT = 16*1024*1024
+    TASK_STDERR_LIMIT = 16*1024*1024
+    stdOut = (task.results["stdOut"][:TASK_STDOUT_LIMIT] + '..') if len(task.results["stdOut"]) > TASK_STDOUT_LIMIT else task.results["stdOut"]
+    stdError = (task.results["stdError"][:TASK_STDERR_LIMIT] + '..') if len(task.results["stdError"]) > TASK_STDERR_LIMIT else task.results["stdError"]
```
sevein: Do you know if the types of `task.results["stdError"]` and `task.results["stdOut"]` are always going to be Py2 `str` (aka `bytes`)? If that's the case I think we'll be fine but if you expect `unicode` strings then we should [encode](https://github.com/artefactual/archivematica/blob/stable/1.5.x/src/archivematicaCommon/lib/archivematicaFunctions.py#L59-L62) them before so we can effectively see its size in bytes when using `len()`. @jrwdunham may know more about this, he's dealt with similar stuff before.

hakamine: @jrwdunham @Hwesta any advice? I noticed stdOut and StdError are converted to unicode with strtoUnicode() after the truncation is done, should it be done before to address what @sevein pointed out?

I also have a question about the truncation limit size, This PR currently proposes to truncate at 16MB. As the purpose of these strings seem to be only to show the output of the command to the user in the dashboard, we may not even need that much? Maybe storing/showing say, the first and last MB of stdout and stderr could be more than enough?

jrwdunham: Yes, I think you should truncate the stdout/stderr strings *after* they are unicode. The strToUnicode func will raise a UnicodeDecodeError (unless you pass obstinate=True). You could easily raise such an exception by truncating bytes and decoding the result to Unicode, e.g., `'a\u0301'.encode('utf8')[:2].decode('utf8')` would do it.

jrwdunham: As @sevein points out, if you truncate a Python Unicode string at index 16x1024x1024, you're not guaranteed to get something that, when encoded (e.g., as UTF-8), is 16x1024x1024 bytes. But I don't think you have a hard requirement (e.g., a database constraint) about that number, right @hakamine? You just want to shorten these strings to some fixed approximate maximum. 

    >>> a = 'Jesu\u0301s'
    >>> print(a)
    Jesús
    >>> a.encode('utf8')
    b'Jesu\xcc\x81s'
    >>> len(a)
    6
    >>> a.encode('utf8')
    b'Jesu\xcc\x81s'
    >>> len(a.encode('utf8'))
    7
    >>> a.encode('utf8')[:5]
    b'Jesu\xcc'
    >>> a.encode('utf8')[:5].decode('utf8')
      File "<stdin>", line 1, in <module>
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcc in position 4: unexpected end of data

sevein: @jrwdunham thanks for clarifying!

hakamine:
Thank you for taking the time to explain me about the problems of truncating before doing the unicode decoding. I changed the PR code so that truncation is done on the decoded string. Also fixed in order to keep the initial and last part of the stdout/stderror (which hopefully contain the most helpful part of the output to the user seeing the log). I did a quick test on a VM and the truncation seems to be working fine.

sevein:
@hakamine, sorry that this has been sitting for so long. I have one question about this.

The process is executed by MCPClient (see details [here](https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/MCPClient/lib/archivematicaClient.py#L185-L198)). Simplified version is:

```
exitCode, stdOut, stdError = executeOrRun("command", command, sInput, printing=False, env_updates=env_updates)
return cPickle.dumps({"exitCode": exitCode, "stdOut": stdOut, "stdError": stdError})
```

Ultimately `executeOrRun` uses the [`launchSubProcess`](https://github.com/artefactual/archivematica/blob/a1bf8ca99799031fbad5204cc93e911bde91a573/src/archivematicaCommon/lib/executeOrRunSubProcess.py#L31-L95) function which is basically doing the following:

```
p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=stdin_pipe, env=my_env)
stdOut, stdError = p.communicate(input=stdin_string)
retcode = p.returncode
return retcode, stdOut, stdError
```

So the MCPClient daemon is buffering all the output of the subprocess in memory and once the subprocess finishes the bytes are written to `stdOut` and `stdErr`. Finally, we use `cPickle.dumps` to serialize the dictionary containing the `exitCode`, `stdOut` and `stdError` keys and the result is sent over the network to gearmand which is forwarding that to MCPServer, which has a thread blocked waiting for the response. Once received (see `taskStandard.py` in MCPServer), `taskCompletedCallBackFunction` is invoked, which finally invokes the function you're editing: `databaseFunctions.logTaskCompletedSQL`.

So I'm wondering if we would achieve better results if we did this optimization early rather than later? We could save MCPClient from buffering all the output of the task to memory and from serializing that using cPickle, sending that over the network to gearmand, etc...

I don't know if we can easily replace `communicate()` (this [solution](http://stefaanlippens.net/python-asynchronous-subprocess-pipe-reading/) is interesting but it would need some tweaks to fit our needs, i.e. limit size), but at least we could try to truncate the output that early, in MCPClient.

sevein (changes requested):
@hakamine is going to look into moving the concat op up in the stack.

sevein:
Something to consider is that the MCP writes the output of some tasks to disk. This is true when `StandardTaskConfig.stdout_file` or `StandardTaskConfig.stderr_file` are not null. These are the different values I've seen, in some cases more than once:

***stdout***

- `%SIPLogsDirectory%filenameCleanup.log`
- `%SIPLogsDirectory%fileFormatIdentification.log`
- `%SIPLogsDirectory%extractContents.log`
- `%SIPLogsDirectory%verifyChecksumsInFileSecOfDSpaceMETSFiles.log`
- `%SIPLogsDirectory%FileUUIDs.log`
- `%SIPLogsDirectory%removedFilesWithNoPremisMetadata.log`
- `%SIPLogsDirectory%removeUnneededFiles.log`

***stderr***

- `%SIPLogsDirectory%filenameCleanup.log`
- `%SIPLogsDirectory%fileFormatIdentification.log`
- `%SIPLogsDirectory%clamAVScan.txt`
- `%SIPLogsDirectory%verifyChecksumsInFileSecOfDSpaceMETSFiles.log`
- `%SIPLogsDirectory%FileUUIDsError.log`
- `%SIPLogsDirectory%removedFilesWithNoPremisMetadata.log`
- `%SIPLogsDirectory%SIPnameCleanup.log`
- `%SIPLogsDirectory%removeUnneededFiles.log`

sevein:
Replaced by #804 and #805.

PR #582, opened 2017-03-23
Hide all units button, refs #595 #10894

jrwdunham:


src/dashboard/src/components/ingest/views_as.py
```diff
@@ -13,9 +14,12 @@
 
 from components.ingest import pair_matcher
 
+logger = logging.getLogger('archivematica.dashboard')
 
 def get_as_system_client():
     repl_dict = models.MicroServiceChoiceReplacementDic.objects.get(description='ArchivesSpace Config')
+    logger.debug('repl_dict.replacementdic')
+    logger.debug(repl_dict.replacementdic)
```
Hwesta: Did you mean to commit this?  Looks like troubleshooting something unrelated.

src/dashboard/tests/test_api.py
```diff
@@ -29,75 +30,89 @@ def test_get_unit_status_processing(self):
         load_fixture(['jobs-processing.json'])
         # Test
         status = views.get_unit_status('3e1e56ed-923b-4b53-84fe-c5c1c0b0cf8e', 'unitTransfer')
+        completed = helpers.completed_units_efficient(unit_type='transfer', include_failed=True)
         # Verify
         assert len(status) == 2
         assert 'microservice' in status
         assert status['status'] == 'PROCESSING'
+        assert len(completed) == 0
```
Hwesta, line 35: Can you add some tests where `completed` isn't 0 or 1?

src/dashboard/src/components/api/views.py
```diff
@@ -221,6 +223,31 @@ def mark_hidden(request, unit_type, unit_uuid):
...
+    """
+    auth_error = authenticate_request(request)
+    response = {}
+    if auth_error is not None:
+        response = {'message': auth_error, 'error': True}
+        return django.http.HttpResponseForbidden(
+            json.dumps(response),
+            content_type='application/json'
+        )
+    return unit_views.mark_completed_hidden(request, unit_type)
```
Hwesta: This will interact with #521 when it gets merged.

src/dashboard/src/components/helpers.py
```diff
@@ -329,3 +330,42 @@ def stream_file_from_storage_service(url, error_message='Remote URL returned {}'
...
+    if unit_type != 'transfer':
+        table_name = 'SIPs'
+        pk_name = 'sipUUID'
+    q = ('SELECT u.{pk_name}, j.jobType, j.microserviceGroup'
+         ' FROM {table_name} as u'
+         ' INNER JOIN Jobs as j'
+         ' ON j.SIPUUID = u.{pk_name}'
+         ' WHERE u.hidden = 0'
+         ' ORDER BY u.{pk_name}, j.createdTime DESC, j.createdTimeDec DESC;'.format(
+             table_name=table_name, pk_name=pk_name))
```
Hwesta, line 362: This may be easier with triple-quote strings.  E.g.

```
q = '''SELECT ...
       FROM ...
       ...
    '''
```

Hwesta (commented):
Looks good! I like the CSS changes to make the popup more readable.  A minor cleanup, and request for tests that touch more than one unit if possible.

src/dashboard/src/components/helpers.py
jrwdunham: I'm gonna leave this as is unless someone strongly objects. I prefer concatenations of single-line/quote strings over triple-quote strings because you can more easily control the whitespace in the final string.

src/dashboard/tests/test_api.py
jrwdunham: Done. See `TestAPI2.test_get_unit_status_multiple`.

sevein:
@jrwdunham, can you rebase?

sevein (changes requested):
@jrwdunham, can you rebase?

src/dashboard/src/components/api/views.py
```diff
@@ -113,7 +113,10 @@ def get_unit_status(unit_uuid, unit_type):
     :return: Dict with status info.
     """
     ret = {}
-    job = models.Job.objects.filter(sipuuid=unit_uuid).filter(unittype=unit_type).order_by('-createdtime', '-createdtimedec')[0]
+    job = models.Job.objects.filter(
+        sipuuid=unit_uuid).filter(
+            unittype=unit_type).order_by(
+                '-createdtime', '-createdtimedec')[0]
```
sevein: I'm assuming you're making this change to increase readability? I don't mind one way or the other. Would you be onboard if we try to agree on a rule and enforce it using a linter? Would that be possible in Python? I'd almost prefer to follow one style consistently even if it's not the more Pythonic way to do it.

src/dashboard/src/components/helpers.py
sevein: Leave it, I think it's good!

We've written SQL queries in other places, e.g. emailFailReport.py, [forms.py](https://github.com/artefactual/archivematica/blob/92d7abd238585e64e6064bc3f1ddfc663c4d3ace/src/dashboard/src/components/administration/forms.py#L435-L454), etc... In the long term I think we should try to stick with whatever is the current convention. We may be inconsistent though, in that case I'd recommend to document the rule that you want us to follow. I don't know if we have started a style guide yet...

jrwdunham: In terms of style guides, we have this: https://wiki.archivematica.org/Contribute_code#Code_Style_Guide_For_Archivematica

src/dashboard/src/components/api/views.py
jrwdunham: Holly and I have been debating this for quite some time. I'm more of a stickler for PEP 8's "Limit all lines to a maximum of 79 characters", cf. https://www.python.org/dev/peps/pep-0008/#maximum-line-length. I have a vim shortcut to make breaking lines easier. And pylint complains about 79-char+ lines by default. And I kinda think you should be able to print your code on a piece of a paper and have it be readable.

That said, pretty much all of the AM codebase ignores this 79 character rule, so being fully consistent would requite a LOT of pointless work. And modern editors can make your long lines quite readable. So I'm open to undoing the above change.

A third option would be for me (and other 79-character sticklers) to not change existing code but continue to insert linebreaks in new code. I can make my editor make long lines readable while still inserting linebreaks in my contributions. I think on this one point of style we can have two opinions/styles co-existing, so this is what I would prefer. Thoughts?

sevein: > A third option would be for me (and other 79-character sticklers) to not change existing code but continue to insert linebreaks in new code. I can make my editor make long lines readable while still inserting linebreaks in my contributions. I think on this one point of style we can have two opinions/styles co-existing, so this is what I would prefer. Thoughts?

That makes sense, thanks for clarifying it.

jrwdunham:
@sevein rebased

sevein (changes requested):
Small change needed: new strings in `src/dashboard/src/media/js/jobs.js` should use the gettext function. You have some examples in that same file. Thank you!

src/dashboard/src/components/api/views.py
sevein: @jrwdunham see #600, I've added linting support in CI.

sevein (approved):
Thanks @jrwdunham, it looks great. I rebased your branch and added a new commit to update the fixture to use the new `Job.currentstep` enum. I needed that for the tests to run (in addition to #601).

PR #584, opened 2017-03-31
Models: stop using loaddata in 0002, fixes #583

### LINKED ISSUE #583: Problem: 0002 migration uses loaddata to import a fixture
`loaddata` can only use existing models. During a migration, we should be using the models provided by the migration system - this avoids a number of issues like making possible to delete existing models (e.g. #576 aims to delete workflow models).

sevein:
Using `loaddata` in a migration is undesirable for various reasons. This commit translates the
original fixture into ORM calls. Fixes #583.


sevein:
I generated the code from a simple script I wrote: https://gist.github.com/sevein/324220395651d89ca04c9376320cc427. I ended up doing an extra cleanup from my editor.

I'm not sure what would be the best way to test that this is working as expected. One idea is to use `mysqldump --extended-insert=FALSE --complete-insert=TRUE` to generate a dump of the table and diff them.

Edit: @Hwesta suggested that `manage.py dumpdata` could also be used for the same effect.

sevein:
I've compared two MySQL dumps, with and without this fix - these are the results: https://gist.github.com/sevein/213c06ee337b8be79e96e03c4bf823d9.

I think the differences are due to the fact that creation timestamps now get a value because of `auto_now` - that didn't seem to happen when using `loaddata`.  In some cases, the creation timestamps are just different which is okay.

sevein:
Rebased.

jhsimpson:
This is a good idea, and this change looks good to me, but I have one question.  I feel a bit odd about rewriting existing migrations - I am not sure why but it seems like a bad idea to me.  What about the idea of squashing the first set of migrations into a single 1.5.1 release migration, and squashing the 2nd set into a 1.6.0 migration?  I don't think we need to have 0002 still around as an individual migration anymore.

Perhaps this suggestion could be done later, after this PR is merged?

sevein:
You can't squash migrations that make use of loaddata AFAIK.

jhsimpson (approved):
I have been looking at this on and off, and testing with it.  Looks good to me, I think it should be merged now.

PR #588, opened 2017-04-06
Added Upload DIP choice to processing config interface.

helenst:
this is @jrwdunham's fix for this, cherrypicked against qa/1.x


axemanyo10:
I have tested this branch and it works - 'Upload DIP' appears as an option in the configuration template

sevein:
Thanks, this is now in qa/1.x (see 5407c5bf).

PR #600, opened 2017-04-23
Problem: there are no linting checks

sevein:
Solution:
- Use tox + flake8, run from Travis CI. Merge #599 first.
- Use [autopep8](https://github.com/hhatto/autopep8) where possible to solve issues reported by linter.
- Use `max-line-length = 160` (see [discussion](https://github.com/artefactual/archivematica/pull/582#discussion_r111644917)).
- Ignore errors that have not been addressed yet, see below.

### Remaining pep8 errors, not addressed in this pull request

Count | Error | Description
----- | ----- | ------------------------------------------
295  | E402 | module level import not at top of file
349 | E501 | line too long (92 > 79 characters)

Total: 644 errors left.


src/archivematicaCommon/lib/externals/HTML/HTML.py
```diff
@@ -283,27 +301,27 @@ def __str__(self):
...
+# except: pass
+# try:
+# if self.col_align[i]:
+#                     col += ' align="%s"' % self.col_align[i]
+# except: pass
+# try:
+# if self.col_valign[i]:
+#                     col += ' valign="%s"' % self.col_valign[i]
+# except: pass
+#             result += '<COL%s>\n' % col
```
helenst: This breaks the formatting of the code inside the comment.

src/archivematicaCommon/lib/externals/HTML/HTML.py
```diff
@@ -362,67 +380,73 @@ def __init__(self, lines=None, ordered=False, start=None, attribs=None):
...
+#
+# def __str__(self):
+#         """return the HTML code for the link as a string"""
+#         attribs_str = ""
+#         if self.url:  self.attribs['href'] = self.url
+# for attr in self.attribs:
+#             attribs_str += ' %s="%s"' % (attr, self.attribs[attr])
+# return '<a%s>%s</a>' % (attribs_str, text)
+
+
```
helenst: again, breaks format of commented out code (maybe it could be removed?)

src/archivematicaCommon/lib/externals/HTML/HTML.py
```diff
@@ -437,58 +461,58 @@ def list(*args, **kwargs):
...
         """
-##        # First, header row:
-##        yield TableRow(('x', 'square(x)'), header=True, bgcolor='blue')
-##        # Then all rows:
-        for x in range(1, n+1):
-            yield (x, x*x)
+# First, header row:
+# yield TableRow(('x', 'square(x)'), header=True, bgcolor='blue')
+# Then all rows:
+        for x in range(1, n + 1):
```
helenst: comment should be indented (though also looks as if it could just be removed)

src/MCPClient/lib/clientScripts/extractContents.py
```diff
@@ -69,17 +74,17 @@ def main(transfer_uuid, sip_directory, date, task_uuid, delete=False):
         try:
             format_id = FileFormatVersion.objects.get(file_uuid=file_.uuid)
         # Can't do anything if the file wasn't identified in the previous step
-        except:
+        except BaseException:
```
helenst: Catching `BaseException` is still very much a catch-all (it'll still mask things like name errors) and while this change keeps the linter from complaining I'm not sure if that's a good thing.

For example, here we could catch `FileFormatVersion.DoesNotExist`. There are several cases of this - deciding which exception types should be handled for each might be out of scope for this piece of work, but perhaps the linter should be allowed to keep note of this until they can be addressed more carefully.

sevein:
@helenst thank you for your feedback!
#599 is now merged and I've updated this PR to target `qa/1.x`.

helenst:
@sevein Just that one whitespace error in `src/archivematicaCommon/lib/custom_handlers.py` and then I think we're good (a local run comes up with no other errors)

jrwdunham:
👍 Looks good to me. We found more potential issues by running pylint, but we'll deal with those in other PR(s)

PR #601, opened 2017-04-24
Problem: tests don't run

sevein:
Problems:
- Permission issue when running tests because logs are written to `/var/log/archivematica` which is owned by `archivematica`. The solution is to log to `stderr` when running tests.
- Fix migration issue that was breaking the test suite, `models.py` was deleted but it turns out that it's important to leave the file there even if it's empty so the migrations are executed. The last migration was to delete the models, which is still important.

This was affecting `qa/1.x` only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@
 
 # coverage data file
 .coverage
+
+# pytest's working directory
+.cache
--- a/src/archivematicaCommon/lib/custom_handlers.py
+++ b/src/archivematicaCommon/lib/custom_handlers.py
@@ -33,7 +33,12 @@ def get_script_logger(name, formatter=SCRIPT_FILE_FORMAT, logfile="/var/log/arch
                 'formatter': 'fmt',
                 'filename': logfile,
                 'maxBytes': logsize,
-            }
+            },
+            'console': {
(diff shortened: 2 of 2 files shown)


sevein:
Just to be clear, this is how you can verify in the Vagrant box:

```
cd /vagrant/src/archivematica
env PYTHONPATH="/vagrant/src/archivematica/src/dashboard/src:/vagrant/src/archivematica/src/archivematicaCommon/lib" py.test -x
```

Before, `sudo -u archivematica` was required and there was an error caused by the fact that the migrations under `components/administration` were not being processed due to the fact that the empty `models.py` file was missing.

sevein:
@hakamine if this is the first time running the test suite for you, you should see this at the very end:

```137 passed, 4 skipped, 2 xfailed in 96.74 seconds```

Don't worry about xfails, those are tests expected to fail :)

hakamine:
great, thanks! with the fixes tests are working for me :+1:

sevein:
Thank you @hakamine. @jrwdunham please let me know if the changes are ok for you.

jrwdunham (approved):
Yep. Works for me too. Thanks for fixing this!

PR #602, opened 2017-04-25
Bring back stderr in tasks!

jrwdunham:
Fixes a typo from 8485853e436db7f3e36a70bfe903c887b5e89b5b that resulted in Task standard error not being displayed in GUI.


sevein:
Thank you for fixing this!

PR #604, opened 2017-04-25
Problem: apache/django.wsgi is unused

sevein:
We went all in with Gunicorn and we don't want to maintain two types of environments. Having `apache/django.wsgi` around sends the wrong message to our users. We shouldn't keep it if we don't have the capacity to support it.

Solution: delete `apache/django.wsgi` and its module `monitor.py` not used by other modules.


hakamine (approved):
👍 I agree, it's better to remove these to avoid confusion. 
I will create an issue in https://github.com/artefactual-labs/ansible-archivematica-src to remove the apache option in deployment.
Do we also need something like this for the Storage Service?

sevein:
Thank you. I think SS got it removed already at some point.

hakamine:
Correcting myself: for the Storage Service I think we never used apache, but used nginx with uwsgi, before nginx with gunicorn.

PR #613, opened 2017-05-11
Update os dependencies in dashboard

### LINKED ISSUE #591: Problem: missing osdeps break pip requirements installation with ansible role (pipeline only install)
When doing a pipeline-only installation with ansible (no storage service), installation of pip requirements fail due to the following missing package dependencies ( getting errors here:https://github.com/artefactual-labs/ansible-archivematica-src/blob/stable/1.6.x/tasks/pipeline-pipdeps-multi-venvs.yml#L35-L47 ) :

- libxml2-dev
- libxslt1-dev
- libssl-dev

(happened in ubuntu 14, probably a similar problem will occur in CentOS/RH)

jhsimpson:
Fixes #591 

Ubuntu os dependencies listed in the debian/control file are used
when building .deb packages.
Other install methods (i.e., ansible from deploy-pub) use the files
in osdeps.

This commit updates all those files to be consistent with one another.
It also move some python dependencies from the debian/control file
and puts them in pip requirements instead.


src/dashboard/src/requirements/base.txt
```diff
@@ -13,13 +14,13 @@ git+https://github.com/artefactual/archivematica-fpr-admin.git@v1.3.0-beta.1#egg
...
 pytz
 pyopenssl
-python-dateutil==2.4.2
+python-dateutil==2.6.0
 ndg-httpsclient
 pyasn1
 requests==2.7.0
-lxml==3.5.0
-django-forms-bootstrap>=3.0.0,<4.0.0
+simplejson==3.10.0
```
sevein: We don't need simplejson anymore since f2e296f4d79156f8bfdb62572afb6ea3d8a065b4 / b6eecbfff776ad61b5711230e788f9c6ef2b1729.

sevein: ... ^ fixed the links to the commit revs.

sevein (changes requested):
simplejson not needed! See above.

sevein:
I've deleted simplejson myself in this branch.

PR #615, opened 2017-05-11
Delete unneeded TravisCI env

sevein:
Combining `python: 2.7` and `matrix` was causing TravisCI to create two
different build jobs. One with the `TOXENV` defined and one without
it. We only need the one defined under the `matrix` section.

To clarify that, I'm trying to avoid the following:

image

The first build in the list is redundant. Having only one job should help to speed up the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ sudo: false
 
 language: python
 
-python:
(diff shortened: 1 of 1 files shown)


jrwdunham:
👍

PR #616, opened 2017-05-13
GPG encryption: encrypted AIPs identified

jrwdunham:
AIPs in GPG-encrypted spaces are marked as "encrypted" in the archival storage tab.

See https://github.com/artefactual/archivematica-storage-service/pull/192

### DIFF
--- a/src/dashboard/src/components/archival_storage/views.py
+++ b/src/dashboard/src/components/archival_storage/views.py
@@ -129,12 +129,12 @@ def search(request):
             }
             index = 'aips'
             doc_type = 'aip'
-            fields = 'name,uuid,size,created,status,AICID,isPartOf,countAIPsinAIC'
+            fields = 'name,uuid,size,created,status,AICID,isPartOf,countAIPsinAIC,encrypted'
             sort = 'name:desc'
         else:
             index = 'aips'
             doc_type = 'aipfile'
-            fields = 'AIPUUID,filePath,FILEUUID'
+            fields = 'AIPUUID,filePath,FILEUUID,encrypted'
             sort = 'sipName:desc'
 
         # To reduce amount of data fetched from ES, use LazyPagedSequence
@@ -208,6 +208,7 @@ def search_augment_aip_results(raw_results, counts):
             'countAIPsinAIC': _get_es_field(fields, 'countAIPsinAIC', '(unknown)'),
             'type': 'AIC' if 'AIC#' in _get_es_field(fields, 'AICID', '') else 'AIP',
             'status': AIP_STATUS_DESCRIPTIONS[_get_es_field(fields, 'status', 'UPLOADED')],
+            'encrypted': _get_es_field(fields, 'encrypted', False),
             'document_id_no_hyphens': item['_id'].replace('-', '____'),
         }
         size = _get_es_field(fields, 'size')
@@ -230,7 +231,7 @@ def search_augment_file_results(es_client, raw_results):
         # try to find AIP details in database
         try:
             # get AIP data from ElasticSearch
-            aip = elasticSearchFunctions.get_aip_data(es_client, clone['AIPUUID'], fields='uuid,name,filePath,size,origin,created')
+            aip = elasticSearchFunctions.get_aip_data(es_client, clone['AIPUUID'], fields='uuid,name,filePath,size,origin,created,encrypted')
 
             # augment result data
             clone['sipname'] = aip['fields']['name'][0]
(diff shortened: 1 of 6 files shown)


src/dashboard/src/templates/archival_storage/list.html
```diff
@@ -96,7 +96,11 @@
       <tbody>
         {% for item in aips %}
           <tr>
-            <td><a href="{% url 'view_aip' item.uuid %}">{{ item.name }}</a></td>
+              {% if item.encrypted %}
```
sevein: Our indentation in HTML files uses two spaces instead of four.

src/dashboard/src/templates/archival_storage/list.html
```diff
@@ -96,7 +96,11 @@
...
+              <td><a href="{% url 'view_aip' item.uuid %}">{{ item.name }} ({% trans "encrypted" %})</a></td>
```
sevein: I think it's better to do as follows:

```django
{% blocktrans with aip_name=item.name %}{{ aip_name }} encrypted{% endblocktrans %}
```

With that the translator sees `{{ aip_name }} encrypted` and they can freely move the placeholder and fully control the structure of the sentence.

sevein (changes requested):

src/dashboard/src/templates/archival_storage/list.html
jrwdunham: Ok. I see you removed the parens too. Did you mean to do that? To me, they indicate that the "encrypted" string is a qualifier on the name of the AIP and not the part of the name itself, i.e., "My AIP (encrypted)" vs "My Other AIP". Are you fine with::

```
    {% blocktrans with aip_name=item.name %}{{ aip_name }} (encrypted){% endblocktrans %}
```

and then the translator can remove the parens if they see fit?

sevein: Oh I didn't see the parens! Sorry about that! `{{ aip_name }} (encrypted)` looks better to me but I think both ways work, up to you.

jrwdunham: Ok. I think I've done that. See 4e74f713e88b1863e6db745ae7b6f20ca3f48e52

sevein:
Rebased.

src/archivematicaCommon/lib/storageService.py
```diff
@@ -238,6 +238,36 @@ def get_files_from_backlog(files):
...
+    params = {
+        'access_protocol': access_protocol,
+        'path': path,
+        'offset': 0,
+    }
+    while True:
+        response = _storage_api_session().get(url, params=params)
+        spaces = response.json()
+        return_spaces += spaces['objects']
+        if not spaces['meta']['next']:
```
jraddaoui: We're trusting in a response field to end this `while` loop, which someday may end up being an endless loop. Do we have a way to get the total count of spaces from the SS?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -566,6 +568,38 @@ def es_pager(page, page_size):
...
+        if not location:
+            continue
+        for spc in spaces:
+            if spc['uuid'] in location['space']:
+                space = spc
+                break
+        if not space:
+            continue
+        if space['access_protocol'] == 'GPG':
+            aip['encrypted'] = True
```
jraddaoui: Could we return this property directly from the SS?

jraddaoui (commented):
Looks good to me @jrwdunham!

But I think it's a little overkill just to know which AIPs are encrypted. Could we add 'encrypted' as a property to the AIPs in the SS, so it's returned directly while fetching them?

src/archivematicaCommon/lib/storageService.py
jrwdunham: I take your point, but we're trusting that the tastypie API to always return a boolean for `['meta']['next']` and to eventually return `False`. I think this is ok. The weaker argument is that this is how this was done previously (this is old code) and how it still is done in other parts of this module. It's moot now though because I'm removing this code in favour of that which indicates encryption status using ES AIP indexing.

src/dashboard/src/components/archival_storage/views.py
jrwdunham: Yes, and this is essentially what sevein advised too. I'll modify this PR to indicate AIP encryption status in the Elasticsearch index. This way it is computed once and will require no extra requests.

jraddaoui (approved):
Great! Thank you @jrwdunham!

PR #621, opened 2017-05-26
TRIM: update to use Django models

jhsimpson:
refs #11161

The TRIM transfer type is normally not available in an Archivematica
dashboard.  If that option is enabled, a workflow is triggered that
uses these four scripts.

These scripts have not been updated since before the 1.4.0 release.
They call getFileUUIDLike(), a method that is defined in
archivematicaCommon/fileOperations.py, and that was updated in
version 1.4.0 to use Django Models.

As a result, the list of permissible values for the 4th parameter
changed.

This commit updates the TRIM workflow to call getFileUUIDLike
correctly.

### DIFF
--- a/src/MCPClient/lib/clientScripts/trimCreateRightsEntries.py
+++ b/src/MCPClient/lib/clientScripts/trimCreateRightsEntries.py
@@ -2,7 +2,7 @@
 
 # This file is part of Archivematica.
 #
-# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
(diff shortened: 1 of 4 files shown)


jhsimpson:
This PR has been made against stable/1.6.x.  The commit here will need to be cherry-picked into qa/1.x as well.

sevein (approved):

jhsimpson:
cherry-picked into qa/1.x branch as 4131fd5cac6f10538e856c686652f5d073b1c32d

nickwilkinson:
[TRIM: update to use Django models](https://trello.com/c/kJP2iejo/3-trim-update-to-use-django-models)

PR #626, opened 2017-06-02
Set filepath for zip files correctly when starting transfer

### LINKED ISSUE #625: Problem: Can't start DSpace Transfer
see https://projects.artefactual.com/issues/11224

When trying to use the start_transfer api endpoint in the dashboard rest api, if the transfer being started contains zip files, a 500 server error is reported. This doesn't happen when the transfer only contains folders.

DSpace Transfers are zip files, so this bug make it impossible to start a transfer of type 'dspace' through the api (i.e. from automation tools).

jhsimpson:
fixes #625 

refs #11224

When using the dashboard rest api to start a transfer, if the contents
of the new transfer contain a zip file, the start_transfer rest api
endpoint was returning a 500 server error, because the path to the
zip file was not getting set correctly.

This commit resolves this issue.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -246,7 +246,9 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
 
         if helpers.file_is_an_archive(path):
             transfer_dir = temp_dir
-            filepath = os.path.join(temp_dir, os.path.basename(path))
+            p = path.split(':', 1)[1]
+            logger.debug('found a zip file, splitting path ' + p)
+            filepath = os.path.join(temp_dir, os.path.basename(p))
         else:
             path = os.path.join(path, '.')  # Copy contents of dir but not dir
             transfer_dir = os.path.join(temp_dir, target)


jhsimpson:
I borrowed this fix from a working production Archivematica system, and tested it on a 2nd install as well.

This should be cherry-picked into the qa/1.x branch also.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -246,7 +246,9 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
 
         if helpers.file_is_an_archive(path):
             transfer_dir = temp_dir
-            filepath = os.path.join(temp_dir, os.path.basename(path))
+            p = path.split(':',1)[1]
+            logging.exception('splitting path ' + p)
```
jrwdunham: I think you should use a `logger` here not `logging`, also should probably use the `info` method instead of the `exception` one. Is there a reason you're using the latter?

jrwdunham (approved):

src/dashboard/src/components/filesystem_ajax/views.py
jhsimpson: I just copied this directly from a working production system and tested it on another.  I will make the change you suggested now.  I think this could be a debug message, it is not really necessary to have at the info level.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -246,7 +246,9 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
 
         if helpers.file_is_an_archive(path):
             transfer_dir = temp_dir
-            filepath = os.path.join(temp_dir, os.path.basename(path))
+            p = path.split(':',1)[1]
```
sevein: If you don't add a space after the comma the linter will fail in `qa/1.x`.

jhsimpson: thanks good catch, I am doing this on a temporary machine, not usual dev environment.

jhsimpson:
cherry-picked to qa/1.x as 5e93dccc7ed6a03d19e84e4e1fd0475763313761

nickwilkinson:
[Set filepath for zip files correctly when starting transfer](https://trello.com/c/kJP2iejo/3-set-filepath-for-zip-files-correctly-when-starting-transfer)

PR #629, opened 2017-06-05
clientscripts: change date format in verifyMD5.sh

scollazo:
Problem: Some AIPs can't be saved to samba network shares.

The default date format uses ":", and that's a reserved character for CIFS (samba) mounts, so files named like "logs/checksum.md5-Check-Tue May 16 16:59:21 UTC 2017" can't be stored on them. This affects saving some aips to network filesystems using samba

This commit replaces colons with dots:
 From: Mon Jun  5 14:38:47 EDT 2017
 To:      Mon Jun  5 14.39.17 EDT 2017

### DIFF
--- a/src/MCPClient/lib/clientScripts/verifyMD5.sh
+++ b/src/MCPClient/lib/clientScripts/verifyMD5.sh
@@ -23,6 +23,7 @@
 # @version svn: $Id$
 
 checkMD5NoGui="`dirname $0`/archivematicaCheckMD5NoGUI.sh"
+dateArgs="-u +%Y%m%dT%H%M%SZ"  # 20170605T220452Z (UTC ISO8601)
 
 target="$1"
 date="$2"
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/verifyMD5.sh
```diff
@@ -23,6 +23,7 @@
 # @version svn: $Id$
 
 checkMD5NoGui="`dirname $0`/archivematicaCheckMD5NoGUI.sh"
+date_format="%a %b %e %_H.%M.%S %Z %Y"
```
sevein: Should we use instead something like the following?

```bash
date="$(date -u '+%Y%m%dT%H%M%SZ')" # 20170605T194004Z
```

It is compliant with ISO8601 and measured in UTC.

src/MCPClient/lib/clientScripts/verifyMD5.sh
```diff
@@ -23,7 +23,7 @@
 # @version svn: $Id$
 
 checkMD5NoGui="`dirname $0`/archivematicaCheckMD5NoGUI.sh"
-date_format="%a %b %e %_H.%M.%S %Z %Y"
+date_args="-u +%Y%m%dT%H%M%SZ"  # 20170605T220452Z (UTC ISO8601)
```
sevein: If you want to be consistent with the other variable names, should't be named `dateArgs` instead?

Why don't you run the command here like in `checkMD5NoGui`?

scollazo: The `checkMD5NoGui` is executed once, while the `date` command runs each time that a file is verified, I didn't want to change current behavior.

sevein: You're right!

nickwilkinson:
[clientscripts: change date format in verifyMD5.sh](https://trello.com/c/PeujJYmZ/1-clientscripts-change-date-format-in-verifymd5sh)

jhsimpson (approved):
this is good to merge 👍

sevein (changes requested):
Nitpick: `date_args` should be `dateArgs`. I believe that consistency matters.

scollazo:
Changed date_args to keep consistency, and rebased prior to merge

sevein (approved):

jhsimpson:
merged in stable/16.x as https://github.com/artefactual/archivematica/commit/ef10d4f532ec7b65ffacf6c8d8068bb3d47ccf75

merged in qa/1.x as https://github.com/artefactual/archivematica/commit/378caf30f6fee62ca3a819d0a2b8427ffc884151

PR #631, opened 2017-06-08
Set currentstep to integer in dashboard fixture.

jrwdunham:
One of the tests in dashboard/tests/test_api.py was failing because dashboard/tests/fixtures/jobs-various.json was using the old-style strings for currentstep values. This PR changes those currentstep values to the appropriate ints.


sevein (approved):
Thank you!

PR #635, opened 2017-06-13
File mod. date update microservice, refs #11063

mcantelon:
Redmine: https://projects.artefactual.com/issues/11063

Microservices to store file modification dates for later indexing and inclusion in METS file.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -386,6 +386,10 @@ def create_premis_object(fileUUID):
     for elem in create_premis_object_characteristics_extensions(fileUUID):
         objectCharacteristics.append(elem)
 
+    creatingApplication = etree.Element(ns.premisBNS + "creatingApplication")
+    etree.SubElement(creatingApplication, ns.premisBNS + "dateCreatedByApplication").text = f.modificationtime.strftime("%Y-%m-%d")
+    objectCharacteristics.append(creatingApplication)
+
     etree.SubElement(object_elem, ns.premisBNS + "originalName").text = escape(f.originallocation)
 
     for elem in create_premis_object_derivations(fileUUID):
--- a/src/MCPClient/lib/clientScripts/storeFileModificationDates.py
+++ b/src/MCPClient/lib/clientScripts/storeFileModificationDates.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python2
+
+# This file is part of Archivematica.
+#
+# Copyright 2010-2017 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+import datetime
+import os
+import sys
+
+import django
+django.setup()
+# dashboard
+from main import models
+# archivematicaCommon
+from custom_handlers import get_script_logger
+
+
+logger = get_script_logger('archivematica.mcp.client.storeFileModificationDates')
+
+
+def get_modification_date(file_path):
+    mod_time = os.path.getmtime(file_path)
+    return datetime.datetime.utcfromtimestamp(int(mod_time))
+
+
+def main(transfer_uuid, shared_directory_path):
+    transfer = models.Transfer.objects.get(uuid=transfer_uuid)
+
+    files = models.File.objects.filter(transfer=transfer)
+    for transfer_file in files:
+        file_path_relative_to_shared_directory = transfer_file.currentlocation.replace('%transferDirectory%', transfer.currentlocation, 1)
+        file_path = file_path_relative_to_shared_directory.replace('%sharedPath%', shared_directory_path, 1)
+        transfer_file.modificationtime = get_modification_date(file_path)
+        transfer_file.save()
+
+    logger.info('Stored modification dates of %d files.', len(files))
+
+
+if __name__ == '__main__':
+    transfer_uuid = sys.argv[1]
+    shared_directory_path = sys.argv[2]
+    main(transfer_uuid, shared_directory_path)
+    sys.exit(0)
--- a/src/MCPClient/tests/test_store_file_modification.py
+++ b/src/MCPClient/tests/test_store_file_modification.py
@@ -1,0 +1,61 @@
+# -*- coding: utf8
+import os
+import shutil
+import sys
+import tempfile
+
+from django.test import TestCase
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.abspath(os.path.join(THIS_DIR, '../lib/clientScripts')))
+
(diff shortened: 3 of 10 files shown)


src/MCPClient/lib/clientScripts/storeFileModificationDates.py
```diff
@@ -0,0 +1,59 @@
...
+
+import django
+django.setup()
+# dashboard
+from main import models
+# archivematicaCommon
+from custom_handlers import get_script_logger
+
+
+logger = get_script_logger('archivematica.mcp.client.storeFileModificationDates')
```
sevein, line 32: You don't need to create a logger if you're not using it.

src/MCPClient/lib/clientScripts/storeFileModificationDates.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_modification_date(shared_directory_path, transfer, transfer_file):
+    file_path_relative_to_shared_directory = transfer_file.currentlocation.replace('%transferDirectory%', transfer.currentlocation, 1)
+    file_path = file_path_relative_to_shared_directory.replace('%sharedPath%', shared_directory_path, 1)
+    mod_time = os.path.getmtime(file_path)
+    return datetime.datetime.fromtimestamp(int(mod_time)).strftime('%Y-%m-%d %H:%M:%S')
```
sevein: I think we should use `fromutctimestamp` instead of `fromtimestamp` because we need to recognize that the user's local datetime is likely not UTC but our database is.

It looks like both functions are going to return a naive datetime.datetime which they seem to be perfectly handled by `DateTimeField`, i.e. no need to convert it into a string as that seems to only force Django to [convert it back into a `datetime.datetime`](https://github.com/django/django/blob/d381914aef50e04ca44b9d7bb9274c8351f5b9bf/django/forms/fields.py#L380-L388).

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -386,6 +386,10 @@ def create_premis_object(fileUUID):
     for elem in create_premis_object_characteristics_extensions(fileUUID):
         objectCharacteristics.append(elem)
 
+    creatingApplication = etree.Element(ns.premisBNS + "creatingApplication")
+    etree.SubElement(creatingApplication, ns.premisBNS + "dateCreatedByApplication").text = str(f.modificationtime)[:10]
```
sevein: I would feel more comfortable using a public function like `strftime("%Y-%m-%d")` where the output format is explicit. You're relying on a `__str__` method of `datetime.datetime`, perhaps that's not as reliable?

`f.modificationtime.isoformat()` may also be okay? Date and time combined seem like a good fit here, isn't it?

src/MCPClient/tests/test_store_file_modification.py
```diff
@@ -0,0 +1,60 @@
...
+                temp_dir + '/'
+            )
+
+            # Assert files have expected modification times
+            assert str(models.File.objects.get(pk='47813453-6872-442b-9d65-6515be3c5aa1').modificationtime) == '2012-06-12 07:21:22+00:00'
+            assert str(models.File.objects.get(pk='60e5c61b-14ef-4e92-89ec-9b9201e68adb').modificationtime) == '2012-06-12 07:21:22+00:00'
+            assert str(models.File.objects.get(pk='791e07ea-ad44-4315-b55b-44ec771e95cf').modificationtime) == '2012-06-12 07:21:22+00:00'
+            assert str(models.File.objects.get(pk='8a1f0b59-cf94-47ef-8078-647b77c8a147').modificationtime) == '2012-06-12 07:21:22+00:00'
+
+        finally:
```
sevein: You may want to use [tearDown](https://docs.python.org/2/library/unittest.html#unittest.TestCase.tearDown) instead.

src/dashboard/src/main/migrations/0033_store_transfer_file_modification_dates.py
```diff
@@ -0,0 +1,112 @@
...
+    MicroServiceChainLinkExitCode.objects.create(
+        id='4eaba00b-30bc-4508-84e4-d9871cd39fc7',
+        microservicechainlink_id=set_transfer_file_mod_dates_cl_pk,
+        nextmicroservicechainlink_id=load_rights_cl_pk,
+        exitcode=0,
+        exitmessage=Job.STATUS_COMPLETED_SUCCESSFULLY
+    )
+
+    # Introduce between "Add processed structMap to METS.xml document" and "Characterize and extract"
+    MicroServiceChainLinkExitCode.objects.filter(microservicechainlink_id='307edcde-ad10-401c-92c4-652917c993ed').update(nextmicroservicechainlink_id=set_transfer_file_mod_dates_cl_pk)
```
sevein: Nitpick, if you're consistently using 79 columns in this file this is a clear violation of the rule! :P

src/MCPClient/lib/clientScripts/storeFileModificationDates.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python2
+
+# This file is part of Archivematica.
+#
+# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
```
sevein: 2017!

src/MCPClient/lib/clientScripts/storeFileModificationDates.py
```diff
@@ -0,0 +1,59 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+# @package Archivematica
+# @subpackage archivematicaClientScript
+# @author Mike Cantelon <mike@artefactual.com>
```
sevein: I think this is so SVN! We really don't need it.

src/MCPClient/lib/clientScripts/storeFileModificationDates.py
```diff
@@ -0,0 +1,59 @@
...
+    mod_time = os.path.getmtime(file_path)
+    return datetime.datetime.fromtimestamp(int(mod_time)).strftime('%Y-%m-%d %H:%M:%S')
+
+
+def main(transfer_uuid, shared_directory_path):
+    transfer = models.Transfer.objects.get(uuid=transfer_uuid)
+
+    files = models.File.objects.filter(transfer=transfer)
+    for transfer_file in files:
+        transfer_file.modificationtime = get_modification_date(shared_directory_path, transfer, transfer_file)
```
sevein: Nitpick, if you passed just a `transfer_file`, then you could access to its transfer with `transfer_file.transfer`?

sevein (changes requested):

sevein:
Small comments but it's looking really good, thanks @mcantelon.

sevein (changes requested):
Looking good. I've added some comments.

src/MCPClient/tests/test_store_file_modification.py
```diff
@@ -0,0 +1,60 @@
...
+    transfer_uuid = 'e95ab50f-9c84-45d5-a3ca-1b0b3f58d9b6'
+
+    def test_store_file_modification_dates(self):
+        """Test storeFileModificationDates.
+
+        It should store file modification dates.
+        """
+
+        # Create files
+        temp_dir = tempfile.mkdtemp()
```
sevein: You should delete this once you're done with it right?

src/MCPClient/lib/clientScripts/storeFileModificationDates.py
mcantelon: True haha. I've added actual logging now.

sevein:
@mcantelon, please notice that the CI checks have failed.

sevein (approved):
Tested, it works great.

PR #637, opened 2017-06-14
Allow 'object count' from SS to be string

### LINKED ISSUE #636: Problem: SS 'object count' is not always an int: ``ungettext`` bug
If a transfer source location contains over 5000 files, then filesystem_ajax/views.py will trigger a ``TypeError`` because the call to ``ungettext`` assumes that the value of the ``'object count'`` attribute returned by the SS will always be an integer. In fact, in this case the value will be a string, i.e., '5000+'.

jrwdunham:
Fixes #636. If a transfer source location contains over 5000 files, then
filesystem_ajax/views.py will trigger a TypeError because the call to
ungettext assumes that the value of the 'object count' attribute
returned by the SS will always be an integer. In fact, in this case the
value will be a string, i.e., '5000+'. This change catches the
``TypeError`` and uses ``ugettext`` with ``%(...)s``.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -83,7 +83,14 @@ def _prepare_browse_response(response):
         if 'levelOfDescription' in prop:
             prop['display_string'] = prop['levelOfDescription']
         elif 'object count' in prop:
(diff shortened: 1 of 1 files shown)


sevein (approved):
Thanks!

PR #639, opened 2017-06-14
Backlog: fix ES query to get all entries

hakamine:
Archivematica uses an ES aggregations query to get backlog entries.
Aggregation queries return the top 10 results only, for this reason the
backlog tab was showing only a maximum of 10 items.

Add a size parameter to terms in the ES query in order to get all
the backlog entries.

Refs. #11252

Also needs to be cherry-picked to qa/1.x


sevein:
I've changed base to `qa/1.x` so I can see the build status.

sevein (approved):
We realize this is not efficient but it's a necessary fix. We'll revisit this later.

sevein:
eb94115e in `stable/1.6.x`.

Issue #640, opened 2017-06-16
Problem: AIP download URL uses SS IP that may not be available to user


helenst:
Dashboard is set up to access the storage service directly using the API, on an IP address or hostname configured in the Admin settings.

Dashboard also provides the user with a generated download URL for downloading the AIP from the storage service. This is generated from the above mentioned configured IP address. However in some setups, that IP address may not be accessible to the user via their browser.

One option would be to have two separate URL settings, one internal and one external. Another would be to give the user a download URL that points to Dashboard, and proxy the content from Storage Service (using something like https://goo.gl/l5bR1F) - this would remove the need to provide the user with an externally available storage URL.

sevein:
Fixed in #641 (20661356).

kellyannewithane:
This can be tested by setting the storage service url to localhost and trying to download an aip

PR #641, opened 2017-06-16
Stream AIP content from SS rather than redirect

helenst:
Redirecting is problematic as the URL redirected to comes from the setting intended for API access, which should be accessible to the Dashboard server but is not necesarily available to the user. By
streaming we can avoid sending the user directly there, and provide indirect access via dashboard instead.

This is for #640


sevein:
Thank you! 20661356

PR #642, opened 2017-06-16
Set up elastic search client in manage.py

helenst:
When running dashboard using `manage.py` (which I find useful for debugging) some areas of the app complain about elastic search not being set up. This is already done in `wsgi.py` but not here.


sevein:
Thank you! 5f47397d

PR #643, opened 2017-06-16
MCPClient: add alternate dependency for libavcodec

scollazo:
This way, the same package can be built for 14.04 and 16.04


sevein:
00593125 in `qa/1.x`.

PR #647, opened 2017-06-22
Microservice to add README.html to AIPs.

mcantelon:
Redmine: https://projects.artefactual.com/issues/11231

* Add "README.html" to shared microservice tasks config files
* Designate "README.html" as an optional SIP file
* Add migration to add "Add README file" microservice chain/link and include "README.html" in AIPs


sevein:
Preferably I think we should stop relying on the shared directory structure to hold assets for our client scripts. I think the purpose of the shared directory should be restricted to hold transient processing data.

If we had to break up archivematica.git today into more than one repo, we would probably have three repos: mcpserver, mcpclient (which could be many) and dashboard. In this PR, you'd be adding a file to mcpserver (shared directory belongs to it) and using it from mcpclient, creating a dependency that I'd rather avoid.

Thinking in the long term, where the shared directory may have other forms, like an object storage, I think we should make an extra effort today towards that kind of scenario, where it's probably better to have self-contained client scripts that can do their job alone. E.g. a client script that looks like the following would do it:

```
#!/usr/bin/env python2

README_HTML = u"""<!DOCTYPE html>
<html lang="en">
  ...
</html>
"""

if __name__ == '__main__':
    # Write it to disk!
```

That's my opinion but I recommend you to talk to @jhsimpson about this. We've discussed this idea before and we've been experimenting with a version of Archivematica without sharedDirectoryStructure/ in RDSS already.

sevein (changes requested):
The solution is being revisited.

jhsimpson (changes requested):
The problem @sevein  has identified is coming from work in the JiscRDSS project.  In order to avoid a conflict with that code, it would be better to store the new readme file inside the MCPClient codebase.

You can put it in src/MCPClient/lib/assets/readme.html, and update the Standard Task Config to use this location, instead of the sharedDirectory location.

sevein:
So if this is the task configuration:

```
StandardTaskConfig.objects.create(
    id=add_readme_file_stc_pk,
    requires_output_lock=False,
    execute='copy_v0.0',
    arguments='"%sharedPath%sharedMicroServiceTasksConfigs/README/README.html" "%SIPDirectory%README.html"')
```

`%sharedPath%` would be whatever `config.get('MCPServer', "sharedDirectory")` returns, and this only works because we implicitly have the sharedDirectory deployed in the same location.

The MCPServer doesn't necessarily know where the assets/  folder is going to be or anything related to MCPClient. We could tweak the MCPClient to introduce its own variables, like `%clientAssets%`, when the task is executed.

sevein (approved):

sevein:
Nicely done, @mcantelon! There was a small problem in the migration after rebasing (your branch was outdated) and there was a UUID collision with 429b35cfeedca6e797df77e6356be29858d169f8 but it was easy to fix.

I've submitted a PR to make the config change in the Ansible role: https://github.com/artefactual-labs/ansible-archivematica-src/pull/116.

mcantelon:
Awesome... thanks! :D

Issue #646, opened 2017-06-18
Problem: AIP download is unnamed


sevein:
In 20661356c8b20dfbf2d5969234ccb6607407b149 we added to Dashboard the ability to stream AIPs to the browser that are served by the Storage Service. We're using [stream_file_from_storage_service](https://github.com/artefactual/archivematica/blob/92d7abd238585e64e6064bc3f1ddfc663c4d3ace/src/dashboard/src/components/helpers.py#L338-L348) which gets the job done but it does not set the filename. In Chrome that results in a file named `download`, in Firefox the file is named  randomly.

In Firefox, the download never finishes.

helenst:
@sevein I've fixed the filename issue in PR #650. I am not sure about the issue of the download not finishing - I am not seeing this happen in my Firefox. Could it be related to the Content-Disposition headers not having been set properly (and if so, does it work now?)

sevein:
> Could it be related to the Content-Disposition headers not having been set properly (and if so, does it work now?)

I'm sure that's the case!

sevein:
Thanks!

PR #650, opened 2017-06-28
Give AIP download a filename

helenst:
This is a fix for #646 - takes the content disposition (which gives the download filename) from storage service response and passes it on to the streaming Dashboard response (and not doing so if no content disposition is set).


sevein (approved):
Nicely done.

PR #655, opened 2017-06-29
Delete unused django-autoslug

sevein:
It is used by fpr-admin which has its own requirements file.


jhsimpson (approved):
:+1:

sevein:
7d0cef67 in `stable/1.6.x`

Issue #664, opened 2017-07-07
Configuration cleanup (merge from Jisc fork)


helenst:
Integrate work from https://github.com/JiscRDSS/archivematica/pull/3 into core Archivematica

This makes improvements to the configuration such that config is no longer shared between components, and it can be configured either in a config file or through environment variables.

This was added to the fork after the Docker work (https://github.com/artefactual/archivematica/issues/663), but might be possible without merging that first.

It will also require consideration of https://github.com/JiscRDSS/archivematica/issues/9 - the work in the Jisc PR changes the way the whitelist is used in authentication.

helenst:
I've started doing this on a branch:

https://github.com/helenst/archivematica/compare/dev/clean-up-config...helenst:dev/auto-deployment

It works and does the install stuff - I wanted to try and address https://github.com/JiscRDSS/archivematica/issues/9 here since the original Jisc PR is the one that changes this behaviour. So far I've set it up as per my suggestion there. I'm not sure if we still want the migration that added - perhaps the default whitelist can go back to being 127.0.0.1?

Also a question of whether the ansible scripts should be changed to (optionally) use this rather than just setting up the test user, similar to how it is done with Docker.

sevein:
> It works and does the install stuff - I wanted to try and address JiscRDSS#9 here since the original Jisc PR is the one that changes this behaviour. So far I've set it up as per my suggestion there. I'm not sure if we still want the migration that added - perhaps the default whitelist can go back to being 127.0.0.1?

That's reasonable. I didn't realize until know that I introduced a migration in the RDSS fork, perhaps I shouldn't have done that. I'm worried that the divergence will make harder to maintain the two, what do you think? I don't want that to influence our decision about the whitelist but at the same time I think that trusting localhost makes very little sense. What if the attacker manages to proxy the request from another service running on the same machine?

> Also a question of whether the ansible scripts should be changed to (optionally) use this rather than just setting up the test user, similar to how it is done with Docker.

I think so, that'd be a really nice improvement.

helenst:
oops, just realised I meant to make that comment on #665 - continuing discussion there...

sevein:
#670 was merged!

PR #670, opened 2017-07-07
Config cleanup

helenst:
**_This is a merge of work from the JiscRDSS fork (see #664) - ~it's WIP~ as it's just at the point where it deploys with Ansible - it's not tested yet and quite possibly doesn't work._**

MCPServer, MCPClient and Dashboard have been changed in this commit with the
following goals:

- Stop sharing configuration between components.
- Use django.conf to manages config.
- Manage config defaults within code.
- Look up environment variables:
  - ARCHIVEMATICA_DASHBOARD_*
  - ARCHIVEMATICA_MCPSERVER_*
  - ARCHIVEMATICA_MCPCLIENT_*

dbsettings, clientConfig.conf or serverConfig.conf are still considered if they
exist - they will be merged with the default INI in code. Environment strings
are applied last.

Client scripts are now executed using MCPClient's `settings.common`. So now
when they look up `django.conf` they're not looking at the Dashboard config.
It wasn't correct to assume that its config was available.


helenst:
This now successfully runs a transfer so I'm de-WIPing it for review. https://github.com/artefactual-labs/ansible-archivematica-src/pull/115 required for Ansible deploy.

sevein:
@helenst, great job! I'm having a meeting with @jrwdunham today to introduce him to these changes. I'm hoping we can get this merged very soon! Thank you!

src/dashboard/src/components/accounts/views.py
```diff
@@ -63,9 +62,6 @@ def add(request):
 
 
 def edit(request, id=None):
-    if get_client_config_value('kioskMode') == 'True':
-        return redirect('main.views.forbidden')
-
```
jrwdunham, line 68: Why was the `kioskMode` config param removed?

jrwdunham (changes requested):
I reviewed the changes and ran the core AM acceptance tests on a VM deployed with these changes (and deployed via https://github.com/artefactual-labs/ansible-archivematica-src/pull/115) and they passed.

After rebasing this on current qa/1.x, I believe you will have to add a call to
``setup_dicts(mcpclient_settings)`` in the new policyCheck.py client script.

I also am wondering why the kioskMode var was removed.

After that, I think this is good to merge.

src/dashboard/src/components/accounts/views.py
sevein: Undocumented and unused. I think this was used at some point when we didn't have non-super users.

jrwdunham:
Also, these changes cause 3 of the tests to fail in archivematicaCommon/tests/test_dicts.py. I think that should be fixed.

src/dashboard/src/components/accounts/views.py
jrwdunham: Ok. Thanks sevein.

helenst:
@jrwdunham Thanks! I've fixed those tests so that they work now.

sevein (approved):

jrwdunham:
@helenst Thanks! Now, unless I'm wrong, I think you just need to rebase and add the `setup_dicts` call in the new client script policyCheck.py.

helenst:
@jrwdunham It's all rebased now and I've updated the policycheck script.

jrwdunham (approved):

PR #672, opened 2017-07-11
Added "Store DIP" choice in config page

jrwdunham:
This required also modifying the workflow (hence the migration) because
the post-DIP-upload "Store DIP?" choice link was being activated AFTER
relocation of the DIP to the uploadedDIPs/ directory, which resulted in
the MCPServer not knowing where the DIP was when it went to look for the
processingMCP.xml config file. Now moving the DIP to uploadedDIPs/ only
occurs after the upload/store DIP steps have completed.

### DIFF
--- a/src/dashboard/src/components/administration/forms.py
+++ b/src/dashboard/src/components/administration/forms.py
@@ -238,7 +238,13 @@ class ProcessingConfigurationForm(forms.Form):
     processing_fields['92879a29-45bf-4f0b-ac43-e64474f0f2f9'] = {
         'type': 'chain_choice',
         'name': 'upload_dip',
-        'label': 'Upload DIP',
+        'label': _('Upload DIP'),
+        'ignored_choices': []
+    }
+    processing_fields['5e58066d-e113-4383-b20b-f301ed4d751c'] = {
+        'type': 'chain_choice',
+        'name': 'store_dip',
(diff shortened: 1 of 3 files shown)


sevein:
So no changes in `src/dashboard/src/components/ingest/views.py` other than cosmetics, right?

jrwdunham:
Yeah, views.py only has cosmetic changes.

sevein (changes requested):
Migration conflicts with `0034_add_readme_to_aips`.

sevein (approved):

jrwdunham:
After discussion with Evelyn and Hector, we have agreed that this PR needs to be modified. The issue is that there are two distinct "Store DIP" paths, which means that processing configuration will only work for one of them. To avoid this, we propose to prune one of the paths. However, this requires additional work beyond just modifying the workflow. Choosing NOT to store a DIP after the "Upload DIP?" decision point has different implications depending on whether the user has or has not Uploaded the DIP: if she has not uploaded, then not storing will lead to rejection of the DIP; however if she has uploaded, then not storing will not lead to rejection. Future commits and CR will are forthcoming for this PR.

jrwdunham:
Closed in favour of https://github.com/artefactual/archivematica/pull/680.

PR #675, opened 2017-07-12
Dev/issue 11352 orphaned link other bugs

jrwdunham:
- Workflow Fixes

  - Makes the "Check if DIP should be generated" link reachable. The previous migration inadvertently orphaned it. Now "Add README file" exits to it.
  - Also modified ``File.modificationtime`` so that it corresponds to the camelCased database column name ``modificationTime``.

- No mod. date created for deleted files

  - Prior to this, if a compressed package was deleted then the storeFileModificationDates micro-service would attempt to assign a modification date to it but would error because the ``File`` would have a ``None`` ``currentLocation`` attr. This fixes that.

- Fixed fixtures to include File.modificationtime vals


sevein:
Related: https://github.com/artefactual/archivematica/pull/647, https://github.com/artefactual/archivematica/pull/635.

mcantelon (approved):
Looks good to me! Thanks for fixing those issues!

PR #678, opened 2017-07-16
Include AT and TB AngularJS projects, refs #11363

jraddaoui:
Since both webpack files are included by default in AM 1.6 and they require
AM to work, this commits introduces the appraisal-tab and transfer-browser
projects into the AM dashboard. The webpack files need to be build apart as
they are not being tracked by Git now.

- Add AT and TB sources.
- Clean-ups in dependencies and webpack builds.
- Pin main angular deps version to avoid upgrading to 1.6.x as the
  angular-route-segment doesn't work in it.
- Upgrade jQuery UI to v1.12.1 in the dashboard.
- Avoid importing jQuery twice in some templates (and other JS files).


jraddaoui:
https://github.com/artefactual-labs/appraisal-tab/pull/159

sevein:
@jraddaoui, I'm wondering if we should repurpose this PR in order to move the sources of the [appraisal tab](https://github.com/artefactual-labs/appraisal-tab) and the [transfer browser](https://github.com/artefactual-labs/transfer_browser) to this repo. What do you think?
I understand that it may require a significant effort but it'll pay off soon. The appraisal-tab and transfer_browser repos served well when we wanted to work on them in parallel but that's not the case anymore.

```
src/dashboard/
├── debian/
├── install/
├── osdeps/
├── src/
│   ├── media/build/appraisal-tab.js
│   ├── media/build/appraisal-tab.css
│   ├── media/build/transfer-browser.js
│   └── media/build/transfer-browser.css
├── tests/
├── .eslint.rc
├── webpack.config.js
└── package.json
```

I think that `src/media/build` can be gitignored for now - we may opt to track them in git if slows down deployments significantly or if it makes packaging too painful. It's also ok not to fix all the things now (e.g. including old assets is ok, stick with webpack and npm probably best, etc...) - we can focus on get it working, figure out the dev/deploy workflow changes, update our Ansible role and dev documentation.

Thoughts?

sevein:
Another benefit would be to stop serving the old `/media/js/transfer_browser.js` (2.6 MB) and `/media/js/appraisal_tab.js` (4.8 MB) in addition to all the extra assets that belong to the Dashboard.

sevein (changes requested):
Please see my comments above.

jraddaoui:
I think that's a great idea. I'll ask for time ;)

src/dashboard/frontend/appraisal-tab/package.json
```diff
@@ -2,9 +2,7 @@
   "name": "appraisal-tab",
   "private": true,
   "version": "0.0.0",
-  "description": "New appraisal tab prototype for Archivematica",
-  "repository": "https://github.com/artefactual-labs/appraisal-tab",
-  "license": "AGPL3",
+  "description": "Appraisal tab prototype for Archivematica",
```
sevein: Excepting it's not a prototype anymore! :D

jraddaoui:
Hi Sevein,

As we talked in the chat, I'll leave you the changes required in the Ansible role. I was using qa/1.7.x for the archivematica-src role with this branch and, when the Vagrant box was ready, all I had to do was:

```
vagrant ssh
curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
sudo apt-get install nodejs build-essential
cd /vagrant/src/archivematica/src/dashboard/frontend/appraisal-tab
npm install
cd /vagrant/src/archivematica/src/dashboard/frontend/transfer-browser
npm install
```

Please, le me know if you need more info and/or if you find anything to change in here.

Thanks!

sevein:
Thanks, Radda! I'll update the Ansible role when I have a chance!

sevein:
I've rebased!

sevein (approved):
I just got this to work with https://github.com/artefactual-labs/ansible-archivematica-src/pull/128. Thanks, @jraddaoui - it's working great!

sevein:
Forgot to mention that I had to add 8543f51 because WhiteNoise won't accept that stylesheet as it was confusing the content of that comment with a real url link which couldn't find because it was URL-encoded. Very confusing, that took me a while!

sevein:
Rebased.

jraddaoui:
Rebased

jraddaoui:
Tested it alongside https://github.com/artefactual-labs/ansible-archivematica-src/pull/128 and it seems to be working as expected. I'm not sure what's the best place to put the 'node_modules' workaround for Vagrant users but, since this is blocking three issues now, I'll merge both PRs and we'll figure out later.

Thanks @sevein!

jraddaoui:
Merged in https://github.com/artefactual/archivematica/commit/40aa9d90ce1e50d7dce080e5fef4fdb374ae42d9

PR #679, opened 2017-07-18
Problem: "Generate AIP METS" fails to manage dspace transfers

sevein:
`Generate AIP METS` may fail with DSpace transfers (error: "No groupID for file") because of an ORM query that is failing to handle multiple results gracefully - it seems to be a regression introduced in 34fd87382020d27f9dfdc40e4853e5dbb1ce6280 when we moved away from SQL, e.g.:

```
mysql> SELECT originalLocation FROM Files WHERE removedTime IS NULL AND fileGrpUse = "original" AND sipUUID = "aa8b1944-3304-404a-9439-f2e940cb6c19" AND originalLocation LIKE '%%transferDirectory%%objects/ITEM_12345.zip-2017-07-18T18:33:06.426589+00:00/%';
+-----------------------------------------------------------------------------------------------+
| originalLocation                                                                              |
+-----------------------------------------------------------------------------------------------+
| %transferDirectory%objects/ITEM_12345.zip-2017-07-18T18:33:06.426589+00:00/bitstream_1086.m4v |
| %transferDirectory%objects/ITEM_12345.zip-2017-07-18T18:33:06.426589+00:00/bitstream_1089.mp4 |
| %transferDirectory%objects/ITEM_12345.zip-2017-07-18T18:33:06.426589+00:00/bitstream_1011.rdf |
+-----------------------------------------------------------------------------------------------+
3 rows in set (0.00 sec)
```

The solution suggested in this PR is to accept the fact that a situation with multiple matches is possible and use the first one in the list.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -773,11 +773,9 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath, baseDirectoryName
                     "filegrpuse": "original",
                     "originallocation__startswith": os.path.dirname(f.originallocation)
                 }
-                try:
-                    original_file = File.objects.get(**kwargs)
+                original_file = File.objects.filter(**kwargs).first()
+                if original_file is not None:
                     GROUPID = 'Group-' + original_file.uuid
-                except (File.DoesNotExist, File.MultipleObjectsReturned):
-                    pass
 
             elif use in ("preservation", "text/ocr"):
                 # Derived files should be in the original file's group


jrwdunham:
👍 
For posterity, we had a fairly lengthy discussion about how a single DSpace license file could correspond to more than one original file and whether it made sense to arbitrarily choose one original to build the GROUPID value. In the end we could find no way of selecting a single best original for this purpose and went back to previous behaviour of arbitrarily choosing one original.

jrwdunham (approved):

sevein:
- `stable/1.6.x`: 11e89646f2708d525ba5f19ccce9b4879b02a619
- `qa/1.x`:  1db2a70f5c36284f2e38250036354530b6ca19c6

PR #680, opened 2017-07-19
Workflow fixes related to DIP Upload & Store

jrwdunham:
Changes to the workflow and related client scripts so that the "Upload DIP" and "Store DIP" decision points make more sense and allow for preconfigured choices for "Store DIP" AFTER "Upload DIP". See RM issue 11195 and closed PR #672.
    
* There used to be 2 distinct "Store DIP" paths—now there is only 1.
* handleUnstoredDIP client script puts an unstored DIP in the rejected/ directory *only* if it has not been copied to the uploadedDIPs/ directory
* Processing config form allows configuration of "Store DIP" decision
* Microsecond precision on SIP status timestamp: Now when ingest/status is called by jobs.js, the dashboard returns microsecond precision on the unix timestamps for SIP last modified. This fixes a bug where the representation of a SIP will not update in the GUI if its final jobs all completed at the same second.

### DIFF
--- a/src/MCPClient/lib/clientScripts/handleUnstoredDIP.py
+++ b/src/MCPClient/lib/clientScripts/handleUnstoredDIP.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+# This file is part of Archivematica.
+#
+# Copyright 2010-2017 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Handle Unstored DIP.
+
+This client script hanles an unstored DIP by moving it to the rejected/
+directory if it has *not* been uploaded and doing nothing otherwise.
+"""
+from __future__ import print_function
+
+import os
+import shutil
+import sys
+
+import django
+django.setup()
+# archivematicaCommon
+from custom_handlers import get_script_logger
+
+
+logger = get_script_logger('archivematica.mcp.client.handleUnstoredDIP')
+
+
+def _get_sip_dirname(sip_path):
+    return os.path.basename(sip_path.rstrip('/'))
+
+
+def _sip_uploaded(sip_path, uploaded_dips_path):
+    """Return ``True`` if the SIP has been uploaded."""
+    return os.path.isdir(os.path.join(
+        uploaded_dips_path,
+        _get_sip_dirname(sip_path)))
+
+
+def _move_to_rejected(sip_path, rejected_path):
+    """Move the SIP at ``sip_path`` to the rejected directory at
+    ``rejected_path``.
+    """
+    dest = os.path.join(rejected_path, _get_sip_dirname(sip_path))
+    shutil.move(sip_path, dest)
+    return dest
+
+
+def main(sip_path, rejected_path, uploaded_dips_path):
+    if _sip_uploaded(sip_path, uploaded_dips_path):
+        msg = ('The SIP has been uploaded so it was NOT moved to the'
+               ' rejected/ directory')
+    else:
+        dest = _move_to_rejected(sip_path, rejected_path)
+        msg = ('The SIP had NOT been uploaded so it was moved to the'
+               ' rejected/ directory at %s' % dest)
+    print(msg)
+    logger.info(msg)
+    return 0
+
+
+if __name__ == '__main__':
+    sip_path, rejected_path, uploaded_dips_path = sys.argv[1:]
+    logger.info('handleUnstoredDIP called with sip directory %s and rejected'
+                ' directory %s', sip_path, rejected_path)
+    sys.exit(main(sip_path, rejected_path, uploaded_dips_path))
--- a/src/dashboard/src/components/ingest/views.py
+++ b/src/dashboard/src/components/ingest/views.py
@@ -32,7 +32,7 @@ import uuid
 from django.conf import settings as django_settings
 from django.contrib import messages
 from django.core.urlresolvers import reverse
-from django.db.models import Max
+from django.db import connection
 from django.forms.models import modelformset_factory
 from django.http import Http404, HttpResponse, HttpResponseBadRequest
 from django.shortcuts import render, redirect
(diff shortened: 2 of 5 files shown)


src/dashboard/src/components/ingest/views.py
```diff
@@ -90,61 +90,70 @@ def post(self, request):
...
-    # Equivalent to: "SELECT SIPUUID, MAX(createdTime) AS latest FROM Jobs WHERE unitType='unitSIP' GROUP BY SIPUUID
-    objects = models.Job.objects.filter(hidden=False, subjobof='').values('sipuuid').annotate(timestamp=Max('createdtime')).exclude(sipuuid__icontains='None').filter(unittype__exact='unitSIP')
+    """Returns the status of the SIPs in ingest as a JSON object with an
+    ``objects`` attribute that is an array of objects, each of which represents
+    a single SIP. Each SIP object has a ``jobs`` attribute whose value is an
+    array of objects, each of which represents a Job of the SIP.
+    """
+    sql = """
+        SELECT SIPUUID,
+                MAX(UNIX_TIMESTAMP(createdTime) + createdTimeDec) AS timestamp
```
sevein, line 100: Nice :)

src/MCPClient/lib/clientScripts/handleUnstoredDIP.py
```diff
@@ -0,0 +1,76 @@
...
+logger = get_script_logger('archivematica.mcp.client.handleUnstoredDIP')
+
+
+def _get_sip_dirname(sip_path):
+    return os.path.basename(sip_path.rstrip('/'))
+
+
+def _sip_uploaded(sip_path, uploaded_dips_path):
+    """Return ``True`` if the SIP has been uploaded."""
+    return _get_sip_dirname(sip_path) in os.listdir(uploaded_dips_path)
```
sevein: Is there any reason why you can't just use `os.path.isdir(path)` or `os.path.exists(path)`? Sounds like it'd be more efficient because it just uses `stat()`.

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
```diff
@@ -0,0 +1,189 @@
...
+    The workflow above is better for the following reaons:
+
+    1. There is only one "Store DIP Location" choice, so it can more easily
+       be made a preconfigured choice within the current system
+    2. It is more clear that when a DIP is neither uploaded nor stored, it is
+       rejected.
+
+    Steps:
+
+    1. Remove "Reject DIP" from options at "Upload DIP" decision point:
```
sevein: `s/:/.`?

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
```diff
@@ -0,0 +1,189 @@
...
+    3. Remove the unused links and their dependencies:
+
+       - "Retrieve DIP Storage Locations" (ed5d8475-3793-4fb0-a8df-94bd79b26a4c)
+       - "Store DIP location" (b7a83da6-ed5a-47f7-a643-1e9f9f46e364)
+       - "Store DIP" (e85a01f1-4061-4049-8922-5694b25c23a2)
+       - "Move to the uploadedDIPs directory"
+         (e3efab02-1860-42dd-a46c-25601251b930)
+
+    4. Change "Completed" link to "Handle unstored DIP". This is handled by a
+       new client script that moves the DIP to the rejected/ directory iff it
```
sevein, line 52: `s/iff/if`?

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
```diff
@@ -0,0 +1,189 @@
...
+    ###########################################################################
+    # 1. Remove "Reject DIP" from options at "Upload DIP" decision point:
+    ###########################################################################
+
+    # Destroy the "Reject DIP" chain and link and attendant models.
+    # For some unknown reason, deleting just these two models will result in a
+    # deletion of the three that follow, despite the fact that this does not
+    # seem to be entailed by Django's foreign key cascade delete strategy.
+    StandardTaskConfig.objects.get(id=reject_dip_stc_uuid).delete()
+    TaskConfig.objects.get(id=reject_dip_tc_uuid).delete()
```
sevein: `TaskConfig` is referenced by `MicroServiceChainLink. currenttask`. `MicroServiceChainLink` is referenced by `MicroServiceChainChoice. choiceavailableatlink` and so on. So I'd assume that the behaviour you're seeing is because of Django is emulating `ON DELETE CASCADE`?

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
```diff
@@ -0,0 +1,189 @@
...
+    ###########################################################################
+
+    completed_link = MicroServiceChainLink.objects.get(
+        id=completed_link_uuid)
+    completed_tc = completed_link.currenttask
+    completed_tc.description = 'Handle unstored DIP'
+    completed_tc.save()
+    completed_stc = StandardTaskConfig.objects.get(id=completed_stc_uuid)
+    completed_stc.arguments = ('"%SIPDirectory%" '
+                               '"%rejectedDirectory%." '
```
sevein: Is the dot necessary?

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
```diff
@@ -0,0 +1,189 @@
...
+        execute='copy_v0.0',
+        arguments='"%SIPDirectory%" '
+                  '"%watchDirectoryPath%uploadedDIPs/." '
+                  '-R')
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [('main', '0037_mediaconch_policy_checks')]
+    operations = [migrations.RunPython(data_migration)]
```
sevein: This is a clear style deviation from Django's suggested way, not that I care but I find interesting that you prefer your way over the convention :D

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
sevein: Joel explained to me that this is the way that biconditionals are abbreviated in English - makes sense! :)

sevein (changes requested):
Please see my comments above.

src/MCPClient/lib/clientScripts/handleUnstoredDIP.py
jrwdunham: You're right—I'll use `isdir`.

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
jrwdunham: ok. fixing punctuation

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
jrwdunham: intentional

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
jrwdunham: Ok. I modified the comment to reference the Django docs. I'm still leaving the other delete attempts in there in case deleting the first two does not delete all of the last three.

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
jrwdunham: No, no I don't believe it is.

src/dashboard/src/main/migrations/0038_fix_workflow_store_dip.py
jrwdunham: I didn't know about Django's style guide on this issue. I'll conform to it. Just thought the whitespace wasn't necessary in this case.

sevein (approved):

PR #686, opened 2017-07-26
clientscripts: rename files with colons

scollazo:
See ef10d4f532ec7 for more info

### DIFF
--- a/src/MCPClient/lib/clientScripts/sanitizeNames.py
+++ b/src/MCPClient/lib/clientScripts/sanitizeNames.py
@@ -32,6 +32,7 @@ from archivematicaFunctions import unicodeToStr
 
 VERSION = "1.10." + "$Id$".split(" ")[1]
 valid = "-_.()" + string.ascii_letters + string.digits
+not_valid = ":"
 replacementChar = "_"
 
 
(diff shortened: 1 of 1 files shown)


sevein:
@scollazo: I've changed the target branch to qa/1.x (since 1.6.1 was released today) and rebased the branch. I've added @jrwdunham to the list of reviewers. Hopefully he can take a look once he is back.

jrwdunham (changes requested):
@scollazo I don't think this is doing what you want it to. That is, your change does not seem to have an effect. The `sanitizeName` function already replaces colons with underscores. Anything that's not in `-_.()abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789` should be getting replaced by an underscore. Can you revise this or explain to me what I am missing?

scollazo:
The problem with semicolons was caused by the output of the verifyMD5.sh (check ef10d4f532ec7b ), but somehow, the AIPs ingested using the old version of verifyMD5.sh can't be re-ingested due to this, as the files in the logs/ folder doesn't seem to be renamed by sanitizeNames.

The problem we were triying to fix is due to using a CIFS (samba) mount point for AIP storage. This doesn't allow for filenames with ":" in them, so the reingest fails to create files from old AIPS with filenames containing the prohibited chars, as they can't be created in the "staging" path.

PR #688, opened 2017-07-27
Fixing typo in xlst-devel package name

### LINKED ISSUE #683: Problem: libxslt1-devel can't be found in CentOS
Here in  archivematica/src/dashboard/osdeps/CentOS-7.json in the list of packages is the line

 { "name": "libxslt1-devel", "state": "latest"},

in going to install this version on archtest.denver.archivematica.org, unable to install the above, didn't even seem to be such a package in the source cache, and indeed the slightly differently libxslt-devel was installed instead. Not sure if package name depends on the RH/CentOS variant, or if the above is a typo, or what, but for moment have made a fork of qa/1.x, dev/issue-11231-CentOS-try, with the line modified to

 { "name": "libxslt-devel", "state": "latest"},

- David H.

jhsimpson:
Fixes #683

This is a bug introduced in #613 

This change will need to be cherry-picked into stable/1.6.x as well.

### DIFF
--- a/src/dashboard/osdeps/CentOS-7.json
+++ b/src/dashboard/osdeps/CentOS-7.json
@@ -15,7 +15,7 @@
   { "name": "policycoreutils-python", "state": "latest"},
   { "name": "libffi-devel", "state": "latest"},
   { "name": "libxml2-devel", "state": "latest"},
-  { "name": "libxslt1-devel", "state": "latest"},
+  { "name": "libxslt-devel", "state": "latest"},
   { "name": "openssl-devel", "state": "latest"},
   { "name": "python-devel", "state": "latest"}
   ]


sevein (approved):
Thanks for fixing this!

sevein:
libxslt-devel is not listed in stable/1.6.x, should we add it?

jhsimpson:
Good point - 1.6.x releases still used the debian/controlfile - libxslt1-dev is listed there https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/dashboard/debian/control#L22

CentOS package Spec files are not kept in this repo - in 1.6.x releases the list of dependencies for CentOS packages is maintained separately in the am-packbuild repo, so this PR doesn't need to be cherry-picked into stable/1.6.x.

Starting with the 1.7.0 release, dependencies will come from the new osdeps folder - so this change is only needed in the qa/1.x branch.

PR #690, opened 2017-07-28
Bind PIDs (persistent identifiers) to files, directories and AIPs

jrwdunham:
This feature allows Archivematica to be configured to assign persistent identifiers (PIDs) to files, directories and AIPs. The PIDs are documented in the AIP/DIP METS and are configured to resolve to user-configured URLs. Auxiliary features:

- Handle server configuration in dashboard
- "Bind PIDs" and "Assign UUIDs to directories" configurable choices
- PID binding works with arranged files and directories, and with extraction
- Accession number from transfer can be used as AIP/DIP PID (instead of default UUID)
- Empty directories in objects/ are now documented in METS file and contain empty .keep files to prevent deletion by BagIt


sevein:
@jrwdunham, I've rebased your branch because the build was failing for a different reason (see #693).

jrwdunham:
See https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/15 to run the applicable acceptance tests (Gherkin features)

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -1039,17 +1126,25 @@ def write_mets(tree, filename):
...
+                      dest="baseDirectoryPathString", default="SIPDirectory")
+    # transferUUID/sipUUID
+    parser.add_option("-f", "--fileGroupIdentifier", action="store",
+                      dest="fileGroupIdentifier", default="")
+    parser.add_option("-t", "--fileGroupType", action="store",
+                      dest="fileGroupType", default="sipUUID")
+    parser.add_option("-x", "--xmlFile", action="store", dest="xmlFile",
+                      default="")
+    parser.add_option("-a", "--amdSec", action="store_true", dest="amdSec",
+                      default=False)
```
sevein, line 1147: This block is only for cosmetic changes?

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -333,11 +333,30 @@ def copy_to_start_transfer(filepath='', type='', accession='', transfer_metadata
...
+    return any([
+        rslt[0] for rslt in
+        models.Transfer.objects.filter(
+            file__uuid__in=file_uuids).values_list(
+                'diruuids').distinct()])
+
+
 def create_arranged_sip(staging_sip_path, files, sip_uuid):
-    shared_dir = django_settings.SHARED_DIRECTORY
+    shared_dir = helpers.get_server_config_value('sharedDirectory')
```
sevein: I think this change should be deleted, `get_server_config_value` was recently deleted.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
jrwdunham: Yeah. Sorry for the noise.

jrwdunham (commented):
The needed mods to the default and automatic processing config constants have been added. See 5149715f6ff40e17e71d63e1ccbc85a0559b4a11.

src/dashboard/src/components/filesystem_ajax/views.py
jrwdunham: Good catch. Should be fixed by 06b38824053cc7fa89e60e37490d8a0d35630113.

sevein:
Rebased.

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@exit_on_known_exception
+def main(transfer_path, transfer_uuid, include_dirs):
+    """Assign UUIDs to all of the directories (and subdirectories, i.e., all
+    unique directory paths) in the absolute system path ``transfer_path``, such
+    being the root directory of the transfer with UUID ``transfer_uuid``. Do
+    this only if ``include_dirs`` is ``True``.
+    """
+    _exit_if_not_include_dirs(include_dirs)
```
sevein, line 117: As a first-timer here I didn't find this function particularly useful. Its name describes the behaviour that the caller expects only with the help of a function decorator. Is this idiomatic in Python?

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
```diff
@@ -0,0 +1,141 @@
...
+        logger.warning('Transfer path %s is not a path to a directory',
+                       root_path)
+        raise DirsUUIDsException
+    # objects/ and root dirs need no UUIDs.
+    exclude_paths = (root_path, os.path.join(root_path, 'objects'))
+    return (format_subdir_path(dir_path, root_path) for dir_path, __, ___ in
+            os.walk(root_path) if dir_path not in exclude_paths)
+
+
+def str2bool(val):
```
sevein: You've also defined this function in `archivematicaFunctions`.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -356,3 +358,17 @@ def deUnicode(str):
     if str is None:
         return None
     return unicode(str).encode('utf-8')
+
+
+def create_directory_models(dir_paths_uuids, unit_mdl, unit_type='transfer'):
```
sevein: Nitpick: you're not creating models, you're creating `Directory` objects I would think.

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -0,0 +1,459 @@
...
+
+def _merge_args_config(args, config):
+    for arg in vars(args):
+        argval = getattr(args, arg, None)
+        if argval or arg not in config:
+            config[arg] = argval
+    return config
+
+
+def get_command_line_params():
```
sevein, line 466: Really nice!

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -0,0 +1,459 @@
...
+    """
+    _validate(kwargs)
+    resolve_url, qualified_resolve_urls, purl_map = _render_url_templates(
+        kwargs)
+    request_body = _render_request_body(kwargs, resolve_url,
+                                        qualified_resolve_urls)
+    response = requests.post(
+        kwargs['pid_web_service_endpoint'],
+        data=request_body,
+        headers={'Content-Type': 'text/xml',  # <= non-paramed assumpton alert
```
sevein: What does the comment mean?

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -0,0 +1,459 @@
...
+                 'Authorization': 'bearer {}'.format(
+                     kwargs['pid_web_service_key'])},
+        verify=False  # TODO: REMOVE IN PRODUCTION!!! Needed in dev because test handle endpoint has bad HTTPS
```
sevein: Why don't you make this configurable?

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -0,0 +1,459 @@
...
+
+Example usage::
+
+    $ ./bindpid.py 58fe1b06-ee88-41ca-9bc9-320dcac6c858 file -c bindpid.cfg
+    Congratulations, you have successfully bound the handle (PID) "58fe1b06-ee88-41ca-9bc9-320dcac6c858". The following persistent URLs (PURLs) are now bound to the indicated URLs:
+    http://<RESOLVER_URL>/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858?locatt=view:access => https://my-domain.org/access/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858
+    http://<RESOLVER_URL>/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858?locatt=view:original => https://my-domain.org/original/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858
+    http://<RESOLVER_URL>/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858?locatt=view:preservation => https://my-domain.org/preservation/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858
+    http://<RESOLVER_URL>/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858 => https://my-domain.org/access/12345/58fe1b06-ee88-41ca-9bc9-320dcac6c858
+    Joels-MacBook-Pro-2:handlepid joeldunham
```
sevein: What does this line mean?

src/archivematicaCommon/tests/test_bindpid.py
```diff
@@ -0,0 +1,131 @@
...
+}
+
+
+class TestBindPID(TestCase):
+
+    def setUp(self):
+        """Set some stuff up...
+        """
+
+    def tearDown(self):
```
sevein: If unused, can we delete them?

src/dashboard/src/components/administration/views.py
```diff
@@ -351,6 +352,23 @@ def processing(request):
...
 
+def handle_config(request):
+    """Display or save the Handle configuration form, which allows for the
+    specification of configuration values for Handle PID creation and binding
+    using the ``bindpid`` module. State is stored in DashboardSettings table.
+    """
+    if request.method == 'POST':
+        form = HandleForm(request.POST)
+        if form.is_valid():
+            models.DashboardSetting.objects.set_dict(
```
sevein, line 363: Happy to see that this is being used!

src/dashboard/src/components/administration/views.py
```diff
@@ -351,6 +352,23 @@ def processing(request):
...
+                'handle', form.cleaned_data)
+            messages.info(request, _('Saved.'))
+    else:
+        form = HandleForm(
+            initial=models.DashboardSetting.objects.get_dict('handle'))
+    return render(request, 'administration/handle_config.html', locals())
```
sevein: Should we replace `locals()` with an explicit dict of things we're passing to the template?

src/dashboard/src/templates/administration/sidebar.html
```diff
@@ -48,7 +49,11 @@
       <li class="{% active request url_user_list %}"><a href="{{ url_user_list }}">{% trans "Users" %}</a></li>
     {% endif %}
 
-    <li class="{% active request url_language %}"><a href="{{ url_language }}">{% trans "Language" %}</a></li>
+    <li class="{% active request handle_config %}"
+        ><a href="{{ handle_config }}">{% trans "Handle server config" %}</a></li>
```
sevein: This `>` in a new line is pretty new to me, never seen this before in the Dashboard. Why are you doing it this way? Do we want to start forcing a 79col rule in HTML templates too from now on?

sevein (changes requested):
Thanks, @jrwdunham. Looking great. I've shared some comments.

src/archivematicaCommon/lib/databaseFunctions.py
jrwdunham: I often call "model class instances" just "models" as a shorthand; given the context, I think it makes sense. Similarly, I informally type-annotate such instances with `_mdl`, e.g., `file_mdl = File()` if I think it improves readability. But I guess you're right that in this case it kinda sounds like I'm dynamically create Django classes, so I'll change it.

jrwdunham: TODO: make this a method on the `Directory` class itself (thanks sevein).

sevein: > I often call "model class instances" just "models" as a shorthand; given the context, I think it makes sense. Similarly, I informally type-annotate such instances with _mdl, e.g., file_mdl = File() if I think it improves readability. But I guess you're right that in this case it kinda sounds like I'm dynamically create Django classes, so I'll change it.

Makes total sense, I like it! 👍

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
jrwdunham: And in bind_pid.py too! I like to practice RY. Just kidding: will fix.

sevein: :dancing_men:

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
jrwdunham: I'm not sure I agree that the name doesn't describe the behaviour: it clearly states that the function should exit (i.e., return) under a certain condition. Just because Python doesn't succinctly let you return from function `f` by calling function `g` from `f`, doesn't mean you shouldn't be able to get around that. Note that the calls to `_get_subdir_paths` and `_get_transfer_mdl` are doing the same thing: they will cause `main` to exit if a known exception is raised too.

Would it help if I renamed `_exit_if_not_include_dirs` to `_return_if_not_include_dirs` or `_assert_include_dirs`? Or do you really just want to see `return` statements in `main`?

src/archivematicaCommon/lib/bindpid.py
jrwdunham: It means I copy-pasted without editing :) (you mean just `Joels-MacBook-Pro-2:handlepid joeldunham`, right?)

src/archivematicaCommon/lib/bindpid.py
jrwdunham: Good idea. Will do.

src/archivematicaCommon/lib/bindpid.py
jrwdunham: I guess I was trying to highlight which parts of the "bind PID" request are baked-in and which are user-configurable. The fact that we have XML in the request body, the fact that we're making a POST request, etc.—these are currently non-configurable. Maybe I'll just remove it.

src/archivematicaCommon/tests/test_bindpid.py
jrwdunham: Sure. Will do.

src/dashboard/src/components/administration/views.py
jrwdunham: I guess so. Using `locals()` like this is new to me, but it seems to be used all over the place in AM/SS when rendering templates. I thought I was just being consistent...

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
sevein: > Note that the calls to _get_subdir_paths and _get_transfer_mdl are doing the same thing: they will cause main to exit if a known exception is raised too.

But you didn't name it as `_exit_or_get_subidr_paths`...

I think it's ok the way it is, clear enough. I think this comes down to the fact that I don't find error handling via exceptions better because the flow seems to be all over the place. This is probably me biased by having spent time using Go where you have to deal with the errors right away.

Thanks for the explanation, please leave it as it is!

src/dashboard/src/templates/administration/sidebar.html
jrwdunham: Haha. Sorry, that's just me. Everything I write has a 79col rule! (emails, notes, html) I know it's crazy. I have line-wrapping turned off in my editor, which is why you see this kind of stuff.

But there's another problem here, no? It looks like I removed the "Language" link!

jhsimpson (approved):
I tested this branch using the acceptance-tests https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/15 and the sampledata https://github.com/artefactual/archivematica-sampledata/pull/4

and the feature works as expected.  I think this is ready to merge.

src/dashboard/src/main/models.py
```diff
@@ -414,17 +439,69 @@ class File(models.Model):
...
+        }
+
+    @classmethod
+    def create_many(cls, dir_paths_uuids, unit_mdl, unit_type='transfer'):
+        """Create ``Directory`` models to encode the relationship between each
+        directory path/UUID pair in ``dir_paths_uuids`` and the ``Transfer`` model
+        that the directories are a part of.
+        """
+        unit_type = {'transfer': 'transfer'}.get(unit_type, 'sip')
+        cls.objects.bulk_create([
```
sevein: Should this return? I'm thinking at some point this could be chained.

src/dashboard/src/templates/administration/handle_config.html
```diff
@@ -0,0 +1,35 @@
...
+      </form>
+
+    </div>
+
+  </div>
+
+</div>
+
+{% endblock %}
+
```
sevein: Can you delete this empty line?

sevein (approved):
👍  tested using am.git!

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
```diff
@@ -0,0 +1,133 @@
...
+    """Return a generator of subdirectory paths in ``root_path``."""
+    if not os.path.isdir(root_path):
+        logger.warning('Transfer path %s is not a path to a directory',
+                       root_path)
+        raise DirsUUIDsException
+    # objects/ and root dirs need no UUIDs.
+    exclude_paths = (root_path, os.path.join(root_path, 'objects'))
+    return (format_subdir_path(dir_path, root_path) for dir_path, __, ___ in
+            os.walk(root_path) if dir_path not in exclude_paths)
+
```
helenst: I find this formatting of the generator expression slightly hard to read, and beginning a new line with the `for` or `if` clause makes things clearer.

```python
return (format_subdir_path(dir_path, root_path) 
        for dir_path, __, ___ in os.walk(root_path)
        if dir_path not in exclude_paths)
```

src/MCPClient/lib/clientScripts/bind_pid.py
```diff
@@ -0,0 +1,156 @@
...
+The idea is to allow for PURL resolution like:
+
+    http://<PID-RESOLVER>/<NAME_AUTH>/<PID>
+        => http://my-org-domain.org/access/files/<PID>
+    http://<PID-RESOLVER>/<NAME_AUTH>/<PID>?locatt=view:preservation
+        => http://my-org-domain.org/preservation/files/<PID>
+    http://<PID-RESOLVER>/<NAME_AUTH>/<PID>?locatt=view:original
+        => http://my-org-domain.org/original/files/<PID>
+
+The sole command-line argument is the File's UUID. If the --pind-pids option
```
helenst: typo here (`--pind-pids`) and in a couple of other places

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -0,0 +1,459 @@
...
+    if not cf:
+        return {}
+    if not os.path.isfile(cf):
+        print('Warning: there is no config file at {}'.format(cf))
+        return {}
+    config = configparser.ConfigParser()
+    with open(cf) as filei:
+        config.read_file(filei)
+    return {key: config.get('Handle', key, fallback=None) for key in CFGABLE_PARAMS}
+
```
helenst: For Python 2, `read_file` needs to be `readfp` and the `fallback` kwarg isn't supported

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -333,11 +333,30 @@ def copy_to_start_transfer(filepath='', type='', accession='', transfer_metadata
...
+    directories. If ``True`` is returned, we assign new UUIDs to all
+    directories in the arranged SIP.
+    """
+    file_uuids = filter(None, [file_.get('uuid') for file_ in files])
+    return any([
+        rslt[0] for rslt in
+        models.Transfer.objects.filter(
+            file__uuid__in=file_uuids).values_list(
+                'diruuids').distinct()])
+
```
helenst: If I am understanding this function right, it could be simplified to remove the need of extra Python code after querying:

e.g.
```python
return Transfer.objects.filter(file__uuid__in=file_uuids, diruuids=True).exists()
```

src/dashboard/src/main/models.py
```diff
@@ -396,11 +399,33 @@ def __str__(self):
...
+                            null=True, blank=False)
+    value = models.TextField(
+        verbose_name=_l('Identifier Value'),
+        help_text=_l('Used for premis:objectIdentifierType and'
+                     ' premis:objectIdentifierValue in the METS file.'),
+        null=True, blank=False)
+
+    def __str__(self):
+        return (u'Identifier {i.identifiervalue} of type'
+                ' {i.identifiertype}'.format(a=self))
```
helenst: should be `i=self`

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
```diff
@@ -1077,26 +1172,69 @@ def write_mets(tree, filename):
...
+        LABEL=sip_dir_name)
+
+    # Get the <dmdSec> for the entire AIP; it is associated to the root
+    # <mets:div> in the physical structMap.
+    sip_mdl = SIP.objects.filter(uuid=fileGroupIdentifier).first()
+    if sip_mdl:
+        aipDmdSec = getDirDmdSec(sip_mdl, sip_dir_name)
+        globalDmdSecCounter += 1
+        dmdSecs.append(aipDmdSec)
+        aip_dmd_id = "dmdSec_" + globalDmdSecCounter.__str__()
```
helenst: instead of calling `__str__` directly you could do `str(globalDmdSecCounter)` which is more idiomatic

src/dashboard/src/components/administration/views.py
helenst: I agree it's preferable to name what we are passing to the template (explicit is better than implicit) - it does make it easier to see what's going on and provides a more clear interface between the view and template. Perhaps we could adopt this more going forwards.

helenst (commented):
Hi @jrwdunham this is awesome! I got the acceptance tests to pass (eventually, since it was the first time I've run them) and played around with the bindpid script. I commented on a couple of small things that were causing errors, plus some style points.

src/dashboard/src/components/administration/views.py
jrwdunham: All right, all right, 2 against 1 and a quote from Zen: will change.

src/MCPClient/lib/clientScripts/bind_pid.py
jrwdunham: Thanks. Found "pind-pids" here and in bind_pids.py too.

src/MCPClient/lib/clientScripts/assignUUIDsToDirectories.py
jrwdunham: yeah, that is more readable.

src/archivematicaCommon/lib/bindpid.py
jrwdunham: Good catch. On my dev machine I must have installed `configparser` in my 2.7.6 Python so I wasn't seeing this error. With the changes coming in, I have tested command-line usage of bindpid.py and it seems to be working on 2.7.6, 2.7.8 and 3.4.2.

src/dashboard/src/main/models.py
jrwdunham: yup. nice catch.

src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
jrwdunham: True. As with the `locals()` issue, this was a result of being consistent with the rest of the codebase, even if it is non-idiomatic, just like when I use camelCase to match the style of legacy code, etc. But I'll change it.

src/dashboard/src/components/filesystem_ajax/views.py
jrwdunham: Right. That looks better. Thanks.

src/dashboard/src/templates/administration/handle_config.html
jrwdunham: sure.

src/dashboard/src/main/models.py
jrwdunham: made it so.

jrwdunham:
Thanks @sevein, @helenst and @jhsimpson for all the great code review; I think it improved this PR quite a bit.

PR #695, opened 2017-08-02
Delete tika from osdeps

sevein:


jhsimpson (approved):
Tika is not being referenced by any fpr rules in Archivematica at the moment. 

:+1:

PR #696, opened 2017-08-02
Multiple fixes

### LINKED ISSUE #699: Problem: antivirus scanning is broken in standard installation
Only affecting qa/1.x (1.7 not released yet).

By standard I mean: Ansible or packages, where MCPClient and clamd run in the same machine.
In a36a9bf we added the ability to talk to a clamd instance over the network but in addition to that we started passing the files to clamd by reference which doesn't always work (depending on how the permissions are set up).

This is being fixed in #696 (9198655a97ec6e388cb13b51a24f5e212d1fe1b1), where we're adding an option so the user can decide if the file is passed to the antivirus by reference or by value (streaming). It defaults to streaming because it doesn't require any special config. Passing by reference may be better in some situations but it also requires certain level of permission orchestration.

sevein:
This fixes #699 (see 9198655) and other small fixes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@
 
 # tox working directory
 .tox/
+
+src/dashboard/src/static/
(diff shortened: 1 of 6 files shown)


jhsimpson (approved):
:+1:

jhsimpson (approved):
I haven't tested this yet, but it looks good to me.

Issue #669, opened 2017-07-07
Add default location option (merge from Jisc fork)


helenst:
Merge work from https://github.com/JiscRDSS/archivematica/pull/22 into core.

This adds a default location option to the storage directories.

Related SS issue: https://github.com/artefactual/archivematica-storage-service/issues/211

helenst:
I'm doing this on the same branch as #667 since it relies heavily on the work there - will PR them together.

Branch here: https://github.com/helenst/archivematica/compare/dev/auto-deployment...helenst:dev/automated-workflow

sallain:
See comment re: testing on the related Storage Service issue: https://github.com/artefactual/archivematica-storage-service/issues/211

sallain:
This has been documented in archivematica-docs/user-manual/administer.rst (_Processing configuration fields_) as well as in the Storage Service documentation.

Issue #667, opened 2017-07-07
Add automated workflow option (merge from Jisc fork)


helenst:
Merge work from https://github.com/JiscRDSS/archivematica/pull/15 into core.

This adds an automated workflow configuration to Archivematica. Uses config so possibly best done after https://github.com/artefactual/archivematica/issues/664

helenst:
Builds on work done in #681 (when the processing config was brought into code rather than from a file)

Branch here: https://github.com/helenst/archivematica/compare/dev/auto-deployment...helenst:dev/automated-workflow

sevein:
#681 is now merged!

PR #698, opened 2017-08-02
Automated workflow

helenst:
Automated workflow option and default location, from JiscRDSS

Bringing in #667 and #669 together as the dependency between these two fixes is quite tight.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@
 
 # tox working directory
 .tox/
+
+src/dashboard/src/static/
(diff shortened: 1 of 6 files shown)


sevein (changes requested):
The automated workflow doesn't seem to include choices for new steps in qa/1.x like `Perform policy checks on originals`, `Perform policy checks on preservation derivatives`, etc...

helenst:
@sevein Thanks! I've rebased against qa/1.x and added those policy check steps to both automated and default workflows (they were missed when that got moved). Github won't let me reopen this and seems confused about the commits so I'll open a new PR.

PR #701, opened 2017-08-03
Automated workflow and default location

helenst:
Another attempt following on from #698

Now based against the qa/1.x branch and including policy check configurations for both automated and default workflows.

I have been testing this with https://github.com/artefactual/archivematica-storage-service/pull/212 and it seems fine. (I am getting store DIP failures but I also get those on qa/1.x)

### DIFF
--- a/src/MCPClient/lib/clientScripts/getAipStorageLocations.py
+++ b/src/MCPClient/lib/clientScripts/getAipStorageLocations.py
@@ -20,6 +20,7 @@ def get_aip_storage_locations(purpose):
     choices = {}
     for storage_dir in storage_directories:
         choices[storage_dir['description']] = storage_dir['resource_uri']
+    choices['Default location'] = '/api/v2/location/default/{}/'.format(purpose)
     print(choices)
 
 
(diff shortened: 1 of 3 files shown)


sevein:
Refs #667 #669

sevein (changes requested):
@helenst small comment, I think we're also missing entries in the processing configuration for `Store DIP location` right?

I would recommend to modify the `automated` workflow from the Dashboard (admin > processing config). Once you do that the new document should be available at `/var/archivematica/sharedDirectory/sharedMicroServiceTasksConfigs/processingMCPConfigs/automatedProcessingMCP.xml`. You can copy that again into `src/MCPServer/lib/archivematicaMCP.py`.

image

I was going to make the edits myself but I don't have access to your repo :) I got too used to put my hands in topic branches that don't belong to me but that's probably not great.

jrwdunham (approved):
👍 lgtm

sevein:
I merged. I'll do the suggested changes in a separate PR.

Issue #704, opened 2017-08-08
Problem: using symlinks breaks Windows dev environments


jhsimpson:
Archivematica cannot be deployed on Windows, but this PR from @minusdavid https://github.com/artefactual/deploy-pub/pull/39 makes it possible to deploy a development environment on Windows, using vagrant to deploy to a linux vm.

That PR is working great, but there is a problem with checking out a git repo that contains symlinks into a windows filesystem (google it, lots of links).  Windows doesn't properly support symlinks, and so checking out a repo with symlinks is difficult, ansible roles choke, you get weird git errors, etc.

In this repo, there are only a few symlinks being used - it would not be hard to remove them altogether.  I think the only place left is in the osdeps folders.  Removing those symlinks and creating duplicate files for now would allow osdeps to differ for each platform, which is fine, and would make developing in a Windows environment much easier, which is a bonus.

jhsimpson:
This is also an issue in the storage service (same osdeps files).

minusdavid:
Alas, it looks like the symlinks in archivematica-storage-service stable/0.10.x are what's preventing the checkout to qa/0.x:

TASK [artefactual.archivematica-src : Checkout archivematica-storage-service code] ***
fatal: [am-local]: FAILED! => {"changed": false, "cmd": "/usr/bin/git checkout --track -b qa/0.x origin/qa/0.x", "failed": true, "msg": "Failed to checkout qa/0.x", "rc": 1, "stderr": "error: Your local changes to the following files would be overwritten by checkout:\n\t
osdeps/RedHat-7.json\n\tosdeps/Ubuntu-16.json\n\tstorage_service/static/js/vendor/base64.js\nPlease, commit your changes or stash them before you can switch branches.\nAborting\n", "stderr_lines": ["error: Your local changes to the following files would be overwritten by
 checkout:", "\tosdeps/RedHat-7.json", "\tosdeps/Ubuntu-16.json", "\tstorage_service/static/js/vendor/base64.js", "Please, commit your changes or stash them before you can switch branches.", "Aborting"], "stdout": "", "stdout_lines": []}
        to retry, use: --limit @/vagrant/singlenode.retry

Because the working directory is dirty, it won't checkout. And even though we specify "force: yes" in Ansible, I think the Ansible git module has a bug where it doesn't force checkouts if the branch being checked out is a remote branch.

minusdavid:
[...]

PR #705, opened 2017-08-08
Remove symlinks

sevein:
Using symlinks cause Windows deployments to become much harder.
See #704 for more details.


sevein:
The diff is not super clear but you can run:

```
$ find . -type l -ls | wc -l
0
```

jhsimpson (approved):
tested and this works for me.

👍 

realized we need a PR in the storage service as well.

PR #711, opened 2017-08-11
Fix checksum verification error/success

### LINKED ISSUE #685: Problem: Verify Checksum “completed successfully” even though not valid
**Verify Transfer Checksum “completed successfully” even though the checksum is not valid**
**Background:** testing in the demo-jisc environment for the Jisc RDSS project. I created a transfer with one file in it, which included an md5 checksum.  The checksum was provided in the ‘request preservation’ message that we mock from the rdss. This is sent using the msgCreator app to the Archivematica Channel Adaptor, which takes the checksum metadata (uuid, checksum type, checksum value) and creates a checksum file that is included in the transfer sent to Archivematica. 

I had already tested with a transfer where the checksum was valid, so in this case, I wanted to test what happened when the checksum provided was incorrect (i.e. would not match a new checksum generated by the same file). I changed 1 digit in the checksum value to do this.

**Expected Behaviour:** When the transfer was processed, I expect the Verify Transfer Checksum to fail; for that failure to trigger the “Failed Transfer” micro service; and then stop processing of the transfer.  

Actual Behaviour: The “Verify Transfer Checksum” indicated that it “completed successfully” and processing of the Transfer continued (straight through to ingest). 

a key detail: when I intentionally corrupted the checksum value, I replaced a hex character with a ’t’ (not considering that the checksum digits must all be hex characters). When I re-ran the test with a hex character instead of a t (but still not the original value), the micro service fails as expected. Normally we expect the checksum to be fine, but for the verification to fail because the file is corrupted. In my scenario, the file was fine but the checksum provided with it was corrupted. So this scenario may not happen frequently.

Output from the task is included below for reference. Discussed the defect with Jesús who wrote the channel adaptor code. His initial assessment is that this has to do with the verify checksum micro service, and is not due to the way the channel adaptor is creating creating a checksum file from metadata. 

— output Job Verify metadata directory checksums
```

Task 63716561-a312-4eb5-89d2-de615bc5a0f3

File UUID
Unknown
File name
test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be
Client
31bf6692ac10_2
Exit code
0
Start time
6/27/2017, 6:15:25 PM
End time
6/27/2017, 6:15:25 PM
Created time
6/27/2017, 6:15:25 PM
Duration
< 1 seconds(s)
Module verifyMD5_v0.0
"%sharedPath%currentlyProcessing/test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be/" "%date%" "%taskUUID%" "a932afa6-e2db-4645-82df-529cc38cc3be"
Standard streams

```
Standard output (stdout)
File Does not exist: /var/archivematica/sharedDirectory/currentlyProcessing/test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be/metadata/checksum.sha1
File Does not exist: /var/archivematica/sharedDirectory/currentlyProcessing/test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be/metadata/checksum.sha256
Standard error (stderr)
md5deep: /var/archivematica/sharedDirectory/currentlyProcessing/test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be/metadata/checksum.md5: Unable to find any hashes in file, skipped.
md5deep: Unable to load any matching files.
Try `md5deep -h` for more information.
md5deep: /var/archivematica/sharedDirectory/currentlyProcessing/test3 for checksum fail-a932afa6-e2db-4645-82df-529cc38cc3be/metadata/checksum.md5: Unable to find any hashes in file, skipped.
md5deep: Unable to load any matching files.
Try `md5deep -h` for more information.

jraddaoui:
Fail the 'Verify metadata directory checksums' job when the checksum
tool fails checking the checksums and when there are failing checksums.

This closes #685.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCheckMD5NoGUI.sh
+++ b/src/MCPClient/lib/clientScripts/archivematicaCheckMD5NoGUI.sh
@@ -34,12 +34,16 @@ integrityReport="$3"
 checksumTool="$4"
 
 tmpDir=`pwd`
+ret=0
+
 cd "$checkFolder"
 #check for passing checksums
 "${checksumTool}" -r -m "$md5Digest" . > $passTmp
+ret+="$?"
 #check for failing checksums
 "${checksumTool}" -r -x "$md5Digest" . > $failTmp
-cd $tmpDir      
+ret+="$?"
+cd $tmpDir
 
 
 
@@ -66,4 +70,6 @@ cat $failTmp 1>&2
 #cleanup
 rm $failTmp $passTmp $reportTmp
 
-exit $numberFail
+ret+="$numberFail"
+
+exit ${ret}


sevein:
Thanks, @jraddaoui! This works as expected. I've tested four transfers: 1) invalid checksum file format (failed), 2) valid checksum file format but wrong checkums (failed), 3) valid checksums (passed), 4) no checksum files (passed).

PR #714, opened 2017-08-12
AT: disable ASpace notes edit

jraddaoui:
To avoid overwriting existing notes until a proper update method is
introduced, the note fields are disabled and the notes not sent while
a record is being updated.

- Update webpack file with fix and dependencies clean-ups
- Avoid adding jQuery (and other JS) twice in some templates

To review the appraisal-tab changes please check: https://github.com/artefactual-labs/appraisal-tab/commit/d6390bf401f0711651365654b12b706b23d90b9d

### DIFF
--- a/src/dashboard/src/templates/appraisal/appraisal.html
+++ b/src/dashboard/src/templates/appraisal/appraisal.html
@@ -13,9 +13,6 @@
 {% endblock %}
 
 {% block js %}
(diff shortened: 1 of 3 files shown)


jraddaoui:
Merged in https://github.com/artefactual/archivematica/commit/d23914ee763fa29c525ebc196a3ca66a273f1fa6

Issue #712, opened 2017-08-11
Problem:  timeouts preventing transfers from starting


jhsimpson:
Changes introduced into the qa/1.x branch in #531  (updated in #565) have changed the way http requests are working.  This is affecting the ability to start large transfers.

The communication between the dashboard and the storage service is synchronous - when a user clicks the 'start transfer' button - the dashboard sends an http request to the storage service to copy files to the activeTransfers watched Directories.   The request blocks until the storage service completes the process of moving the files around.  For large this this can take several minutes.

Ideally the  storage service would be modified to support asynchronous calls (accept a callback with the  original request, send 201, and call the callback when the work is completed).   For the 1.7.0 release, it may be necessary to revert the #531 change.

sevein:
Fixed in https://github.com/artefactual/archivematica/pull/747.

sromkey:
@currmie I believe @ross-spencer might have some advice on how to qa this (e.g. how to generate a large transfer)

currmie:
Verification status: I can start large file transfers but tests still fail, this time during the Index AIP job.

PR #715, opened 2017-08-16
Timeout setting for large file transfers

mamedin:
There was an issue with large file transfers that were not being loaded into the dashboard. With this change in timeout the transfer gets to the dashboard to be accepted or rejected.

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -54,7 +54,7 @@ def _storage_service_url():
     return storage_service_url
 
 
-def _storage_api_session():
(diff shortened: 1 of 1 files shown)


sevein:
See #712.

mamedin:
This issue is also being followed in:

JiscRDSS/rdss-archivematica#56

The timeout setting has been changed to the copy_files function instead of changing it globally in the storage_api_session. We could add timeout=300 in this call instead, and then it would only affect the move_files operation, not all usage of the storage service API.

sevein:
See https://github.com/JiscRDSS/archivematica/pull/30.

PR #717, opened 2017-08-17
MCPClient: fixes in normalization process

jraddaoui:
- Don't return NO_RULE_FOUND or RULE_FAILED when exists and doesn't fail
- Avoid exception getting default rule
- Don't output to stderr when no rule was found
- Don't report normalization failures when no rule was found

### DIFF
--- a/src/MCPClient/lib/clientScripts/normalize.py
+++ b/src/MCPClient/lib/clientScripts/normalize.py
@@ -332,30 +332,33 @@ def main(opts):
             )
         return SUCCESS
 
+    do_fallback = False
     format_id = get_object_or_None(
         FileFormatVersion,
         file_uuid=opts.file_uuid
     )
 
     # Look up the normalization command in the FPR
-    if format_id is None:
-        rule = get_default_rule(opts.purpose)
-        print(os.path.basename(file_.currentlocation), "not identified - falling back to default", opts.purpose, "rule")
-    else:
+    if format_id:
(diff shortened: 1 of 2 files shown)


sevein:
I realize it must be really hard for the user to understand the origin of this pull request. The original issue ticket is only encoded in the name of the branch and even if someone could figure that it's accessible from https://projects.artefactual.com/issues/11273, the issue happens to be private, only available for Artefactual staff.

Could you file an issue in this repo describing the problem or add some more context to the description of this pull request? What's the problem that you're trying to solve?

jraddaoui:
Sure, np!

https://github.com/artefactual/archivematica/issues/718

jrwdunham (approved):
The logic of this looks sound and I tested it as follows. I created a transfer with this branch which contained a CSV file that was normalized for preservation: the normalization micro-services returned exit code 2 for this file as expected, the absence of a rule was written to stdout not stderr, and there was no failure listed in the failures page under the admin tab (cf.
https://projects.artefactual.com/issues/11273#note-12). I ran the same transfer through a qa/1.x deploy and saw the reported error: a failure was listed under the admin tab, an undesired failure report was emailed, and the absence of a rule was (contrary to desire) written to stderr.

jraddaoui:
Thanks @jrwdunham, merged in https://github.com/artefactual/archivematica/commit/533ba3dc4e37ada678dee22aef0df57eb1d2dd64

PR #720, opened 2017-08-18
Fixes #719: ES broken in backlog

### LINKED ISSUE #719: ElasticSearch indexing broken on send to backlog
To recreate: on an AM qa/1.x install, follow these steps:

1. Destroy the ES indices: `curl -XDELETE localhost:9200/*`
2. Create an AIP all the way through to local storage. 
3. Create a Transfer and put it in backlog.

At step (3) you should see a failure at micro-service "Create SIP from Transfer: Index transfer contents". Looking at the Task/Job, you will see that the empty string is the value for `modification_date` for the METS.xml file: this raises `IllegalArgumentException[Invalid format: ""]`, which raises `MapperParsingException`. 

See https://github.com/artefactual/archivematica/pull/635 and https://github.com/artefactual/archivematica/pull/692. Refs RM issue https://projects.artefactual.com/issues/11437

jrwdunham:
See https://github.com/artefactual/archivematica/issues/719 for how to recreate the issue on qa/1.x. Follow the same steps with this branch and expect to no longer see the issue.

I'm not sure if this is the best way to fix this issue, but it appears to be working. Maybe it would be better to ensure that transfer METS files DO have a modification time somehow instead loosening the Elasticsearch validation/type checking.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -339,6 +339,13 @@ def set_up_mapping_transfer_index(client):
         'status': MACHINE_READABLE_FIELD_SPEC,
         'origin': MACHINE_READABLE_FIELD_SPEC,
         'ingestdate': {'type': 'date', 'format': 'dateOptionalTime'},
+        # METS.xml files in transfers sent to backlog will have '' as their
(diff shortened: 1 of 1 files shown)


mcantelon (approved):
LGTM! :+1:

PR #722, opened 2017-08-22
Update link

kant:


jhsimpson:
Thanks for submitting this pull request, it's nice to have extra eyes to catch typos like that.

This is a just a small grammatical fix, so it is not necessary, but if you're willing to would you consider submitting a contributor's agreement?  https://wiki.archivematica.org/Contributor_Agreement  In case you submit larger pull requests in the future.

jhsimpson (approved):
:+1:

PR #738, opened 2017-09-27
Mirror locations and AIP replication

jrwdunham:
Introduces functionality related to mirror/replicator locations and AIP replication. Since the Storage Service (SS) is now replicating AIPs into user-configured replicator locations, it must now also create pointer files for the replicas and add replication events (and other details) to the replicated AIPs' pointer files. As a result, pointer file creation has been moved to the SS, which further necessitates that AM provide information about the AIP (e.g., compression event-related info) when requesting AIP storage from the SS.

1. The storeAIP micro-service now passes events, agents and aip_subtype in the package create call to the SS.
2. Removed the createPointerFile client script and micro-service (migration) since pointer file creation is now performed by the SS.
3. Adds the compression algorithm to the compression event's eventDetail in a parseable way so that the SS can extract it and use it.
4. Uses a dev branch of mets-reader-writer in requirements files. **Will need to be updated to a tag once metsrw's version is bumped.**


sevein (approved):
Excellent! 👍

PR #741, opened 2017-10-03
MCPClient: check uploadDIP/ dir in Store DIP (Fixes #729)

### LINKED ISSUE #729: Store DIP fails
On a new AM 1.x VM, the following condition fails at Store DIP:

- DIP upload: Do not upload
- Store DIP: Store in standard Archivematica directory

The error output reads:
`OSError: [Errno 2] No such file or directory: '/var/archivematica/sharedDirectory/watchedDirectories/uploadedDIPs/tester-img-b161664e-dfd3-45a0-87bd-9cc36c14643e'`

May be related to #680.

jrwdunham:
If the user chooses "Do not upload DIP" at the "Upload DIP?" choice point, then the DIP will not be moved to the uploadedDIPs/ directory but will stay in its original location in uploadDIP/. However, the "Store DIP" micro-service expects the DIP to be in the uploadedDIPs/ directory. This changes the storeAIP.py micro-services so that if the aip_path does not exist, the path is modified so that the first uploadedDIPs/ directory in the AIP path is replaced by uploadDIP. Not the most elegant solution, but it works.  Fixes #729.

### DIFF
--- a/src/MCPClient/lib/clientScripts/storeAIP.py
+++ b/src/MCPClient/lib/clientScripts/storeAIP.py
@@ -23,7 +23,6 @@
 from __future__ import print_function
 from annoying.functions import get_object_or_None
 import argparse
-import logging
 import os
 import sys
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


kellyannewithane:
@nickwilkinson Can you find someone to CR this?

sevein:
I've rebased.

sevein (changes requested):
@jrwdunham, I'ves tested it and it's working.  👍 

I think the commit message could include the same description that you've added to the pull request, don't you think? I'd also like to ask you to change the subject line. We normally use the [imperative mode](https://chris.beams.io/posts/git-commit/#imperative) and prefix the commit with the component being modified (e.g. `MCPClient: ...`).

jrwdunham:
@sevein I changed the commit msg and the PR title and initial comment. Are we good to merge or were there changes requested on the code?

sevein (approved):

sevein:
Rebased. I'm planning to merge.

PR #744, opened 2017-10-05
MCPClient: fix README microservice in Docker

sevein:
The `assets/` directory is by default expected to be at
`/usr/lib/archivematica/MCPClient/assets/`. However this is different in the
Docker build. This commit uses the corresponding environment variable to
redefine its location.

---

Screenshot of the error that this issue is trying to address:

image


mamedin (approved):
I have tested in a docker build and it fixes the issue.

PR #754, opened 2017-10-12
Enable re-login after Shibboleth logout

helenst:
django-shibboleth-user sets a session flag to prevent immediate bounce back to logged in state after logout. This is useful in some setups but is unnecessary here, and it prevents logging back in again until cookies are cleared.

I've added a custom logout which unsets that session flag after the standard logout. This should enable users to log back in.

Fixes https://github.com/artefactual/archivematica-storage-service/issues/238

See also https://github.com/artefactual/archivematica-storage-service/pull/245 for SS PR


helenst:
oops... I'll recreate this as a PR from my fork.

PR #756, opened 2017-10-12
Enable re-login after Shibboleth logout

helenst:
django-shibboleth-user sets a session flag to prevent immediate bounce back to logged in state after logout. This is useful in some setups but is unnecessary here, and it prevents logging back in again until cookies are cleared.

I've added a custom logout which unsets that session flag after the standard logout. This should enable users to log back in.

Fixes artefactual/archivematica-storage-service#238

See also artefactual/archivematica-storage-service#247 for SS PR


sevein:
:+1:

lwo:
Excellent.... we'll try it out tomorrow.

Issue #624, opened 2017-05-31
Problem: email report script fails to close email client


sevein:
`emailFailReport.py` calls quit() inside a finally block which is not working if the client could not be instantiated in the first place, e.g. see https://stackoverflow.com/q/984526.

```
File "/src/MCPClient/lib/clientScripts/emailFailReport.py", line 75, in send_email
    server.quit()
AttributeError: 'str' object has no attribute 'quit'
```

jhsimpson:
In a production deployment, the file would be in /usr/lib/archivematica/MCPClient/lib/clientScripts, rather than src/MCPClient/...

sevein:
In #757 I suggest a solution based on the django.core.mail package, instead of using `smtplib` directly. One benefit is that it lets us define a It also brings the benefit of letting us set up different email configuration settings via our settings modules. Currently, MCPClient relies on the framework defaults which results in the same behaviour AM16 has:

- `EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend`
- `EMAIL_HOST=localhost`

That means that AM17 still expects to have locally available a MTA listening on 25/tcp. I believe our packages and Ansible role deploy a Postfix server. In the RDSS fork of Archivematica Richard added support for Amazon SES and Helen added better configuration support via environment variables or the MCPClient INI file, see: https://github.com/JiscRDSS/archivematica/pull/35.

sromkey:
@sevein could you please explain how to QA, then reassign to @currmie. Thanks!

sevein:
The problem was that when email delivery was unsuccessful (e.g. MTA is unavailable) the client script would fail to handle the error gracefully. For this to be tested I guess we would need to reproduce a couple of scenarios:

1. The email settings are set up properly pointing to the SMTP server. In Ansible/packages I think this is typically done with a local Postfix MTA. In this case, emails should be delivered properly to the user.

2. The email settings point to a SMTP server that isn't available. In this case, the email report should return an error (i.e. `Report email was not delivered`) instead of an uncontrolled exception (e.g. `AttributeError: 'str' object has no attribute 'quit') .

[...]

PR #757, opened 2017-10-13
MCPClient: use django.core.mail in emailFailReport

sevein:
This solves #624.

### DIFF
--- a/src/MCPClient/lib/clientScripts/emailFailReport.py
+++ b/src/MCPClient/lib/clientScripts/emailFailReport.py
@@ -2,7 +2,7 @@
 
 # This file is part of Archivematica.
 #
-# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
+# Copyright 2010-2017 Artefactual Systems Inc. <http://artefactual.com>
 #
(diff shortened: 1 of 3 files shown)


jrwdunham (commented):
Overall this looks good to me (nice to see tests showing success and error cases). 

One thing I'm wondering about is why the use of bare exceptions? Can't we continue to catch `smtplib.SMTPException`? From the django docs at https://docs.djangoproject.com/en/1.8/topics/email/: "send_mail will raise an smtplib.SMTPException. See the smtplib docs for a list of possible exceptions, all of which are subclasses of SMTPException."

sevein:
> One thing I'm wondering about is why the use of bare exceptions? Can't we continue to catch smtplib.SMTPException? From the django docs at https://docs.djangoproject.com/en/1.8/topics/email/: "send_mail will raise an smtplib.SMTPException. See the smtplib docs for a list of possible exceptions, all of which are subclasses of SMTPException."

Because depending on the email backend we may or we may not see that exception, unfortunately. E.g. we've confirmed that the Amazon SES backend we're using @ RDSS raises an exception that belongs to boto3 the library and it doesn't subclass SMTPException.

jrwdunham (approved):

PR #761, opened 2017-10-14
Workflow: directory UUIDs for all transfer types

### LINKED ISSUE #742: Problem: UUIDs not assigned to directories for non-standard transfer types
We are testing the new microservice 'Assign UUIDs to directories' with the latest code changes from the qa/1.0 branch. However, even though we configured the default processing configuration to assign UUIDs to directories, we don't see this microservice called during transfer. I assume this job should have been called right before or right after the job 'Assign file UUIDs to objects' was called, but all I see in the dashboard for 'Approve transfer' is this:
```
Microservice: Approve transfer
- Job Include default Transfer processingMCP.xml 
- Job Verify bag, and restructure for compliance 
- Job Assign checksums and file sizes to objects 
- Job Assign file UUIDs to objects 
- Job Rename with transfer UUID 
- Job Extract zipped bag transfer
- Job Approve zipped bagit transfer
```
Finally, when we obtain the METS from the AIP, we only see UUIDs for files. So no UUIDs for directories here in the METS either.

jrwdunham:
1. Add a migration that creates "Assign UUIDs to directories?" and "Assign UUIDs to directories" micro-service chain links at appropriate locations in the workflow for the following transfer types: Bag, DSpace, maildir, and TRIM.
2. Unify distinct but identical choices. Assume that "decision point" means "a situation where one has options" and "decision" means "the course of action that is ultimately pursued". Archivematica allows for the creation of distinct decision points and decisions (i.e., model instances or db rows) which are in fact semantically identical. This commit allows one decision (point) to be designated as canonical. The configuration of a canonical decision (point) can now be made to apply to all of its counterparts.
    Introduces the utils.py::choice_unifier dict that maps decision point and decision UUIDs to their designated canonical equivalents. This allows a processing MCP XML configuration file to reference UUIDs for canonical decision points/decisions and have the equivalent points/decisions be configured also.

Fixes #742 

For relevant acceptance test mods, see https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/26

### DIFF
--- a/src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
+++ b/src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
@@ -28,6 +28,7 @@ import os
 import threading
 import time
 
+from utils import choice_unifier
 from linkTaskManager import LinkTaskManager
 from linkTaskManagerChoice import choicesAvailableForUnits, choicesAvailableForUnitsLock, waitingOnTimer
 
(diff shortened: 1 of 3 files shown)


src/dashboard/src/components/administration/forms.py
```diff
@@ -32,6 +33,9 @@
 import storageService as storage_service
 
 
+logger = logging.getLogger('archivematica.dashboard')
```
sevein: Why?

sevein (commented):
@jrwdunham, looks great to me. I'd suggest to put the MCPServer changes into their own commit because they're really different things and they don't depend on each other right?

I like how the terms choice points and options sound. I'm confortable with those being used but shouldn't we at least make some references (perhaps in the code comments) to how they're usually named in our docs ([MCP](https://wiki.archivematica.org/MCP))?

I'm glad we're not adding another table to encode these equivalents.

src/dashboard/src/components/administration/forms.py
jrwdunham: Because I was using it to debug and forgot to delete it. I'm surprised flake8 didn't complain. Maybe it did?

jrwdunham:
@sevein How are my "choice point" and "options" usually named in our docs? While useful, I find many things in the [MCP](https://wiki.archivematica.org/MCP) doc confusing. (On a separate note, it needs revision). 

I myself am not consistent in what I call these things. I think maybe a better guide would be the official user manual. Essentially, we are talking about the user's understanding of the /administration/processing/edit/ page, so maybe the [Dashboard administration tab](https://www.archivematica.org/en/docs/archivematica-1.6/user-manual/administer/dashboard-admin/#dashboard-processing) docs are a better guide. That page uses "decision point" quite consistently to describe the the lefthand side column. However, it uses "decision", "choice", "selection", "option" to describe the righthand side column.

It's made confusing by the fact that "choice" in English is ambiguous between the situation where one has options and the course of action that is ultimately pursued. I think the least ambiguous might be to use "choice point" and "decision", respectively, but maybe we should use "decision point" and "decision" to be more consistent with the Dashboard admin tab docs. What do you think?

jrwdunham:
@sevein I've split up the commit into Workflow: and MCPServer: and I've used "decision point" and "decision" consistently in the code comments, commit messages and the GH metadata.

sevein (commented):
Thanks for the thoughtful response.
It's perfect now. 👍

PR #762, opened 2017-10-17
Delete database connection pool

sevein:
Redmine: https://projects.artefactual.com/issues/10355

---

This is a sequel of #505 which was originally opened to replace the `django_mysqlpool` backend with the one provided by Django: `django.db.backends.mysql`. I believe it was based on the erroneous assumption that `django.db.backends.mysql` offered a database connection pool. What it does is to provide persistent connections which are not a solution for web servers using greenlets (Dashboard) or threaded applications like MCPServer or MCPClient.

I think that deleting `django_mysqlpool` is still a good thing because it doesn't support Py3 and it's pretty much abandoned. In this PR, MCP{Server,Client} start using `django.db.backends.mysql` again. The connection pool allowed us to recycle connections (opening connections is expensive) but it didn't solve what I think is the biggest scalability issue: the total number of connections opened simultaneously (see #553 for context). That's an issue that needs to be addressed separately, e.g. implementing contention before the thread opens a connection.

In this PR we also set up the Dashboard to use the `django.db.backends.mysql` backend again (I think that the change to `django_mysqlpool` was only an accident during the config cleanup in `qa/1.x`) and explicitly give `CONN_MAX_AGE` a value of 0 because that's needed when using gevent (greenlets), otherwise database connections will accumulate.

### DIFF
--- a/src/MCPServer/lib/settings/common.py
+++ b/src/MCPServer/lib/settings/common.py
@@ -45,7 +45,6 @@ CONFIG_MAPPING = {
     'db_password': {'section': 'client', 'option': 'password', 'type': 'string'},
     'db_host': {'section': 'client', 'option': 'host', 'type': 'string'},
     'db_port': {'section': 'client', 'option': 'port', 'type': 'string'},
-    'db_pool_max_overflow': {'section': 'client', 'option': 'max_overflow', 'type': 'string'},
 }
 
 
@@ -71,9 +70,8 @@ user = archivematica
 password = demo
 host = localhost
 database = MCP
-max_overflow = 40
 port = 3306
-engine = django_mysqlpool.backends.mysqlpool
+engine = django.db.backends.mysql
 """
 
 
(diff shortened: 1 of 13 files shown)


sevein:
I've processed a transfer with with 100 files of 8k of random data and also tried processing two of them simultaneously (i.e. 200 files across two transfers) and in both cases the number of total threads used by MCPServer wasn't higher than 75 threads which corresponds to the `limitTaskThreads` setting meaning that the number of database connections opened is normally going to be lower than that - each watched directory has its own and it's also counted.

src/MCPServer/lib/settings/common.py
```diff
@@ -93,17 +91,15 @@
...
         'PORT': config.get('db_port'),
-    }
-}
 
-MYSQLPOOL_BACKEND = 'QueuePool'
-MYSQLPOOL_ARGUMENTS = {
-    'use_threadlocal': False,
-    'max_overflow': config.get('db_pool_max_overflow'),
+        # CONN_MAX_AGE is irrelevant in MCPServer because Django's database
+        # connection reciclyng mechanism is only used in the web context, i.e.
```
jrwdunham, line 96: nitpick: typo

helenst (approved):
Nice work. Looks pretty sound to me. :+1:

jhsimpson (approved):
This looks quite sound to me.  I'd like to test it at scale, but I think we can do that after merging to qa/1.x - it has been tested enough to justify merging now.

sevein:
Thank you! I forgot to mention that our need for greenlets in the Dashboard or the Storage Service is pretty much imperative because our views do frequent IO blocking, e.g. waiting for SS to do something synchronously.

PR #765, opened 2017-10-19
Update fido to v1.3.7

### LINKED ISSUE #755: Problem: fiwalk characterization is broken
We have a class in [src/archivematicaCommon/lib/externals/fiwalk_plugins/pronom_ident.py](https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/externals/fiwalk_plugins/pronom_ident.py#L33) that subclasses `fido.Fido` (from https://github.com/openpreserve/fido).

```python
class FiwalkFido(fido.Fido):
    """docstring for FiwalkFido"""
    # [...]
    def identify_file(self, filename):
        """Identify the type of @param filename."""
        # [...]
        bofbuffer, eofbuffer = self.get_buffers(f, size, seekable=True)
```

The method `get_buffers` is treated as it returned a tuple of two items however the method changed a long time ago starting to return a tuple of three items: https://github.com/openpreserve/fido/blob/master/fido/fido.py#L536-L582.

It seems that this change was introduced in v1.3.4-84: https://github.com/openpreserve/fido/commit/d3d5e2368cdc633fd8125d2c1b3f234d9ccbda54.

---

Redmine: https://projects.artefactual.com/issues/11256 (private)

sevein:
Also taking this opportunity to fix #755 which relates to fido (see ecd3aab).

---

The build fails because opf-fido has not been published yet.
See https://github.com/openpreserve/fido/pull/115 - needs CR @jrwdunham.

### DIFF
--- a/src/MCPClient/requirements/base.txt
+++ b/src/MCPClient/requirements/base.txt
@@ -10,4 +10,4 @@ lxml==3.5.0
 metsrw==0.2.0
 requests==2.18.4
 unidecode==0.04.19
-opf-fido==1.3.6
+opf-fido==1.3.7
--- a/src/archivematicaCommon/lib/externals/fiwalk_plugins/pronom_ident.py
+++ b/src/archivematicaCommon/lib/externals/fiwalk_plugins/pronom_ident.py
@@ -30,7 +30,7 @@ class FiwalkFido(fido.Fido):
             f = open(filename, 'rb')
             size = os.stat(filename)[6]
             self.current_filesize = size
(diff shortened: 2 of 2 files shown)


jrwdunham (approved):

jhsimpson (approved):
openpreserve/fdo#115 is merged now and 1.3.7 release of fido is tagged, as soon as the pypi package is available, travis can be rerun here and it should pass.

other than that, :+1:

PR #769, opened 2017-10-24
Add LDAP deps to Dockerfile

sevein:
#766 added the build deps of LDAP to dashboard.Dockerfile. This PR adds the same deps to MCPServer.Dockerfile and MCPClient.Dockerfile. This is necessary because the latter depend on `src/dashboard/requirements/base.txt` which is unfortunate and it'll be eventually fixed but it's beyond the scope of this release.


jhsimpson (approved):
looks good to me.  I tested this locally and it resolves the issue, I had just ran into it before @sevein pointed out this PR to me.

PR #772, opened 2017-10-27
MCPClient: use both clamscan and clamdscan

### LINKED ISSUE #753: Problem: antivirus scanning errors if file is too big
Some defaults in [clamd.conf](https://linux.die.net/man/5/clamd.conf) are causing the antivirus scanning client script to fail. I believe these are the attributes involved:

```
■ MaxScanSize SIZE
Sets the maximum amount of data to be scanned for each input file. Archives and other containers are recursively extracted and scanned up to this value. Warning: disabling this limit or setting it too high may result in severe damage to the system.
Default: 100M

■ MaxFileSize SIZE
Files larger than this limit won't be scanned. Affects the input file itself as well as files contained inside it (when the input file is an archive, a document or some other kind of container). Warning: disabling this limit or setting it too high may result in severe damage to the system.
Default: 25M

■ StreamMaxLength SIZE
Clamd uses FTP-like protocol to receive data from remote clients. If you are using clamav-milter to balance load between remote clamd daemons on firewall servers you may need to tune the Stream* options. This option allows you to specify the upper limit for data size that will be transfered to remote daemon when scanning a single file. It should match your MTA's limit for a maximum attachment size.
Default: 10M
```

I tried to change their values to:
```
StreamMaxLength 0
MaxScanSize 0
MaxFileSize 0
```

Zero means unlimited. However the value of `StreamMaxLength` seems to be hard-coded to 4G (read [this](http://lists.clamav.net/pipermail/clamav-win32/2012-May/000807.html)).

With the previous config, I've made the following tests:

image

The largest one (4.1G) failed.

The output of the client script is:

```
archivematicaClamscan.py: ERROR     2017-10-11 22:05:26,861  archivematica.mcp.client.clamscan:main:95:  Unexpected error scanning: /var/archivematica/sharedDirectory/currentlyProcessing/Test-4.1G-File-With-Accents_2-31a95a67-a0e0-4546-815d-fdbeebd41da6/objects/Volcán.jpg
  File "/usr/lib/archivematica/MCPClient/clientScripts/archivematicaClamscan.py", line 93, in main
    result = client.instream(open(target))
  File "/usr/share/python/archivematica-mcp-client/local/lib/python2.7/site-packages/clamd/__init__.py", line 190, in instream
    self.clamd_socket.send(size + chunk)
error: [Errno 104] Connection reset by peer
archivematicaClamscan.py: INFO      2017-10-11 22:05:26,867  archivematica.mcp.client.clamscan:record_event:56:  Recording event for fileUUID=29ffd407-35c6-41bf-ae5a-30dee0589962 outcome=Fail
```

In /var/log/clamav/clamav.log:

```
WARNING: INSTREAM: Size limit reached, (requested: 1024, max: 1023)
```

sevein:
Added new `CLAMAV_CLIENT_THRESHOLD` setting to MCPClient in order to let the
user define a threshold in MB so when the size of a file being scanned for
viruses is too big the scanner can fall back to clamscan because clamd can't
handle big files not even when the file is passed by reference.

The default value: `CLAMAV_CLIENT_THRESHOLD=20` (1024^2 bytes).

This closes #753 and also https://github.com/artefactual-labs/ansible-archivematica-src/issues/140.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
@@ -2,7 +2,7 @@
 
 # This file is part of Archivematica.
 #
-# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
+# Copyright 2010-2017 Artefactual Systems Inc. <http://artefactual.com>
 #
 # Archivematica is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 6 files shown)


sevein:
@jrwdunham, this is ready for CR!

sevein:
The transfer in the following screenshot included three small files that didn't exceed the threshold so they used clamd (one of them had a virus). The fourth one is a 30M file that was scanned with clamscan instead.

127 0 0 1-62080-tasks-71c5e705-c80b-4f61-9861-d50bf58291bf-

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
 
 logger = get_script_logger("archivematica.mcp.client.clamscan")
 
 
-def get_client(addr):
-    if ':' in addr:
-        host, port = addr.split(':')
-        return ClamdNetworkSocket(host=host, port=int(port), timeout=mcpclient_settings.CLAMAV_CLIENT_TIMEOUT)
-    return ClamdUnixSocket(path=addr)
+def clamav_version_parts(str):
```
jrwdunham: Probably not a good idea to redefine the builtin `str`.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
+
+    @abc.abstractmethod
+    def scan(self, path):
+        """Scan a file and return a tuple of three elements reporting the
+        results. These are the three elements expected:
+            1. passed (bool)
+            2. state (str - "OK", "ERROR", or "FOUND")
+            3. details (str - extra info when ERROR or FOUND)
+        """
+        return
```
jrwdunham: Technically, I don't think you need a return statement in an abstract method if you have a docstring.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
+
```
jrwdunham: It seems redundant that all of these abstract properties have the same implementation in the subclasses. Why not just define them as concrete methods on `ScannerBase` and have `_version_attrs` be an abstract method?

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
+        return subprocess.check_output((self.COMMAND,) + args)
+
+    def _version_attrs(self):
+        try:
+            return clamav_version_parts(self._call('-V'))
+        except subprocess.CalledProcessError:
+            return None, None
+
+    def scan(self, path):
+        passed, state, details = (True, 'OK', None)
```
jrwdunham: Is it dangerous to have `True` as the default passed value here, given the underlying issue that this PR addresses? If it's possible for your `_call` to raise some other exception, then maybe `passed=True` should be set in an `else:` block after the `except:`.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
+        'file_uuid',
+        metavar='fileUUID')
+    parser.add_argument(
+        'path',
+        metavar='PATH',
+        help='File or directory location')
+    parser.add_argument(
+        'date',
+        metavar='DATE',
+        help='E.g.: ')
```
jrwdunham: Should this help string be fleshed out or just removed?

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -23,95 +23,300 @@
...
+        try:
+            self._version
+        except AttributeError:
+            self._version = self.client.version()
+        return clamav_version_parts(self._version)
+
+    def get_client(self):
+        if ':' not in self.addr:
+            return ClamdUnixSocket(path=self.addr)
+        else:
```
jrwdunham: Nitpick: don't need the `else:` here.

src/MCPClient/lib/settings/common.py
```diff
@@ -74,6 +75,7 @@
 storage_service_client_timeout = 86400
 agentarchives_client_timeout = 300
 clamav_client_timeout = 86400
+clamav_client_threshold = 20
```
jrwdunham: Maybe add a comment here so that the user knows the unit is MB.

src/MCPClient/tests/test_clamscan.py
```diff
@@ -0,0 +1,344 @@
...
+from clamd import ClamdNetworkSocket, ClamdUnixSocket
+import pytest
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.abspath(os.path.join(THIS_DIR, '../lib/clientScripts')))
+
+import archivematicaClamscan
+
+
+@pytest.mark.parametrize("version, want", [
```
jrwdunham, line 20: This is awesome.

src/MCPClient/tests/test_clamscan.py
```diff
@@ -0,0 +1,344 @@
...
+        return_value=("ClamAV 0.99.2", "23992/Fri Oct 27 05:04:12 2017"))
+
+    assert scanner.program == "ClamAV (clamd)"
+    assert scanner.version == "ClamAV 0.99.2"
+    assert scanner.virus_definitions == "23992/Fri Oct 27 05:04:12 2017"
+
+
+def test_clamdscanner_get_client(settings):
+    scanner = setup_clamdscanner(settings, addr="/var/run/clamav/clamd.ctl")
+    assert type(scanner.client) == ClamdUnixSocket
```
jrwdunham: Nitpick: more pythonic to use `isinstance`

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
jrwdunham: Also, details is always `None` with ClamScanner? Is there nothing useful in `err`?

src/MCPClient/tests/test_clamscan.py
```diff
@@ -0,0 +1,344 @@
...
+
+    assert exit_code == 0
+    re.assert_called_once_with(
+        args['file_uuid'],
+        args['date'],
+        sc,
+        True)
+
+
+def test_scan_file_not_passed(mocker):
```
jrwdunham: I think this func is identical to `test_scan_file_passed` above.

src/MCPClient/tests/test_clamscan.py
```diff
@@ -0,0 +1,344 @@
+# -*- coding: utf8 -*-
```
jrwdunham, line 1: These test are blowing my mind man! Very thorough, very cool—learned lots.

src/MCPClient/tests/test_clamscan.py
jrwdunham: Is there some way to be more DRY in a similar way to this when it comes to the tests that use `setup_test_scan_file_mocks` below? Just wondering.

jrwdunham (commented):
This is great. I have some minor comments and questions.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
sevein: Right! 💡

sevein:
Thanks @jrwdunham!
I've addressed your comments in 24987a7 and f5287367.

src/MCPClient/tests/test_clamscan.py
sevein: I gave it a shot, take a look please.

jrwdunham:
👍 This can be merged imo. Could use a couple thousand more namedtuples, but hey.

src/MCPClient/tests/test_clamscan.py
jrwdunham: Very nice. Very clever. I'm worried it may be a little hard to understand and I want it to be easy to understand because I want other AM contributors to write tests like this.

I know I'm being opinionated here, but I think your `record_event_fn_params` would be a perfect candidate for a namedtuple, as would the return value of `setup_test_scan_file_mocks`. This would allow you to be terse yet explicit without adding more comments.

The above might make it easier to include the `test_scan_file_already_scanned` test into the parametrize-decorated test (i.e., by passing `'file_already_scanned_patch'` or `'record_event_patch'` to `test_scan_file` and telling it what to assert against.

sevein: You totally read my mind. I considered it while I was coding but then I was lazy I guess. See dce412c3!

sevein:
Thanks for all the feedback!

PR #773, opened 2017-10-27
Add ADR 1: unification of METS creation in AM

jrwdunham:
Added a first Architectual Decision Record (ADR) entitled
"Unification of METS Creation in Archivematica". Currently status of
"proposed" should be changed to "accepted" once code reviewers /
stakeholders have agreed to it (perhaps after modification) or it should
be rejected. It can always be accepted and then later overturned or
"superseded" by a subsequent ADR.

### DIFF
--- a/docs/arch/adr-001.md
+++ b/docs/arch/adr-001.md
@@ -1,0 +1,59 @@
+# Architectural Decision Record (ADR) 1: Unification of METS Creation in Archivematica
+
+## Context
+
+### Technological forces
+
+- Mets-reader-writer, i.e., [metsrw](https://github.com/artefactual-labs/mets-reader-writer) already exists, is being used in Archivematica, and seems like a good tool for abstracting METS file creation.
+
+- Currently, [Archivematica](https://github.com/artefactual/archivematica) (AM) and the [Archivematica Storage Service](https://github.com/artefactual/archivematica-storage-service) (SS) both create or modify METS files (pointer files are METS files too), using mets-rw or various lxml APIs. This inconsistency makes it difficult to reliably create predictable METS files.
+
+- Artefactual has created a [METS validator](https://github.com/artefactual/mets-validator) project, which contains a Schematron file that validates AM-generated AIP METS files.
+
+- Metsrw (v. 0.1.1) cannot create pointer files and it does not support METS file validation, either via XMLSchema or Schematron.
+
+- Metsrw (v. 0.2.0) does have support for pointer file creation and for METS file validation generally, both via XMLSchema and Schematron.
+
+- Changing how we write METS files may inadvertently affect the METS files produced; METS validation and related testing will need to prevent this, or formally recognize it if it is intentional.
+
+- At present in AM and SS, METS/PREMIS data structures are encoded at various points as XML strings, MySQL/SQLite database rows, and Python objects (i.e., metsrw or lxml instances.
+
+### Project forces
+
+- In the context of the project "AIP Encryption via Mirror Locations", mirror locations in the Storage Service must replicate stored AIPs and give them their own pointer files; if replica pointer files are to be generated in the SS then it makes sense to generate all pointer files there.
+
+- An [analysis of pointer file creation in Archivematica](https://docs.google.com/document/d/1iyEz47TN0zmhPiOi8QVDfy0HxJPK-Ar8-8S9IR0lRC4/) has been performed.
+
+### Political forces
+
+- Other projects under discussion or on the horizon will involve the need to recognize various "flavours" or versions of AM-generated and externally-generated (e.g., DSpace, from-LOCKSS) METS files and process them accordingly. One such project relates to re-ingesting old AM-generated AIPs whose METS files may be significantly different from their current counterparts. This implies more sophisticated METS file validation and recognition capabilities than AM currently implements.
+
+- The Mirror Locations project’s budget does not cover wholesale refactoring of how AM creates METS files, or even significant alterations to metsrw.
+
+
+## Decision
+
+We will use METS-reader-writer (metsrw) to create and modify METS files---including pointer files---in all future code. We will not rewrite existing METS creation code to use metsrw unless a clear opportunity arises to do so, e.g., direct funding for this purpose or a project that requires non-trivial changes to such code.
+
+In the AIP Mirror project, metsrw will be used to create METS pointer files, both when AIPs are stored and when they are replicated. This will involve removing the Create Pointer File micro-service in AM and calling a `create_pointer_file` method of Package at the end of AIP storage, reingest and replication.
+
+
+## Status
+
+Proposed.
+
+## Consequences
+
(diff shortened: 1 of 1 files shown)


jrwdunham:
For context on what an ADR is, see http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions.

evelynPM:
"AIP compression is still performed in AM (and stored in AM’s db) but now pointer file creation will be performed by metsrw in the SS. This means that either AM will somehow have to communicate compression event facts to SS or AIP compression should be moved to the SS." If AIP compression were moved to SS how would that affect a setup where multiple AM processing pipelines were using the same SS? Would users be able to make different compression choices for the different pipelines?

jrwdunham:
@evelynPM I would guess that compression configuration decisions could be passed from AM to SS similarly to how AM's compression events (and agents) are currently passed to SS.

In the AIP encryption work, I ended up taking the route where AM still compresses but it sends PREMIS events and agents to SS so that SS can correctly construct the pointer file.

docs/arch/adr-001.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Artefactual has created a [METS validator](https://github.com/artefactual/mets-validator) project, which contains a Schematron file that validates AM-generated AIP METS files.
+
+- Metsrw (v. 0.1.1) cannot create pointer files and it does not support METS file validation, either via XMLSchema or Schematron.
+
+- Metsrw (v. 0.2.0) does have support for pointer file creation and for METS file validation generally, both via XMLSchema and Schematron.
+
+- Changing how we write METS files may inadvertently affect the METS files produced; METS validation and related testing will need to prevent this, or formally recognize it if it is intentional.
+
+- At present in AM and SS, METS/PREMIS data structures are encoded at various points as XML strings, MySQL/SQLite database rows, and Python objects (i.e., metsrw or lxml instances.
```
sevein: Nitpick: missing closing parenthesis.

docs/arch/adr-001.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Project forces
+
+- In the context of the project "AIP Encryption via Mirror Locations", mirror locations in the Storage Service must replicate stored AIPs and give them their own pointer files; if replica pointer files are to be generated in the SS then it makes sense to generate all pointer files there.
+
+- An [analysis of pointer file creation in Archivematica](https://docs.google.com/document/d/1iyEz47TN0zmhPiOi8QVDfy0HxJPK-Ar8-8S9IR0lRC4/) has been performed.
```
sevein, line 25: This link doesn't seem to be public.

sevein (commented):
Nitpick: in this document I've seen metsrw been named in too many forms:

- Mets-reader-writer
- METS-reader-writer
- mets-reader-writer
- Metsrw
- metsrw

Should this be fixed?

sevein (commented):
Should we choose a different repository to host these proposals? I've seen a couple of projects doing this:

- https://github.com/yarnpkg/rfcs
- https://github.com/golang/proposal

We could have our own github.com/archivematica/adrs.

docs/arch/adr-001.md
jrwdunham: should be public now.

docs/arch/adr-001.md
jrwdunham: fixed.

jrwdunham:
I have tried to make the naming of metsrw more consistent in the document. I am now using "METS Reader & Writer" as the canonical long form name and "metsrw" everywhere else. The forms "Mets-reader-writer", "METS-reader-writer", and "Metsrw" should now no longer be present. The "mets-reader-writer" string is part of the github URL for the project, so it must remain.

sevein:
Cool, thanks @jrwdunham. Are we still considering other venus for this kind of documentation? I think Justin was thinking on giving some use to the github.com/archivematica organization. What about a github.com/archivematica/adrs repository? It reminds me to https://github.com/golang/proposal.

joel-simpson:
This looks great @jrwdunham 
I worked for a while at Lloyds bank where ADRs were standard practice. We had to create them as part of the design for every project. I have a couple of minor comments / suggestions:

Could we include a section on "alternatives considered"? You've described some of the alternatives in the tech forces section, but they don't pop out. One of the key use cases at Lloyds was looking at old ADRs to find out "why did anyone ever decide that we should use x (insert any old, crappy, non-trendy tech)".  A big part of the ADRs I've seen are really just describing options considered with pros and cons (and we've done some of the same approaches in Jisc a few times). 

naming: Perhaps you could name this ADR "Standard component for writing METS".  You can imagine over time you’ll get many ADRs that start with “Standard component for…x" which should make the list of ADRs pretty easy to scan quickly. Like the name you used, this approach also doesn't really include the decision in the ADR title, which is helpful because often you want to set up an ADR to collaborate and get opinions well in advance of actually making a decision.

I like @sevein suggestion of having a separate repo for these. But in the meantime I don't think that should stop us from creating more of these using the structure you've started here. we should probably have more than one ADR written before we worry about setting up a new repo.

docs/arch/adr-001.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Artefactual has created a [METS validator](https://github.com/artefactual/mets-validator) project, which contains a Schematron file that validates AM-generated AIP METS files.
+
+- metsrw (v. 0.1.1) cannot create pointer files and it does not support METS file validation, either via XMLSchema or Schematron.
+
+- metsrw (v. 0.2.0) does have support for pointer file creation and for METS file validation generally, both via XMLSchema and Schematron.
+
+- Changing how we write METS files may inadvertently affect the METS files produced; METS validation and related testing will need to prevent this, or formally recognize it if it is intentional.
+
+- At present in AM and SS, METS/PREMIS data structures are encoded at various points as XML strings, MySQL/SQLite database rows, and Python objects (i.e., metsrw or lxml instances).
```
joel-simpson, line 19: I wasn't clear from the decision whether this problem is solved going forward? If I'm a developer in future do I just need to know that as long as I use metsrw I won't be making this problem worse? (I'm assuming this inconsistency is a problem)

ablwr:
I agree with @sevein's proposal. The Rust language chooses to do this as well for RFCs for new proposed changes to the language: http://rust-lang.github.io/rfcs/ and they are a great, active, and thoughtful group, with a lot of wisdom into how things haven't worked in the past. I know a language isn't the same as an application, but even still. For us, I think a dedicated repo would encourage the larger community to see and feel comfortable giving feedback, instead of getting lost in a very active sea of code changes. I think even for Artefactual it will give us a dedicated space to have open, public communication and people can see how decisions are made.

sevein:
I'm going to close this pull request because the process to record and discuss design decisions is under development. I've created a temporary venue in our artefactual-labs organization: https://github.com/artefactual-labs/archivematica-rfcs-test.

Issue #774, opened 2017-10-27
Problem: StoreDIP is stuck until refresh screen


kellyannewithane:
When the DIP is stored after the AIP AM dashboard gets stuck at StoreDIP location.  Refreshing the screen the job reports that the DIP is stored successfully. 

Can this be resolved or should we document a workaround for 1.7?

kellyannewithane:
https://projects.artefactual.com/issues/11476#change-56586

jrwdunham:
Fixed by https://github.com/artefactual/archivematica/pull/777

ablwr:
In regression-testing for DIP workflow in am17x.qa and centos17x.qa, this has not been a problem. I think this is resolved.

PR #777, opened 2017-10-30
Dashboard: more nuanced job view update logic

jrwdunham:
When deciding whether to update its view, the `Sip` Backbone model now
does so if some jobs are in progress or all jobs have just recently finished.
Previously, the update occurred only when there were in-progress jobs.
This fixes issue #774 where "Store DIP" was finished but the GUI was
still displaying "Executing command(s)".


sevein (approved):
Great!

PR #778, opened 2017-10-30
Dashboard: delete media block from nginx sample

sevein:
That block was unnecessary since we started using WhiteNoise.


jhsimpson (approved):
this matches a working config file I just tested with.

PR #782, opened 2017-11-03
Backward-compatible logging

### LINKED ISSUE #776: Problem:  can't configure logging location easily in 1.7
the 1.7 release redirects logs from the dashboard, mcp client and mcp server to stdout and stderr.  These get captured at the moment by systemd (on systems with that) or upstart (on ubuntu 14.04).  

In order to maintain backwards compatibility with how logs are stored in Archivematica 1.6.1 and earlier, there needs to be a way to direct logs to /var/log/archivematica, and to decide what logging level goes into what file.

sevein:
AM 1.7 brought a new simpler logging configuration with less responsibilities -
log writing to files, debugging log file or automatic file rotation are gone
now. The new default is to send the events to the standard streams which is
more convenient when you're running Archivematica in a cluster.

In this pull request I add the ability to change the new logging configuration
defaults via a JSON configuration file. The examples included mimic the old
behaviour - this is important for deb/pkg/ansible installations that still
prefer to have the old defaults in place.

- `/etc/archivematica/dashboard.logging.json` (see
[install/README.md](https://github.com/artefactual/archivematica/blob/dev/backward-compatible-logging/src/dashboard/install/README.md))
- `/etc/archivematica/clientConfig.logging.json` (see
[install/README.md](https://github.com/artefactual/archivematica/blob/dev/backward-compatible-logging/src/MCPClient/install/README.md))
- `/etc/archivematica/serverConfig.logging.json` (see
[install/README.md](https://github.com/artefactual/archivematica/blob/dev/backward-compatible-logging/src/MCPServer/install/README.md))

This closes #776.

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -16,6 +16,10 @@
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
 import StringIO
+import json
+import logging
+import logging.config
+import os
 
 from appconfig import Config
 
@@ -129,6 +133,13 @@ INSTALLED_APPS = (
     'fpr',
 )
 
+# Configure logging manually
+LOGGING_CONFIG = None
+
+# Location of the logging configuration file that we're going to pass to
+# `logging.config.fileConfig` unless it doesn't exist.
+LOGGING_CONFIG_FILE = '/etc/archivematica/clientConfig.logging.json'
+
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
@@ -155,6 +166,12 @@ LOGGING = {
         'level': 'WARNING',
     }
 }
+
+if os.path.isfile(LOGGING_CONFIG_FILE):
+    with open(LOGGING_CONFIG_FILE, 'rt') as f:
+        LOGGING = logging.config.dictConfig(json.load(f))
+else:
+    logging.config.dictConfig(LOGGING)
 
 
 SHARED_DIRECTORY = config.get('shared_directory')
--- a/src/MCPServer/install/README.md
+++ b/src/MCPServer/install/README.md
@@ -1,0 +1,31 @@
+## Configuration files
+
+This directory contains the following files:
+
+- [`config.ini`](./config.ini) - a sample configuration file that the user can place in
+`/etc/archivematica/serverConfig.conf` to tweak the configuration of MCPServer.
+
+- [`logging.json`](./logging.json) - read the
+[logging configuration section](#logging-configuration) for more details.
+
+## Logging configuration
+
+Archivematica 1.6.1 and earlier releases are configured by default to log to
+subdirectories in `/var/log/archivematica`, such as
+`/var/log/archivematica/MCPServer/MCPServer.debug.log`. Starting with
+Archivematica 1.7.0, logging configuration defaults to using stdout and stderr
+for all logs. If no changes are made to the new default configuration logs
+will be handled by whichever process is managing Archivematica's services. For
+example on Ubuntu 16.04 or Centos 7, Archivematica's processes are managed by
+systemd. Logs for the MCPServer can be accessed using
+`sudo journalctl -u archivematica-mcp-server`. On Ubuntu 14.04, upstart is used
+instead of systemd, so logs are usually found in `/var/log/upstart`. When
+running Archivematica using docker, `docker-compose logs` commands can be used
+to access the logs from different containers.
+
+The MCPServer will look in `/etc/archivematica` for a file called
+`serverConfig.logging.json`, and if found, this file will override the default
+behaviour described above.
+
+The [`logging.json`](./logging.json) file in this directory provides an example
+that implements the logging behaviour used in Archivematica 1.6.1 and earlier.
--- a/src/dashboard/install/README.md
+++ b/src/dashboard/install/README.md
@@ -1,0 +1,34 @@
+## Configuration files
+
+This directory contains the following files:
+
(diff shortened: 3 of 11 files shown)


jhsimpson:
The example json files being provided by this pull request, they are being put in the source code at:

`src/MCPClient/logging.json
src/MCPServer/logging.json
src/dashboard/install/logging.json`

in stable/1.6.x there is a directory at src/MCPClient/etc/, but that is gone now in qa/1.x.  What do you think about storing the MCPClient's logging.json in src/MCPClient/etc/logging.json ?  or src/MCPCLient/etc/example.logging.json ?

The dashboard one is in src/dashboard/install/.  Maybe all three should be in an install directory?

src/MCPClient/lib/settings/common.py
```diff
@@ -127,6 +130,13 @@
     'fpr',
 )
 
+# Configure logging manually
+LOGGING_CONFIG = None
+
+# Location of the logging configuration file that we're going to pass to
+# `logging.config.fileConfig` unless it doesn't exist.
+LOGGING_CONFIG_FILE = '/etc/archivematica/clientConfig.logging.json'
```
jhsimpson, line 141: given this convention, which Archivematica is following, of looking for config in /etc/archivematica, I suggest adding a readme.md at src/MCPClient/install/readme.md that describes what to do with the contents of install (and same thing in src/dashboard/install/.

Does that make sense?  Those readme's would become the documentation for how to set the logging config

src/dashboard/install/README.md
```diff
@@ -0,0 +1,3 @@
+- `logging.json` is the configuration file that used to have Archivematica 1.6
```
jhsimpson: I suggest the following text:

Archivematica 1.6.1 and earlier releases are configured by default to log to subdirectories in /var/log/archivematica, such as `/var/log/archivematica/dashboard/dashboard.debug.log`.  
Starting with Archivematica 1.7.0, logging configuration defaults to using stdout and stderr for all logs.  If no changes are made to the new default configuration logs will be handled by whichever process is managing Archivematica's services.  For example on Ubuntu 16.04 or Centos 7, Archivematica's processes are managed by systemd.  Logs for the MCP Server can be accessed using `sudo journalctl -u archivematica-dashboard`.  On Ubuntu 14.04, upstart is used instead of systemd, so logs are usually found in /var/log/upstart.  When running Archivematica using docker, `docker-compose logs` commands can be used to access the logs from different containers.

The dashboard will look in /etc/archivematica for a file called dashboard.logging.json, and if found, this file will override the default behaviour described above.  

The `logging.json` file in this directory provides an example that implements the logging behaviour used in Archivematica 1.6.1 and earlier.

sevein:
Good ideas - I've added docs and moved to `install/`.

src/MCPServer/install/README.md
```diff
@@ -0,0 +1,31 @@
...
+`/etc/archivematica/serverConfig.conf` to tweak the configuration of MCPServer.
+
+- [`logging.json`](./logging.json) - read the
+[logging configuration section](#logging-configuration) for more details.
+
+## Logging configuration
+
+Archivematica 1.6.1 and earlier releases are configured by default to log to
+subdirectories in `/var/log/archivematica`, such as
+`/var/log/archivematica/dashboard/dashboard.debug.log`. Starting with
```
jhsimpson: nitpick change reference here to mcp-server, instead of dashboard

sevein:
I haven't tested this in a virtual machine but I did test it in Docker. Once I had my environment running I edited the MCPServer container:

```shell
 [host] $ docker-compose exec --user=root archivematica-mcp-server bash

[guest] $ apt && apt install nano
[guest] $ mkdir -p /etc/archivematica
[guest] $ nano /etc/archivematica/serverConfig.logging.conf
        # <<< PASTE YOUR CONFIG; SAVE; EXIT >>>
[guest] $ exit

 [host] $ docker-compose restart archivematica-mcp-server
 [host] $ docker-compose logs -f archivematica-mcp-server
```

Restarting the container doesn't destroy it meaning that the changes we made to the filesystem were persisted and survived to the restart.

I pasted the following config:

```json
{
  "loggers": {
    "archivematica": {
      "level": "DEBUG"
    }
  },
  "disable_existing_loggers": false,
  "handlers": {
    "console": {
      "formatter": "detailed",
      "class": "logging.StreamHandler",
      "level": "DEBUG"
    }
  },
  "formatters": {
    "detailed": {
      "datefmt": "%Y-%m-%d %H:%M:%S",
      "format": "---> %(levelname)-8s  %(asctime)s  %(name)s:%(module)s:%(funcName)s:%(lineno)d:  %(message)s"
    }
  },
  "version": 1,
  "root": {
    "level": "DEBUG",
    "handlers": [
      "console"
    ]
  }
}
```

The config above is similar to the default one shipped by MCPServer in Archivematica 1.7. The only difference is that it sets the level of the root logger to `DEBUG` and changes the formater to print an arrow `--->` for each event.

Run the following to start MCPServer in a new fresh container based on the original image defined in the `MCPServer.Dockerfile`:

    $ docker-compose up -d --force-recreate --no-deps archivematica-mcp-server

jhsimpson (approved):
I was able to reproduce seveins results using a docker install. I wasn't able to test in an ansible deployment, but that is a problem outside the scope of this PR. 

This is good to merge I think.

sevein:
@scollazo @hakamine @mamedin @DavidHume-ArtefactualSystems: I think this is ready for you guys to integrate this as part of https://github.com/artefactual-labs/am-packbuild/issues/63 and https://github.com/artefactual-labs/ansible-archivematica-src/issues/147.

PR #795, opened 2017-11-13
Fixed DIP upload UI JS issue, refs #11213

mcantelon:
Fixed issue preventing DIP upload to AtoM from working.

### DIFF
--- a/src/dashboard/src/media/js/ingest.js
+++ b/src/dashboard/src/media/js/ingest.js
@@ -364,7 +364,7 @@ $(function()
                       $.ajax({ type: 'POST', url: url, data: { target: target } })
                     ).done(function(resp1, resp2) {
 
-                      if (resp2.ready) {
+                      if (resp2[0].ready) {
                         executeCommand(self);
                       }
 


sevein (commented):
@mcantelon, what was the problem? Has the API changed?

sevein (changes requested):
@mcantelon could you please add some context? I'd like to understand the reason why the code stopped working if it's possible. Has the API of jQuery changed?

sevein:
@mcantelon before merging is it possible to make a comment on the solution please?

mcantelon:
Hi Jesus... sorry, I forgot about this PR. I'll prepare a summary of the issue.

sevein:
Thank you, no problem! You must have the notifications disabled or something like that.

mcantelon:
At the bottom of this jQuery page there's an example similar to the fix in the PR:

https://api.jquery.com/jquery.when/

The "when"-reliant code isn't present in our stable/1.6.x branch so the bug might not have been encountered if noone's tested the qa/1.7.x AtoM DIP uploading functionality since the change was added:

https://github.com/artefactual/archivematica/commits/stable/1.6.x/src/dashboard/src/media/js/ingest.js

sevein (approved):
Thanks!

Issue #803, opened 2017-11-21
Problem: when DISABLE_SEARCH_INDEXING is True, Backlog, Appraisal and Archival storage break


jrwdunham:
If an Archivematica instance is deployed with `DISABLE_SEARCH_INDEXING` set to `True` in dashboard/src/settings/base.py and MCPClient/lib/settings/common.py, then the Backlog, Appraisal and Archival storage tabs will display errors as described below.

- Backlog: `ConnectionError` from `elasticSearchFunctions.get_client()` in dashboard/src/components/backlog/views.py
- Appraisal: clicking "Search transfer backlog" results in notification reading "Unable to retrieve transfer data from Archivematica."
- Archival storage: `ConnectionError` from `elasticSearchFunctions.get_client()` in dashboard/src/components/archival_storage/views.py

Instead of returning HTTP responses with 500 status codes or exposing non-functional search interfaces, these tabs should display a warning to the user that they are non-functional and why.

jrwdunham:
This should be fixed by https://github.com/artefactual/archivematica/pull/804

jrwdunham:
Fixed by https://github.com/artefactual/archivematica/pull/804

sromkey:
This was qa'd as part of the 1.7 release, closing.

PR #804, opened 2017-11-21
Dashboard: GUI reflects DISABLE_SEARCH_INDEXING

jrwdunham:
If Archivematica has been deployed with Elasticsearch indexing disabled, then the following changes will result:

1. The Backlog, Appraisal, and Archival storage tabs will now no longer be displayed in the navigation bar. If the user navigates to them using their URLs, the pages will indicate their non-operationality.
2. The SIP Arrange pane of the Ingest tab will no longer be displayed.
3. The disabled status of ES indexing will be indicated under Administration > General.
4. The "Send to backlog" option of "Create SIP(s)" will not be available in the Processing Configuration edit pages.
5. When the "Create SIP(s)" choice is displayed to the user during transfer, it will no longer contain a "Send to backlog" option.

Fixes https://github.com/artefactual/archivematica/issues/803 and implements feature file at https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/50.

Possible TODO: If the user supplies a processingConfig.xml file in a transfer that causes "Create SIP(s)" to default to "Send to backlog", then MCPServer should silently ignore this.

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -38,7 +38,10 @@ CONFIG_MAPPING = {
     'client_modules_file': {'section': 'MCPClient', 'option': 'archivematicaClientModules', 'type': 'string'},
     'elasticsearch_server': {'section': 'MCPClient', 'option': 'elasticsearchServer', 'type': 'string'},
     'elasticsearch_timeout': {'section': 'MCPClient', 'option': 'elasticsearchTimeout', 'type': 'float'},
-    'disable_search_indexing': {'section': 'MCPClient', 'option': 'disableElasticsearchIndexing', 'type': 'boolean'},
+    'search_enabled': [
+        {'section': 'MCPClient', 'option': 'disableElasticsearchIndexing', 'type': 'boolean'},
+        {'section': 'MCPClient', 'option': 'search_enabled', 'type': 'iboolean'},
+    ],
     'removable_files': {'section': 'MCPClient', 'option': 'removableFiles', 'type': 'string'},
     'temp_directory': {'section': 'MCPClient', 'option': 'temp_dir', 'type': 'string'},
     'secret_key': {'section': 'MCPClient', 'option': 'django_secret_key', 'type': 'string'},
@@ -58,6 +61,8 @@ CONFIG_MAPPING = {
     'db_port': {'section': 'client', 'option': 'port', 'type': 'string'},
 }
 
+SEARCH_ENABLED_DEFAULT = False
+
 CONFIG_DEFAULTS = """[MCPClient]
 MCPArchivematicaServer = localhost:4730
 sharedDirectoryMounted = /var/archivematica/sharedDirectory/
@@ -71,7 +76,7 @@ LoadSupportedCommandsSpecial = True
 numberOfTasks = 0
 elasticsearchServer = localhost:9200
 elasticsearchTimeout = 10
-disableElasticsearchIndexing = True
+search_enabled = {search_enabled_default}
 temp_dir = /var/archivematica/sharedDirectory/tmp
 removableFiles = Thumbs.db, Icon, Icon\r, .DS_Store
 clamav_server = /var/run/clamav/clamd.ctl
@@ -88,7 +93,7 @@ host = localhost
 database = MCP
 port = 3306
 engine = django.db.backends.mysql
-"""
+""".format(search_enabled_default=SEARCH_ENABLED_DEFAULT)
 
 config = Config(env_prefix='ARCHIVEMATICA_MCPCLIENT', attrs=CONFIG_MAPPING)
 config.read_defaults(StringIO.StringIO(CONFIG_DEFAULTS))
@@ -184,7 +189,6 @@ LOAD_SUPPORTED_COMMANDS_SPECIAL = config.get('load_supported_commands_special')
 GEARMAN_SERVER = config.get('gearman_server')
 NUMBER_OF_TASKS = config.get('number_of_tasks')
 CLIENT_MODULES_FILE = config.get('client_modules_file')
-DISABLE_SEARCH_INDEXING = config.get('disable_search_indexing')
 REMOVABLE_FILES = config.get('removable_files')
 TEMP_DIRECTORY = config.get('temp_directory')
 ELASTICSEARCH_SERVER = config.get('elasticsearch_server')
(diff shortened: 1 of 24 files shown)


sevein:
@jrwdunham we can't merge this into qa/1.x because AM17 hasn't been released yet. Should't you be targetting dev/1.8.x instead?

src/MCPServer/lib/linkTaskManagerChoice.py
```diff
@@ -57,6 +57,9 @@ def __init__(self, jobChainLink, pk, unit):
 
         choices = MicroServiceChainChoice.objects.filter(choiceavailableatlink_id=str(jobChainLink.pk))
         for choice in choices:
+            if (django_settings.DISABLE_SEARCH_INDEXING and
+                    choice.chainavailable.description == 'Send to backlog'):
```
sevein: Could we instead have a number of abilities and hide these items based on that? E.g. pipeline `X` has the `search` ability disabled, choice `Y` depends on `search` ability so it's ignored.

src/dashboard/src/templates/appraisal/appraisal.html
```diff
@@ -20,9 +20,14 @@
...
-    <ng-include src="'front_page/content.html'"></ng-include>
-</div>
-{% endverbatim %}
+  {% if disable_search_indexing %}
+    <h1>Elasticsearch Indexing Disabled</h1>
+    <p class="es-indexing-disabled">Sorry, Elasticsearch indexing has been disabled in this Archivematica installation. The appraisal tab is non-functional when indexing is turned off.</p>
+  {% else %}
+    {% verbatim %}
+    <div ng-app="appraisalTab">
+          <ng-include src="'front_page/content.html'"></ng-include>
```
sevein: Fix indentation.

src/dashboard/src/templates/archival_storage/list.html
```diff
@@ -25,103 +25,110 @@
...
 
-  <ul class="breadcrumb">
-    {% trans "Archival storage" as archival_storage_label %}
-    {% trans "Search" as search_label %}
-    {% breadcrumb_url archival_storage_label 'components.archival_storage.views.overview' %}
-    {% breadcrumb search_label %}
-  </ul>
+  {% if disable_search_indexing %}
+    <h1>Elasticsearch Indexing Disabled</h1>
+    <p class="es-indexing-disabled">Sorry, Elasticsearch indexing has been disabled in this Archivematica installation. The archival storage tab is non-functional when indexing is turned off.</p>
```
sevein: Fix i18n.

sevein (changes requested):

src/MCPServer/lib/linkTaskManagerChoice.py
jrwdunham: Yes, I think that's a good idea. I've tried to implement what you've described in the most recent commit. Let me know what you think.

src/dashboard/src/templates/appraisal/appraisal.html
jrwdunham: fixed

src/dashboard/src/templates/archival_storage/list.html
jrwdunham: Fixed here and also in all newly added template strings.

sevein:
@jrwdunham, I've rebased `dev/1.8.x` because it was behind qa/1.x missing some important fixes.

I've also created `stable/1.7.x` so you can start submitting pull requests against `qa/1.x` for AM18 stuff. Perhaps you can repurpose this PR to include what's in `dev/1.8.x` plus changes in `dev/issue-11442-indexless-GUI`. We should delete `dev/1.8.x`, `dev/1.8.x-indexless-GUI` and `dev/1.8.x-without-logs` I assume?

jrwdunham:
@sevein I've rebased this against the new dev/1.8.x. Can you CR again?

The dev/1.8.x-indexless-GUI and dev/1.8.x-without-logs branches are gone.

sevein (changes requested):
Two observations:

- I think that `Add common abilities with choice dependencies` should be updated
so it doesn't include the setting rename because it's unrelated.

- Regarding to `Dashboard: GUI reflects DISABLE_SEARCH_INDEXING`, I think as a
developer my only setting interface should be `django.conf.settings`. I don't
want to remember what `get_search_enabled` is, I want to simply look up
`django_settings.SEARCH_ENABLED`. I think your motivation is good because the
existing name sucks but I'm not confortable solving that problem this way
because it's not cohesive. IMO it would be a win win (for users and for us) to
implement it in `appconfig.Config` and add some tests - and if you don't have
time for that then I would just ignore this for now. Can we have something like
this in `CONFIG_MAPPING`?

```
'search_enabled': [
    {'section': 'MCPClient', 'option': 'disableElasticsearchIndexing', 'type': 'boolean'},
    {'section': 'MCPClient', 'option': 'search_enabled', 'type': 'iboolean'},
}
```

Two changes would be required:

- `appconfig.Config.get` would need to handle lists.
- `env_configparser.EnvConfigParser` would need to have a new `iboolean` method
added (you could subclass `EnvConfigParser` too).

src/MCPClient/lib/settings/common.py
```diff
@@ -58,6 +61,8 @@
     'db_port': {'section': 'client', 'option': 'port', 'type': 'string'},
 }
 
+SEARCH_ENABLED_DEFAULT = False
```
sevein: Do we still need this?

src/archivematicaCommon/lib/env_configparser.py
```diff
@@ -70,3 +70,7 @@ def getfloat(self, *args, **kwargs):
     @fallback_option
     def getboolean(self, *args, **kwargs):
         return ConfigParser.SafeConfigParser.getboolean(self, *args, **kwargs)
+
+    @fallback_option
+    def getiboolean(self, *args, **kwargs):
+        return not self.getboolean(*args, **kwargs)
```
sevein, line 76: `tests/test_env_configparser.py` has already good coverage, should we include a test for this new method?

jrwdunham: I've added new tests for appconfig that test the new "multiple equivalent options" feature.

src/MCPClient/lib/settings/common.py
jrwdunham: No, I guess not

sevein (approved):

PR #805, opened 2017-11-21
Disable logging from mcp client

scollazo:
Based on changes made for jisc project


scollazo:
While working on the changes proposed by joel in https://github.com/JiscRDSS/archivematica/pull/42#issuecomment-342634961 I found some interesting side effects:
 - Mysql rows "stdOut" and "stdErr" are non nullable, I changed that on the stop with:
<pre>
flter table Tasks modify stdErr longtext null;
flter table Tasks modify stdOut longtext null;
</pre>

But it should be part of a migration.

Also, the self.writeOutputs() call in https://github.com/artefactual/archivematica/blob/dev/1.8.x-without-logs/src/MCPServer/lib/taskStandard.py#L99 fails

PR #813, opened 2017-11-28
Set up email settings via config

sevein:
With this PR we make the email settings highly configurable and set up the default backend to `django.core.mail.backends.console.EmailBackend` instead of a local MTA listening on `127.0.0.1:25`.

This is connected to #1128.

### Backward-compatibility

The new default is to use the backend `django.core.mail.backends.console.EmailBackend` (send to the standard streams) because we don't want to make an assumption on the email delivery solution used by the user. Before we used to assume that there was a MTA available at `127.0.0.1:25`.

To maintain backward-compatibility we will have to introduce the old change to our packages and the Ansible role via configuration. CC @scollazo @hakamine. The old behaviour can be reproduced passing the following environment variables to `archivematica-mcp-client`:

```
ARCHIVEMATICA_MCPCLIENT_EMAIL_BACKEND=django.core.mail.backends.smtp.EmailBackend
ARCHIVEMATICA_MCPCLIENT_EMAIL_HOST=127.0.0.1
ARCHIVEMATICA_MCPCLIENT_EMAIL_PORT=25
ARCHIVEMATICA_MCPCLIENT_EMAIL_USE_TLS=False
```

### Configuration

| Environment variable | Default |
| - | - |
| `ARCHIVEMATICA_MCPCLIENT_EMAIL_BACKEND` | `"django.core.mail.backends.console.EmailBackend"` |

##### File backend

(https://docs.djangoproject.com/en/dev/topics/email/#file-backend)

| Environment variable | Default |
| - | - |
| `ARCHIVEMATICA_MCPCLIENT_FILE_PATH` | `""` |

##### SMTP backend

(https://docs.djangoproject.com/en/dev/topics/email/#smtp-backend)

| Environment variable | Default |
| - | - |
| `ARCHIVEMATICA_MCPCLIENT_HOST` | `"smtp.gmail.com"` |
| `ARCHIVEMATICA_MCPCLIENT_HOST_PASSWORD` | `""` |
| `ARCHIVEMATICA_MCPCLIENT_HOST_USER` | `"your_email@example.com"` |
| `ARCHIVEMATICA_MCPCLIENT_PORT` | `"587"` |
| `ARCHIVEMATICA_MCPCLIENT_SSL_CERTFILE` | `""` |
| `ARCHIVEMATICA_MCPCLIENT_SSL_KEYFILE` | `""` |
| `ARCHIVEMATICA_MCPCLIENT_USE_SSL` | `"False"` |
| `ARCHIVEMATICA_MCPCLIENT_USE_TLS` | `"True"` |

##### General settings, not backend-specific

| Environment variable | Default |
| - | - |
| `ARCHIVEMATICA_MCPCLIENT_DEFAULT_FROM_EMAIL` | `"webmaster@example.com"` |
| `ARCHIVEMATICA_MCPCLIENT_SUBJECT_PREFIX` | `"[Archivematica]"` |
| `ARCHIVEMATICA_MCPCLIENT_TIMEOUT` | `"300"` |

---

Same envs are made available in Dashboard and MCPServer although effectively neither of them send emails at all at this moment.

### DIFF
--- a/src/MCPClient/install/README.md
+++ b/src/MCPClient/install/README.md
@@ -1,4 +1,4 @@
-# MCPClient Configuration 
+# MCPClient Configuration
 
 ## Table of contents
 
@@ -10,21 +10,21 @@
 
 ## Introduction
 
-Archivematica components can be configured using multipe methods.  All 
+Archivematica components can be configured using multipe methods.  All
 components follow the same pattern:
 
-1. **Environment variables** - setting a configuration parameter with an 
+1. **Environment variables** - setting a configuration parameter with an
    environment variable will override all other methods.
-1. **Configuration file** - if the parameter is not set by an environment 
+1. **Configuration file** - if the parameter is not set by an environment
    variable, the component will look for a setting in its default configuration file.
-1. **Application defaults**  - if the parameter is not set in an environment 
+1. **Application defaults**  - if the parameter is not set in an environment
    variable or the config file, the application default is used.
 
 Logging behaviour is configured differently, and provides two methods:
 
 1. **`logging.json` file** - if a JSON file is present in the default location,
     the contents of the JSON file will control the components logging behaviour.
-1. **Application default** - if no JSON file is present, the default logging 
+1. **Application default** - if no JSON file is present, the default logging
    behaviour is to write to standard streams (standard out and standard error).
 
 MCPClient specific details are provided below.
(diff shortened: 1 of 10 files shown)


jraddaoui (approved):
👍 +1

sevein (commented):
This needs to be updated once #845 is merged.

PR #814, opened 2017-11-29
Dashboard: update fpr-admin to v1.7.4

### LINKED ISSUE #792: Problem: migration fpr.0015_pronom_92 is invalid
The "slug" field in fpr_format table is a varchar(50), and there are slugs with more than 50 chars in this migration.

In centos, mysql drops the characters that exceed the field size silently, but in ubuntu xenial the migration fails to aply.

sevein:
This closes #792.


scollazo (approved):
:+1:

PR #818, opened 2017-11-30
getfilesize return now gets compared with None (antivirus)

### LINKED ISSUE #808: Clamav fails to scan empty files
During testing with the latest code changes from the qa branch, we ran into a problem with the virus scan. When clamav encounters an empty file, it fails with the error message: "Unexpected file size", with a failing transfer as a result. This is new behavior we haven't encountered before and it occurs both in our acceptance environment and a local vagrant installation, so we assume something was changed in the clamav configuration in the commits from the last couple of weeks.

ross-spencer:
This fix ensures that the comparison for a valid file size of zero is not a negation, but will still return 1 as a failure if trying to find file size returns None. 

A fix that helps unblock @sromkey QA testing. 

Fixes #808

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
@@ -259,8 +259,8 @@ def scan_file(file_uuid, path, date, task_uuid):
 
     try:
         size = get_size(file_uuid, path)
-        if not size:
-            logger.error('Unexpected file size')
+        if size is None:
+            logger.error('Getting file size returned: %s', size)
             return 1
 
         scanner = get_scanner(size)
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -259,8 +259,8 @@ def scan_file(file_uuid, path, date, task_uuid):
 
     try:
         size = get_size(file_uuid, path)
-        if not size:
-            logger.error('Unexpected file size')
+        if size is None:
+            logger.error('Getting file size returned: ' + str(size))
```
sevein: I think it's better to do: `logger.error('Getting file size returned: %s', size)`. What do you think?

ross-spencer:
@sevein - I will fix these failing checks. Apologies.

sevein (approved):
Small observation if you have a chance to fix it but 👍. Also the subject line of the commit exceeds the 50chr limit because of that `resolves #...`, you could move it to the body of the message instead.

Thank you!

ross-spencer:
@sevein I think that's okay to re-review.

sevein (approved):
👍

ross-spencer:
Closes #808

PR #820, opened 2017-12-04
Add missing exit code to DIP upload chain link.

mcantelon:
The choose DIP upload config chain link was indicating failure after a choice was selected.


sevein (approved):
Uh I'm sure that wasn't easy! The fix looks good. Something to remember is that because you're editing an existing migration you have to remind people to flush their QA sites and make sure you do because I've seen that becoming a recurrent issue during deployments. `manage.py migrante` won't do it. An alternative is to create a new migration instead but I'm personally ok with this.

PR #824, opened 2017-12-05
Parse the created date from the METS XML

marktriggs:
We can't rely on the mtime of the METS file on disk because it can be a temporary file.  In this case, the mtime will always be "now".

Fixes issue #49.

One other issue we wanted to check: we're parsing the timestamp coming from the METS file as a UTC timestamp.  This seemed to be consistent with what we could see in other parts of the code, but if it's intended to be local time can change it.  Thanks!


sevein:
c24b518b in `stable/1.7.x`. Thanks @marktriggs.

PR #828, opened 2017-12-06
MCPClient: refactor archivematicaClient.py

jrwdunham:
Adds docstrings, more Pythonic formatting, and some cosmetic structural
changes to archivematicaClient.py

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -1,8 +1,34 @@
 #!/usr/bin/env python2
 
+"""Archivematica Client (Gearman Worker)
+
+This executable does the following.
+
+1. Loads tasks from config. Loads a list of performable tasks (client scripts)
+   from a config file (typically that in lib/archivematicaClientModules) and
+   creates a mapping from names of those tasks (e.g., 'normalize_v1.0') to the
+   full paths of their corresponding (Python or bash) scripts (e.g.,
+   '/src/MCPClient/lib/clientScripts/normalize.py').
+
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/archivematicaClient.py
```diff
@@ -1,8 +1,34 @@
 #!/usr/bin/env python2
 
+"""Archivematica (Gearman) Client
```
sevein: To be accurate, it's a Gearman worker.

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -21,166 +47,190 @@
...
+        script, task_uuid, capture_output = _process_gearman_job(
+            gearman_job, gearman_worker)
+    except ProcessGearmanJobError as error:
+        return cPickle.dumps(error)
     except OSError:
         logger.exception('Execution failed')
-        output = ["Archivematica Client Error!", traceback.format_exc()]
-        exitCode = 1
-        return cPickle.dumps({"exitCode": exitCode, "stdOut": output[0], "stdError": output[1]})
+        return cPickle.dumps({'exitCode': 1,
```
sevein, line 190: ^ this `OSError` was used to control the exceptions raised by subprocess (executeOrRun), should be done inside _process_gearman_job instead?

sevein (changes requested):
I have a couple of small observations but looking very good! It should target qa/1.x, this is definitely something we want to include!

jrwdunham:
Replaced by https://github.com/artefactual/archivematica/pull/829

PR #832, opened 2017-12-08
Implements Antivirus (AV) Configuration Options

### LINKED ISSUE #779: Problem: python-clamd fails to detect INSTREAM error from clamd
This is a sequel of #753. When we're using python-clamd to talk to clamd (new in AM17) and we're streaming the file (INSTREAM) and we hit the max filesize limit clamd returns an error and closes the connection - this behaviour is the expected as it's documented in the clamd docs.

However python-clamd raises an Exception under this situation. I think that python-clamd needs to be update to handle it gracefully.

```
In [18]: c.instream(open('/tmp/big-file'))
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-18-26788bd2b854> in <module>()
----> 1 c.instream(open('/tmp/big-file'))

/home/jesus/env-test-py/local/lib/python2.7/site-packages/clamd/__init__.pyc in instream(self, buff)
    188             while chunk:
    189                 size = struct.pack(b'!L', len(chunk))
--> 190                 self.clamd_socket.send(size + chunk)
    191                 chunk = buff.read(max_chunk_size)
    192

error: [Errno 104] Connection reset by peer
```

This is not a priority for AM17 because #772 mitigates the issue by not sending big files to clamd at all and use the local clamscan command-line tool instead.

This was never caught before because before using python-clamd we were using clamdscan (the official clamd cli client) which gracefully managed the situation mentioned before but it was still passing the file as clean even when it was a big file with viruses in it.

### LINKED ISSUE #819: MCPClient options to configure antivirus scanner and thresholds
First-cut of a feature file is described here:

https://github.com/artefactual-labs/archivematica-acceptance-tests/commit/3ad52531e2a4d7b78825114fb1c48a22066e5505

Will require merging into RDSS work, likely superseding https://github.com/JiscRDSS/archivematica/pull/38

Out of scope: Run-time configuration within the AM dashboard/FPR

ross-spencer:
* Separates the responsibilities of AV unit tests
* Provides user options for which AV scanner to use
* Provides user options for max_file_size and max_scan_size
* Ensures PREMIS events are not created for unscanned, and ostensibly unscanned files
* Fixes #779
* Resolves #819


src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -114,6 +115,11 @@ def scan(self, path):
         try:
             result = getattr(self, method_name)(path)
             state, details = result[result_key]
+        except IOError as err:
+            if err.errno == errno.EPIPE:
+                logger.error(
+                    '[Errno 32] Broken pipe. File not scanned. Check Clamd StreamMaxLength')
```
sevein: This is a good idea!

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -258,30 +278,60 @@ def scan_file(file_uuid, path, date, task_uuid):
...
+        max_file_size = (
+            mcpclient_settings.CLAMAV_CLIENT_MAX_FILE_SIZE * 1024 * 1024)
+        max_scan_size = (
+            mcpclient_settings.CLAMAV_CLIENT_MAX_SCAN_SIZE * 1024 * 1024)
+
+        valid_scan = True
+
+        if size > max_file_size:
+            logger.info(
+                "File will not be scanned. Size %s bytes greater than scanner max file size %s bytes", size, max_file_size)
```
sevein: You may have noticed in our flake8 settings that our max length permitted is 160 characters:

https://github.com/artefactual/archivematica/blob/f527d53a3b61164a327038084b007beb2f9240a7/tox.ini#L32

However we're trying hard to keep it down to 79. You can even break strings into multiple lines if you need to, e.g.:

```python
Class Example(object)
        help_text=_l('Other locations that will be used to create replicas of'
                     ' the packages stored in this location'))
```

It's good if you set up your IDE so it highlights lines when they're longer so it's easy to tell.
Blame @jrwdunham, he converted me. Haha!

src/MCPClient/tests/test_clamscan.py
```diff
@@ -171,14 +62,21 @@ def test_clamscanner_version_attrs(mocker, settings):
...
 
 
-def test_clamscanner_scan(mocker):
+def test_clamscanner_scan(mocker, settings):
     scanner = setup_clamscanner()
     mock = mocker.patch.object(
         scanner, '_call',
         return_value='Output of clamscan')
 
+    # thresholds set within AM environment due to AV limitations
```
sevein: Code comments should be sentences, i.e. start it with a capital letter and end it with a period.

src/MCPClient/tests/test_antivirus.py
```diff
@@ -0,0 +1,211 @@
...
+    (
+        {'scanner_passed': True},
+        0,
+        RecordEventParams(scanner_is_None=False, passed=True),
+    ),
+])
+def test_scan_file(mocker, setup_kwargs, exit_code, record_event_params, settings):
+    deps = setup_test_scan_file_mocks(mocker, **setup_kwargs)
+
+    # Thresholds determined by scanner limitations
```
sevein: Code comments should be sentences, i.e. start it with a capital letter and end it with a period.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -226,13 +239,20 @@ def get_parser():
...
+def get_scanner():
+    """ Return the ClamAV client configured by the user and found in the
+    installation's environment variables. Clamdscanner may perform quicker
+    than Clamscanner given a larger number of objects. Return clamdscanner
+    object as a default if no other, or an incorrect value is specified. """
+
+    CLAMAV_CLIENT_BACKEND_CLAMD = 'clamdscanner'
+    CLAMAV_CLIENT_BACKEND_CLAMSCAN = 'clamscanner'
+
+    if str(mcpclient_settings.CLAMAV_CLIENT_BACKEND).lower() == CLAMAV_CLIENT_BACKEND_CLAMSCAN:
```
sevein: You don't need to convert into a string because we know that's already its type as that's how we defined it in the `CONFIG_MAPPING`:

    {'section': 'MCPClient', 'option': 'clamav_client_backend', 'type': 'string'}
                                                                ^^^^^^^^^^^^^^^^

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -226,13 +239,20 @@ def get_parser():
...
         return ClamScanner()
+    elif str(mcpclient_settings.CLAMAV_CLIENT_BACKEND).lower() != CLAMAV_CLIENT_BACKEND_CLAMD:
```
sevein: Same here, it's already a string.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -226,13 +239,20 @@ def get_parser():
...
+        logger.info(
+            "Unexpected value specified for CLAMAV CLIENT, returning Clamsdscanner()")
```
sevein: `CLAMAV CLIENT` should be `CLAMAV_CLIENT_BACKEND` perhaps?

sevein: Similarly, I think you're returning a `ClamsdScanner` not a `Clamsdscanner()`.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -226,13 +239,20 @@ def get_parser():
...
+    than Clamscanner given a larger number of objects. Return clamdscanner
+    object as a default if no other, or an incorrect value is specified. """
+
+    CLAMAV_CLIENT_BACKEND_CLAMD = 'clamdscanner'
+    CLAMAV_CLIENT_BACKEND_CLAMSCAN = 'clamscanner'
+
+    if str(mcpclient_settings.CLAMAV_CLIENT_BACKEND).lower() == CLAMAV_CLIENT_BACKEND_CLAMSCAN:
         return ClamScanner()
+    elif str(mcpclient_settings.CLAMAV_CLIENT_BACKEND).lower() != CLAMAV_CLIENT_BACKEND_CLAMD:
+        logger.info(
```
sevein: Perhaps this should be a warning instead (`logger.warning(...)`) to give it more visibility. I'm assuming the system administrator is going to be only looking at WARNING/ERROR/CRITICAL and ignore INFO/DEBUG but still be interested in discovering that a config attribute has an unknown value.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -226,13 +239,20 @@ def get_parser():
     return parser
 
 
-def get_scanner(size):
-    """ Return the ClamAV scanner that is more appropiate given the size of the
-    file and the threshold setting. The goal is to use ClamScanner instead of
-    ClamdScanner because the latter won't take files that are too big. """
-    threshold = mcpclient_settings.CLAMAV_CLIENT_THRESHOLD * 1024 * 1024  # Convert to bytes
-    if size > threshold:
+def get_scanner():
```
sevein: I wrote a fancier version of `get_scanner`, let me know what you think: https://gist.github.com/sevein/2f99b51c1bffcbc6d5703a3df04b5bf9. Mine feels more pythonic but hey feel free to ignore me - yours works too. It's easy to get picky when you're looking at someone else's code, I hope you understand!

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -258,30 +278,60 @@ def scan_file(file_uuid, path, date, task_uuid):
     scanner, passed = None, False
 
     try:
+
         size = get_size(file_uuid, path)
+
```
sevein: You may find these extra lines clearer and that's fine. However in AM you'll see the habit is to write more condensed code. Not a big deal but you're warned that's what you're going to see! :)

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -154,12 +160,17 @@ def _call(self, *args):
     def scan(self, path):
         passed, state, details = (False, 'ERROR', None)
         try:
-            self._call(path)
+            max_file_size = "--max-filesize=" + \
+                str(int(mcpclient_settings.CLAMAV_CLIENT_MAX_FILE_SIZE)) + "M"
+            max_scan_size = "--max-scansize=" + \
+                str(int(mcpclient_settings.CLAMAV_CLIENT_MAX_SCAN_SIZE)) + "M"
```
sevein: I really like the formatting solutions in Python as `%` or `.format()` (see [pyformat.info](https://pyformat.info/)), you don't have to change this (it's more of a taste preference) but I personally find the following more readable:

```python
max_file_size = '--max-filesize=%dM' % 23.54
```

`.format` is nicer but its support for things like value conversion in Py2 is very limited I believe.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -194,16 +205,18 @@ def record_event(file_uuid, date, scanner, passed):
                 scanner.virus_definitions(),
             )
 
-    outcome = 'Pass' if passed else 'Fail'
-    logger.info('Recording new event for file %s (outcome: %s)', file_uuid, outcome)
+    if passed is not None:
```
sevein: This is more of a style preference but why don't you make this check (`passed is not None`?) at the top of the function where we're also checking `file_uuid`? Better IMO to return early and avoid running the extra code or having nested blocks.

```python
def record_event(file_uuid, date, scanner, passed):
    if passed is None or file_uuid == "None":
        return
    # […] Omitted
```

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -258,30 +278,60 @@ def scan_file(file_uuid, path, date, task_uuid):
...
+            logger.info(
+                "File will not be scanned. Size %s bytes greater than scanner max file size %s bytes", size, max_file_size)
+            valid_scan = False
+        elif size > max_scan_size:
+            logger.info(
+                "File will not be scanned. Size %s bytes greater than scanner max scan size %s bytes", size, max_scan_size)
+            valid_scan = False
+
+        if valid_scan:
+
```
sevein: This blank line would make sense if the `else` was consistent with it. I'd probably delete it, it doesn't particularly help the code to be more readable or it feel like the `if/else` is unbalanced. And this is my biggest nitpick of the year, sorry.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -258,30 +278,60 @@ def scan_file(file_uuid, path, date, task_uuid):
...
+                         passed, state, details)
     finally:
         record_event(file_uuid, date, scanner, passed)
 
-    return 0 if passed else 1
+    if passed is False:
+        return 1
+    # if True or None, then we have no error, the file
+    # can move through the process as expected...
+    return 0
```
sevein: I liked the condensed version, e.g.:

```python
# If True or None, then we have no error, the file can move through the process
# as expected...
exit_code = 1 if passed is False else 0
return exit_code
```

Or...

```python
# If True or None, then we have no error, the file can move through the process
# as expected...
return 1 if passed is False else 0
```

Again it's just a matter of taste, similar to the `outcome = 'Pass' if passed else 'Fail'` in `record_event`.

Notice that the code comment isn't wrapped arbitrarily.

src/MCPClient/tests/test_antivirus.py
```diff
@@ -0,0 +1,211 @@
...
+import pytest
+import test_clamdscan
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(
+    os.path.abspath(os.path.join(THIS_DIR, '../lib/clientScripts')))
+
+import archivematicaClamscan
+
+# Antivirus Tests
```
sevein: This sounds like documentation at the module level, right? If that's the case, you already have a docstring for the module in L2 that you may want to update instead. I don't know if that was me or you who wrote it but worth fixing it either way :)

src/MCPClient/tests/test_antivirus.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def test_get_scanner(settings):
+    """ Test that get_scanner returns the correct instance of antivirus
+    per the user's configuration. Test return of clamdscanner by default. """
+
+    # doubly ensure that clamd env settings are available to the test
```
sevein: Comments should be sentences: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html (from the guy who wrote coverage.py).

src/MCPClient/tests/test_antivirus.py
```diff
@@ -0,0 +1,211 @@
...
+
+args = OrderedDict()
+args['file_uuid'] = 'ec26199f-72a4-4fd8-a94a-29144b02ddd8'
+args['path'] = '/path'
+args['date'] = '2019-12-01'
+args['task_uuid'] = 'c380e94e-7a7b-4ab8-aa72-ec0644cc3f5d'
+
+
+class FileMock():
+
```
sevein, line 68: You added this blank line but flake8 seemed to be happy with it.

src/MCPClient/tests/test_clamdscan.py
```diff
@@ -0,0 +1,120 @@
...
+
+from collections import namedtuple
+from clamd import ClamdNetworkSocket, ClamdUnixSocket
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.abspath(os.path.join(THIS_DIR, '../lib/clientScripts')))
+
+import archivematicaClamscan
+
+# ClamdScanner tests
```
sevein: Same here, I think we should use the module docstring.

sevein (changes requested):
It's looking good, @ross-spencer. I've shared a bunch of suggestions but mostly about cosmetics.

I think it's good that you broke the tests into three files. Do you think maybe `src/MCPClient/tests/test_clamscan.py` should be renamed to `src/MCPClient/tests/test_antivirus_clamscan.py` and same with `test_clamdscan.py`? Just so the reader can easily find the relationship between the three files. We have to recognize that tests also have to be maintained so anything that makes it easier is welcome.

src/MCPClient/tests/test_clamdscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: If we keep it like this then we can bullet proof the code to ensure that if the setting ever arrives here, `None` then an error isn't thrown when we call `lower()` which is bullet proofing in itself ensuring the user can make a mistake.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: If we keep it like this then we can bullet proof the code to ensure that if the setting ever arrives here, None then an error isn't thrown when we call lower() which is bullet proofing in itself ensuring the user can make a mistake.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
sevein: That makes sense.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
sevein: Ok.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Thx!

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved. (I think).

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: I think that what makes the condensed code difficult is that natural language places the subject and the object the other way around, the clause and the result. We're asked here to ask the question the other way around. I am sure it has been discussed ad nauseum in the programming forums, and happy to start using more AM style to be consistent with the code base.

ross-spencer: Resolved.

src/MCPClient/tests/test_antivirus.py
ross-spencer: Resolved.

src/MCPClient/tests/test_antivirus.py
ross-spencer: Resolved.

src/MCPClient/tests/test_antivirus.py
ross-spencer: Resolved.

src/MCPClient/tests/test_clamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Resolved.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
sevein: I thought it was more pythonic but I've looked it up and PEP8 actually discourages compound statements (multiple statements on the same line). Something I'm sure we could enforce at the linter level hehehe. Please do as you prefer.

src/MCPClient/tests/test_antivirus.py
ross-spencer: Flake8 3.2.1. output with either configuration of line:

```
ross-spencer@artefactual:~/git/artefactual-labs/am/src/archivematica/src/MCPClient/tests:$ flake8 *.py
test_antivirus.py:66:1: E302 expected 2 blank lines, found 1
ross-spencer@artefactual:~/git/artefactual-labs/am/src/archivematica/src/MCPClient/tests:$ flake8 *.py
test_antivirus.py:72:1: E302 expected 2 blank lines, found 1
```

sevein: So before and after the class right?

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: You'll see I've borrowed some of your style in the revised version it's neat and I hadn't considered using vars/strings like this before in Python so thanks! 

The try/catch I've avoided. Personal choice is not to deliberately cause an exception (that's more java like - less Golang like - Python in-between?? :) ) where we can deliberately encode our decision making. Hope you like the compromise.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: :+1:

ross-spencer:
@sevein should be good to further code-review. :+1:

sevein (approved):
@ross-spencer this is great, thank you!

Congrats on your first large contribution to Archivematica.

sevein:
e5ac4978 in `qa/1.x`.

Issue #822, opened 2017-12-04
Problem: extra structMap breaks dip upload


jhsimpson:
Archivematica 1.7.0 has a new feature that allows unique identifiers to be assigned to all files and folders in a SIP, even when some of the directories have no contents.

Empty directories are described in a new, 2nd logical structmap (labelled 'normative directory structure') that is added to the METS file.  

The feature is described here:

https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/master/features/core/pid-binding.feature#L17-L29

This new logical structmap breaks dip upload in Atom 2.4 and earlier.  It should be possible to revert to the previous (1.6.x) behaviour and not include the new structmap if it is not required.  Possibly setting a new optional parameter passed to the create METS microservice?

jrwdunham:
I think the solution would be to add a new boolean option like `--create-normative-structmap` to MCPClient/lib/clientScripts/archivematicaCreateMETS2.py with a default value (of `False`?). 

Then we would want this to be user-configurable in some way. The easy way would be to have normative structmap creation turned off by default and then have admin users use MySQL to modify the `arguments` column of the `StandardTasksConfigs` table with execute = `createMETS_v2.0`. 

Better would be to allow runtime configuration via the processing config, but that would likely require a migration and modifications to the workflow...

sromkey:
As we were just discussing in Slack, a user-configurable micro-service named something like "Document empty directories?" would be ideal.

sromkey:
This has been closed by #833 .

currmie:
Kicked off transfers in augustus.qa, am17x.qa, and centos17x.qa.
Successfully sent DIPs from augustus to atom-24rc3-test server, but Rsync quits unexpectedly for transfers that started in am17x.qa and centos17x.qa.

Here's the error message:
Standard error (stderr)

upload-qubit.py: ERROR     2018-02-10 03:05:47,165  archivematica.upload.qubit:log:64:  [uploadDIP] Target: file-1
upload-qubit.py: ERROR     2018-02-10 03:05:47,165  archivematica.upload.qubit:log:64:  [uploadDIP] rsync --protect-args -rltz -P --chmod=ugo=rwX /var/archivematica/sharedDirectory/watchedDirectories/uploadDIP/loris3-2145cef3-9f6b-48a4-9ae9-f884aed6833d 158.69.72.17:/tmp
[...]

PR #833, opened 2017-12-10
Make METS normative structMap creation optional.

mcantelon:
Add a chain choice link that allows the user to select whether
or not they want a normative structMap created when METS is
generated. Normative structMaps in METS break DIP upload
compatibility with AtoM 2.4.0.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
@@ -441,17 +441,14 @@ def update_mets(sip_dir, sip_uuid, keep_normative_structmap=True):
     add_new_files(mets, sip_uuid, sip_dir)
     delete_files(mets, sip_uuid)
 
+    serialized = mets.serialize()
     if not keep_normative_structmap:
         # Remove normative structMap
-        serialized = mets.serialize()
         structmaps = serialized.findall(
             'mets:structMap[@LABEL="Normative Directory Structure"]', namespaces=ns.NSMAP)
         for structmap in structmaps:
             structmap.getparent().remove(structmap)
             print("Removed normative structMap")
-
-    # Delete original METS
-
     return serialized
 
 
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
@@ -426,7 +426,7 @@ def _get_old_mets_rel_path(sip_uuid):
         'METS.' + sip_uuid + '.xml')
 
 
-def update_mets(sip_dir, sip_uuid):
+def update_mets(sip_dir, sip_uuid, keep_normative_structmap=True):
 
     old_mets_path = os.path.join(sip_dir, _get_old_mets_rel_path(sip_uuid))
     print('Looking for old METS at path', old_mets_path)
@@ -441,9 +441,18 @@ def update_mets(sip_dir, sip_uuid):
     add_new_files(mets, sip_uuid, sip_dir)
     delete_files(mets, sip_uuid)
 
+    if not keep_normative_structmap:
+        # Remove normative structMap
+        serialized = mets.serialize()
+        structmaps = serialized.findall(
+            'mets:structMap[@LABEL="Normative Directory Structure"]', namespaces=ns.NSMAP)
+        for structmap in structmaps:
+            structmap.getparent().remove(structmap)
+            print("Removed normative structMap")
+
     # Delete original METS
 
-    return mets.serialize()
+    return serialized
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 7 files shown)


src/dashboard/src/components/administration/forms.py
```diff
@@ -343,6 +343,13 @@ class ProcessingConfigurationForm(forms.Form):
         'yes_option': '1e79e1b6-cf50-49ff-98a3-fa51d73553dc',
         'no_option': 'fcfea449-158c-452c-a8ad-4ae009c4eaba',
     }
+    processing_fields['d0dfa5fc-e3c2-4638-9eda-f96eea1070e0'] = {
+        'type': 'boolean',
+        'name': 'normative_structmap',
+        'label': _('Add normative structMap to METS XML'),
```
jrwdunham: Can you change this so that it is something that will make more sense to an end user? I think it should be `'Document empty directories'` (cf. https://github.com/artefactual/archivematica/issues/822#issuecomment-349145732) because the purpose of the normative logical structMap is to allow AM to document empty directories.

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -112,6 +112,11 @@
       <appliesTo>05357876-a095-4c11-86b5-a7fff01af668</appliesTo>
       <goToChain>fcfea449-158c-452c-a8ad-4ae009c4eaba</goToChain>
     </preconfiguredChoice>
+    <!-- Add normative structMap to METS XML: Yes -->
```
jrwdunham, line 115: Was it agreed that "Yes" should be the default here? I'm not disagreeing with this default, I'm just double-checking.

mcantelon:
Oops. Just noticed this failed checks. I'll fix.

jrwdunham (changes requested):
Overall, this looks good. I have confirmed that it works both with a
pre-configured choice and by making the choice in the GUI.

I have noticed that if you create an AIP that DOES document empty dirs and then
set the config to NOT document empty dirs and then re-ingest the AIP you will
still get the normative structMap in your METS. I don't know if this is an
issue, but I'm just pointing it out here.

Once the user-facing choice text is changed to "Document empty directories" and
the unit tests are made to pass, I think this can be merged.

src/MCPServer/lib/archivematicaMCP.py
mcantelon: I'm not sure if we decided on this. I'll ask on chat.

src/dashboard/src/components/administration/forms.py
mcantelon: Sounds good.

src/MCPServer/lib/archivematicaMCP.py
jhsimpson: default of Yes is good

mcantelon:
@jrwdunham Okay this is ready for a final review, but now it's showing it has conflicts. I think I did the dev against qa/1.x, maybe I should have against stable/1.7.

mcantelon:
@jrwdunham I've added some logic to archivematicaCreateMETSReingest.py to delete any existing normative structmaps before a new one is created. If you could test to see if it fixes the issue you ran into, that'd be awesome (I'm having trouble getting reingest to work on my AM dev VM).

src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
```diff
@@ -441,9 +441,17 @@ def update_mets(sip_dir, sip_uuid):
...
     delete_files(mets, sip_uuid)
 
+    # Remove normative structMap
+    serialized = mets.serialize()
+    structmaps = serialized.findall(
+        'mets:structMap[@LABEL="Normative Directory Structure"]', namespaces=ns.NSMAP)
+    for structmap in structmaps:
+        structmap.getparent().remove(structmap)
+        print("Removed normative structMap")
+
```
jrwdunham: @mcantelon this approach looks better/simpler than modifying metsrw, but doesn't this need to be contingent on `createNormativeStructmap`? I mean, I think the `update_mets` function needs a kwarg like `keep_normative_structmap=True` which you set to `False` when `createNormativeStructmap` is `False` in archivematicaCreateMETS2.py.

jrwdunham (changes requested):
I think the changes to archivematicaCreateMETSReingest.py will work but I think the METS update should only remove normative structmaps if the `createNormativeStructmap` option is `False`.

mcantelon:
@jrwdunham Makes sense... I've updated this branch to make the removal of normative structmaps conditional.

jrwdunham:
I rebased this against the current stable/1.7.x, bumped the migration and made a fix to the "keep normative structmap" logic. My testing has confirmed that you can now create an AIP whose METS file has a normative logical structMap, fully reingest that AIP and choose not to document empty directories (in which case the normative structMap is removed from the revised METS), and then again fully reingest that AIP and choose to again document empty dirs (in which case the normative structMap is restored).

I think this can be merged now.

jrwdunham:
👍

jrwdunham (approved):

jrwdunham:
Cherry-picked to qa/1.x as 642c3fbc29b82401267dcfc5599af0b5cf08e33d, df18fe78426ef52564f7f4b370fa4b39c9713c3f, and 2c47d3736fc5d5bef13c37017010af10d9aa89d5.

PR #836, opened 2017-12-12
MCPClient: stop passing capture_output to executeOrRun

jrwdunham:
Commit f527d53a3b61164a327038084b007beb2f9240a7 added call `executeOrRun('command', script, stdIn='', printing=True, capture_output=capture_output)`, which is problematic because `executeOrRun` does not recognize the `capture_output` kwarg, which is currently only available in dev/1.8.x.


sevein (approved):

sevein:
Weird that this didn't break any test, should we add some?

PR #840, opened 2017-12-13
Fix MCP{Server,Client} config path

sevein:
We have a backward-compatibility strategy to support the old configuration
files but the paths were not correct. This commit fixes that.

It should also be included in `qa/1.x`.

### DIFF
--- a/src/MCPClient/install/README.md
+++ b/src/MCPClient/install/README.md
@@ -3,8 +3,8 @@
 This directory contains the following files:
 
 - [`clientConfig.conf`](./clientConfig.conf) - a sample configuration file that
-the user can place in `/etc/archivematica/clientConfig.conf` to tweak the
-configuration of MCPClient.
+the user can place in `/etc/archivematica/MCPClient/clientConfig.conf` to tweak
+the configuration of MCPClient.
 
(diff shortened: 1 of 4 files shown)


scollazo:
One question, does the clientConfig and serverConfig have priority over the environment vars?

From my point of view, those files are the old way of configuring archivematica, and we support them for environments updated to 1.7, but for fresh installs, the env vars method configuration should be preferred over the legacy one.

One way or other, can you add a note regarding the existence of environment vars to the README?

sevein:
The order of precedence is the following, (with the last listed item winning prioritization):

1. Application defaults
2. Configuration files
3. Environment variables (always win precedence)

I'll add that detail to https://github.com/artefactual/archivematica/pull/845.

sevein:
@scollazo ^ I've updated my comment with a more precise description of the behaviour.

scollazo (changes requested):
Can you add a note regarding the existence of the env vars, and that they take precendence over values in clientConfig and serverConfig?

sevein:
@scollazo I'm doing that in #845. The scope of this PR is solely to fix the path.

scollazo (approved):

sevein:
75838066 in `qa/1.x`

Issue #746, opened 2017-10-09
Problem: Appraisal tab, upload DIP to ArchivesSpace fails


jraddaoui:
On AM17xqa

Transfer 'archivesspacetest'

Failed at Job: Upload to ArchivesSpace.

STDERR:
```
usage: upload-archivesspace.py [-h] [--host host] [--port port] [--user USER]
[--passwd PASSWD] [--dip_location DIP_LOCATION]
[--dip_name DIP_NAME] [--dip_uuid DIP_UUID]
[--restrictions {yes,no,premis}]
[--object_type OBJECT_TYPE]
[--xlink_actuate {none,onLoad,other,onRequest}]
[--xlink_show {embed,new,none,other,replace}]
[--use_statement USE_STATEMENT]
[--uri_prefix URI_PREFIX]
[--access_conditions ACCESS_CONDITIONS]
[--use_conditions USE_CONDITIONS]
[--inherit_notes INHERIT_NOTES] [--version]
upload-archivesspace.py: error: argument --restrictions: invalid choice: '%restrictions%' (choose from 'yes', 'no', 'premis')
```

jhsimpson:
@jraddaoui is this being actively worked on right now?

jraddaoui:
It's my next priority, I'll start with it tomorrow ;)

jraddaoui:
Hi @jhsimpson,

I have fixed a couple of issues in https://github.com/artefactual/archivematica/pull/844 and now the task is executed without errors. However, the following warning appears for each of the files in the DIP in the stderr output:

```
WARNING   2017-12-14 12:02:33  __main__:upload-archivesspace:upload_to_archivesspace:79:  Skipping file /var/archivematica/sharedDirectory/watchedDirectories/uploadDIP/test-0fef53b0-0573-4398-aa4f-ebf04fe711cf/objects/65b7894b-eb17-4efb-b54d-ae06c3ab8fdb-funky_breakbeat_4.mp3 (65b7894b-eb17-4efb-b54d-ae06c3ab8fdb) - no pairing found
```

Which is raised in here: https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPClient/lib/clientScripts/upload-archivesspace.py#L79

I've tested the upload with a DIP created from a normal transfer and with one created from the appraisal-tab and in both cases I got the same result.

No changes are shown in the ASpace instance over the record targeted in the upload.

[...]

PR #844, opened 2017-12-14
Fix ASpace DIP upload

jraddaoui:
The ASpace config dict is used in linkTaskManagerReplacementDicFromChoice
to set the parameters sent to the upload-archivesspace script, which is
expecting restrictions instead of premis. This field doesn't seem to
be used anywhere else, so it only needs to be updated in the related form
fields that are used to update the config dict.

This also fixes an issue where the config port was not being used in the upload.

Fixes https://github.com/artefactual/archivematica/issues/746

### DIFF
--- a/src/MCPClient/lib/clientScripts/upload-archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload-archivesspace.py
@@ -225,6 +225,6 @@ if __name__ == '__main__':
 
     args.inherit_notes = args.inherit_notes.lower() in INHERIT_NOTES_CHOICES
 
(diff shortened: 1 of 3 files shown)


sevein:
@jraddaoui are you still working on this?

jraddaoui:
I've asked for feedback in #746

jhsimpson (approved):
I tested this with @jraddaoui and it looks good.  :+1:

jraddaoui:
fd6947c437ac3bd35a481bc023fddeaee3c3043a in qa/1.x

PR #847, opened 2017-12-15
Fix handle form

### LINKED ISSUE #834: Problem: Verify SSL certificates in Handle server config defaults back to checked
When you have unchecked the "Verify SSL certificates" box on the Handle server config page, it appears to have saved. However, if you then navigate away from the page and come back, it is checked again.

### LINKED ISSUE #835: Problem: Some required fields are unmarked in Handle server config
Bind PIDs will fail if Archive resolve URL template and File resolve URL template are left blank, but they are not marked as required in the user interface.

jrwdunham:
1. Ensures that the `verify_ssl_certs` HTML widget matches the state of the configuration setting in the database. The `pid_request_verify_certs` field/attribute of the Handle config settings is stored as a `DashboardSetting` instance/row. However, it is a boolean and the dashboard settings key/value pairs are untyped. Therefore it is passed to the template as a string `'False'`. This commit is a hack that converts it to boolean `False`. A more thorough fix would add type information to the dashboard settings table (or assume the value column is JSON, which would be parsed to Python). Fixes #834.

2. Marks required Handle fields as required. Fixes #835.

### DIFF
--- a/src/dashboard/src/components/administration/forms.py
+++ b/src/dashboard/src/components/administration/forms.py
@@ -112,7 +112,7 @@ class HandleForm(SettingsForm):
     )
 
     resolve_url_template_archive = forms.CharField(
-        required=False,
+        required=True,
         label=_('Archive resolve URL template'),
         help_text=_('Template (Django or Jinja2) for the URL that a unit\'s'
                     ' PURL should resolve to. Has access to "pid" and'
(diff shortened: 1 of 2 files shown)


jrwdunham:
If approved, this will be cherry-picked to stable/1.7.x.

jrwdunham:
The problem/solution described in https://github.com/artefactual/archivematica/issues/848 would be a more general fix for https://github.com/artefactual/archivematica/issues/834 (which motivates this PR) but is out of scope for this fix.

sevein (approved):
LGTM

jrwdunham:
Cherry-picked to stable/1.7.x as ec94237a953a6b4ad7955b1ffea0bf083cdf007f and fb275f9fabc2acb8893be80553d87bb159f2a235.

PR #852, opened 2017-12-16
Frontend: update vulnerable deps

sevein:
We haven't been able to keep up with the frequent updates seen in the dependencies of our transfer-browser and appraisal-tab projects. This pull request attempts to upgrade just those that addressed vulnerability issues. FWIW they're not even regular dependencies, they're just development dependencies.

In the future we should consider a massive upgrade of dependencies (most of them still backward-compatible) and also we should consider the idea of merging transfer-browser and appraisal-tabs into a single project - they're separate today because they were developed separately in two different git repos.

### DIFF
--- a/src/dashboard/frontend/appraisal-tab/package-lock.json
+++ b/src/dashboard/frontend/appraisal-tab/package-lock.json
@@ -936,16 +936,6 @@
       "integrity": "sha1-KAOY5dZkvXQDi28JBRU+borxvCA=",
       "dev": true
     },
-    "camel-case": {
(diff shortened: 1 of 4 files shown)


sevein:
Build broke, I'll look into it. :-(

sevein:
Actually it seems to be working, a saw an error but I can't reproduce anymore - using `docker build --no-cache`.

jraddaoui (approved):

sevein:
d0441aff in `qa/1.x`.

PR #853, opened 2017-12-18
Update version to 1.7

sevein:
This is in addition to 1f9adf8fae46211459fe910b81e6c18e18ad853b.

Not including the `PATCH` component because we don't include database migrations in patch releases. Once we move it out of the database (#657) we can fix that.

This migration is important for users already running AM16. The agent is published during the installation process but users upgrading to AM17 shouldn't be running that step so we need to bump the version manually from the migration.


sevein:
Rebased. It should be ready for review again.

sevein:
7a10307d in qa/1.x

Issue #760, opened 2017-10-13
Problem: Icons missing from ArchivesSpace tab on Appraisal pane


kellyannewithane:
Should have (according to the 1.6.1 docs)

-Folder icons to indicate ArchivesSpace resources and archival objects which have child records nested under them. 
-Page icons to indicate resources and archival objects that do not have child records. 
-Parallel lines icons for digital object components

ross-spencer:
Hi @kellyannewithane what is the best way for me to begin to recreate this issue?

ross-spencer:
The documentation shows the following: 

![image](https://user-images.githubusercontent.com/1880412/34175659-da486730-e4c2-11e7-85f9-e660b45634cf.png)

vs. qa17x server:

![image](https://user-images.githubusercontent.com/1880412/34180890-e0ccfbee-e4d5-11e7-8b89-5fce59cd5ffe.png)

**Issue:**

The tree is driven by this AngluarJS Control:

- Source: https://github.com/wix/angular-tree-control
- Demo: http://jsfiddle.net/8ApLX/5/

And we have configured tree-light for ArchivesSpace:

https://github.com/artefactual/archivematica/blob/d0441aff02e5e5da24d2351881edc4f4eaf455d8/src/dashboard/frontend/appraisal-tab/app/front_page/content.html#L244

Other parts of the Appraisal window are configured to use tree-classic:

https://github.com/artefactual/archivematica/blob/d0441aff02e5e5da24d2351881edc4f4eaf455d8/src/dashboard/frontend/appraisal-tab/app/front_page/content.html#L318

Switching these various options in and out and rebuilding the `bootstrap-dashboard-frontend` components fixes the issue. We just need to confirm that we don't see any unanticipated side effects. It will be useful to understand the regression if possible as well.

ross-spencer:
This fix should also ensure that the other icons are consistent with what we're expecting too as described in #859. Described as:

> When a new resource shows up on the ArchivesSpace pane of the Appraisal tab, the icon is a document. When a digital object is added to that resource the icon changes from a document to a folder.

ross-spencer:
Some screenshots from **1.6.1**.

![image](https://user-images.githubusercontent.com/1880412/34268487-80cc4ad4-e646-11e7-8d1c-d84a7dfff0b7.png)
![image](https://user-images.githubusercontent.com/1880412/34268445-50b97204-e646-11e7-94e8-ff8d573f8455.png)
[...]

PR #858, opened 2017-12-20
Makes the Tree Control configuration consistent

ross-spencer:
Following analysis of the issue it seems that one of our configuration options for the tree control that we use inside the appraisal tab was modified and so didn't show the default icons that were anticipated. This regression has been fixed with this pull-request. Documented in #760. 

Before merging @kellyannewithane and myself will check that the behaviour is consistent in an AM VM that @mamedin will configure in the morning. 

**N.B.** Because of the regression we should take a look at the functionality of the tree component in a bit more detail too to ensure we're interacting with it as expected.

### DIFF
--- a/src/dashboard/frontend/appraisal-tab/app/front_page/content.html
+++ b/src/dashboard/frontend/appraisal-tab/app/front_page/content.html
@@ -241,7 +241,7 @@
     </div>
     <div class="transfer-tree panel-body">
       <treecontrol id="archivesspace-tree"
-             class="tree-light"
+             class="tree-classic"
              tree-model="data"
              options="options"
              selected-node="selected"


sevein:
Can you prefix the commit with the component that it changes?
E.g.: `Appraisal Tab: make Tree Control config consistent`

You may have noticed by looking at the commit history that we do that regularly.

sevein:
Also it'd be nice if you add a bit of background in the body of the commit message. Your first paragraph of the PR would be enough.

sevein:
One more nitpick (sorry!!!!!), remember in the subject line to use the imperative mood (_If applied, this commit will **Refactor subsystem X for readability**_) and limit it (when possible) to 50 characters. It'd be cool to check that kind of stuff in Travis, I know some projects do it. Perhaps we should add a mention to https://chris.beams.io/posts/git-commit/ in [CONTRIBUTING.md](https://github.com/artefactual/archivematica/blob/stable/1.7.x/CONTRIBUTING.md). If you think that's a good idea I'd suggest to add it right in this PR unless you want to do it separately.

ross-spencer:
@sevein the length of the commit message before the addition of which part of the system was affected 48 characters and so i have to question the trade off when the files referenced in the commit neatly describe the context and modules. I am happy to record information like this in code-review guidelines so that we are consistently consistent for sure. I am working on these for myself in the background as I learn. They might be useful to contribute back in future. 

Pushing some of this stuff to Travis is useful up to a point, but it still relies on us getting it right on the command line locally first. (Maybe we need to write a AM Style Checker (AM Lint?) in Go for the CLI? ;) ) Whatever we do has to be usefully applicable to the community. 

_RE:_ Imperative mood. My mistake, i also thought `"this commit makes the tree control configuration consistent"` was a useful heading, and also thought it fitted into the imperative style but i re-read that link and see that it isn't described as such. I forgot `will` is part of the construct. My grammar is a bit rusty.

ross-spencer:
Hi @sevein the commit message has been changed. Just seeking confirmation that it's okay to merge.

sevein (approved):

Issue #854, opened 2017-12-18
Problem: Handle server config page displays error


sromkey:
The Handle server config page is displaying:

"Internal server error
Sorry! Something wrong happened. Please, contact with your administrator."

jrwdunham:
@sromkey This should be fixed by https://github.com/artefactual/archivematica/pull/864 which is in stable/1.7.x and qa/1.x

sallain:
Confirming that this looks to be fixed on our test servers.

PR #864, opened 2017-12-20
Fix Handle config form bug

jrwdunham:
Before rendering the Handle config form in the dashboard, the view used to expect a ``'pid_request_verify_certs'`` key in ``settings_dict``. However, this key will not be present if the form has never been saved and a ``KeyError`` will be raised. This change sets the default for this key to ``True``.

Should be cherry-picked to stable/1.7.x after merge.

Fixes https://github.com/artefactual/archivematica/issues/854


sevein (approved):

jrwdunham:
Cherry-picked to stable/1.7.x as ff2f9c2a65c71a187d1c680acef53cd5a63750b1

PR #877, opened 2018-01-10
Antivirus: Ensures that known exceptions are managed

### LINKED ISSUE #875: Issue: Clamdscan fails when files bigger than the StreamMaxLength are sent to the daemon
The antivirus microservice is failing if a user attempts to send a file bigger than the clamdscanner config value for StreamMaxLength.

From the clamscan manual pages on Ubuntu:

```
StreamMaxLength SIZE
              Close the STREAM session when the data size limit is exceeded.
              The value should match your MTA's limit for the maximum attachment size.
              Default: 25M

```

Defaults in docker are 100M and in standard Ubuntu deployments 25M. 

We need to have a look at the expected behaviour for antivirus, but the behaviour that would be consistent with the other limit settings would be for the service to return 0, in that it behaved as expected but we would record no PREMIS events for the file. We do not want to suggest that virus scanning resulted in a pass or a fail - simply that ostensibly it did not occur.

ross-spencer:
This pull request will manage exceptions known to be returned from the clamd library. Specifically, it ensures that these exceptions do not result in a transfer failing. Each exception that is managed will report that virus scanning did not occur. Because of this, no PREMIS event will be recorded as per previous discussions. 

* Introduces an exception handler to manage known exceptions from clamd
* Additional unit tests mocking the various different exceptions

Resolves #875.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaClamscan.py
@@ -32,7 +32,7 @@ import django
 from django.conf import settings as mcpclient_settings
 
 from archivematicaFunctions import cmd_line_arg_to_unicode
-from clamd import ClamdUnixSocket, ClamdNetworkSocket
+from clamd import ClamdUnixSocket, ClamdNetworkSocket, BufferTooLongError, ConnectionError
 from custom_handlers import get_script_logger
 from databaseFunctions import insertIntoEvents
 from main.models import Event, File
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/archivematicaClamscan.py
```diff
@@ -115,19 +115,38 @@ def scan(self, path):
...
             state, details = result[result_key]
-        except IOError as err:
-            if err.errno == errno.EPIPE:
-                logger.error(
-                    '[Errno 32] Broken pipe. File not scanned. Check Clamd '
-                    'StreamMaxLength')
-                return None, state, details
         except Exception as err:
-            logger.error('Virus scanning failed: %s', err, exc_info=True)
+            return self.clamd_exception_handler(Exception, err), state, details
```
jrwdunham: There's no need to pass in the `Exception` class here. It's just a globally available var anyways and you're not using it. Also, I think you should make this a function or a `@staticmethod` and remove the `self` since it's not being used.

Also, in general catching `Exception` is frowned upon in Python. Why not do `except (IOError, BufferTooLongError, ConnectionError) as err:` instead, assign to `passed` in the scope of that `except`, and then put the sole `return` call of `scan` in a `finally:` block.

src/MCPClient/tests/test_antivirus_clamdscan.py
```diff
@@ -68,6 +69,10 @@ def test_clamdscanner_scan(mocker, settings):
     FOUND_RET = ('FOUND', 'Eicar-Test-Signature')
 
     def patch(scanner, ret=OKAY_RET, excepts=False):
+        """Patch the scanner function and enable testing of exceptions raised
+        by clamdscanner that we want to control. excepts can take an argument
+        of True to pass a generic exception. excepts can also take an exception
+        as an argument for better granularity."""
```
jrwdunham: Nitpick: the final `"""` should be on its own line, cf. https://www.python.org/dev/peps/pep-0008/#documentation-strings.

src/MCPClient/tests/test_antivirus_clamdscan.py
```diff
@@ -112,8 +119,43 @@ def patch(scanner, ret=OKAY_RET, excepts=False):
...
+    # Testing a generic IOError that is not a broken pipe error that we're
+    # expecting to be able to manage from clamdscan.
+    patch(scanner, ret=OKAY_RET, excepts=IOError("Testing a generic IO Error"))
+    passed, state, details = scanner.scan('/file')
+    assert passed is False
+    assert state is None
+    assert details is None
+
+    # Broken pipe is a known error from the clamd library.
+    brokenpipeError = IOError("Testing a broken pipe error")
```
jrwdunham: Another nitpick, but I think you should use `snake_case` for Python vars consistently in new code (i.e., `brokenpipeError`, `instreamError`, etc., as per PEP 8. (I know the code base is a mix of snake and camel, but going forward it would be nice if we follow Python conventions.)

jrwdunham (changes requested):
I think this is good. I have a couple of requests for (primarily cosmetic) changes.

src/MCPClient/lib/clientScripts/archivematicaClamscan.py
ross-spencer: Good catch, I kept saying to myself to remove that! - Have added a finally statement. Hopefully looking a lot better. Much appreciated.

ross-spencer:
@jrwdunham changes made :+1:

jrwdunham (approved):

jrwdunham:
I think this is good to merge. (I still think catching `Exception` could be avoided, but I'm fine with it as is.)

PR #883, opened 2018-01-11
Add MediaConch install to MCPClient.Dockerfile

### LINKED ISSUE #882: Problem MCPClient Dockerfile does not install MediaConch
Fixed by https://github.com/artefactual/archivematica/pull/883

jrwdunham:
This commit adds a Docker `Run` command to download the .deb files to install MediaConch v. 16.12 and then install it using dpkg.

Fixes #882

### DIFF
--- a/src/MCPClient.Dockerfile
+++ b/src/MCPClient.Dockerfile
@@ -40,6 +40,8 @@ RUN set -ex \
 		inkscape \
 		jhove \
 		libimage-exiftool-perl \
+		libevent-dev \
+		libjansson4 \
 		libxml2-utils \
 		md5deep \
 		mediainfo \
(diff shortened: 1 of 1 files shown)


ross-spencer (approved):
I don't see any concerns with this script. Have confirmed it runs and installs MediaConch successfully. Have pointed Joel in the direction of another MediaConch issue on a new install. This might be due to having an empty string in the default policy entry but is unrelated to this issue.

jrwdunham:
@ross I have just updated the Dockerfile so that the MCPClient container will have a home directory for the archivematica user, which is required by MediaConch. See https://github.com/MediaArea/MediaConch_SourceCode/issues/634#issuecomment-357918480.

You should now be able to rebuild the MCPClient container with:

    $ docker-compose up -d --force-recreate --build --no-deps archivematica-mcp-client

The correct operation of MediaConch should be verifiable by running something like:

    $ docker-compose exec archivematica-mcp-client mediaconch -mc -fx /path/to/some/mkv-file.mkv

PR #886, opened 2018-01-16
Auto-select DIP upload configuration, refs #11213

mcantelon:
If pulling replacement dict data from dashboard settings, automatically advance
to the next link rather than requiring the user to manually select the config.


mcantelon:
I'm presuming that, given we haven't released this, it's okay to modify an existing migration, but if that's going to be a problem I can make it into a separate migration.

jrwdunham (approved):

PR #889, opened 2018-01-18
Fix typo "perservation"

### LINKED ISSUE #888: Problem: validateFile client script reports "access" instead of "preservation"
When the validateFile_v1.0 client script validates preservation derivatives and finds a file that is not a preservation derivative, it prints to stdout "File <UUID> is not an access derivative; not validating". This should print "File <UUID> is not a preservation derivative; not validating".

jrwdunham:
Fixing the above typo fixes the erroneous stdout message printed by validateFile when an original is not validated during preservation derivation: "File is not an access derivative; not validating".

Fixes #888.

### DIFF
--- a/src/MCPClient/lib/clientScripts/validateFile.py
+++ b/src/MCPClient/lib/clientScripts/validateFile.py
@@ -219,7 +219,7 @@ class Validator(object):
 
     def _not_derivative_msg(self):
         """Return the message to print if the file is not a derivative."""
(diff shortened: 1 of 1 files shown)


jhsimpson (approved):
👍

jrwdunham:
Cherry-picked to stable/1.7.x as 8f4fe4076664629f5f779d1f94514dd74cddcc87

Issue #866, opened 2017-12-20
Problem: Unable to fetch levels of description from AtoM


sromkey:
To reproduce:

From Administration/ AtoM, click on Levels of Description then on Fetch from AtoM. There are no errors shown but the levels do not appear. I tested this on the augustus test server.

jraddaoui:
Ready for code review in #893

jraddaoui:
Fix added to qa/1,x and stable/1.7.x

nickwilkinson:
Reopening since @jraddaoui is still working on this.

jraddaoui:
Update:

We fixed this issue by removing auth. in the endpoints were there wasn't in stable/1.7.x. But in https://github.com/artefactual/archivematica/pull/900 we re-added auth. to those endpoints and added session auth. to the entire dashboard API.

After that, we've realized that session auth. is failing in POST request, for example to '.../api/filesystem/metadata/' when a LoD is set in the 'Edit metadata' pop-out of the arrange panel.

jraddaoui:
Ready for code review in https://github.com/artefactual/archivematica/pull/917

jraddaoui:
Fix added to qa/1.x and stable/1.7.x

PR #893, opened 2018-01-23
Fix fetch LoD from AtoM

jraddaoui:
- Fix button URL, the reverse method was not obtaining the URL used
in the onClick function, naming the url fixes the issue.
- Fix dict access, use the get method to obtain the values from the dict,
which defaults to 'None'.
- Add missing double quotes in up arrow link.
- Remove auth from LoDs get and fetch endpoints, called from the
dashboard GUI to fetch and get the AtoM LoDs, but the auth. method
used in the API doesn't take in consideration the current user session,
only the API key.

Refs #866.

### DIFF
--- a/src/dashboard/src/components/api/urls.py
+++ b/src/dashboard/src/components/api/urls.py
@@ -39,7 +39,7 @@ urlpatterns = [
     url(r'^ingest/copy_metadata_files/$', views.copy_metadata_files_api),
 
     url(r'administration/dips/atom/levels/$', views.get_levels_of_description),
-    url(r'administration/dips/atom/fetch_levels/$', views.fetch_levels_of_description_from_atom),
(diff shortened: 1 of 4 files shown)


jraddaoui:
Ideally, we should take in consideration the user session or send the auth. parameters from the JS call made in:

https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/dashboard/src/media/js/administration/atom_levels_of_description.js#L7

But this should be added to stable/1.7.x and I don't think we have time to do that. Removing the auth. from those two methods leave them as they were in stable/1.6.x, but I have kept the note about if they should have auth.

jraddaoui:
I'll cherry pick to stable/1.7.x after it's reviewed and merged in qa/1.x.

sevein (approved):

jraddaoui:
ef35d8f in stable/1.7.x

jraddaoui:
Thanks @sevein!

PR #894, opened 2018-01-23
Dashboard: add rebuild-transfer-backlog

sevein:
[`rebuild-transfer-backlog`](https://github.com/artefactual/archivematica-devtools/blob/master/tools/rebuild-transfer-backlog) used to belong to archivematica-devtools. This
pull request implements it as a Dashboard management command for the
following reasons:

- Its implementation does not make use of formal APIs.
- It was designed to run in the environment where Archivematica was installed.
- The command MUST be used during AM upgrades, i.e.: devtools is for devs?

The installation docs will need to be upgraded accordingly.

In [Compose](https://github.com/artefactual-labs/am/tree/master/compose), you can run the command as follows:

    $ docker-compose run \
          --no-deps --rm \
          --entrypoint /src/dashboard/src/manage.py \
              archivematica-dashboard \
                  rebuild_transfer_backlog --no-prompt

Or in a running container:

    $ docker-compose exec archivematica-dashboard /src/dashboard/src/manage.py rebuild_transfer_backlog

You can pass an argument `--transfer-backlog-dir`. If missing, `/var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog` will be used.

You should see something like the following:

```
Rebuilding "transfers" index from /var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog/originals.
Connected to Elasticsearch node TestNode (v1.7.6).
Deleting all transfers in the "transfers" index...
Creating index...
Indexing 0bbdb151-167c-4c81-ab5b-12014cbafe7b (Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b)
Skipping indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/processingMCP.xml
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/metadata/directory_tree.txt (UUID: )
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/metadata/submissionDocumentation/METS.xml (UUID: )
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/objects/go-outline (UUID: f9b6b334-7843-4ea2-b48d-9a9f17c1d1f7)
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/logs/fileFormatIdentification.log (UUID: )
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/logs/clamAVScan.txt (UUID: )
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/logs/filenameCleanup.log (UUID: )
Indexing Prueba001-0bbdb151-167c-4c81-ab5b-12014cbafe7b/logs/FileUUIDsError.log (UUID: )
transferfile UUID: 0bbdb151-167c-4c81-ab5b-12014cbafe7b
Files indexed: 7
Request tranfer reindex in Storage Service.
```

It fixes https://github.com/artefactual/archivematica-devtools/issues/52.
Related: https://github.com/artefactual/archivematica-devtools/issues/43.

### DIFF
--- a/src/dashboard/src/main/management/commands/__init__.py
+++ b/src/dashboard/src/main/management/commands/__init__.py
@@ -1,0 +1,23 @@
+from django.core.management.base import BaseCommand
+from django.utils.six.moves import input
+
+
+class DashboardCommand(BaseCommand):
+    def success(self, message):
+        self.stdout.write(self.style.MIGRATE_SUCCESS(message))
+
+    def error(self, message):
+        self.stdout.write(self.style.ERROR(message))
+
+    def warning(self, message):
+        self.stdout.write(self.style.WARNING(message))
+
+
+def boolean_input(question, default=None):
+    question += '\n\nType "yes" to continue, or "no" to cancel: '
+    result = input('%s ' % question)
+    if not result and default is not None:
+        return default
+    while len(result) < 1 or result.lower() not in ('yes', 'no'):
+        result = input('Please answer "yes" or "no": ')
+    return result.lower() == 'yes'
--- a/src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
+++ b/src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
@@ -1,0 +1,137 @@
+"""Recreate the ``transfers`` Elasticsearch index from the Transfer Backlog.
+
+This is useful if the Elasticsearch index has been deleted or damaged
+or during software upgrades where the document schema has changed.
+
+It also requests the Storage Service to reindex the transfers.
+
+This must be run on the same system that Archivematica is installed on, since
+it uses code from the Archivematica codebase.
+
+The one required parameter is the path to the directory where Transfer Backlog
+location is stored.
+
+Copied from https://git.io/vN6v6.
+"""
+
+import logging
+import os
+import sys
+import time
+
+from django.conf import settings as django_settings
+from django.core.management.base import CommandError
+
+import elasticSearchFunctions
+import storageService
+from main.management.commands import boolean_input, DashboardCommand
+from main.models import Transfer
+
+
+logger = logging.getLogger('archivematica.dashboard')
+
+
+class Command(DashboardCommand):
+    """Recreate the ``transfers`` Elasticsearch index from the Transfer Backlog."""
+
+    help = __doc__
+
+    DEFAULT_TRANSFER_BACKLOG_DIR = \
+        '/var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog'
+
+    CHECKSUM_TYPE = 'sha256'
+    CHECKSUM_UTIL = 'sha256sum'
+
+    def add_arguments(self, parser):
+        """Entry point to add custom arguments."""
+        parser.add_argument('--transfer-backlog-dir',
+                            default=self.DEFAULT_TRANSFER_BACKLOG_DIR)
+        parser.add_argument('--no-prompt', action='store_true')
+
+    def handle(self, *args, **options):
+        """Entry point of the rebuild_transfer_backlog command."""
+        if not self.confirm(options['no_prompt']):
+            sys.exit(0)
+
+        transfer_backlog_dir = self.prepdir(options['transfer_backlog_dir'])
+        if not os.path.exists(transfer_backlog_dir):
+            raise CommandError('Directory does not exist: %s',
+                               transfer_backlog_dir)
+        self.success('Rebuilding "transfers" index from {}.'.format(
+            transfer_backlog_dir))
+
+        # Connect to Elasticsearch.
+        elasticSearchFunctions.setup_reading_from_conf(django_settings)
+        es_client = elasticSearchFunctions.get_client()
+        try:
+            es_info = es_client.info()
+        except Exception as err:
+            raise CommandError("Unable to connect to Elasticsearch: %s" % err)
+        else:
+            self.success('Connected to Elasticsearch node {} (v{}).'.format(
+                es_info['name'], es_info['version']['number']))
+
+        self.delete_index(es_client)
+        self.create_index(es_client)
+        self.populate_index(es_client, transfer_backlog_dir)
+        self.success('Indexing complete!')
+
+    def confirm(self, no_prompt):
+        """Ask user to confirm the operation."""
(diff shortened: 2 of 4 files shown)


jhsimpson:
The name of this PR doesn't really reflect the full scope of the change, I don't think.  This PR refactors the sanitize names client script, in addition to adding a new management command.

Is there some risk that this is altering the behaviour of the 'Clean Up Names' microservice group?  What do you think about splitting this PR into two (one to alter sanitize names, one to add the management command) or else rename this PR to reflect the scope of the suggested change?

sevein:
That was not a change introduced arbitrarily, I just moved the sanitization functions to a new module in archivematicaCommon so they can be used by rebuild-transfer-backlog. Otherwise we would need to install the MCPClient sources into the Dashboard environment.

sevein:
@jhsimpson I've removed the use of the sanitize function.

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -0,0 +1,209 @@
...
+        transfer_path = elasticSearchFunctions.index_files(
+            es_client,
+            'transfers',
+            'transferfile',
+            transfer_uuid,
+            os.path.join(transfer_path, ''),  # Expected by index_files().
+            status='backlog')
+
+    def ss_reindex(self, transfer_uuid):
+        """Request tranfer reindex in Storage Service."""
```
jrwdunham: typo "tranfer".

src/dashboard/src/main/management/commands/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+        self.stdout.write(self.style.WARNING(message))
+
+
+def boolean_input(question, default=None):
+    result = input('%s ' % question)
+    if not result and default is not None:
+        return default
+    while len(result) < 1 or result[0].lower() not in 'yn':
+        result = input('Please answer yes or no: ')
+    return result[0].lower() == "y"
```
jrwdunham: Kinda weird that I could enter "yno" here and this would be interpreted as "yes".

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -0,0 +1,209 @@
...
+                currentlocation=os.path.join(
+                    transfer_path.replace(django_settings.SHARED_DIRECTORY,
+                                          '%sharedPath%', 1),
+                    transfer_dir))
+        except django.db.utils.IntegrityError:
+            pass
+
+        # Extract list of files from METS data so as to populate the MySQL
+        # Files table.
+        mets_bns = '{http://www.loc.gov/METS/}'
```
jrwdunham: Maybe this should be taken from archivematicaCommon/lib/namespaces.py so it is centralized. Same with xlink.

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -0,0 +1,209 @@
...
+                file_basename)
+            checksum = self.get_file_checksum(real_file_path)
+
+            # Attempt to insert row in MySQL Files table if it doesn't already
+            # exist.
+            try:
+                File.objects.create(
+                    uuid=id_,
+                    originallocation=location_with_token,
+                    currentlocation=location_with_token,
```
jrwdunham: Given @jhsimpson's [comment](https://github.com/artefactual/archivematica/pull/894#issuecomment-359978178), I'm wondering if what you were doing here previously was getting the original (unsanitized) file name/path from the METS file. In any case, shouldn't we be doing that? I mean, given how this works right now, if I backlog a transfer containing files with sanitizable names, then delete File rows and rebuild the index using this command, then won't the `originallocation` value be technically incorrect. Probably a minor issue, but just pointing it out

jrwdunham (changes requested):
I have confirmed this works as advertised. I have a few requests for minor changes and a question.

jhsimpson (approved):
the refactoring looks good to me.  Other than responding to @jrwdunham comments, this looks good to me now.

sevein:
@jrwdunham please take another look. 97911e0 addresses the issues you mentioned.

jrwdunham:
Hey @sevein. Yes, 97911e056e509beb6acc18f18416059d402dca44 addresses most of the issues. However, I'm still tripping up on the sanitization part. I created a transfer from a directory that contains both a directory and a file whose names will be sanitized, specifically it looked like sanitizeme/bi!son/buf!falo.txt. I put this transfer in backlog. Then I manually deleted everything in the Files table:

    mysql> SET foreign_key_checks = 0;
    mysql> delete from Files;
    mysql> SET foreign_key_checks = 1;

Then I ran your `rebuild_transfer_backlog` command and got this (line 159):

    File not found: /var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog/originals/wed1-1eb01134-5476-4602-88ed-bcf6d3fa1f67/objects/bi!son/buf!falo.txt

It seems to me that you need to sanitize all the elements of the path in addition to just the basename. Also, you need to do it earlier on because the file/path on disk is sanitized:

    $ docker-compose exec archivematica-storage-service ls /var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog/originals/wed1-1eb01134-5476-4602-88ed-bcf6d3fa1f67/objects/
    bi_son
    $ docker-compose exec archivematica-storage-service ls /var/archivematica/sharedDirectory/www/AIPsStore/transferBacklog/originals/wed1-1eb01134-5476-4602-88ed-bcf6d3fa1f67/objects/bi_son
    buf_falo.txt

Also, when you call `File.objects.create` I think you need to give `originallocation` the unsanitized path and `currentlocation` the sanitized path.

My larger concern is that I wonder what's the point of just reconstructing the File table. As shown above, I had remove MySQL's foreign key checks in order to simulate the state of affairs where that table would need to be re-created. This seems highly artificial. If it did need to be re-created, then presumably the Events and Agents tables would also need fixing. This seems like a rabbithole. Maybe reconstructing the Files table is not worth the effort...

sevein:
951f5efe in stable/1.7.x

PR #896, opened 2018-01-24
Change 'None' to '' in default AtoM config

### LINKED ISSUE #879: Problem: DIP upload rsync command populated with "None" on deployment
In the AtoM DIP upload configuration screen, there's a parameter for adding an rsync command. It looks like this field is being automatically populated with the text string "None", which causes DIP upload to AtoM to fail.
I suspect that this value is present when Archivematica is deployed, but this should be checked, as well as if it reappears after an upgrade. If it does, we should fix this so that the field is left blank.

jrwdunham:
Fixes migration 0032 so that when an AtoM setting has no value in the previous state of the database it will be set to the empty string instead of `str(None)`, i.e., 'None'.

Fixes #879.

### DIFF
--- a/src/dashboard/src/main/migrations/0048_fix_upload_qubit_setting.py
+++ b/src/dashboard/src/main/migrations/0048_fix_upload_qubit_setting.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def data_migration(apps, schema_editor):
+    """Migration 0032 inadvertently set previously non-existent "Upload DIP to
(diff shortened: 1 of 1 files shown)


sevein (changes requested):
Could you make a new migration instead? 0032 is already used in production in the JiscRDSS fork.

jrwdunham:
@sevein I went ahead and made this into a migration. Any objections to merging?

sevein (approved):

jrwdunham:
Cherry-picked to stable/1.7.x as a9e836d62e19e8d3e65a8077334e01c97b377b09.

Issue #861, opened 2017-12-20
Problem: unable to update ss url in dashboard


scollazo:
In certain conditions, after accidentally misentered the storage service URL in the dashboard, you can't update it.

The dashboard log shows the following errors:
<pre>
HTTPError: 404 Client Error: NOT FOUND for url: http://localhost:8001/api/v2/pipeline/823c3af0-7940-46cb-8ec2-87a8759566fc/
ERROR     2017-12-05 18:39:15  archivematica.dashboard:views:transfer_source_locations:59:  Error retrieving source directories
Traceback (most recent call last):
  File "/usr/share/archivematica/dashboard/components/transfer/views.py", line 56, in transfer_source_locations
    return helpers.json_response(storage_service.get_location(purpose="TS"))
  File "/usr/lib/archivematica/archivematicaCommon/storageService.py", line 149, in get_location
    pipeline = _get_pipeline(get_setting('dashboard_uuid'))
  File "/usr/lib/archivematica/archivematicaCommon/storageService.py", line 123, in _get_pipeline
    response.raise_for_status()
  File "/usr/share/python/archivematica-dashboard/lib/python2.7/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 404 Client Error: NOT FOUND for url: http://localhost:8001/api/v2/pipeline/823c3af0-7940-46cb-8ec2-87a8759566fc/
WARNING   2017-12-05 18:39:21  archivematica.common:storageService:_get_pipeline:122:  This Archivematica instance is not registered with the storage service or has been disabled.
WARNING   2017-12-05 18:39:21  archivematica.common:storageService:_get_pipeline:125:  Error fetching pipeline
Traceback (most recent call last):
  File "/usr/lib/archivematica/archivematicaCommon/storageService.py", line 123, in _get_pipeline
    response.raise_for_status()
  File "/usr/share/python/archivematica-dashboard/lib/python2.7/site-packages/requests/models.py", line 935, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 404 Client Error: NOT FOUND for url: http://localhost:8001/api/v2/pipeline/823c3af0-7940-46cb-8ec2-87a8759566fc/
WARNING   2017-12-05 18:40:24  archivematica.common:storageService:_get_pipeline:122:  This Archivematica instance is not registered with the storage service or has been disabled.
[...]

PR #899, opened 2018-01-24
Dashboard: register pipeline from config page

sevein:
Register pipeline when SS returns 404. The error message has also been updated
to suggest the user that the pipeline isn't registered or it's inaccessible
based on the status code.

It's already possible to register it from `/installer/storagesetup/` but it's not intuitive for the user.

This closes https://github.com/artefactual/archivematica/issues/861.

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -36,6 +36,7 @@ from components.administration.forms import AgentForm, HandleForm, StorageSettin
 import components.administration.views_processing as processing_views
 import components.decorators as decorators
 import components.helpers as helpers
+from installer.steps import setup_pipeline_in_ss
 import storageService as storage_service
 
 from version import get_full_version
@@ -445,8 +446,15 @@ def _intial_settings_data():
 
 def general(request):
     initial_data = _intial_settings_data()
-    storage_form = StorageSettingsForm(request.POST or None, prefix='storage', initial=initial_data)
-    checksum_form = ChecksumSettingsForm(request.POST or None, prefix='checksum algorithm', initial=initial_data)
+    initial_data['storage_service_use_default_config'] = {
+        'False': False}.get(
+            initial_data.get('storage_service_use_default_config', True),
+            True)
+    storage_form = StorageSettingsForm(request.POST or None,
+                                       prefix='storage', initial=initial_data)
+    checksum_form = ChecksumSettingsForm(request.POST or None,
+                                         prefix='checksum algorithm',
+                                         initial=initial_data)
 
     if storage_form.is_valid() and checksum_form.is_valid():
         storage_form.save()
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/administration/views.py
```diff
@@ -456,11 +457,14 @@ def general(request):
     dashboard_uuid = helpers.get_setting('dashboard_uuid')
     try:
         pipeline = storage_service._get_pipeline(dashboard_uuid)
-    except Exception:
-        messages.warning(request, _("Storage server inaccessible. Please contact an administrator or update storage service URL below."))
-    else:
-        if not pipeline:
-            messages.warning(request, _("This pipeline is not registered with the storage service or has been disabled in the storage service. Please contact an administrator."))
+    except Exception, err:
```
jrwdunham: I think you should use the more modern syntax `except Exception as err` so that this will work with python3.

jrwdunham: You need to add some logic to make sure that the `'storage_service_use_default_config'` key of `initial_data` is a boolean. Right now it's a string and not a boolean because dashboard settings values are untyped. This is the same issue addressed by https://github.com/artefactual/archivematica/pull/847. Using that approach, I would add a line like:

    initial_data['storage_service_use_default_config'] = {
        'False': False}.get(
            initial_data.get('storage_service_use_default_config', True),
            True)

src/dashboard/src/components/administration/views.py
```diff
@@ -456,11 +457,14 @@ def general(request):
...
+        if err.response is not None and err.response.status_code == 404:
+            if storage_form.is_valid():
+                setup_pipeline_in_ss(storage_form.cleaned_data['storage_service_use_default_config'])
```
jrwdunham: Should there be some exception handling here? I noticed that if I simply disable the pipeline in the SS and then try to submit this form I'm getting an uncaught `HTTPError/requests.exceptions.RequestException` because the SS is returning 500 when you try to create a pipeline that already exists: `{"error_message": "(1062, \"Duplicate entry ...`.

jrwdunham (changes requested):
I have some requests for changes. I have verified that the basic requirement, however, is met: if you delete the pipeline from the SS and then submit the form at /administration/general/ the pipeline (and its locations) will be created in the SS.

sevein:
Ok @jrwdunham please give it another look and tell me what you think. Very tempted to test the hell out of this function but I have a couple of giant dev branches waiting for me on JiscRDSS.

jrwdunham (approved):

jrwdunham:
Thanks @sevein. This fixes the issues I saw. lgtm 👍

sevein:
34717735 (nice rev!) in `stable/1.7.x`

PR #900, opened 2018-01-26
Add session auth. to dashboard API

jraddaoui:
And restore auth. checks removed in 520eb63

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -31,7 +31,7 @@ from django.conf import settings as django_settings
 
 # External dependencies, alphabetical
 from annoying.functions import get_object_or_None
-from tastypie.authentication import ApiKeyAuthentication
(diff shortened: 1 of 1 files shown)


sevein:
Is it safe to do this? Could an authenticated client take advantage of it and move files around for example?

jraddaoui:
Hi @sevein, I guess it's not safe, but it was like that in stable/1.6.x and I'm not sure if the dashboard and/or other client apps use these endpoints. I could try to add user session auth. in them for the dashboard requests but it may break requests made from other places.

sevein:
Then I would call that a bug and I think it would justify the change even if it's backward incompatible.
@jhsimpson @ross-spencer @jrwdunham, what do you think?

jrwdunham:
I think these endpoints should allow both api key- and session-based authentication.

sevein (approved):
Excellent, thanks for doing this @jraddaoui!

jraddaoui:
0af67bee24acf19e62cbb3ada28356393bb9e400 in stable/1.7.x

PR #912, opened 2018-02-06
Make decision notification display on correct tab

### LINKED ISSUE #910: Problem: decision count notification circle on wrong tab in indexless deploy
When Archivematica is deployed with indexing turned off, the backlog and appraisal tabs are not displayed. This causes the red circle that indicates awaiting decisions to be located on the access tab when it should be on the ingest tab. This is presumably caused by JavaScript logic assuming that the items in the navbar &lt;ul&gt; are constant.

jrwdunham:
Fixes #910

### DIFF
--- a/src/dashboard/src/media/js/status.js
+++ b/src/dashboard/src/media/js/status.js
@@ -27,8 +27,8 @@ $(function()
         {
           this.poll();
 
-          this.transfer = this.$('ul.nav > li').eq(0);
-          this.sip = this.$('ul.nav > li').eq(3);
+          this.transfer = this.$('ul.nav > li > a[href="/transfer/"]').eq(0);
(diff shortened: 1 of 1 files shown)


sevein (approved):
I think it would be more reliable to rely on the class of ID of the element, e.g. if you're serving the site from a subpath that query wouldn't work, although I think there are many other places where that's not possible atm.

jrwdunham:
Cherry-picked to stable/1.7.x as f85e3126e5cdd8b416a05a28c66624a2a4989793.

jrwdunham:
Was actually merged into qa/1.x as ab7d00c5c2b1da96558abef8e09c48d1e9b8016e. GH shows as closed because the rebased dev branch was pushed after the qa/1.x push.

PR #913, opened 2018-02-06
Fix position of <creatingApplication> in METS

### LINKED ISSUE #907: Issue: METS is invalid due to premis:creatingApplication misplaced
In qa.1.7 we have added the PREMIS element creatingApplication to record the date an object was created. It needs to be placed in the METS file before the premis:objectCharacteristicsExtension, rather than after, in order to pass PREMIS in METS validation.

I have examples of how it is now as well as how it would look fixed.

jrwdunham:
Fixes #907


sevein (approved):

jrwdunham:
Cherry-picked to stable/1.7.x as a5c62732e87f35f5cdfcc953d708656f084e13ff.

PR #915, opened 2018-02-06
MCPClient: osdeps, delete python-unidecode

### LINKED ISSUE #897: Problem: python-unidecode osdep is unneeded
It's listed in `MCPClient/osdeps/CentOS-7.json` and `MCPClient/osdeps/RedHat-7.json`.

Not needed because it's installed via pip.

sevein:
Installed via pip.

This closes #897.


scollazo:
The package should also be removed from https://github.com/artefactual-labs/am-packbuild/blob/master/rpm/archivematica/archivematica.spec#L53

scollazo (changes requested):
There is another place where this change needs to be made

sevein:
> There is another place where this change needs to be made

And how does that block this PR?

scollazo (approved):

PR #917, opened 2018-02-08
Fix saveLevelOfDescription request authentication

jraddaoui:
Tastypie SessionAuthentication requires to send the CSRFToken in the
X-CSRFToken header for some request methods. This is the only request
not using GET made from the dashboard frontend (using session auth.).

Refs #866


jraddaoui:
Related to #900, fixes #866.

sevein (approved):

jraddaoui:
163e3dd4b3f89cac2e96f9b8fe285c790c9c6042 in stable/1.7.x

PR #931, opened 2018-02-19
MCPClient: add Docker base image

sevein:
As a developer my docker-compose environment attempts to build the docker images from scratch too often. Most of the images take not too long to build excepting the MCPClient image because it installs a bunch of dependencies.

This pull request introduces a base Docker image for MCPClient already published as `artefactual/archivematica-mcp-client-base:20180219.01.52dc9959`. This is going to make our builds significantly faster.

However this should only be considered a temporary solution. In the long term we should probably be looking at things like automated builds, split mcpclient tools in separate images, update to Ubuntu 16.04, consume osdeps.json, etc...

### DIFF
--- a/src/MCPClient-base.Dockerfile
+++ b/src/MCPClient-base.Dockerfile
@@ -1,0 +1,96 @@
+FROM ubuntu:14.04
+
+# Set the locale
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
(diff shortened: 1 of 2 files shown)


jrwdunham (approved):
Looks good to me. It might be useful to add some notes to the README of am.git (and other repos that use these docker files) explaining how devs should handle cases where they have to add OS-level dependencies to the MCPClient container.

PR #935, opened 2018-02-22
MCPClient: Delete empty directories after registering them

### LINKED ISSUE #934: Problem: not all directories recorded in structMap
When testing PIDs using TestTransfers/acceptance-tests/pid-binding/hierarchy-with-empty-dir on am17x.qa

-physical and normative structMap don't record presence of dir2aiii, which should have ds_store object in it.  Instead, dir2aiii shows up in exifTool raw output

jrwdunham:
Moves the "delete empty directories" step in archivematicaCreateMETS2 to after the call to `get_normative_structmap` so that the normative logical structMap will contain the empty directories.

Fixes #934


sevein (approved):

jrwdunham:
Cherry-picked to stable/1.7.x as 14bdc74181aa9525777a4bd601b6072bd5d0d3aa.

PR #936, opened 2018-02-22
Add package API (async transfers)

sevein:
[RDSSARK-474](https://jiscdev.atlassian.net/browse/RDSSARK-474)

New package API - it allows to create new packages.

It's an alternative to the [existing methods](https://gist.github.com/sevein/5615e0df0ff9645f17ed06672a271633) to start a transfer. JiscRDSS's channel adapter (deposit packages directly into `activeTransfers/*`) and automation tools (uses existing `/api/{start_transfer,approve,unapproved}` APIs) could adopt this new API but the existing functionality isn't going to stop working. This PR updates the transfer browser widget to adopt the new API - before it used internal endpoints (`/filesystem/transfer/`, `/mcp/list`, `/mcp/execute`).

Request example:

```
$ curl \
	-v \
	-d "@send-transfer.json" \
	-H "Content-type: application/json" \
	-H "Authorization: ApiKey test:test" \
	-X POST \
		http://127.0.0.1:62080/api/v2beta/package/
```

This is the document we sent in the payload:

```json5
{
  "name": "Test12345",
  // The value of `path` is the result of `base64(location-uuid:relative-path)`
  "path": "MWQ0ZmU3NTctNGM2ZS00MmQxLWE5Y2ItZmQ4YTBkMjNlZWRkOi9ob21lL2FyY2hpdmVtYXRpY2EvYXJjaGl2ZW1hdGljYS1zYW1wbGVkYXRhL1NhbXBsZVRyYW5zZmVycy9JbWFnZXMvcGljdHVyZXM="
}
```

It starts the transfer without approval - as MCPServer knows when the files have been copied and the package is ready for processing. `POST /package` returns the UUID of the package/transfer immediately.

### DIFF
--- a/src/MCPServer/lib/RPCServer.py
+++ b/src/MCPServer/lib/RPCServer.py
@@ -1,8 +1,8 @@
-#!/usr/bin/env python2
+"""Archivematica MCPServer RPC."""
 
 # This file is part of Archivematica.
 #
-# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
+# Copyright 2010-2018 Artefactual Systems Inc. <http://artefactual.com>
 #
 # Archivematica is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
@@ -17,74 +17,70 @@
 # You should have received a copy of the GNU General Public License
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
-# @package Archivematica
-# @subpackage MCPServer
-# @author Joseph Perry <joseph@artefactual.com>
-
-import cPickle
-import gearman
 import logging
 import lxml.etree as etree
 from socket import gethostname
 import time
 
-from linkTaskManagerChoice import choicesAvailableForUnits
-
 from django.conf import settings as django_settings
+import gearman
+
+from linkTaskManagerChoice import choicesAvailableForUnits
+from package import create_package
+from worker_util import log_exceptions, unpickle_payload, pickle_result
 
 
 LOGGER = logging.getLogger("archivematica.mcp.server.rpcserver")
 
 
-def rpcError(code="", details=""):
-    ret = etree.Element("Error")
-    etree.SubElement(ret, "code").text = code.__str__()
-    etree.SubElement(ret, "details").text = details.__str__()
-    return ret
+@pickle_result
+@unpickle_payload
+@log_exceptions(LOGGER)
+def job_approve_handler(payload, gearman_worker, gearman_job):
+    job_id = payload["jobUUID"]
+    chain = payload["chain"]
+    user_id = str(payload["uid"])
+    LOGGER.debug("Approving: %s %s %s", job_id, chain, user_id)
+    if job_id in choicesAvailableForUnits:
+        choicesAvailableForUnits[job_id].proceedWithChoice(chain, user_id)
+    return "approving: ", job_id, chain
 
 
-def getJobsAwaitingApproval():
-    ret = etree.Element("choicesAvailableForUnits")
-    for UUID, choice in choicesAvailableForUnits.items():
+@pickle_result
+@log_exceptions(LOGGER)
+def job_awaiting_approval_handler(gearman_worker, gearman_job):
+    ret = etree.Element('choicesAvailableForUnits')
+    for uuid, choice in choicesAvailableForUnits.items():
         ret.append(choice.xmlify())
     return etree.tostring(ret, pretty_print=True)
 
 
-def approveJob(jobUUID, chain, user_id):
-    LOGGER.debug("Approving: %s %s %s", jobUUID, chain, user_id)
-    if jobUUID in choicesAvailableForUnits:
-        choicesAvailableForUnits[jobUUID].proceedWithChoice(chain, user_id)
-    return "approving: ", jobUUID, chain
-
-
-def gearmanApproveJob(gearman_worker, gearman_job):
-    try:
-        data = cPickle.loads(gearman_job.data)
-        jobUUID = data["jobUUID"]
-        chain = data["chain"]
-        user_id = str(data["uid"])
-        return cPickle.dumps(approveJob(jobUUID, chain, user_id))
-    # catch OS errors
-    except Exception:
-        LOGGER.exception('Error approving job')
-        raise
-
-
-def gearmanGetJobsAwaitingApproval(gearman_worker, gearman_job):
-    try:
-        return cPickle.dumps(getJobsAwaitingApproval())
-    # catch OS errors
-    except Exception:
-        LOGGER.exception('Error getting jobs awaiting approval')
-        raise
+@pickle_result
+@unpickle_payload
+@log_exceptions(LOGGER)
+def package_create_handler(payload, gearman_worker, gearman_job):
+    """Handle create package request."""
+    create_package_args = (
+        payload.get('name'),
+        payload.get('type'),
+        payload.get('accession'),
+        payload.get('path'),
+        payload.get('metadata_set_id'),
+        payload.get('auto_approve', True),
+    )
+    return create_package(*create_package_args).pk
 
 
 def startRPCServer():
     gm_worker = gearman.GearmanWorker([django_settings.GEARMAN_SERVER])
     hostID = gethostname() + "_MCPServer"
     gm_worker.set_client_id(hostID)
-    gm_worker.register_task("approveJob", gearmanApproveJob)
-    gm_worker.register_task("getJobsAwaitingApproval", gearmanGetJobsAwaitingApproval)
+
+    # The tasks registered in this worker should not block.
+    gm_worker.register_task("approveJob", job_approve_handler)
+    gm_worker.register_task("getJobsAwaitingApproval", job_awaiting_approval_handler)
+    gm_worker.register_task("packageCreate", package_create_handler)
+
     failMaxSleep = 30
     failSleep = 1
     failSleepIncrementor = 2
--- a/src/MCPServer/lib/archivematicaMCP.py
+++ b/src/MCPServer/lib/archivematicaMCP.py
@@ -60,6 +60,7 @@ from unitFile import unitFile
 from unitTransfer import unitTransfer
 from utils import isUUID
 import RPCServer
+import worker
 
 from archivematicaFunctions import unicodeToStr
 from databaseFunctions import auto_close_db, createSIP, getUTCDate
@@ -659,6 +660,16 @@ if __name__ == '__main__':
     t = threading.Thread(target=flushOutputs)
     t.daemon = True
     t.start()
+
+    # Start internal workers to process background operations. Workers were
+    # introduced when the non-blocking `packageCreate` method was added to the
+    # RPC service and we needed to schedule background jobs. If the workers are
+    # busy the jobs are queued by Gearman.
+    for i in range(django_settings.WORKERS):
+        t = threading.Thread(target=worker.start_worker, args=(i + 1,))
+        t.daemon = True
+        t.start()
+
     cleanupOldDbEntriesOnNewRun()
     watchDirectories()
 
--- a/src/MCPServer/lib/package.py
+++ b/src/MCPServer/lib/package.py
@@ -1,0 +1,401 @@
+"""Package management."""
+
+# This file is part of Archivematica.
+#
+# Copyright 2010-2018 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+import cPickle
+import collections
+import logging
+import os
+import shutil
+from tempfile import mkdtemp
+from uuid import uuid4
+
+from django.conf import settings as django_settings
+import gearman
+
+from archivematicaFunctions import unicodeToStr
+from jobChain import jobChain
+from main.models import Transfer, TransferMetadataSet
+import storageService as storage_service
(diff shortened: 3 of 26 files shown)


src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -659,6 +660,13 @@ def _except_hook_log_everything(exc_type, exc_value, exc_traceback):
     t = threading.Thread(target=flushOutputs)
     t.daemon = True
     t.start()
+
+    # TODO: how many workers should we have? Should this be configurable?
```
sevein: I'd like to get your thoughts here!

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+def _copy_from_transfer_sources(paths, relative_destination):
+    """Copy files from source locations to the currently processing location.
+
+    Any files in locations not associated with this pipeline will be ignored.
+
+    :param list paths: List of paths. Each path should be formatted
+                       <uuid of location>:<full path in location>
+    :param str relative_destination: Path relative to the currently processing
+                                     space to move the files to.
+    :returns: Tuple of (boolean error, message)
```
helenst: It doesn't return anything. Given that the calling function doesn't expect a return value, I am guessing it is the docstring that is incorrect?

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+            err = e
+        else:
+            # No exceptions!
+            return
+        # At this point we know that the transfer has failed and we want to do
+        # our best effort to update the state without further interruptions.
+        try:
+            Transfer.objects.get(pk=args[0])
+
+            #
```
helenst: does this also need adding to the TODO list in the main PR comments?

src/MCPServer/lib/archivematicaMCP.py
helenst: I guess this comes down to how many package transfers might need to be started simultaneously? (at least for now with that being the only endpoint served by the workers). I don't know what is a good number to choose there but making it configurable seems sensible to me.

helenst (commented):
I guess there are still things to be done here but so far this seems to be working and looks good to me - just a couple of minor comments.

src/MCPServer/lib/package.py
sevein: I had some code here that was updating the `Transfer.state` FSM. We decided to think about it a bit further so I deleted it. I'm going to add a comment to make that clear. I wanted to leave the decorator in place because I thought it would be clearer to have it even if it's a no-op right now.

sevein:
@helenst I've addressed your comments, thanks!

src/MCPServer/lib/RPCServer.py
```diff
@@ -31,60 +25,62 @@
 from linkTaskManagerChoice import choicesAvailableForUnits
```
jrwdunham: Nitpick: shouldn't this go below the third-party imports?

src/MCPServer/lib/worker_util.py
```diff
@@ -0,0 +1,55 @@
...
+                internal_err = isinstance(err, RPCServerError)
+                logger.error('Exception raised by handler %s: %s',
+                             fn.__name__, err, exc_info=not internal_err)
+                raise  # So GearmanWorker knows that it failed.
+        return wrap
+    return decorator
+
+
+def unpickle_payload(fn):
+    """Worker handle decorator to unpickle the incoming payload.
```
jrwdunham: Nitpick: should "Worker handle" here and "Working handle" below both be "Worker handler" as above?

src/MCPServer/lib/RPCServer.py
```diff
@@ -31,60 +25,62 @@
...
+LOGGER = logging.getLogger("archivematica.mcp.server.rpcserver")
 
 
-def getJobsAwaitingApproval():
-    ret = etree.Element("choicesAvailableForUnits")
-    for UUID, choice in choicesAvailableForUnits.items():
+@pickle_result
+@unpickle_payload
+@log_exceptions(LOGGER)
+def job_approve_handler(payload, gearman_worker, gearman_job):
```
jrwdunham, line 39: This is very cool. Although, it's a bit confusing because these Gearman task handlers don't have the expected signature. Also, these 3 funcs take args that are never used. Could you replace all of the `gearman_worker, gearman_job` in these 3 funcs with just `*args`? Would that make it more readable? Same question applies to the handler in lib/worker.py.

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+            # TODO: update state once we have a FSM.
+            #
+
+        except Exception:
+            pass
+        raise err
```
jrwdunham: It's possible to raise `None` here, which we should not allow I think. Pylint says "Raising NoneType while only classes or instances are allowed (raising-bad-type)"

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+        watched_dir=os.path.join(
+            django_settings.WATCH_DIRECTORY,
+            'activeTransfers/TRIM'),
+        chain='e4a59e3e-3dba-4eb5-9cf1-c1fb3ae61fa9',
+        link='2483c25a-ade8-4566-a259-c6c37350d0d6',
+    ),
+}
+
+
+def _file_is_an_archive(file):
```
jrwdunham: Nitpick: shouldn't use `file` as a var name.

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+def _check_filepath_exists(filepath):
+    error = None
+    if filepath == '':
+        error = 'No filepath provided.'
+    # check if exists
+    if error is None and not os.path.exists(filepath):
+        error = 'Filepath ' + filepath + ' does not exist.'
+    # check if is file or directory
+    # check for trickery
+    try:
```
jrwdunham: This is odd. Why not:

    def _check_filepath_exists(filepath):
        if filepath == '':
            return 'No filepath provided.'
        if not os.path.exists(filepath):
            return 'Filepath ' + filepath + ' does not exist.'
        if '..' in filepath:  # check for trickery
            return 'Illegal path.'
        return None

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+
+    Any files in locations not associated with this pipeline will be ignored.
+
+    :param list paths: List of paths. Each path should be formatted
+                       <uuid of location>:<full path in location>
+    :param str relative_destination: Path relative to the currently processing
+                                     space to move the files to.
+    """
+    processing_location = storage_service.get_location(purpose='CP')[0]
+    transfer_sources = storage_service.get_location(purpose='TS')
```
jrwdunham, line 154: Too bad we need to perform 2 requests here. Another argument for revising the SS API.

jrwdunham: Also too bad that this is basically copy-pasted from filesystem_ajax/views.py. It would be nice if we could consolidate these and then just change how the func is called in filesystem_ajax/views.py and catch the exceptions raised here.

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+    # background so we don't make the caller wait. We persist the job handle so
+    # we can use it later to attempt to know the status of the background job.
+    _submit_create_package_internal_bg_job(transfer.pk, name, type_, path)
+    return transfer
+
+
+def _submit_create_package_internal_bg_job(id_, name, type_, path):
+    """Submit package creation background job to queue.
+
+    It returns the job handle assigned by Gearman.
```
jrwdunham: Why does it return anything if the return value is never used?

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+    })
+    gm_client = gearman.GearmanClient([django_settings.GEARMAN_SERVER])
+    request = gm_client.submit_job('MCPServerInternalPackageStartTransfer',
+                                   payload, background=True,
+                                   wait_until_complete=False)
+    gm_client.shutdown()
+    return request.job.handle
+
+
+def _create_temporary_directory():
```
jrwdunham: Again, I know this is taken from elsewhere in AM, but this seems like a good candidate for a context manager which yields the tmpdir and calls `shutil.rmtree` on it at the end so the caller doesn't have to. Maybe an enhancement that is out of scope for this work. Just an idea.

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+    logger.debug('Package %s: starting transfer (%s)', id_, (name, type_, path))
+    transfer = Transfer.objects.get(pk=id_)
+
+    tmpdir = _create_temporary_directory()
+    logger.debug('Package %s: temp directory created: %s', id_, tmpdir)
+
+    if _file_is_an_archive(path):
+        transfer_dir = tmpdir
+        p = path.split(':', 1)[1]
+        logger.debug('Package %s: found a zip file, splitting path %s', id_, p)
```
jrwdunham: Seems odd to call it a "zip" file when it could be different types of archive. Also, doesn't it sound odd to say we're splitting the result of the split? Shouldn't `p` here be `path`?

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+        filepath = os.path.join(tmpdir, os.path.basename(p))
+    else:
+        path = os.path.join(path, '.')  # Copy contents of dir but not dir
+        transfer_dir = os.path.join(tmpdir, name)
+        filepath = os.path.join(tmpdir, name)
```
jrwdunham: Again, I see lots of this is copy/paste from filesystem_ajax/views.py, but couldn't this just be `transfer_dir = filepath = os.path.join(tmpdir, name)`?

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,391 @@
...
+                 id_, filepath)
+    dest = _move_to_processing_dir(filepath)
+
+    # Similar to what's happening in archivematicaMoveTransfer.py.
+    # TODO: is this needed?
+    logger.debug('Package %s: updating Transfer.currentlocation=%s', id_, dest)
+    src = dest
+    dest = dest.replace(django_settings.SHARED_DIRECTORY, '%sharedPath%', 1)
+    if dest.endswith("/"):
+        dest = os.path.join(dest, os.path.basename(src))
```
jrwdunham: This seems really weird. I'd like to understand if this code path ever gets run and what is expected to happen when it does, and why. From the snippet below, it looks like it would do nothing...

jrwdunham: >>> dest = '/a/b/c/d/e/f/'
    >>> src = dest
    >>> dest = dest.replace('/a/b', '%sharedPath%', 1)
    >>> dest
    '%sharedPath%/c/d/e/f/'
    >>> src
    '/a/b/c/d/e/f/'
    >>> if dest.endswith('/'):
    ...     dest = os.path.join(dest, os.path.basename(src))
    ... 
    >>> dest
    '%sharedPath%/c/d/e/f/'

src/MCPServer/lib/settings/common.py
```diff
@@ -168,3 +172,5 @@
 LIMIT_GEARMAN_CONNS = config.get('limit_gearman_conns')
 RESERVED_AS_TASK_PROCESSING_THREADS = config.get('reserved_as_task_processing_threads')
 SEARCH_ENABLED = config.get('search_enabled')
+STORAGE_SERVICE_CLIENT_TIMEOUT = config.get('storage_service_client_timeout')
```
jrwdunham, line 175: Is this setting ever used in the MCPServer? I searched for it and couldn't find it. Is it needed?

src/MCPServer/lib/worker_error.py
```diff
@@ -0,0 +1,26 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
+
+
+class RPCServerError(Exception):
+    """Base exception of RPCServer."""
+    pass
```
jrwdunham: Nitpick: you don't need `pass` if you have a docstring.

src/dashboard/src/contrib/mcp/client.py
```diff
@@ -56,3 +63,35 @@ def notifications(self):
...
+        })
+        response = gm_client.submit_job('packageCreate', data,
+                                        background=False,
+                                        wait_until_complete=True,
+                                        poll_timeout=INFLIGHT_POLL_TIMEOUT)
+        gm_client.shutdown()
+        if response.state == gearman.JOB_COMPLETE:
+            return cPickle.loads(response.result)  # Transfer ID (pickled)
+        elif response.state == gearman.JOB_FAILED:
+            raise RPCError('MCPServer returned an error (check the logs)')
```
jrwdunham, line 86: Are other `state` values not also possible here? What about

    JOB_UNKNOWN  = 'UNKNOWN'  # Request state is currently unknown, either unsubmitted or connection failed
    JOB_PENDING  = 'PENDING'  # Request has been submitted, pending handle
    JOB_CREATED  = 'CREATED'  # Request has been accepted

jrwdunham (commented):
I have some comments. I've confirmed that I can use the example curl command to start different types of transfers. I guess that the path to the packages resources will be changed, but for now the description of this PR is incorrect in that you have to make a POST request to `/api/v2beta/package/` and not the stated `/api/package/`.

I also noticed that if you start a Bag or Zipped Bag transfer, the first half dozen or so micro-services are listed under the group "Approve transfer" which doesn't really make sense any more because approval is gone. I think this should probably be changed and it will require a migration I think.

src/MCPServer/lib/settings/common.py
sevein: Yes, it's an implicit requirement when you import storageService because that module looks up `django_settings`. I realize it's not great, I will think about it.

src/dashboard/src/contrib/mcp/client.py
sevein: I think they are but I don't know what I would do with them. We have other RPC methods doing the same and it's never been a problem. We're hoping to use gRPC soon but the first iteratin is just reusing this ugly RPC mechanism we invented on top of Gearman. It's ugly but I wouldn't try to make it much better, kind of reinventing the wheel when we know gRPC is awesome for this purpose. I guess this is one of these situations where I can use the "walking skeleton" card hoping that we can ignore this for now... ;-)

sevein:
For now only merged in JiscRDSS. The version here is older and needs to be updated.

PR #937, opened 2018-02-22
MCPClient: add CAPTURE_CLIENT_SCRIPT_OUTPUT setting

### LINKED ISSUE #763: Problem: there is no way to stop passing output streams from MCPClient worker processes to the MCPServer scheduler
Columbia University Library (CUL) would like to increase scalability and reduce performance problems by making it possible to configure Archivematica to stop MCPClient worker processes from passing very large stdout/stderr streams back to the MCPServer (Gearman) task scheduler.

While it would be nice to send the MCPClient worker output streams to some other storage service (e.g., a document store), CUL and Artefactual have opted to forego this option and instead simply allow for a binary configuration setting, allowing for the following two options:

1. retain output streams, i.e., by continuing to send them from worker to scheduler and storing them in the MCP database; or
2. discard output streams entirely, except for stderr which should be saved in the same way as before for debugging failures; see https://github.com/artefactual/archivematica/issues/909.

Expected results would be: 
- decrease in total processing time. This was borne out by [AM acceptance tests PR 47](https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/47) which showed a 6% reduction in total processing time for small AIPs (with larger AIPs expected to exhibit greater performance improvements).
- increase in number of files processed simultaneously
- increase in size of the largest file that can be processed

### LINKED ISSUE #816: Problem: capture_output defaults on subprocess functions are unclear
The recent [merge of PR 805](https://github.com/artefactual/archivematica/pull/805/files) allows for output streams from subprocesses to be captured or not. However, the defaults values for `capture_output` for the affected functions are inconsistent. This needs to be addressed before the dev/1.8.x branch can be merged into qa/1.x. 

@lower29, this work may be relevant to your https://github.com/JiscRDSS/archivematica/pull/42

### LINKED ISSUE #909: Problem: deploys with `capture_output` set to `no` lose task error logs
If an MCPClient client script task fails (non-zero exit code), then stderr is lost under a deploy where "capture output" is set to "no". It should be relatively easy to cause stderr to be passed from gearman worker (MCPClient) to gearman master (MCPServer) only in those cases where something has gone wrong (non-zero exit code) while still maintaining the performance benefits in the successful cases.

jrwdunham:
Adds a CAPTURE_CLIENT_SCRIPT_OUTPUT setting to the MCPClient's Django settings. This controls what is passed to `executeOrRun`'s newly introduced `capture_output` kwarg. Similar kwargs added to subprocess-calling functions in executeOrRunSubProcess.py. Includes tests for correct behaviour of archivematicaCommon/lib/executeOrRunSubProcess.py::launchSubProcess.

Note: output capturing of the compression call in the compressAIP.py client script is turned off here uniformly because a lot of useless output is generated here for very large AIPs, which hampers performance.

Connected to #763 
Connected to #909 
Connected to #816 

Acceptance tests that confirm correct behaviour and performance enhancement of this feature are in [acceptance tests PR 75](https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/75).

### DIFF
--- a/src/MCPClient/install/README.md
+++ b/src/MCPClient/install/README.md
@@ -248,6 +248,11 @@ This is the full list of variables supported by MCPClient:
     - **Type:** `string`
     - **Default:** `3306`
 
+- **`ARCHIVEMATICA_MCPCLIENT_MCPCLIENT_CAPTURE_CLIENT_SCRIPT_OUTPUT`**:
+    - **Description:** controls whether or not to capture stdout from client script subprocesses.  If set to `true`, then stdout is captured; if set to `false`, then stdout is not captured. If set to `true`, then stderr is captured; if set to `false`, then stderr is captured only if the subprocess has failed, i.e., returned a non-zero exit code.
+    - **Config file example:** `MCPClient.capture_client_script_output`
+    - **Type:** `boolean`
+    - **Default:** `true`
 
 ## Logging configuration
 
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -185,7 +185,9 @@ def execute_command(gearman_worker, gearman_job):
                 task_uuid, script)
     try:
         exit_code, std_out, std_error = executeOrRun(
-            'command', script, stdIn='', printing=True)
+            'command', script, stdIn='',
+            printing=django_settings.CAPTURE_CLIENT_SCRIPT_OUTPUT,
+            capture_output=django_settings.CAPTURE_CLIENT_SCRIPT_OUTPUT)
     except OSError:
         logger.exception('Execution failed')
         return cPickle.dumps({'exitCode': 1,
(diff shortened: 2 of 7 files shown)


jraddaoui (commented):
I guess we can't get the setting value from the MCPClient in `archivematicaCommon/lib/executeOrRunSubProcess.py` to avoid passing it through all the other methods, right?

jrwdunham:
@jraddaoui The `CAPTURE_CLIENT_SCRIPT_OUTPUT` should only affect how these subprocess-calling functions behave when the MCPClient makes use of them. Since these functions are used elsewhere in Archivematica, we need to make sure that this `CAPTURE_CLIENT_SCRIPT_OUTPUT` flag does not affect those other uses, which may be depending on more verbose return values.

jraddaoui:
Great, thanks @jrwdunham! This looks great to me, but I'd wait to hear @sevein and @ross-spencer feedback ;)

sevein (approved):
@jrwdunham this lgtm but I think we need to update the install/README.md file with the new setting that you're adding.

jrwdunham:
@sevein lemme know if there's an issue with the changes to the install readme. Otherwise I'll merge.

sevein:
Merge away!

Issue #598, opened 2017-04-20
Problem: quarantine delay is not reliable


sevein:
When a transfer is sent to quarantine the user will be prompted to remove it from quarantine manually. But also, by default, the transfer is removed from quarantined automatically after 28 days. This is a delay that can be configured by the user in the processing configuration.

The purpose of the delay is to allow virus definitions to update, before virus scan.

There are two modules in MCP implementing the processing delay ([1](https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/MCPServer/lib/linkTaskManagerChoice.py#L97-L121), [2](https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py#L87-L116)).

It's done with a timer from the threading module which doesn't seem to be provide real guarantees. What would happen if the process is interrupted before the timer finishes?
AMQP or Redis seem to offer primitives that allow implementing scheduling. Gearman doesn't seem to provide any.

### Solution 1

Deprecate quarantine delay functionality. Only the user would be able to remove it from quarantine.

### Solution 2

Update virus definitions before antivirus checking?

### Solution 3

Implement delayed jobs using Redis or similar. Once a new scheduled job comes in, MCP would persist it somewhere. In a loop, the tasks would be polled frequently and throw new jobs when needed. The following is a library that could be used for this purpose or as a reference: https://github.com/josiahcarlson/rpqueue/ (it uses Python + Redis).

Remember that Redis is already used as a Gearman backend, we've tried this successfully in our local Docker developer setup. Adding Redis to our stack would be also beneficial for other purposes like caching in Django.

jhsimpson:
I prefer deprecating the quarantine functionality altogther (i.e. remove the microservice).  This requires more discussion, @sallain @kellyannewithane and others.  

Quarantine is a feature that was created before there was an aip reingest workflow, or a transfer backlog.

It is possible to put transfers into backlog and do a full reingest later, which achieves the same purpose as quarantine.  The quarantine storage space is not managed by the storage service, and is not being monitored, so it is not a good place to store content for any length of time.

sromkey:
Justin's comment:
"It is possible to put transfers into backlog and do a full reingest later, which achieves the same purpose as quarantine."
[...]

PR #948, opened 2018-03-04
Remove magic links

### LINKED ISSUE #947: Problem: MCPServer magic links are deprecated
We have some chains in the workflow that make use of [magic links](https://wiki.archivematica.org/MCPServer/TaskTypes#Magic_Links_-_Deprecated).

Because magic links were deprecated we should replace the chain links to make use of unit variables instead. This would simplify the code significantly.

sevein:
This closes #947.

MCPServer's workflow system has [ten different types of chain links with different purposes](https://wiki.archivematica.org/MCPServer/TaskTypes), e.g. "run once", "run for each file", "get user choice", etc...

One of the chain links known as "Set Unit Variable" is used by Archivematica to persist data associated to a particular unit using the `UnitVariable` model. Later this data can be looked up using "Get Unit Variable MicroServiceChainLink" chain links. Combined, this can be used to determine the path of the workflow in a future decisioning point.

When this mechanism was introduced it deprecated magic links. The latter is inferior because the associated value is persisted in a field of the model (Transfer/SIP) meaning that you can only have one associated value at a time.

This pull request replaces the bits in the workflow still using magic links with the superior UnitVariable-based mechanism. As a result, we can remove one model, four fields and bunch of unreliable code.


src/dashboard/src/main/migrations/0049_remove_magic_links.py
```diff
@@ -0,0 +1,244 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='taskconfigassignmagiclink',
+            name='execute',
+        ),
+        migrations.RemoveField(
+            model_name='taskconfigassignmagiclink',
+            name='replaces',
+        ),
```
jrwdunham, line 220: Why is it necessary to remove the field of the model that you are deleting below? Is it because these fields point to related tables and must be required before Django's ORM or the RDBMS will allow the deletion?

src/dashboard/src/components/administration/management/commands/graph_links.py
```diff
@@ -0,0 +1,754 @@
+# -*- coding: utf-8 -*-
```
jrwdunham, line 1: How is this conceptually related to the removal of magic links and their replacement by unit variables? Do these two new commands differ in any significant way from their devtools counterparts?

src/dashboard/src/components/administration/management/commands/graph_links.py
```diff
@@ -0,0 +1,754 @@
...
+    g.set('transform', 'matrix(0.13429683446884155,0,0,0.13429683446884155,-878.8951971178758,2349.1367317996337)')
+    script = etree.Element('script')
+    script.set('type', 'text/ecmascript')
+    script.text = SVGPAN_SCRIPT
+    g.addprevious(script)
+    g.find('{http://www.w3.org/2000/svg}title').text = 'Archivematica MCP Workflows'
+    xml.write(filename)
+
+
+def main(format, verbosity):
```
jrwdunham: Nitpick: probably shouldn't redefine builtin `format`.

src/dashboard/src/components/administration/management/commands/graph_links.py
jrwdunham: In an effort to answer my own questions, I've run a diff on the devtools version and this version and see the following changes:

- `edge_magic_chain_links` is removed as well as the `bridgeMagicChainLinks*` funcs, which shows the relationship to the "remove magic links" work.
- names changed to be more pythonic (camelCase to snake_case) and other cosmetic changes.
- MySQL password no longer being passed around.

I've tested running both Django commands and have confirmed that they work. I'm happy for these two to be included in this PR even though they seem tangential.

src/dashboard/src/components/administration/management/commands/json_links.py
```diff
@@ -0,0 +1,316 @@
...
+    try:
+        ttype = TASK_TYPES[task_config.tasktype_id]
+    except KeyError:
+        # I could just make this an error but it's pretty bad
+        raise ValueError('Unknown task type: %s', task_config.tasktype_id)
+    else:
+        model, description, manager_class_name = ttype
+
+    link_config = {
+        '@model': model.__name__ if model is not None else None,
```
jrwdunham: Can `model` ever be `None` here? If not, I don't think you need the `if ...`.

src/dashboard/src/components/administration/management/commands/json_links.py
```diff
@@ -0,0 +1,316 @@
...
+            link_config['filter_subdir'] = stdtask_config.filter_subdir
+            link_config['filter_file_start'] = stdtask_config.filter_file_start
+            link_config['filter_file_end'] = stdtask_config.filter_file_end
+            link_config['stdout_file'] = stdtask_config.stdout_file
+            link_config['stderr_file'] = stdtask_config.stderr_file
+            link_config['requires_output_lock'] = stdtask_config.requires_output_lock
+
+    elif model == MicroServiceChainChoice:
+        choices = MicroServiceChainChoice.objects.filter(choiceavailableatlink_id=link.id)
+        link_config['chain_choices'] = list(item.chainavailable_id for item in choices)
```
jrwdunham: Nitpick: Why are you `list`ifying a generator? Why not just use a list comp: `[item.chainavailable_id for item in choices]`?

src/dashboard/src/main/migrations/0049_remove_magic_links.py
```diff
@@ -0,0 +1,244 @@
...
+    ###########################################################################
+
+    TaskConfig.objects.filter(
+        pk='c6f9f99a-0b60-438f-9a8d-35d4989db2bb',
+        tasktype__description='assign magic link',
+    ).delete()
+    TaskConfig.objects.filter(
+        pk='38324d67-8358-4679-902d-c20dcdfd548b',
+        tasktype__description='goto magic link',
+    ).delete()
```
jrwdunham: What's the point of adding the `tasktype__description=` condition here? Is it just to make it clearer what you're deleting? It's not strictly necessary right?

src/dashboard/src/components/administration/management/commands/graph_links.py
```diff
@@ -0,0 +1,754 @@
...
+
+    return ret
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('-f', '--format', choices=['png', 'svg'], default='svg')
+
+    def handle(self, *args, **options):
+        main(options['format'], options['verbosity'])
```
jrwdunham, line 754: Note: to run this command and get the SVG on my host, I had to do the following. If this could be done more easily, please advise.

One-time dependency installation:

    $ docker-compose exec --user=root archivematica-dashboard bash -c "apt-get update && apt-get install --yes graphviz libgraphviz-dev"
    $ docker-compose exec --user=root archivematica-dashboard pip install pygraphviz

Run the `graph_links` command and copy the SVG file from container to host:    

    $ docker-compose exec archivematica-dashboard /src/dashboard/src/manage.py graph_links
    /var/archivematica/sharedDirectory/tmp/chainlinks-2018-03-06.svg
    $ docker-compose ps -q archivematica-dashboard
    88cf70f6d27b68be4d7b84018059b4c93b6384c5d76be5fa73c72c3d285e466d
    $ docker cp 88cf70f6d27b68be4d7b84018059b4c93b6384c5d76be5fa73c72c3d285e466d:/var/archivematica/sharedDirectory/tmp/chainlinks-2018-03-06.svg ./

src/MCPServer/lib/linkTaskManagerUnitVariableLinkPull.py
```diff
@@ -30,11 +30,23 @@
...
 
-        # Update the unit
-        if link is not None:
-            self.jobChainLink.setExitMessage(Job.STATUS_COMPLETED_SUCCESSFULLY)
-            self.jobChainLink.jobChain.nextChainLink(link, passVar=self.jobChainLink.passVar)
+        # Determine the next link.
+        link = self.unit.getmicroServiceChainLink(
+            var.variable,
+            var.variablevalue,
+            var.defaultmicroservicechainlink_id)
```
jrwdunham: This is an error I believe. `var.defaultmicroservicechainlink_id` here should be `var.defaultmicroservicechainlink`. Otherwise MCPServer triggers `AttributeError: 'unicode' object has no attribute 'id'` when the default is returned by `getmicroServiceChainLink()` and the `id` attr is accessed on a Unicode object.

This manifests as micro-service "Determine which files to identify" stalling/hanging.

jrwdunham: Another instance where mypy and type-checking would have caught this automatically.

jrwdunham (changes requested):
As indicated in the comments, I think there is an error in linkTaskManagerUnitVariableLinkPull.py. Once I fixed this as suggested in the comment, I confirmed that I could create a transfer, send it to quarantine, de-quarantine it, and see it running as I expected.

The magic link and workflow changes here are complex and difficult to reason about. But that was already the case. I also have some other more nitpicky comments.

src/dashboard/src/main/migrations/0049_remove_magic_links.py
sevein: It was generated by Django like that, I've only added the `data_migration`.

src/dashboard/src/main/migrations/0049_remove_magic_links.py
sevein: You're right, not necessary but I thought it would be useful for the reader. I guess I could just replace it with a comment.

src/dashboard/src/components/administration/management/commands/json_links.py
sevein: Heh no idea, thanks for the tip. I wrote this a long time ago when I did not know what a generator was!

src/dashboard/src/components/administration/management/commands/graph_links.py
sevein: I actually added some notes to the module docstring but I guess you didn't see them.

> docker cp 88cf70f6d27b68be4d7b84018059b4c93b6384c5d76be5fa73c72c3d285e466d:/var/archivematica/sharedDirectory/tmp/chainlinks-2018-03-06.svg ./

Or `/tmp/am-pipeline-data/tmp/chainlinks-2018-03-06.svg` if you're using the standard am.git external volumes.

src/MCPServer/lib/linkTaskManagerUnitVariableLinkPull.py
sevein: Ok sorry for the oversight, I think I probably missed this because I was only testing the code path used in that first magic link during transfer but not quarantine. I agree, mypy could be really helpful, unless you misuse union types haha.

sevein:
@jrwdunham, give it another try please. I've run some tests and it's working. Your comments were addressed. What's funny is that we may be getting rid of quarantine at some point (related https://github.com/artefactual/archivematica/issues/598)... oh well! Maybe I should have done that in the first place and get rid of that `autoRestructureForCompliance` watched directory.

One step at a time.......

jrwdunham (approved):

jrwdunham:
All my comments have been addressed. I think this is good to merge. 

@sevein I saw your module docstring for graph_links.py explaining how to install the needed extra deps, but I thought it might be helpful to add a command there showing how to actually run the command, e.g.,:

    $ docker-compose exec archivematica-dashboard /src/dashboard/src/manage.py graph_links

PR #950, opened 2018-03-04
Stop using django-annoying

### LINKED ISSUE #949: Problem: get_object_or_None isn't idiomatic
We depend on this Django package called `django-annoying` for the only purpose of using one of its mixins called `get_object_or_None`. I've looked at the way we use it and it does not seem idiomatic and sometimes obscure.

Here's the function:

```python
def get_object_or_None(klass, *args, **kwargs):
    """
    Uses get() to return an object or None if the object does not exist.
    klass may be a Model, Manager, or QuerySet object. All other passed
    arguments and keyword arguments are used in the get() query.
    Note: Like with get(), a MultipleObjectsReturned will be raised if more than one
    object is found.
    """
    queryset = _get_queryset(klass)
    try:
        return queryset.get(*args, **kwargs)
    except queryset.model.DoesNotExist:
        return None
```

It looks like there's been some attempts to include this function in Django, e.g. https://code.djangoproject.com/ticket/11352. But it's always been rejected.

My suggestion is to stop using it which saves us from having to install the dependency.

sevein:
This closes #949.

Similar work in fpr-admin: https://github.com/artefactual/archivematica-fpr-admin/pull/72.

+6 LoC but one less dependency!

### DIFF
--- a/src/MCPClient/lib/clientScripts/storeAIP.py
+++ b/src/MCPClient/lib/clientScripts/storeAIP.py
@@ -21,7 +21,7 @@
 # @subpackage archivematicaClientScript
 # @author Joseph Perry <joseph@artefactual.com>
 from __future__ import print_function
-from annoying.functions import get_object_or_None
+
 import argparse
 import os
 import sys
@@ -125,8 +125,13 @@ def store_aip(aip_destination_uri, aip_path, sip_uuid, sip_name, sip_type):
             print('Noting DIP UUID {} related to AIP so relationship can be created when AIP is stored.'.format(uuid))
     else:
         uuid = sip_uuid
-        related_package = get_object_or_None(UnitVariable, unituuid=sip_uuid, variable='relatedPackage')
-        related_package_uuid = related_package.variablevalue if related_package is not None else None
+        try:
+            related_package = UnitVariable.objects.get(
+                unituuid=sip_uuid, variable='relatedPackage')
+        except UnitVariable.DoesNotExist:
+            pass
+        else:
+            related_package_uuid = related_package.variablevalue
 
     # If AIP is a directory, calculate size recursively
     if os.path.isdir(aip_path):
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -30,7 +30,6 @@ import django.http
 from django.conf import settings as django_settings
 
 # External dependencies, alphabetical
-from annoying.functions import get_object_or_None
 from tastypie.authentication import ApiKeyAuthentication, MultiAuthentication, SessionAuthentication
 
 # This project, alphabetical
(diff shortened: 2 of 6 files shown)


src/MCPClient/lib/clientScripts/storeAIP.py
```diff
@@ -125,8 +125,13 @@ def store_aip(aip_destination_uri, aip_path, sip_uuid, sip_name, sip_type):
             print('Noting DIP UUID {} related to AIP so relationship can be created when AIP is stored.'.format(uuid))
     else:
         uuid = sip_uuid
-        related_package = get_object_or_None(UnitVariable, unituuid=sip_uuid, variable='relatedPackage')
-        related_package_uuid = related_package.variablevalue if related_package is not None else None
+        try:
+            related_package = UnitVariable.objects.get(
+                unituuid=sip_uuid, variable='relatedPackage')
+        except UnitVariable.DoesNotExist:
+            related_package_uuid = None
```
jrwdunham: This line seems redundant given line 110. Why not just `pass` here?

src/dashboard/src/components/api/views.py
```diff
@@ -655,10 +660,8 @@ def path_metadata(request):
     path = request.GET.get('path', '') if request.method == 'GET' else request.POST.get('path', '')
 
     # Get current metadata, if any
-    file_lod = get_object_or_None(models.SIPArrange, arrange_path=path, sip_created=False)
-    if file_lod is None:
-        # Try with trailing / to see if it's a directory
-        file_lod = get_object_or_None(models.SIPArrange, arrange_path=path + '/', sip_created=False)
+    file_lod = models.SIPArrange.objects \
+        .get(arrange_path__in=(path, path + '/'), sip_created=False)
```
jrwdunham: Shouldn't there still be a `try/except` here? I mean, can't there still be an unmatching `SIPArrange`? Also, is it possible for there to be matches for both `arrange_path`s and for your new code to return the `/`-suffixed one where the original code would have returned the `/`-less one?

jrwdunham (commented):
I have a minor comment and a question. See below.

src/dashboard/src/components/api/views.py
```diff
@@ -655,33 +660,33 @@ def path_metadata(request):
...
-    file_lod = get_object_or_None(models.SIPArrange, arrange_path=path, sip_created=False)
-    if file_lod is None:
-        # Try with trailing / to see if it's a directory
-        file_lod = get_object_or_None(models.SIPArrange, arrange_path=path + '/', sip_created=False)
+    files = models.SIPArrange.objects.filter(
+        arrange_path__in=(path, path + '/'),
+        sip_created=False)
+    if not files:
+        raise django.http.Http404
+    file_lod = files.first()
```
sevein, line 668: So I don't know if this is ever going to happen but I think this pretty much mimics the behavior found before this change.

sevein:
@jrwdunham, try again please!

src/dashboard/src/components/api/views.py
jrwdunham: Right... previously attribute access on `None` would have triggered an exception and Django probably would have returned a 500 response (?)

jrwdunham:
👍

jrwdunham (approved):

PR #960, opened 2018-03-07
Dashboard: update frontend dependencies

sevein:
Update "moment" package after vulnerability report from GitHub.


jraddaoui (approved):
It looks like after the build a couple of dependencies more have been updated in both package-lock.json files. Which shouldn't be a problem ;)

sevein:
> It looks like after the build a couple of dependencies more have been updated in both package-lock.json files. Which shouldn't be a problem ;)

Thank you. Yeah, I think it should be ok because the version constraints are loose but they use the caret `^` so I guess we can assume that there are no backward incompatibility changes.

sevein:
8c046c1c in `stable/1.7.x`

PR #973, opened 2018-03-10
MCPClient: fix matching originals to manual derivatives

### LINKED ISSUE #972: Problem: manually normalized files cannot be prefixes of one another
If the user provides manually normalized files in the transfer source, then Archivematica will break (inconsistently/variably, it seems) if one of those files is a prefix of another AND normalize for preservation is run. For example, a transfer with the following structure

    ├── logs
    ├── objects
    │   ├── abc.NEF
    │   ├── abc_1.NEF
    │   └── manualNormalization
    │       └── preservation
    │           ├── abc.tif
    │           ├── abc_1.tif

will fail because the normalize.py client script will fail to recognize that abc.NEF has a manually created preservation derivative in abc.tif. See https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPClient/lib/clientScripts/normalize.py#L152-L155. 

The failure arises because normalize.py finds two matches (both abc.tif and abc_1.tif) and can't handle that so it returns `None`. This causes normalize.py to re-normalize a file (abc.NEF) that has already been normalized. And this causes problems down the road, which ultimately *can* cause the entire AIP to fail. But not always.

Refs internal https://projects.artefactual.com/issues/11996

jrwdunham:
Fixes the matching of original files to their manually created preservation derivatives in a manualNormalization/preservation/ directory. Previous behaviour would find a match and all other Files that matched by having a path that was a superfix of the original path. This would trigger an exception. This fixes the issue by returning the shortest match. It also adds print statements to better understand what has happened through the Tasks GUI.

Fixes #972 

Should resolve internal https://projects.artefactual.com/issues/11996.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -890,7 +890,12 @@ def createFileSec(directoryPath, parentDiv, baseDirectoryPath,
 
             elif use in ("preservation", "text/ocr"):
                 # Derived files should be in the original file's group
-                d = Derivation.objects.get(derived_file_id=f.uuid)
+                try:
+                    d = Derivation.objects.get(derived_file_id=f.uuid)
+                except Derivation.DoesNotExist:
+                    print('Fatal error: unable to locate a Derivation object'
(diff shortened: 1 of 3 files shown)


sevein (approved):
👏 👏 👏 👏

Issue #938, opened 2018-02-22
Problem: starting Python and importing Django in many gearman workers is inefficient and unnecessary


jrwdunham:
Profiling of Archivematica's MCPClient indicates that it takes roughly 1.25 seconds for a (Python-based) client script to be started in a new Python process and for it to import Django (and other heavy modules) and create a database connection. Given a package creation event with very many files each requiring several Pythonic client script process forks, this can result in significant inefficiencies.

In many cases this "cold startup" cost is unnecessary when the following could be done instead:

1. Python client scripts could be directly (dynamically) imported into the long-running Python (MCP Client) Gearman worker process instead of forking new Python processes for the short-lived client script.

2. If the client script must run in a separate process, then we could avoid the Django import and db connection by having the long-running MCPClient process provide to the client scripts the state (e.g., database data, Django settings, unit variables) that they need. Once done, the client script could pass back (e.g., as JSON printed to stdout) directives for the MCPClient process to mutate the database appropriately.

For evidence of the problem, see the following figure and the [AM acceptance tests issue 62](https://github.com/artefactual-labs/archivematica-acceptance-tests/issues/62#issuecomment-358220844):
![normalize-flame-graph-analysis jpg 001](https://user-images.githubusercontent.com/2504271/35063177-29cde2a2-fb7b-11e7-8d17-290cd0533d75.jpeg)

---

Dependencies:
- Parent of https://github.com/artefactual/archivematica-docs/issues/182.

marktriggs:
This issue has morphed into looking at end-to-end performance of transfers and ingests, with an eye to making the improvements Joel describes above.  We've had a lot of discussion via email, and @jhsimpson suggested moving the discussion here, so I'm going to try to give a potted history.

The crux of the proposal above was to switch MCP Client scripts from forking processes to loading them into a long-running Python instance where possible.  To check the feasibility of the approach, we (@payten, @jambun, @marktriggs) put together a quick proof-of-concept (https://github.com/hudmol/archivematica/tree/inline-mcp-client) that used `runpy` (https://docs.python.org/2/library/runpy.html) to load the client scripts, performing a bit of mangling along the way to deal with command-line arguments and capturing stdout/stderr.

Testing this with a 500 file transfer produced a surprising result: CPU time reduced in line with Joel's analysis above, but the overall duration of the transfer and ingest (in wall clock time) didn't budge.  The CPU did less work in the same amount of time.

I put together some numbers and a screencast to show what I think is happening:

[...]

PR #974, opened 2018-03-12
When creating one task per file, batch the inserts

marktriggs:
Otherwise tasks end up staggered by the time it takes to perform a DB commit, which in many cases may be longer than the task's execution time.

Yields a small performance improvement, but should also improve the effectiveness of future performance improvements in the MCP Client scripts.

### DIFF
--- a/src/MCPServer/lib/linkTaskManagerFiles.py
+++ b/src/MCPServer/lib/linkTaskManagerFiles.py
@@ -36,6 +36,7 @@ from dicts import ReplacementDict
 from main.models import StandardTaskConfig, UnitVariable
 
 from django.conf import settings as django_settings
+from django.db import transaction
 
 LOGGER = logging.getLogger('archivematica.mcp.server')
 
@@ -85,6 +86,10 @@ class linkTaskManagerFiles(LinkTaskManager):
         # Escape all values for shell
         for key, value in SIPReplacementDic.items():
             SIPReplacementDic[key] = archivematicaFunctions.escapeForCommand(value)
+
+        threads_to_run = []
+        tasks_to_log = []
+
         self.tasksLock.acquire()
         for file, fileUnit in unit.fileList.items():
             if filterFileEnd:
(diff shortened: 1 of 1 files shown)


sevein:
Why do you think that `logTaskCreatedSQL` is so slow?

marktriggs:
It all comes down to the database commits.  The inserts themselves are cheap, but each one runs in its own transaction, auto-committing after each statement.  The commits are slow because they force MySQL to flush to disk--exactly how long that takes depends on the specs of the storage system and how busy it is, but running an ingest against docker on an otherwise unloaded system I see times of around 150ms in general.

Wrapping the inserts in a transaction just batches those inserts into a single disk write, which saves all of that overhead.

Here's a little sample program that shows the difference:

  https://gist.github.com/marktriggs/d68db4a6fe76e39d693151093317839d

running against my Docker setup:

    Time to insert 1000 rows with autocommit = True: 58.905809
    Time to insert 1000 rows with autocommit = False: 0.634376

marktriggs:
Sorry, I also meant to note for posterity that this PR came out of the analysis for https://github.com/artefactual/archivematica/issues/938

sevein:
That's impressive! Do you think that it would be a good alternative to set `AUTOCOMMIT=False` (https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DATABASE-AUTOCOMMIT) in `src/MCPServer/lib/settings/common.py`? I don't understand how transactions on single statements are beneficial. Originally when the MCPServer was written I think it was using raw SQL queries, later it was updated to use the ORM but for reasons like clarity or reusability of the models.
Also FWIW connections are not shared or recycled. Django's ORM doesn't let you do that. And we also have that decorator in place to make sure that the connections are destroyed when a thread terminates - this was necessary because Django would only do that cleanup after the end of a web request and we don't have those in MCPServer so we have to do it manually. I'm not sure if this is relevant for your work though.

marktriggs:
Yeah, amazing what a difference it makes :)

I don't think turning off autocommit would necessarily be safe... the trouble is, if you don't autocommit then you have to manually do the commits yourself at the right times (and if you miss one, your updates get rolled back), so it's fiddly to retrofit.

I think the main thing is to keep this issue in mind in the interface redesign work we do for https://github.com/artefactual/archivematica/issues/938.  Right now, the architecture of MCP Client makes effective use of transactions quite difficult: if you have 10,000 files that each need a row inserted into the DB, the optimal thing to do is to open a transaction, do your 10,000 inserts, commit the transaction.  But since currently each of those 10,000 files will be processed in a separate thread (with its own DB connection), there's not that opportunity to batch the inserts like that.

Anyway, I'll move the rest of this discussion over to #938 :)

sevein:
@marktriggs, does https://github.com/artefactual/archivematica/pull/1105 deprecate this pull request?

marktriggs:
@sevein Ah, yes!  It does.  Closing this one out :)

PR #981, opened 2018-03-12
archivematicaCommon: stop accessing content attribute of exception

### LINKED ISSUE #962: Problem: we access non-existent content attribute on RequestException instance
When a POST request to the Storage Service triggers a `requests.exceptions.RequestException`, we are attempting to access a non-existent `.content` attribute on that exception, which raises another exception:  `AttributeError: 'RequestException' object has no attribute 'content'`. See https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/archivematicaCommon/lib/storageService.py#L223

Maybe the intention was to access `e.response.content`? Cf. http://docs.python-requests.org/en/master/_modules/requests/exceptions/.

jrwdunham:
Stop accessing a non-existent `.content` attribute of an exception instance in the Storage Service client.

Fixes #962

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -220,7 +220,7 @@ def copy_files(source_location, destination_location, files):
         response = _storage_api_session().post(url, json=move_files)
         response.raise_for_status()
     except requests.exceptions.RequestException as e:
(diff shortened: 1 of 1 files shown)


sevein (approved):

jrwdunham:
Cherry-picked to qa/1.x as cd08753a25ddf69480a53a8a46056a7f7aa07cbe

PR #982, opened 2018-03-12
MCPServer: fix arg passed to linkProcessingComplete

### LINKED ISSUE #980: Problem: UnitVariableLinkPull jobs are marked as failed
My cleanup in 45b52bc2d8b4fcd871a9746282c4d6f68b124fb2 introduced a problem where the jobs for UnitVariableLinkPull chain links where marked as failed.

sevein:
linkProcessingComplete takes an exit code not a job status.

Connects to https://github.com/artefactual/archivematica/issues/980.

### DIFF
--- a/src/MCPServer/lib/linkTaskManagerSetUnitVariable.py
+++ b/src/MCPServer/lib/linkTaskManagerSetUnitVariable.py
@@ -25,7 +25,7 @@ from linkTaskManager import LinkTaskManager
 
 choicesAvailableForUnits = {}
 
-from main.models import TaskConfigSetUnitVariable, Job
+from main.models import TaskConfigSetUnitVariable
 
 
 class linkTaskManagerSetUnitVariable(LinkTaskManager):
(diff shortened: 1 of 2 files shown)


ross-spencer (changes requested):
Good spot. One small point on style, rather than passing a zero literal, I wonder if we can use a variable instead, so that what we pass to `linkProcessingComplete` is a little more self-describing, e.g. https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPServer/lib/linkTaskManagerFiles.py#L48 

There might be a more idiomatic way of achieving something like this?

sevein:
> There might be a more idiomatic way of achieving something like this?

Would be better if I use the name of the argument? e.g.

    self.jobChainLink.linkProcessingComplete(exitCode=0)

Or you're thinking on something like the following?

    self.jobChainLink.linkProcessingComplete(self.EXIT_CODE_SUCCESS)

ross-spencer:
@sevein I'm used to doing the second one, but I also like this one `self.jobChainLink.linkProcessingComplete(exitCode=0)` as I think that's pretty descriptive too.

sevein:
Ok fixed!

ross-spencer (approved):

Issue #959, opened 2018-03-07
Problem: migration 0033_store_transfer_file_modification_dates failed


scollazo:
This might be a corner case, or caused by a deployment problem in the past.

This happened installing 1.7 migrations, in an AM 1.6.0 install using ansible on trusty, that was updated to 1.6.1,  and has been dist-upgraded to xenial, before deploying AM 1.7 using the ansible playbooks.

MySQL version is mysql-server-core-5.7

<pre>
root@test-161ubuntutrusty-updated-from-16withtransfers:/usr/share/archivematica/dashboard# /usr/share/python/archivematica-dashboard/bin/python manage.py migrate
Operations to perform:
  Synchronize unmigrated apps: django_forms_bootstrap, staticfiles, messages, webdesign, accounts, installer, mcp
  Apply all migrations: sessions, administration, tastypie, auth, contenttypes, fpr, longerusername, main
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
  Installing custom SQL...
Running migrations:
  Rendering model states... DONE
  Applying main.0031_job_currentstep_choices... OK
  Applying main.0032_dashboardsetting_scope... OK
  Applying administration.0007_delete_unused_models... OK
  Applying fpr.0011_mediaconch_validation... OK
  Applying fpr.0012_mediaconch_policy_checks... OK
  Applying fpr.0013_normalization_rules_mov... OK
  Applying fpr.0014_fix_fits_command... OK
  Applying fpr.0015_pronom_92...Installed 164 object(s) from 1 fixture(s)
 OK
  Applying fpr.0016_update_idtools... OK
  Applying fpr.0017_ocr_unique_names... OK
  Applying fpr.0018_slug_unique... OK
  Applying longerusername.0001_initial... OK
  Applying longerusername.0002_add_username_error_message... OK
  Applying main.0033_store_transfer_file_modification_dates...Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/share/python/archivematica-dashboard/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/share/python/archivematica-dashboard/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
[...]

PR #986, opened 2018-03-13
Dashboard: fix migration 0033, make field nullable

sevein:
If the field is not nullable and because of the lack of a default, Django
attempts to use the `0000-00-00 00:00:00` value which is considered invalid in
MySQL when using the strict mode.

This solves #959.

### DIFF
--- a/src/dashboard/src/main/migrations/0033_store_transfer_file_modification_dates.py
+++ b/src/dashboard/src/main/migrations/0033_store_transfer_file_modification_dates.py
@@ -108,7 +108,8 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='file',
             name='modificationtime',
-            field=models.DateTimeField()
+            field=models.DateTimeField(auto_now_add=True, null=True,
+                                       db_column=b'modificationTime'),
         ),
         migrations.RunPython(data_migration),
     ]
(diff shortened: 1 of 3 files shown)


sevein:
@scollazo @mamedin I'm hoping that this PR solves the problem that you shared in #959. After the upgrade, the `File.modificationTime` column should be `NULL` now that it's nullable. Let me know if it works for you.

mamedin:
Thanks @sevein , it fixes the migration-0033 issue.

The migration 0035_fix_workflow_dip_generation needs the same change:
```
field=models.DateTimeField(auto_now_add=True, null=True, db_column=b'modificationTime'))                        
```
After changing these two migrations the issue is fixed.

sevein:
Cool, thanks, I'll fix it!

sevein:
@mamedin ok looks like 0035 was altering the field, i.e. 0033 created it, 0035 improved/fixed it. I've consolidated so it's all now done from 0033. I'm hoping that's not a problem. I remember recommending @jrwdunham at some point that we shouldn't edit migrations but I also realize that qa/dev instances are ephemeral and it's very hard for us to guarantee that migrations are inmutable. I think it's better this way, but @jrwdunham I'd be know if you're ok with this change before we merge.

jrwdunham (approved):
I think this is good to merge. It seems like there is no alternative to modifying migrations here.

sevein:
76d3cbba in `stable/1.7.x`

PR #990, opened 2018-03-15
Reenables the preview capability of appraisal tab

### LINKED ISSUE #884: Problem: Appraisal tab file preview doesn't show file
When you try to preview a file in Appraisal Tab > Analysis > Preview File, instead of showing the file it just immediately gets downloaded and the file never appears in the preview pane. Tested on augustus and am17x.qa.

ross-spencer:
* Creates a new url to resolve links of type:preview
* Content disposition to inline for preview only if explicitly requested
* Resolves preview for standard transfers or those with examine contents set
* Resolves #884

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -928,22 +928,41 @@ def download_fs(request):
         raise django.http.Http404
 
 
-def download_by_uuid(request, uuid):
+def preview_by_uuid(request, uuid):
+    """Simple wrapper to the download_by_uuid function to create a route
+    through the code to enable the preview of a file in the browser, not
+    automatically download.
     """
-    Download a file from the Storage Service, given its UUID.
+    return download_by_uuid(request, uuid, preview_file=True)
 
-    This view will stream the response directly from the storage service, so, unlike download_ss, this will work even if the Storage Service is not accessible to the requestor.
 
-    Returns 404 if a file with the requested UUID cannot be found, and 400 if the storage service fails to retrieve the record.
+def download_by_uuid(request, uuid, preview_file=False):
+    """Download a file from the Storage Service, given its UUID.
+
+    This view will stream the response directly from the storage service,
+    so, unlike download_ss, this will work even if the Storage Service is
+    not accessible to the requestor.
+
+    Returns 404 if a file with the requested UUID cannot be found. Otherwise
+    the status code is returned via the call to
+    ``stream_file_from_storage_service``
+
+    ``preview_file`` is an instruction to be applied to the response headers
+    to enable the file to be seen inside the browser if it is capable of being
+    rendered. On receiving this instruction, the content-disposition header
+    will be set in the stream_file_from_storage_service to 'inline'.
     """
     try:
         f = models.File.objects.get(uuid=uuid)
     except models.File.DoesNotExist:
         response = {
             'success': False,
-            'message': _('File with UUID %(uuid)s could not be found') % {'uuid': uuid},
+            'message': _('File with UUID %(uuid)s '
+                         'could not be found') % {'uuid': uuid},
         }
         return helpers.json_response(response, status_code=404)
     relative_path = f.currentlocation.replace('%transferDirectory%', '')
-    redirect_url = storage_service.extract_file_url(f.transfer_id, relative_path)
-    return helpers.stream_file_from_storage_service(redirect_url, 'Storage service returned {}; check logs?')
+    redirect_url = storage_service.extract_file_url(f.transfer_id,
+                                                    relative_path)
+    return helpers.stream_file_from_storage_service(
+        redirect_url, 'Storage service returned {}; check logs?', preview_file)
--- a/src/dashboard/src/components/helpers.py
+++ b/src/dashboard/src/components/helpers.py
@@ -304,14 +304,27 @@ def processing_config_path():
     )
 
 
-def stream_file_from_storage_service(url, error_message='Remote URL returned {}'):
-    stream = requests.get(url, stream=True, timeout=django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT)
+def stream_file_from_storage_service(url,
+                                     error_message='Remote URL returned {}',
+                                     preview_file=False):
+    # Repetetive or constant values to use below when seeting the headers.
+    header_content_disposition = 'Content-Disposition'
+    content_disposition_inline = 'inline'
+    storage_timeout = django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT
+    stream = requests.get(url,
+                          stream=True,
+                          timeout=storage_timeout)
     if stream.status_code == 200:
         content_type = stream.headers.get('content-type', 'text/plain')
-        content_disposition = stream.headers.get('content-disposition')
         response = StreamingHttpResponse(stream, content_type=content_type)
-        if content_disposition:
-            response['Content-Disposition'] = content_disposition
+
+        # Provide a content-disposition so the browser knows how to handles
+        # the response.
+        content_disposition = stream.headers.get('content-disposition')
+        if preview_file:
+            response[header_content_disposition] = content_disposition_inline
+        elif content_disposition:
+            response[header_content_disposition] = content_disposition
         return response
     else:
         response = {
--- a/src/dashboard/frontend/appraisal-tab/app/preview/preview.controller.js
+++ b/src/dashboard/frontend/appraisal-tab/app/preview/preview.controller.js
@@ -15,7 +15,8 @@ controller('PreviewController', ['$scope', 'gettextCatalog', '$routeSegment', 'A
 
   vm.set_file_data = file => {
     $scope.file = file;
-    $scope.url = '/filesystem/' + file.id + '/download';
+    $scope.download = '/filesystem/' + file.id + '/download';
+    $scope.preview = '/filesystem/' + file.id + '/preview';
   };
 
   $scope.$routeSegment = $routeSegment;
--- a/src/dashboard/frontend/appraisal-tab/app/preview/preview.html
+++ b/src/dashboard/frontend/appraisal-tab/app/preview/preview.html
@@ -1,6 +1,6 @@
-<div class="preview-frame"><iframe ng-src="{{ url }}" name="preview-frame"></iframe></div>
+<div class="preview-frame"><iframe ng-src="{{ preview }}" name="preview-frame"></iframe></div>
 <p ng-if='file'>
   <span translate>Previewing {{ file.title }}</span>
-  <a ng-href="{{ url }}" download="filename">({{ "Download" | translate }})</a>
+  <a ng-href="{{ download }}" download="filename">({{ "Download" | translate }})</a>
 </p>
 <p ng-if='!file'>{{ "No file selected" | translate }}</p>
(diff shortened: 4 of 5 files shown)


jraddaoui (approved):
Nice! Looks great to me.

src/dashboard/src/components/helpers.py
```diff
@@ -304,14 +304,24 @@ def processing_config_path():
...
     if stream.status_code == 200:
         content_type = stream.headers.get('content-type', 'text/plain')
         content_disposition = stream.headers.get('content-disposition')
         response = StreamingHttpResponse(stream, content_type=content_type)
-        if content_disposition:
-            response['Content-Disposition'] = content_disposition
+        if preview_file:
+            response[header_content_disposition] = content_disposition_inline
+        elif content_disposition:
+            response[header_content_disposition] = content_disposition
```
sevein: Nitpick: I think the style here is rather verbose. There's a lot of this in Archivematica but you can also find more recent code that it's more concise and perhaps more Pythonic, probably introduced by Holly and Joel. E.g. you could go from 6 to 2 lines of code with:

    response['Content-Disposition'] = \ 
        'inline' if preview_file else content_disposition

I believe that this is ultimately up to you. But you're going to find that a lot of the new code follows that inline approach like the following:

https://github.com/artefactual/archivematica/blob/4e84fa7bf4392b8cecf7bce8abb6009ed71d50ea/src/archivematicaCommon/lib/bindpid.py#L307-L357

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
         raise django.http.Http404
 
 
-def download_by_uuid(request, uuid):
+def preview_by_uuid(request, uuid):
+    """Simple wrapper to the download_by_uid function to create a route
```
sevein: Typo: `s/download_by_uid/download_by_uuid`.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
+    This view will stream the response directly from the storage service,
+    so, unlike download_ss, this will work even if the Storage Service is
+    not accessible to the requestor.
+
+    Returns 404 if a file with the requested UUID cannot be found, and
+    400 if the storage service fails to retrieve the record.
+
+    preview_file is an instruction to be applied to the response headers to
+    enable the file to be seen inside the browser if it is capable of being
+    rendered.On receiving this instruction, the content-disposition header
```
sevein: Typo: `rendered.On`.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
 
 
-def download_by_uuid(request, uuid):
+def preview_by_uuid(request, uuid):
+    """Simple wrapper to the download_by_uid function to create a route
+    through the code to enable the preview of a file in the browser, not
+    automatically download.
     """
-    Download a file from the Storage Service, given its UUID.
+    return download_by_uuid(request, uuid, True)
```
sevein: Should be:

```python
return download_by_uuid(request, uuid, preview_file=True)
```

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
     relative_path = f.currentlocation.replace('%transferDirectory%', '')
-    redirect_url = storage_service.extract_file_url(f.transfer_id, relative_path)
-    return helpers.stream_file_from_storage_service(redirect_url, 'Storage service returned {}; check logs?')
+    redirect_url = storage_service.extract_file_url(f.transfer_id,
+                                                    relative_path)
+
+    response_err_message = 'Storage service returned {}; check logs?'
+    return helpers.stream_file_from_storage_service(redirect_url,
+                                                    response_err_message,
+                                                    preview_file)
```
sevein: Another nitpick about inline vs extra vars - you could write:

```python
return helpers.stream_file_from_storage_service(
    redirect_url, 'Storage service returned {}; check logs?', preview_file)
```

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
-    Returns 404 if a file with the requested UUID cannot be found, and 400 if the storage service fails to retrieve the record.
+def download_by_uuid(request, uuid, preview_file=False):
+    """Download a file from the Storage Service, given its UUID.
+
+    This view will stream the response directly from the storage service,
+    so, unlike download_ss, this will work even if the Storage Service is
+    not accessible to the requestor.
+
+    Returns 404 if a file with the requested UUID cannot be found, and
+    400 if the storage service fails to retrieve the record.
```
sevein: It doesn't seem right that we're sending a 400 code to the browser (aka "your request is broken") when SS doesn't return a 200. It's good that we're documenting this, we should probably change it at some point. Maybe a TODO note here would be wise.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
+
+    preview_file is an instruction to be applied to the response headers to
```
sevein: I'm in the process myself of getting used to a pattern that Joel introduced: using reStructuredText in docstrings and things like rst inline literals, e.g.:

    ``preview_file`` is an introduction...

src/dashboard/src/components/helpers.py
```diff
@@ -304,14 +304,24 @@ def processing_config_path():
...
-def stream_file_from_storage_service(url, error_message='Remote URL returned {}'):
-    stream = requests.get(url, stream=True, timeout=django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT)
+def stream_file_from_storage_service(url,
+                                     error_message='Remote URL returned {}',
+                                     preview_file=False):
+    # Repetetive or constant values to use below when seeting the headers.
+    header_content_disposition = 'Content-Disposition'
+    content_disposition_inline = 'inline'
+    storage_timeout = django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT
+    stream = requests.get(url,
```
sevein, line 314: A note for the future about this function (not for stable): it looks like the caller expects at all times that the func returns a valid response. If this is really the case I think that we should run get() inside a `try/except` combined with `raise_for_status()`. This way we could handle both `ConnectionError` (connectivity issues) and `HTTPError` (status_code isn't 2xx).

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
+    will be set in the stream_file_from_storage_service to 'inline'.
     """
     try:
         f = models.File.objects.get(uuid=uuid)
     except models.File.DoesNotExist:
         response = {
             'success': False,
-            'message': _('File with UUID %(uuid)s could not be found') % {'uuid': uuid},
+            'message': _('File with UUID %(uuid)s '
+                         'could not be found') % {'uuid': uuid},
```
sevein, line 961: Good idea!

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -928,22 +928,44 @@ def download_fs(request):
...
         }
         return helpers.json_response(response, status_code=404)
+
```
sevein: I realize you like blank lines, me too actually. Something to keep in mind is the following statement in PEP8 about blank lines:

    Use blank lines in functions, sparingly, to indicate logical sections.

I think that we're a team with different styles. Joel's code is denser inside functions. I'm probably not very constant. I often thing that it would be nice to have a convention and a way to force it, even if that was something like "zero blank lines inside functions".

sevein (changes requested):
@ross-spencer, I like what you did here. I've added a bunch of annoying comments about code style, take it with a grain of salt.

sevein (changes requested):
The commit subject line doesn't use the imperative form (https://chris.beams.io/posts/git-commit/#imperative).

> If applied, this commit will "**Reenables the preview capability of appraisal tab**"

src/dashboard/src/components/helpers.py
ross-spencer: I think for an `if` the compound style goes somewhat against the readability of Python, we chatted before and PEP8 seems to be a bit in between on it. https://www.python.org/dev/peps/pep-0008/#other-recommendations I will try and keep it in mind where I can become more compact elsewhere for sure.

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: Good spot!

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: Gah!! Again good spot :)

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: Ahah, this is good. Getting better at PEP8/Flake8 fixes like this.

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: I was about to log this in a Github issue, rather than a `todo` but I am looking at the response, I can't see the 400 - what do you think? The docstring is wrong?

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: I think this is a good idea. I'll try and remember it when adding snippets like this to comments. Q. Do our docstrings render anywhere? These can be easier to reinforce (like GoDocs) when there's something tangible to look at.

src/dashboard/src/components/helpers.py
ross-spencer: Seems to be more idiomatic to our code base the way you suggest for sure.

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: I think I have them when writing or working out the function for the first time, I make my own logical sections. But having just removed these, it still looks fine to me in Sublime, so something I'll try and keep in mind. Happy to find the middle ground for sure.

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: :+1:

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: Have logged a ticket here. https://github.com/artefactual/archivematica/issues/994 Will update the docstring to be more specific about how it's returning here. It's a little ambiguous in that it actually returns whatever stream from storage service returns - I'm guessing from a previous refactor.

ross-spencer:
Thanks @sevein. It's all appreciated. I've made a handful of changes. The commit message should be fixed too.

sevein (approved):

ross-spencer:
7fb96ee8a143a8c4ef24a64c061873793ff67309 in `stable/1.7.x`

Issue #874, opened 2018-01-08
Problem: "xml" file appearing in normalization reports


jraddaoui:
The normalization report table shows all the files related with the SIP in the database. That includes, in some cases, the `pointer.xml` file generated after the normalization process. Moreover, the filename obtained from the database and displayed in the table doesn't have the first 23 chars (the SIP name and the objects directory) but, because this file is outside the objects directory, it appears in the table as "xml", causing even more confusion.

jraddaoui:
Two possible solutions:

- Remove only the SIP part of the filename. It will make clear it's referencing to the pointer file appearing as "pointer.xml" instead of "xml". It will make all the others show the "objects/" part before the filename. This could be done by directly showing the location instead of the filename in:

  
  https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/dashboard/src/templates/ingest/normalization_report.html#L31

- Show only the SIP files where a normalization task was executed (the actual normalization may not be attempted in the end, but the task is executed for all the transfer files). This could be done by adding the following clause to the query:
    ```
    AND (c.taskUUID is not NULL OR b.taskUUID is not NULL)
    ``` 
    After:

  
  https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/dashboard/src/components/ingest/views_NormalizationReport.py#L152

jraddaoui:
Hi @jhsimpson,

I'd like to know what do you think before going forward. This comes from https://projects.artefactual.com/issues/11631, so it may be needed in stable/1.6.x too.

jraddaoui:
Chatted with @jhsimpson a while ago and decided that we should change how the pointer file is added to the database and use `submissionDocumentation` instead of the default (`original`) as its `use` in:

https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/MCPClient/lib/clientScripts/createPointerFile.py#L217

However, this shouldn't be an issue in stable/1.7.x as the pointer file creation has been moved to the SS; check 665470ec50a40af31e6e6e5aacbf4e764a4b6bd3. But we should create a migration to change the `use` of existing pointer files to fix old reports and change how the pointer file is created in stable/1.6.x.

jraddaoui:
In order to test the migration you'll need to:

First, in a 1.6 instance of AM, create one or more AIPs and check the normalization report(s) looking for a failing file called 'xml'. I believe this entry doesn't show-up until the AIP is stored (and the pointer file created) so it won't be present when approving normalization.

[...]

PR #992, opened 2018-03-15
Change filegrpuse of SIP pointer files

jraddaoui:
In AM 1.6, the pointer file was being created in the MCPClient and added
to the SIP files with the default filegrpuse: 'original'. This was making
this file to appear, named 'xml', in the normalization reports as an
error. This file should have been added to the SIP files with
'submissionDocumentation' as its filegrpuse and changing it in this
migration will make them disappear from existing reports.

Refs #874.

### DIFF
--- a/src/dashboard/src/main/migrations/0049_change_pointer_file_filegrpuse.py
+++ b/src/dashboard/src/main/migrations/0049_change_pointer_file_filegrpuse.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def data_migration_up(apps, schema_editor):
+    """
+    Change filegrpuse of pointer files:
+
+    In AM 1.6, the pointer file was being created in the MCPClient and added
+    to the SIP files with the default filegrpuse: 'original'. This was making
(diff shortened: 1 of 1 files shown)


jraddaoui:
I'll cherry-pick to qa/1.x with different filename and dependencies for the migration if there are more migrations in that branch. I'm not completely sure that's needed in Django migrations, but I think it's clearer that way. Please, let me know if you have any objections.

sevein (approved):
This looks good to me and I like that you implemented the downgrade path too, I find that useful when I'm testing my own migrations.

I would suggest to wait for a second opinion from @jrwdunham because he knows that model very well! But if you're confident then go ahead and merge!

jraddaoui:
Thanks @sevein! @jrwdunham, I'm thinking now, can `currentlocatiom` change after the pointer file creation?

jrwdunham:
@jraddaoui I don't think the `currentlocation` of a pointer file can change after its creation. I ran a transfer through to storage and then reingested it and the pointer file's current location in the db seemed to stay constant. I think this is good to merge.

jraddaoui:
Thanks! 1a4d48d in qa/1.x

PR #995, opened 2018-03-16
MediaConch: update to 17.12 and reduce verbosity

jrwdunham:
Fixes #966 

- Makes validateFile print less to stdout/stderr so the Tasks GUI is less cluttered.
- Updates MCPClient to use MediaConch v. 17.12 so that .mkv implementation checks to do not hang indefinitely on certain files.
- Modifies the MCPClient Dockerfile base image (required by above step).

- Related PRs:
  - [AM-fpr-admin PR 75](https://github.com/artefactual/archivematica-fpr-admin/pull/75) which adds migrations to reduce update the MediaConch version and reduced the verbosity of MediaConch-based validation and policy check micro-services.
  - [AM-acceptance-tests PR 77](https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/77) which adds a feature that confirms that the changes in this PR fix #966.

PR #1004, opened 2018-03-20
Stable 1.7: MediaConch: update to 17.12 and reduce verbosity

jrwdunham:
This PR is functionally equivalent to https://github.com/artefactual/archivematica/pull/995 but because qa/1.x and stable/1.7.x have diverged in areas relevant to that PR, I am opening a new one instead of cherry-picking a mixture of qa/1.7.x commits and adding new ones. Better safe than sorry.


sevein (approved):
LGTM

PR #1012, opened 2018-03-23
De-duplicate checksum verification

### LINKED ISSUE #918: Problem: checksum verification happens twice
**Goal**: improve the performance of Archivematica in processing large transfers (many files and/or large files).

**Context**: AM's global `checksum_type` setting can be set via the GUI at administration/general/ where the options are MD5, SHA-1, SHA-256, and SHA-512. On a standard AM install, the default value appears to be SHA-256.

## Overview: relevant micro-services in order of occurrence

1. Checksums are calculated near the start of Transfer and stored in the db.
2. Checksums are verified (re-calculated and compared to db values) near the end of ingest.
3. Bag is created, near the end of ingest, which involves calculating checksums and storing them in manifest-<ALGORITHM>.txt.
4. Bag is verified, near the end of ingest, which involves re-calculating checksums and confirming they match what is in manifest-<ALGORITHM>.txt.

### "Assign checksums and file sizes to objects" (Transfer)

- A.k.a. `updateSizeAndChecksum_v0.0` or `archivematicaUpdateSizeAndChecksum.py`.
- Creates a new gearman worker task for each file which:
  - calculates the checksum of each file and stores it in the `Files` table in the database; and
  - uses the checksum algorithm specified in the `checksum_type`-named row in the `DashboardSettings` table (which **defaults to `'sha256'`**).

### "Verify checksums generated on ingest" (Ingest)

- A.k.a. `verifyPREMISChecksums_v0.0` or `verifyPREMISChecksums.py`.
- Creates a new gearman worker task for each file which:
   - fetches the file's checksum (and type/algorithm) from the db;
   - re-calculates the file's checksum (with the given algorithm);
   - verifies that the db checksum and the just-calculated one match;
   - creates a `'fixity check'` type event in the database to document that the checksum of the file made early on in transfer has not changed by the end of ingest.

### "Prepare AIP" (Ingest)

- A.k.a. `bagit_v0.0` or `archivematicaBagWithEmptyDirectories.py`.
- Calls `bag create` which "creates a bag from supplied files/directories, completes the bag, and then writes in a specified format."
- The checksum algorithm passed to `bag` is taken from the `checksum_type`-named row in the
`DashboardSettings` table.
- **`'sha512'` is the default algorithm if there is no `checksum_type`-named row in `DashboardSettings` table.**

### "Verify AIP" (Ingest)

- A.k.a. `verifyAIP_v0.0` or `verifyAIP.py`.
- Micro-service uses the `bag` (BagIt) CLI and calls `bag verifypayloadmanifests`, which re-calculates checksums for all files in the AIP/bag and verifies that they match what is documented in  `manifest-<ALGORITHM>.txt`, e.g., `manifest-sha256.txt`.

## Proposed Solution

1. Remove the "Verify checksums generated on ingest" micro-service.
2. Enhance "Verify AIP" to bulk query the db for Transfer-generated checksums and then verify that they match what is documented in the `bag`-generated `manifest-<ALGORITHM>.txt`.

## Problem with Proposed Solution

The "Verify checksums generated on ingest" micro-service creates events in the db that must end up in the AIP METS. However, the "Verify AIP" micro-service necessarily occurs after the "Generate METS.xml document".

## Solution 1 (preferred) to the Problem with Proposed Solution

1. Continue to do (1) and (2) of the proposed solution above.
2. Have "Verify AIP" create an AIP-level `"fixity check"` PREMIS:EVENT that it can pass to the Storage Service, which will document this verification (fixity check of the AIP as a whole) in the **pointer file**.

Problems with this:

1. Uncompressed AIPs have no pointer files. Thus uncompressed AIPs would have not AIP-level fixity check events, unless we also begin creating pointer files for uncompressed AIPs, which is motivated here but also by the fact that encrypted uncompressed AIPs need pointer files and replicated uncompressed AIPs need pointer files.

## Solution 2 to the Problem with Proposed Solution

Instead of removing the "Verify checksums generated on ingest" micro-service, convert it to a once-per-unit type micro-service which simply optimistically creates the `'fixity check'` PREMIS events in the database and specifies `bag` as the tool used. Then, if "Verify AIP" ultimately fails, the  AIP as a whole has failed so the inaccurate METS is not an issue.

Problems with this:

1. It will result in the same number of `'fixity check'` PREMIS:EVENTs in the METS file as there are currently. It would be good to reduce this number, if possible, so that the METS file is not so huge for large transfers and, correspondingly, so that the time needed to write the METS file is decreased.
2. It seems like a bad idea to "optimistically" create PREMIS:EVENTs that are later invalidated and document them in a METS file. (Though we could invalidate them in the database post hoc, if necessary.)

jrwdunham:
Fixes #918 

In short, removes the once-per-file verifyPREMISChecksums_v0.0 micro-service and replaces it with once-per-SIP equivalent functionality in a new verifyAIP_v1.0 client script (which supersedes verifyAIP_v0.0). **Question:** I have deleted verifyAIP_v0.0 because other superseded client scripts appear to have been deleted. Should it be retained?

The new verifyAIP_v1.0 parses the BagIt manifest file for the checksums that `bag` has just calculated instead of needlessly re-calculating them.

The functional consequences of this change are as follows.

1. AIP METS files will no longer have `fixity check` PREMIS events for each file.
2. The AIP pointer file will now contain a single `fixity check` PREMIS event for the entire AIP:

       <mets:mdWrap MDTYPE="PREMIS:EVENT">
         <mets:xmlData>
           <premis:event xmlns:premis="info:lc/xmlns/premis-v2" xsi:schemaLocation="info:lc/xmlns/premis-v2 http://www.loc.gov/standards/premis/v2/premis-v2-2.xsd" version="2.2">
             <premis:eventIdentifier>
               <premis:eventIdentifierType>UUID</premis:eventIdentifierType>
               <premis:eventIdentifierValue>bdf5c772-1923-4241-a88b-1f949ab542c2</premis:eventIdentifierValue>
             </premis:eventIdentifier>
             <premis:eventType>fixity check</premis:eventType>
             <premis:eventDateTime>2018-03-23T16:57:37.602859+00:00</premis:eventDateTime>
             <premis:eventDetail>program="python, bag"; module="hashlib.sha256()"</premis:eventDetail>
             <premis:eventOutcomeInformation>
               <premis:eventOutcome>Pass</premis:eventOutcome>
               <premis:eventOutcomeDetail>
                 <premis:eventOutcomeDetailNote>
                   All 3 checksums generated at start of transfer match those generated by BagIt (bag).
                 </premis:eventOutcomeDetailNote>
               </premis:eventOutcomeDetail>
             </premis:eventOutcomeInformation>
             <premis:linkingAgentIdentifier>
               <premis:linkingAgentIdentifierType>preservation system</premis:linkingAgentIdentifierType>
               <premis:linkingAgentIdentifierValue>Archivematica-1.7</premis:linkingAgentIdentifierValue>
             </premis:linkingAgentIdentifier>
             <premis:linkingAgentIdentifier>
               <premis:linkingAgentIdentifierType>repository code</premis:linkingAgentIdentifierType>
               <premis:linkingAgentIdentifierValue>test</premis:linkingAgentIdentifierValue>
             </premis:linkingAgentIdentifier>
           </premis:event>
         </mets:xmlData>
       </mets:mdWrap>

3. AIPs without pointer files will have no record of the SIP/AIP-level fixity check. [This was determined to be an acceptable trade-off](https://github.com/artefactual/archivematica/issues/918#issuecomment-364611999) and should be remedied once we [begin creating pointer files for uncompressed AIPs] (artefactual/archivematica-storage-service#324), which is motivated by other considerations as well.


src/MCPClient/lib/archivematicaClientModules
```diff
@@ -130,11 +130,10 @@ trimVerifyChecksums_v0.0 = %clientScriptsDirectory%trimVerifyChecksums.py
 trimVerifyManifest_v0.0 = %clientScriptsDirectory%trimVerifyManifest.py
 updateSizeAndChecksum_v0.0 = %clientScriptsDirectory%archivematicaUpdateSizeAndChecksum.py
 validateFile_v1.0 = %clientScriptsDirectory%validateFile.py
-verifyAIP_v0.0 = %clientScriptsDirectory%verifyAIP.py
+verifyAIP_v1.0 = %clientScriptsDirectory%verify_AIP.py
```
sevein, line 133: https://www.python.org/dev/peps/pep-0008/#package-and-module-names
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

src/MCPClient/lib/clientScripts/verify_AIP.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env python2
+from __future__ import print_function
+import os
+from pprint import pformat
+import shutil
+import sys
+
+import django
+django.setup()
```
sevein: Can you move this call inside the `if __name__ == '__main__':` block? I find it's more explicit and clearer. If there was a case where someone is using this module then the caller should be responsible to initialize Django. Probably not very important, but I've been doing it this way recently.

src/MCPClient/lib/clientScripts/verify_AIP.py
```diff
@@ -0,0 +1,244 @@
...
+
+from main.models import File
+
+
+class VerifyChecksumsError(Exception):
+    """Checksum verification has failed."""
+
+
+def extract_aip(aip_path, extract_path):
+    os.makedirs(extract_path)
```
sevein: The default mode is ok? Just checking.

src/MCPClient/lib/clientScripts/verify_AIP.py
```diff
@@ -0,0 +1,244 @@
...
+    them here was redundant.
+
+    sys.argv[1] = UUID
+      UUID of the SIP, which will become the UUID of the AIP
+    sys.argv[2] = current location
+      Full absolute path to the AIP's current location on the local filesystem
+    """
+
+    sip_uuid = sys.argv[1]  # %sip_uuid%
+    aip_path = sys.argv[2]  # SIPDirectory%%sip_name%-%sip_uuid%.7z
```
sevein: Nice way to put it with the examples, that's useful :)

src/MCPClient/lib/clientScripts/verify_AIP.py
```diff
@@ -0,0 +1,244 @@
...
+    return_code = 0
+    for command in verification_commands:
+        print("Running test: ", command)
+        exit_code, _, _ = executeOrRun("command", command, printing=True)
+        if exit_code != 0:
+            print("Failed test: ", command, file=sys.stderr)
+            return_code = 1
+
+    if return_code == 0:
+        return_code = verify_checksums(bag, sip_uuid)
```
sevein: One could argue that it's not `verify_checksums`'s concern to know about exit codes - it'd be more pythonic to have it raise an exception right? Feel free to ignore this. Just a random thought, I've done the same so many times.

src/MCPClient/lib/clientScripts/verify_AIP.py
```diff
@@ -0,0 +1,244 @@
...
+    else:
+        print('Not verifying checksums because other tests have already'
+              ' failed.')
+
+    # cleanup
+    if not is_uncompressed_aip:
+        shutil.rmtree(extract_dir)
```
sevein: Do we want to make the script report an error if the clean up did not succeed? You may want to trap `OSError` if that's not the case.

sevein (changes requested):

src/MCPClient/lib/clientScripts/verify_AIP.py
jrwdunham: ok

src/MCPClient/lib/clientScripts/verify_AIP.py
jrwdunham: Seems to be ok. Extraction works

src/MCPClient/lib/clientScripts/verify_AIP.py
jrwdunham: I didn't do it, but I'll take the compliment

src/MCPClient/lib/clientScripts/verify_AIP.py
jrwdunham: Yeah, I think you're right. I'll change.

src/MCPClient/lib/clientScripts/verify_AIP.py
jrwdunham: sounds good

sevein:
Sorry I didn't realize that it wasn't really your code what I was looking at!

sevein (approved):

Issue #1026, opened 2018-03-27
Problem: rebuild-elasticsearch-aip-index-from-files is only accessible via devtools


sevein:
This tool belongs to devtools but we've moving it to this repo.

Related issues:
- https://github.com/artefactual/archivematica-devtools/issues/43
- https://github.com/artefactual/archivematica-docs/issues/105
- ~https://github.com/artefactual/archivematica-devtools/pull/50~

The original one is here: https://github.com/artefactual/archivematica-devtools/blob/master/tools/rebuild-elasticsearch-aip-index-from-files.

PR #1028, opened 2018-03-28
Dashboard: add cmd rebuild-elasticsearch-aip-index-from-files

sevein:
Import `rebuild-elasticsearch-aip-index-from-files` from devtools.

See https://github.com/artefactual/archivematica/issues/1026 for more details.

---

@jhsimpson I hope this works for you. Not sure what's the best way to test this, I don't have access to big datasets. @scollazo maybe you can test this during your integration tests? Perhaps you can revive the case that brought you to file https://github.com/artefactual/archivematica-devtools/pull/50 and let me know if this works for you.

### DIFF
--- a/src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py
+++ b/src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py
@@ -1,0 +1,287 @@
+"""Recreate the Elasticsearch index from AIPs stored on disk.
+
+This is a copy of: https://github.com/artefactual/archivematica-devtools/blob/1fd49faf2b415a4f4229da832445d22b35c262f4/tools/rebuild-elasticsearch-aip-index-from-files.
+
+This is useful if the Elasticsearch index has been deleted or damaged, but you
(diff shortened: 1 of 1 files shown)


src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py
```diff
@@ -0,0 +1,287 @@
+"""Recreate the Elasticsearch index from AIPs stored on disk.
+
+This is a copy of: https://git.io/vxadg.
```
jhsimpson: The shortened url here doesn't work for me.  I suggest putting in the full url, rather than use a shortener.  I am guessing you are trying to link to something like https://github.com/artefactual/archivematica-devtools/blob/master/tools/rebuild-elasticsearch-aip-index-from-files

sevein: Fixed.

jrwdunham (approved):

sevein:
0a6dbb20 in qa/1.x - not tested though

PR #1049, opened 2018-04-08
storageService.create_file: raise HTTP errors, refactor

### LINKED ISSUE #815: Problem: failure in "Store the AIP" not reflected in the GUI
When the storeAIP_v0.0 micro-service runs and makes an API call to the SS to store the AIP and the SS returns a bad status code, the micro-service gives no indication to the user that the AIP storage has failed. Often the AIP is in fact stored but it may not have been stored correctly and completely (e.g., a pointer file may not have been created). AIP storage should probably occur in a transaction of sorts and should be rolled back if not completed fully and correctly (which is an issue for the SS).

Related to https://github.com/artefactual/archivematica/issues/809.

hakamine:
Fixes #815 
Note that this only fixes the issue of the "store AIP" job/micro-service not failing if getting an error from the SS (probably also need to roll back changes in the SS database in another PR)

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -246,16 +246,21 @@ def create_file(uuid, origin_location, origin_path, current_location,
                 current_path, package_type, size, update=False,
                 related_package_uuid=None, events=None, agents=None,
                 aip_subtype=None):
-    """Creates a new file. Returns a tuple of (resulting dict, None) on
-    success, (None, error) on failure. Note: for backwards compatibility
+    """Creates a new file. Note: for backwards compatibility
     reasons, the SS API calls "packages" "files" and this function should be
     read as ``create_package``.
 
     origin_location and current_location should be URIs for the storage service.
+
+    Returns:
+        Dict with the JSON response from the SS API
+
+    Raises:
+        RequestException: if the SS API call fails
     """
     pipeline = _get_pipeline(get_setting('dashboard_uuid'))
     if pipeline is None:
-        return (None, 'Pipeline not available, see logs.')
+        raise ResourceNotFound('Pipeline not available')
     if events is None:
         events = []
     if agents is None:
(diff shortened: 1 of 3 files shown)


src/archivematicaCommon/lib/storageService.py
```diff
@@ -285,11 +285,11 @@ def create_file(uuid, origin_location, origin_path, current_location,
         else:
             url = _storage_service_url() + 'file/'
             response = session.post(url, json=new_file)
-    except requests.exceptions.RequestException as e:
+        # raise an HTTPError exception if getting HTTP 4xx or 5xx error
+        response.raise_for_status()
+    except (requests.exceptions.RequestException, requests.exceptions.HTTPError) as e:
```
sevein: `requests.RequestException` is the base class in Request library so we're already catching all the subclasses too. See https://github.com/requests/requests/blob/master/requests/exceptions.py.

sevein: `requests.RequestException` is the base class in Request library so we're already catching all the subclasses too. See https://github.com/requests/requests/blob/master/requests/exceptions.py.

src/archivematicaCommon/lib/storageService.py
```diff
@@ -285,11 +285,11 @@ def create_file(uuid, origin_location, origin_path, current_location,
         else:
             url = _storage_service_url() + 'file/'
             response = session.post(url, json=new_file)
-    except requests.exceptions.RequestException as e:
+        # raise an HTTPError exception if getting HTTP 4xx or 5xx error
+        response.raise_for_status()
```
sevein: 👍

src/archivematicaCommon/lib/storageService.py
```diff
@@ -285,11 +285,11 @@ def create_file(uuid, origin_location, origin_path, current_location,
...
+    except (requests.exceptions.RequestException, requests.exceptions.HTTPError) as e:
         LOGGER.warning("Unable to create file from %s because %s", new_file, e)
         return (None, e)
```
sevein: FYI I think that at some point we should change the way this function returns - it happens with other functions in this module. We shouldn't be returning the error, the idiomatic way in Python is to return the expected value or raise an exception instead. Pretty much like requests is doing too! An exception can carry information about the error too. To be clear, I'm not asking to change it - just saying! :)

sevein (changes requested):
Thanks for doing this, @hakamine.

hakamine:
Thank you @sevein. As suggested, create_file has been refactored to return the expected value or raise an exception instead (also parts of the code that call this function have been fixed accordingly). Did a quick test storing to aipstore and backlog, and it seems working fine.

sevein:
@hakamine thanks, I've added a couple of fixes: I changed the signature of a return we left behind in the create_file function and an extra commit that it's more of a cosmetic change where we avoid the `else` and its nesting if the `except` is a catch-all.

hakamine:
Thank you @sevein!  Would it be ok if I squash and merge?

sevein (approved):

PR #1054, opened 2018-04-12
WIP: Dashboard: fix review_matches to use access derivatives

### LINKED ISSUE #1034: Problem: archivespaceids.csv attempts to match derivatives to originals
In the context of Archivematica/ArchivesSpace integration, a user-supplied archivesspaceids.csv file encodes how to match user-supplied **access derivatives** to ArchivesSpace resource IDs. However, when AM attempts to upload a DIP to ArchivesSpace, it parses the METS file for **originals** and attempts to match those original UUIDs to ArchivesSpace resource IDs but it fails (raises a `KeyError` exception) because there are only access derivative UUIDs to match against.

Archivematica users at RAC confirm that they are looking to match individual access files from a DIP to an archival_object in ArchivesSpace and believe that this feature has worked for them in the past.

This issue could be fixed by modifying the "upload DIP" logic to parse the METS file for access derivatives (instead of originals).

jrwdunham:
Was relevant to #1034

### DIFF
--- a/src/dashboard/src/components/ingest/pair_matcher.py
+++ b/src/dashboard/src/components/ingest/pair_matcher.py
@@ -146,16 +146,34 @@ def match_dip_objects_to_resource_component_levels(client, request, resource_com
 
     return render(request, match_template, locals())
 
+def remove_prefix(string, prefix):
+    if string.startswith(prefix):
+        return string[len(prefix):]
+    return string
+
+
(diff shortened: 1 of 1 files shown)


jrwdunham:
Closing this in favour of https://github.com/artefactual/archivematica/pull/1153; the fix will be present in the 1.7.2 release.

PR #1056, opened 2018-04-17
Restore SS's file status check

sevein:
This commit restores a check we used to make on the returned value of `storage_service.create_file` (before 833bbc97b2b65f77335cc514b1f7c321533a21ed) because it's possible that the `status` equals to `FAIL` and if that's the case we should also have the client script fail.

This is in addition to #1049.
It relates to #815.


hakamine (approved):
👍 Thank you @sevein for restoring the check. Looks good, successfully quick-tested two cases (store AIP success, store AIP failure with a http 500 error). In order to test a 'status' == 'FAIL' condition I guess we would need to add a unit test, but I believe this is out of scope.

PR #1069, opened 2018-04-27
MCPClient: ignore deleted derivatives in checksum assignment

### LINKED ISSUE #1068: Problem: double full re-ingest breaks at "Assign checksums and file sizes to objects"
To re-create:

1. Create an AIP, normalizing for preservation.
2. Do a full re-ingest on that AIP.
3. Do another full re-ingest on that AIP.
4. Expect the "Assign checksums and file sizes to objects" micro-service to fail.

The cause seems to be that "Assign checksums and file sizes to objects" is finding `USE="deleted"` preservation derivatives in the METS file and is trying to create `Derivation` model instances in the db referencing those derivatives, which raises a Django `IntegrityError`. See
https://github.com/artefactual/archivematica/pull/1052/files#diff-7306ab3f3cc47a171581d7e2e512ba3bR92

jrwdunham:
Modifies archivematicaUpdateSizeAndChecksum.py so that deleted derivatives are not mapped to Derivative model instances, thus avoiding the Django ORM `IntegrityError` exception bug of #1068.

Fixes #1068

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
@@ -89,7 +89,11 @@ def get_file_info_from_mets(shared_path, file_):
             relationship.related_event_identifier_value)
         if (not event) or (event.type != 'normalization'):
             continue
-        related_object_uuid = relationship.related_object_identifier_value
+        rel_obj_uuid = relationship.related_object_identifier_value
+        related_object_fsentry = mets.get_file(file_uuid=rel_obj_uuid)
+        if getattr(related_object_fsentry, 'use', None) != 'preservation':
+            continue
+        related_object_uuid = rel_obj_uuid
         break
 
     premis_object_doc = [


jrwdunham:
I tested that this fix works by creating an AIP from the Images/ sampledata directory and double full re-ingesting it both with this fix and without it. With this fix, the second full re-ingest did not fail. Without this fix, the second full re-ingest failed at "Assign checksums and file sizes to objects", as expected given #1068.

src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
```diff
@@ -89,7 +89,11 @@ def get_file_info_from_mets(shared_path, file_):
             relationship.related_event_identifier_value)
         if (not event) or (event.type != 'normalization'):
             continue
-        related_object_uuid = relationship.related_object_identifier_value
+        rel_obj_uuid = relationship.related_object_identifier_value
+        related_object_fsentry = mets.get_file(file_uuid=rel_obj_uuid)
+        if related_object_fsentry.use != 'preservation':
```
jhsimpson: I tested and this throws an error.

I think it is because the call to mets.get_file() is returning None where the file belongs to a fileGrp with use='deleted'.  Those files do not have a mets:FLocat element, maybe that is why get_file() returns None?

I changed this to the code here and it worked:
```
if related_object_fsentry is None:
            continue  # no fsentry for 'deleted' files 
```

Note sure if that is the best solution.

jhsimpson (changes requested):

src/MCPClient/lib/clientScripts/archivematicaUpdateSizeAndChecksum.py
jrwdunham: @jhsimpson I think your solution probably works, but just in case it's possible for an original to be linked to something as a derivative that DOES have an "fsentry" in the METS but which does not have use="preservation", I've done the following instead (which will avoid the `AttributeError`):

    >>> if getattr(related_object_fsentry, 'use', None) != 'preservation':
    ...     continue

jrwdunham:
@jhsimpson @sevein I have modified this PR as detailed in the outdated conversation with Justin. I have tested that a double full re-ingest on Images/pictures/ (with normalization for preservation) fails without this fix and passes with it.

In the process I also discovered the bug at https://github.com/artefactual/archivematica/issues/1072. I believe issue 1072 is too edge case to be worthy of the 1.7 release at this stage.

However, I have updated this PR to target the 1.7 release, given our most recent discussion.

sevein (approved):

jrwdunham:
Cherry-picked to qa/1.x as 9c080be19bfe221efd3faa7ea4019bcac05576c2

PR #1075, opened 2018-05-02
MCPClient: refactor archivematicaClient.py

jrwdunham:
Adds docstrings, more Pythonic formatting, and some cosmetic structural changes to archivematicaClient.py

Note: these same changes have already been approved and merged into qa/1.x. See https://github.com/artefactual/archivematica/pull/829.

Issue #1074, opened 2018-05-02
Problem: archivematicaClient.py is difficult to read


jrwdunham:
There are a lot of un-Pythonic conventions being used in this core module. Because it is quite an isolated piece of code, it would be relatively easy to refactor it in order to make it more readable.

jrwdunham:
Addressed by https://github.com/artefactual/archivematica/pull/829/

PR #1077, opened 2018-05-02
MCPClient: remove capture_output kwarg

### LINKED ISSUE #1076: Problem: archivematicaClient.py passes unexpected capture_output kwarg
Commit f527d53a3b61164a327038084b007beb2f9240a7 added this call:

    executeOrRun('command', script, stdIn='', printing=True, capture_output=capture_output)

The problem is that `executeOrRun` does not recognize the `capture_output` kwarg, which is currently only available in dev/1.8.x.

jrwdunham:
Fixes #1076

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -1,8 +1,34 @@
 #!/usr/bin/env python2
 
+"""Archivematica Client (Gearman Worker)
+
+This executable does the following.
+
+1. Loads tasks from config. Loads a list of performable tasks (client scripts)
+   from a config file (typically that in lib/archivematicaClientModules) and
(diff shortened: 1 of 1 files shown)


sevein:
Is this PR trying to solve the problem being introduced by #1075 which we haven't merged yet?

jhsimpson:
@sevein I think:
Issue #1074 is fixed by PR #1075 
This PR includes the commit from #1075, so it looks like this is trying to fix both, but if 1075 is merged first, then this PR will only have 1 new commit, which fixes issue #1076 
#1078 contains both of these commits, if it is merged last out of these 3 PR's, it will introduce 2 new commits, to fix #763.

sevein (approved):

jrwdunham:
Closing this because the revised https://github.com/artefactual/archivematica/pull/1078 avoids the creation of the issue that this targeted.

PR #1084, opened 2018-05-10
Reingest: fix AttributeError MDRef

### LINKED ISSUE #1083: Reingesting DSpace AIPs errors at Generate METS
#### How to reproduce

1. Create DSpace transfer, e.g. ITEM@2429-2701 from sampledata
2. Full re-ingest

#### Results

image

image

sevein:
metsrw assumes that the METS is self-contained, MDRef pointers are not
followed. It would be up to `archivematicaCreateMETSReingest.py` to do so but
instead this commit ignores them. We're doing this because any rights updates
performed during re-ingest would result in the AIP METS file being edited, not
the DSpace one.

Connects to https://github.com/artefactual/archivematica/issues/1083.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
@@ -158,6 +158,9 @@ def update_rights(mets, sip_uuid):
     for fsentry in original_files:
         rightsmds = [s for s in fsentry.amdsecs[0].subsections if s.subsection == 'rightsMD']
         for r in rightsmds:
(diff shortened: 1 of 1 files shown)


jrwdunham (approved):
The code looks good to me. I'm a little concerned about whether a full re-ingest on a DSpace transfer will have the correct metadata (METS) file after this change, but at least now it will work. I think addressing the above concern is something that an analyst needs to look at as part of QA for the release where this is included.

sromkey:
See https://github.com/artefactual/archivematica/issues/1083#issuecomment-429452094

PR #1114, opened 2018-05-30
Dashboard: fix review_matches to avoid KeyError

### LINKED ISSUE #1112: Problem: upload DIP to ArchivesSpace not working
When try to pair a resource in ArchivesSpace with a DIP in AM the object and associated metadata does not show up in ArchivesSpace.  

Archivematica log files show that even though a match was made, the match is ignored or skipped and no information is sent to ArchivesSpace.

jrwdunham:
The pair_matcher.py::review_matches function has been modified so that the 'object_path' key of each `match` dict in the `matches` list valuates to the current location (with '%SIPDirectory%' and 'objects/' prefixes removed) of the matching file as recorded in the database. Previous functionality was attempting to valuate this key to the path of an original file as extracted from the DIP METS file but that approach was faulty because access derivatives are not documented in the METS.

Related to #1034

### DIFF
--- a/src/dashboard/src/components/ingest/pair_matcher.py
+++ b/src/dashboard/src/components/ingest/pair_matcher.py
@@ -152,15 +152,34 @@ def match_dip_objects_to_resource_component_levels(client, request, resource_com
     return render(request, match_template, locals())
 
 
-def review_matches(client, request, template, uuid, matches=[]):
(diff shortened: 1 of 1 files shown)


jrwdunham:
Closing in favour of duplicate PR https://github.com/artefactual/archivematica/pull/1153.

PR #1115, opened 2018-05-31
Correct default behaviour of extract packages and adds Unicode formatted strings

### LINKED ISSUE #1104: Problem: Extract contents crashes due to UnicodeEncodeError
We have come across a transfer where the "Extract contents from compressed archives" job seems to run fine, until it comes across a new compressed object where it fails with the following message in the task overview in the dashboard:
```
....
Not extracting contents from Cotu_K.doc  - No rule found to extract
Not extracting contents from UMT_24.02.12.pdf  - No rule found to extract
Not extracting contents from Rapport_d_activite_Alen.niger_version2.doc  - No rule found to extract
Not extracting contents from Wawan_10.04.17.pdf  - No rule found to extract

extractContents.py: INFO      2018-05-17 20:56:32,786  archivematica.mcp.client.extractContents:get_dir_uuids:240:  Assigning UUID d425717b-eadf-45fc-b5d7-ab13cf550682 to directory path %transferDirectory%objects/SEMINAIRES_2010/MID_TERM_EVALUATION/revised_ghislaine.zip-2018-05-17T20:54:14.968039+00:00/
  File "/usr/lib/archivematica/MCPClient/clientScripts/extractContents.py", line 188, in <module>
    sys.exit(main(transfer_uuid, sip_directory, date, task_uuid, delete=delete))
  File "/usr/lib/archivematica/MCPClient/clientScripts/extractContents.py", line 164, in main
    transfer_mdl)
  File "/usr/share/archivematica/dashboard/main/models.py", line 502, in create_many
    for dir_path, dir_uuid in dir_paths_uuids])
  File "/usr/lib/archivematica/archivematicaCommon/archivematicaFunctions.py", line 237, in get_dir_uuids
    dir_uuid, dir_path)
UnicodeEncodeError: 'ascii' codec can't encode character u'\x82' in position 119: ordinal not in range(128)
```

The script crashes here, with an exit code of 1 according to the task overview dashboard. In the Transfer dashboard however, the job is incorrectly marked as 'Completed successfully'. 

Furthermore, it seems to have skipped the jobs 'Sanitize extracted objects' file and directory names', 'Scan for viruses on extracted files', etc. that would normally run after extraction of packages. Instead it simply moves forward to the 'Update METS.xml document' as if no packages to extract were found. 

This finally results in a 'real' error during METS creation during ingest:
```
  File "/usr/lib/archivematica/MCPClient/clientScripts/archivematicaCreateMETS2.py", line 1314, in <module>
    baseDirectoryPath, objectsDirectoryPath, directories)
  File "/usr/lib/archivematica/MCPClient/clientScripts/archivematicaCreateMETS2.py", line 1182, in get_normative_structmap
    add_normative_structmap_div(all_fsitems, normativeStructMapDiv, directories)
  File "/usr/lib/archivematica/MCPClient/clientScripts/archivematicaCreateMETS2.py", line 1220, in add_normative_structmap_div
    LABEL=basename)
  File "src/lxml/lxml.etree.pyx", line 3112, in lxml.etree.SubElement (src/lxml/lxml.etree.c:81786)
  File "src/lxml/apihelpers.pxi", line 203, in lxml.etree._makeSubElement (src/lxml/lxml.etree.c:18358)
  File "src/lxml/apihelpers.pxi", line 198, in lxml.etree._makeSubElement (src/lxml/lxml.etree.c:18281)
  File "src/lxml/apihelpers.pxi", line 302, in lxml.etree._initNodeAttributes (src/lxml/lxml.etree.c:19840)
  File "src/lxml/apihelpers.pxi", line 316, in lxml.etree._addAttributeToNode (src/lxml/lxml.etree.c:20196)
  File "src/lxml/apihelpers.pxi", line 1439, in lxml.etree._utf8 (src/lxml/lxml.etree.c:32441)
ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters
```
Further investigation into this error reveals it failed to insert the filename of an extracted file into a normative structmap, due to the fact that the file was of course never normalized after extraction.

ross-spencer:
This commit adds Unicode format strings to archivematicaCommonFunctions and corrects the fallback behaviour of the extract_packages microservice if the task should fail whatsoever.

@mcantelon as this fix impacts the dspace chainlinks as well, can you confirm whether the behaviour I have implemented here (fallback to fail transfer) is acceptable for that work as well or if there is additional nuance you'd like to see implemented. 

I made some pylint changes to the archivematicaCommonFunctions too. You will see this in a separate commit. 

Tested by re-deploying my docker implementation.

Additional test files can be found in #1104. 

Resolves #1104.

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -233,7 +233,7 @@ def get_dir_uuids(dir_paths, logger=None):
     """
     for dir_path in dir_paths:
         dir_uuid = str(uuid4())
-        msg = 'Assigning UUID {} to directory path {}'.format(
+        msg = u'Assigning UUID {} to directory path {}'.format(
             dir_uuid, dir_path)
         print(msg)
         if logger:
@@ -288,7 +288,7 @@ def reconstruct_empty_directories(mets_file_path, objects_path, logger=None):
     if (not os.path.isfile(mets_file_path) or
             not os.path.isdir(objects_path)):
         if logger:
-            logger.info('Unable to construct empty directories, either because'
+            logger.info(u'Unable to construct empty directories, either because'
                         ' there is no METS file at {} or because there is no'
                         ' objects/ directory at {}'.format(mets_file_path,
                                                            objects_path))
@@ -299,7 +299,7 @@ def reconstruct_empty_directories(mets_file_path, objects_path, logger=None):
             NORMATIVE_STRUCTMAP_LABEL), NSMAP)
     if logical_struct_map_el is None:
         if logger:
-            logger.info('Unable to locate a logical structMap labelled {}.'
+            logger.info(u'Unable to locate a logical structMap labelled {}.'
                         ' Aborting attempt to reconstruct empty'
                         ' directories.'.format(NORMATIVE_STRUCTMAP_LABEL))
         return
@@ -307,7 +307,7 @@ def reconstruct_empty_directories(mets_file_path, objects_path, logger=None):
         'mets:div/mets:div[@LABEL="objects"]', NSMAP)
     if root_div_el is None:
         if logger:
-            logger.info('Unable to locate a logical structMap labelled {}.'
+            logger.info(u'Unable to locate a logical structMap labelled {}.'
                         ' Aborting attempt to reconstruct empty'
                         ' directories.'.format(NORMATIVE_STRUCTMAP_LABEL))
         return
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -21,9 +21,8 @@
 # @subpackage archivematicaCommon
 # @author Joseph Perry <joseph@artefactual.com>
 
-"""archivematicaFunctions.py
-
-Provides various helper functions across the different Archivematica modules.
+"""archivematicaFunctions provides various helper functions across the
+different Archivematica modules.
 """
 
 from __future__ import print_function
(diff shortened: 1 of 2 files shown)


src/dashboard/src/main/migrations/0052_correct_extract_packages_fallback_link.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+"""0052_correct_extact_packages_fallback_link.py
```
sevein: typo: extact.

Generally I'd say that using the filename as the first line of the module docstring isn't great - but It's hard to come up with a good one-line subject at least for me. It seems that what we've doing so far is to ignore that requirement (pep257 multi-line docstrings) so the paragraph below could replace this.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -21,6 +21,11 @@
 # @subpackage archivematicaCommon
 # @author Joseph Perry <joseph@artefactual.com>
 
+"""archivematicaFunctions.py
+
+Provides various helper functions across the different Archivematica modules.
```
sevein: This could be the module one-line summary.

sevein (approved):
LGTM.

Something that we usually ignore is the fact that our migrations could also have the downgrade path implemented, e.g.:

    migrations.RunPython(data_migration_up, data_migration_down)

I've found this useful during testing too but it's even more important when a deployment goes wrong in production and you want to revert the changes. I'm not asking you to do it but it's something to keep in mind. Some of our migrations implement it.

I've made a couple of comments about docstrings.

ross-spencer:
Hi @sevein Thanks for the feedback. I had a look at making a downgrade path. I feel like it might be a good precedent to set. Can you have a look at my changes and let me know if they're okay?

sevein (approved):

ross-spencer:
To use the migration that was created as part of this pull-request, users need to run the following from the dashboard django admin:

`python manage.py migrate main 0052_correct_extract_packages_fallback_link`

PR #1126, opened 2018-06-13
MCPClient: add hashdeep to osdeps for Xenial

mamedin:
This fixes the issue:

https://github.com/artefactual-labs/ansible-archivematica-src/issues/207

It is needed to use the `hashdeep` package from the Archivematica
repository on Ubuntu 16 and this change will force its updated if the
system has an old version installed from the Ubuntu repository.

No changes are needed for Ubuntu 14, because it uses the `md5deep`
package instead of `hashdeep`.

### DIFF
--- a/src/MCPClient/osdeps/Ubuntu-16.json
+++ b/src/MCPClient/osdeps/Ubuntu-16.json
@@ -23,6 +23,7 @@
   { "name": "fits", "state": "latest"},
   { "name": "gearman", "state": "latest"},
   { "name": "ghostscript", "state": "latest"},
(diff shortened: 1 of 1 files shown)


scollazo:
I'm ok with this, but we might need to involve @sevein @jrwdunham or @jhsimpson to be sure we are following the expected workflow regarding tags/waffle and so

mamedin:
There is no rush, it just happens when upgrading on Ubuntu 16, so it does not affect in production.

sevein (approved):

sevein:
f4c94931 in `qa/1.x`.

Issue #1140, opened 2018-06-13
Problem: AM doesn't have a notion of public URL


sevein:
The self-registration process that AM performs against SS (during the installation) is sometimes insufficient because the pipeline details does not include an accessible URL so SS can reach the pipeline.

The request made by `storageService` during the registration is similar to the one produced with the following command:

```
curl -s -d '{
        "uuid": "99354557-9e6e-4918-9fe3-a65b00ecb199",
        "description": "Test pipeline",
        "create_default_locations": true,
    }' \
        -X POST \
        -H "Authorization: ApiKey test:test" \
        -H "Content-Type: application/json" \
            "http://127.0.0.1:8000/api/v2/pipeline/"
```

The `remote_name` value is not used. When unset, SS populates that property after the value found in the `REMOTE_ADDR` HTTP header - this is okay as a fallback but we want AM to learn a new `site_url` setting that the user can set. This value would be used during the self-registration process.

In RDSS, this was solved differently. See https://github.com/JiscRDSS/archivematica-storage-service/pull/22. for more details.

sromkey:
This is connected to https://github.com/artefactual/archivematica/pull/1142 and since the setting is very much there (under Administration, General) I'm calling this verified!

sromkey:
Whoops,  jumped the gun, we still need to document this! Assigning myself.

sallain:
Documented and added to release notes.

PR #1142, opened 2018-06-14
Add site_url setting

sevein:
This is connected to #1140.

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -29,10 +29,11 @@ from django.db.models import Max, Min
 from django.http import Http404, HttpResponseNotAllowed, HttpResponseRedirect
 from django.shortcuts import redirect, render
 from django.template import RequestContext
+from django.utils.six.moves import map
 from django.utils.translation import ugettext as _
 
 from main import models
-from components.administration.forms import AgentForm, HandleForm, StorageSettingsForm, ChecksumSettingsForm, TaxonomyTermForm
+from components.administration.forms import AgentForm, HandleForm, GeneralSettingsForm, StorageSettingsForm, ChecksumSettingsForm, TaxonomyTermForm
 import components.administration.views_processing as processing_views
 import components.decorators as decorators
 import components.helpers as helpers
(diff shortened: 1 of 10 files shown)


src/dashboard/src/installer/forms.py
```diff
@@ -22,8 +22,21 @@
 from django.forms.widgets import TextInput
 from django.utils.translation import ugettext_lazy as _l
 
+from main.models import DashboardSetting
+
+
+site_url_field = forms.CharField(
```
jraddaoui, line 28: Could we use `URLField` to improve validation when it's not read-only?

jraddaoui (approved):
Looks great, just a minor comment.

src/dashboard/src/installer/forms.py
sevein: I wanted to but `URLField` won't validate values like `http://archivematica-storage-sevice:8000`. I guess it's looking for fully qualified domain names but if you're on a local network I'm not sure why that should be a requirement.

src/dashboard/src/installer/forms.py
```diff
@@ -22,8 +22,21 @@
...
+    label=_l('Site URL'),
+    help_text=_l('It defines the public address of this service. This is'
+                 ' needed so the service (which is part of the pipeline)'
+                 ' is accessible by Storage Service.'),
```
jrwdunham: Nitpick: "is accessible by" sounds weird here to me. Maybe "can be accessed by the".

src/dashboard/src/installer/forms.py
```diff
@@ -32,6 +45,10 @@ class Meta:
         model = User
         fields = ['org_name', 'org_identifier', 'username', 'first_name', 'last_name', 'email', 'password1', 'password2']
 
+    def __init__(self, *args, **kwargs):
+        super(OrganizationForm, self).__init__(*args, **kwargs)
```
jrwdunham: Wrong first argument to `super`; it should be `SuperUserCreationForm` I think.

src/dashboard/src/templates/administration/general.html
```diff
@@ -34,6 +34,10 @@ <h3>{% trans "General configuration" %}</h3>
 
       <form method='POST'>
 
+        {% include "_form.html" with form=general_form %}
+
+        <hr />
+
```
jrwdunham: Seems a bit inconsistent that there's a horizontal rule here and then there are `<h4>`s for "Storage Service options" and "Elasticsearch indexing", etc. but no `<h4>` header for this section. Super nitpicky yet vague formatting comment, so feel free to ignore.

src/dashboard/src/components/administration/views.py
```diff
@@ -450,15 +451,17 @@ def general(request):
...
                                        prefix='storage', initial=initial_data)
     checksum_form = ChecksumSettingsForm(request.POST or None,
                                          prefix='checksum algorithm',
                                          initial=initial_data)
 
-    if storage_form.is_valid() and checksum_form.is_valid():
-        storage_form.save()
-        checksum_form.save()
+    forms = (general_form, storage_form, checksum_form)
+    if map(lambda form: form.is_valid(), forms):
```
jrwdunham: I'm not sure this condition is correct. No matter how many of the forms are valid or invalid, the return value of `map` is still going to be something that is `True` when `bool` is applied to it.

jrwdunham (changes requested):
I think there's one or two problems with this PR and I have some other minor comments.

src/dashboard/src/components/administration/views.py
sevein: Ops I meant to use `all(map(...))`

jrwdunham (approved):

PR #1153, opened 2018-06-20
Dashboard: fix review_matches to avoid KeyError

jrwdunham:
The `pair_matcher.py::review_matches` function has been modified so that the `'object_path'` key of each `match` dict in the `matches` list valuates to the current location (with `'%SIPDirectory%'` and `'objects/'` prefixes removed) of the matching file as recorded in the database. Previous functionality was attempting to valuate this key to the path of an original file as extracted from the DIP METS file but that approach was faulty because access derivatives are not documented in the METS.

This is connected to #1034 (should fix this issue).
Also related to #1112

### DIFF
--- a/src/dashboard/src/components/ingest/pair_matcher.py
+++ b/src/dashboard/src/components/ingest/pair_matcher.py
@@ -152,15 +152,34 @@ def match_dip_objects_to_resource_component_levels(client, request, resource_com
     return render(request, match_template, locals())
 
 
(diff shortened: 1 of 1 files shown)


jrwdunham:
Note: we are currently waiting on RAC to see if this fix works for them in the most recent incarnation of their dev environment, so CR can wait until I update this PR with news from them.

sevein (approved):

sevein:
e4ce4639 in `qa/1.x`

Issue #1124, opened 2018-06-11
Problem: can't build MCPServer from Dockerfile


jhsimpson:
This is the same issue as: https://github.com/JiscRDSS/archivematica/issues/75

The MCPServer Dockerfile is based on the python-2.7 docker image.  That recently changed from being based on Debian Jessie to Stretch.  

It might be good to update the Dockerfile to work with Stretch, but the solution used in the JiscRDSS issue referenced above was to use python:2.7-jessie instead of FROM python:2.7 at MCPServer.Dockerfile:

```diff --git a/src/MCPServer.Dockerfile b/src/MCPServer.Dockerfile
index 91d02009..d18bf84f 100644
--- a/src/MCPServer.Dockerfile
+++ b/src/MCPServer.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7
+FROM python:2.7-jessie

 ENV DEBIAN_FRONTEND noninteractive
 ENV DJANGO_SETTINGS_MODULE settings.common
```

sromkey:
![](https://github.trello.services/images/mini-trello-icon.png) [Artefactual (internal)](https://trello.com/c/AxNvL2Bv/11-artefactual-internal)

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Jisc RDSS](https://trello.com/c/iIwIebKY/21-jisc-rdss)

joel-simpson:
I have deployed using docker this week and had no problem building the MCP Server container (and several other people doing QA & dev have this week), so I'm moving this issue to verified.

PR #1154, opened 2018-06-21
Docker: use Debian jessie base images

sevein:
Our images were based on `python:2.7`. This image moved to Debian stretch
causing an issue in our build. For further details see https://git.io/fCV6c.

This commit moves the base images to `python:2.7-jessie` to ensure that we
continue using Debian jessie until we have time to evaluate the upgrade.

This is connected to #1124.


jraddaoui (approved):

sevein:
b6dcfb07 in `stable/1.7.x`

Issue #1158, opened 2018-06-21
Problem: ghostscript to-PDF normalization command claims to create PDF/A 1a when in fact it creates PDF/A 1b


jrwdunham:
Our "Transcoding to pdfa with Ghostscript" FPR normalization command says its output format is "Acrobat PDF/A 1a". See `/fpr/fpcommand/d6a33093-85d5-4088-83e1-b7a774a826bd/`. However, that command actually generates "Acrobat PDF/A 1b" derivatives. See [PDF/A 1b on PRONOM](http://www.nationalarchives.gov.uk/pronom/fmt/354) and [PDF/A 1a on PRONOM](http://www.nationalarchives.gov.uk/pronom/fmt/95).

I have verified that we create 1b not 1a with our `gs` command

    gs -dPDFA -dBATCH -dNOPAUSE -sDEVICE=pdfwrite -dPDFACompatibilityPolicy=1 -sOutputFile="%outputDirectory%%prefix%%fileName%%postfix%.pdf" "%fileFullName%"

It sounds like it is actually quite difficult to generate PDF/A 1a from arbitrary textual source files. It has much more stringent requirements than PDF/A 1b. See this forum post where a developer (creator?) of Ghostscript says (in 2009) that he [will not support PDF/A 1a creation because it is too hard, if not impossible in practice](https://ghostscript.com/pipermail/gs-devel/2009-August/008556.html).

Then there's also this document where [we talk about creating PDF/A 1b](https://wiki.archivematica.org/PDF_to_PDF/A_using_Ghostscript.) with that same `gs` FPR command.

My conclusion from all of this is that we probably just need to change our FPR command to be more honest about what its true output format is, viz. "Acrobat PDF/A 1b", i.e., PRONOM fmt/354.

See internal redmine 12265.

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Artefactual (internal)](https://trello.com/c/AxNvL2Bv/11-artefactual-internal)

sromkey:
The output file format in the command Transcoding to pdfa with Ghostscript now correctly reads Acrobat PDF/A 1b.

PR #1160, opened 2018-06-21
Fix PDF normalization command output

jrwdunham:
This changes the archivematica-fpr-admin requirement to point to a dev branch which contains a migration that fixes the Ghostscript PDF normalization FPCommand so that it states the correct output FormatVersion as PDF/A 1b (not PDF/A 1a). This will need to be updated to a new AM-fpr-admin v.1.7.7 tag/release once such is available.

This is connected to https://github.com/artefactual/archivematica/issues/1158.


sevein (approved):
Final version should be v1.7.7.

👍

jrwdunham:
Cherry-picked to stable/1.7.x as 6fd0076d08357d028096b41f61d28387263e9eef

Issue #1138, opened 2018-06-13
Problem: processing configurations can't be reset


sevein:
Pull request @ JiscRDSS: https://github.com/JiscRDSS/archivematica/pull/67.

joel-simpson:
Adding some context to this problem because some of it was documented in the Jisc RDSS wiki and isn't publicly available: 

In the RDSS we recommended configuration settings that all of the HEIs participating in the pilot could use. One is an "Automated" workflow configuration, which requires no user input. We also provided recommended settings for the "default" workflow configuration that requires some user input. Any setting relating to external integrations are set to 'no' or it's equivalent. 

We found in testing that users often change configuration settings, and it is then a laborious task to reset them. 

We also wanted to address a general problem that there was no easy way to download a processingMCP.xml file (the configuration file) so that users could choose to use specific configurations with certain transfers. 

The solution we implemented was to implement a download button as well as a 'reset' button. The download button simply downloads the file in the browser. The reset button puts all of the configuration settings back to our recommended settings. 

These changes can be seen on the Processing Configuration screen: 
![screen shot 2018-06-18 at 9 06 15 pm](https://user-images.githubusercontent.com/17187309/41570450-e674135c-733c-11e8-8cc5-3f6e9748b4b8.png)

Note that if users add their own processing configuration (using the "Add" button), there is no reset feature (instead they will just a 'delete' button). 

Broadly I believe this is a very useful feature but there are some considerations for implementing in the public project: 

a) we may want to revisit the actual settings chosen by the Jisc project for both the Automated and Default workflows. 

b) we should publish in the documentation the rationale for each setting in these two configurations. This is useful to do in it's own right, but also important to help users understand the impact of selecting the "Reset" button.

sallain:
@joel-simpson has this been documented?

joel-simpson:
@sallain this was documented on the Jisc project with two wiki pages: 
1) [What is a 'processing configuration' and what settings should I use?](https://jiscdev.atlassian.net/wiki/spaces/RDSSP/pages/260440092/What+is+a+processing+configuration+and+what+settings+should+I+use)
This page maps to the [processing configuration](https://www.archivematica.org/en/docs/archivematica-1.8/user-manual/administer/dashboard-admin/#dashboard-processing) section in docs and is more about recommendations for settings (so would need quite a bit of rewording at the least, and is potentially not necessary at all). 

[...]

PR #1166, opened 2018-06-27
Allow users to reset processing configuration

sevein:
This is connected to #1138.


jraddaoui (approved):
Looks great to me! Nice addition.

Issue #1135, opened 2018-06-13
Problem: processing of datasets containing no files stalls at the end of the Transfer stage 


sevein:
Original issue @ JiscRDSS: https://github.com/JiscRDSS/rdss-archivematica/issues/76, https://github.com/JiscRDSS/rdss-archivematica/issues/59.
Pull request @ JiscRDSS: https://github.com/JiscRDSS/archivematica/pull/59.

joel-simpson:
This is working as expected. 
To test, I created an empty directory and selected it in the Dashboard to start a transfer.
The Transfer fails Verify Transfer Compliance. 
![screenshot from 2018-07-25 19-09-49](https://user-images.githubusercontent.com/17187309/43232351-67512b0e-903e-11e8-861a-fe0ee47487f4.png)

And an error is captured in the task details screen
![screenshot from 2018-07-25 19-09-16](https://user-images.githubusercontent.com/17187309/43232365-745cefcc-903e-11e8-9a7b-fa2360ced739.png)

ross-spencer:
This is an interesting one @joel-simpson @sevein. I wonder if Archivematica could do more earlier to prevent the user from having to enter to Verify Transfer Compliance screen to understand the failure?

PR #1167, opened 2018-06-27
Add checks for empty transfers

sevein:
This goes into verifyTransferCompliance microservice. Fail the transfer
at this stage if there are no files in the transfer.

This is connected to #1135.

### DIFF
--- a/src/MCPClient/lib/clientScripts/verifyTransferCompliance.py
+++ b/src/MCPClient/lib/clientScripts/verifyTransferCompliance.py
@@ -53,11 +53,20 @@ def verifyNothingElseAtTopLevel(SIPDir, ret=0):
     return ret
 
 
+def verifyThereAreFiles(SIPDir, ret=0):
+    """Make sure there are files in the transfer."""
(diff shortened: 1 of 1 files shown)


src/MCPClient/lib/clientScripts/verifyTransferCompliance.py
```diff
@@ -53,11 +53,26 @@ def verifyNothingElseAtTopLevel(SIPDir, ret=0):
...
 
 
+def verifyThereAreFiles(SIPDir, ret=0):
+    """
+    Make sure there are files in the transfer
+    """
+    if not any(
+        files
+        for (_, _, files) in os.walk(SIPDir)
+    ):
```
jraddaoui: This could fit in one line.

jraddaoui (approved):

src/MCPClient/lib/clientScripts/verifyTransferCompliance.py
sevein: Done, thanks.

Issue #1129, opened 2018-06-13
Problem: transfer names can't start with a dash


sevein:

sevein:
@ross-spencer do you think issue is now fixed by https://github.com/artefactual/archivematica/pull/1350?

ross-spencer:
LGTM :+1:  

![image](https://user-images.githubusercontent.com/1880412/53534621-f6036580-3aff-11e9-9901-2cde87516e08.png)

But https://github.com/archivematica/Issues/issues/380 will need to be looked at separately. 

@sromkey as this is merged, and looks to work, should we add a `1.9` milestone and mark it as verified so that it's picked up in release notes? Additionally get an analyst double-check?

sromkey:
ok, I've labeled appropriately!

sevein:
Woohoo!

currmie:
Fails for me on am19xenial. I just tried a standard transfer with a dash at the start of the transfer name, and AM threw this error:
```--unitType "SIP" --unitIdentifier "3446d975-dcb8-40f1-8b8b-87001688af5a" --unitName "-mc-dash-in-name"```

**Comment by currmie** [comment]
Oh wait, maybe this build hasn't included the fix yet?

**Comment by sevein** [comment]
Thanks @currmie, you're right, it's still broken.

I think that position arguments must use the extra double dash which is what Helen was doing in the original PR. More here: https://stackoverflow.com/questions/14693718/how-to-parse-positional-arguments-with-leading-minus-sign-negative-numbers-usi.

```
./compress_aip.py 1 2 3 4 5
Namespace(compression='1', compression_level='2', sip_directory='3', sip_name='4', sip_uuid='5')

./compress_aip.py 1 2 3 "the-name" 5
Namespace(compression='1', compression_level='2', sip_directory='3', sip_name='the-name', sip_uuid='5')

./compress_aip.py 1 2 3 "--the-name" 5
usage: compress_aip.py [-h]
                       compression compression_level sip_directory sip_name
                       sip_uuid
compress_aip.py: error: too few arguments

./compress_aip.py -- 1 2 3 "--the-name" 5
Namespace(compression='1', compression_level='2', sip_directory='3', sip_name='--the-name', sip_uuid='5')
```

sromkey:
I took the milestone off, since this wasn't meant to be in scope for this release anyway.

PR #1169, opened 2018-06-27
Make sure transfer names can begin with dash

sevein:
This is connected to #1129.

### DIFF
--- a/src/dashboard/src/main/migrations/0053_quote_args.py
+++ b/src/dashboard/src/main/migrations/0053_quote_args.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def data_migration_put_quotes_in_compress_aip_args(apps, schema_editor):
+    StandardTaskConfig = apps.get_model('main', 'StandardTaskConfig')
+
+    tasks = (
+        # upload-archivesspace.py
+        (
+            '10a0f352-aeb7-4c13-8e9e-e81bda9bca29',
+            u'--host="%host%" --port="%port%" --user="%user%"'
+            u' --passwd="%passwd%" --dip_location="%SIPDirectory%"'
+            u' --dip_name="%SIPName%" --dip_uuid="%SIPUUID%"'
+            u' --restrictions="%restrictions%" --object_type="%object_type%"'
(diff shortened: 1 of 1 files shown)


src/dashboard/src/main/migrations/0053_quote_args.py
```diff
@@ -0,0 +1,70 @@
...
+    tasks = (
+        # upload-archivesspace.py
+        (
+            '10a0f352-aeb7-4c13-8e9e-e81bda9bca29',
+            u'--host="%host%" --port="%port%" --user="%user%"'
+            u' --passwd="%passwd%" --dip_location="%SIPDirectory%"'
+            u' --dip_name="%SIPName%" --dip_uuid="%SIPUUID%"'
+            u' --restrictions="%restrictions%" --object_type="%object_type%"'
+            u' --xlink_actuate="%xlink_actuate%" --xlink_show="%xlink_show%"'
+            u' --use_statement "%use_statement%" --uri_prefix "%uri_prefix%"'
```
jrwdunham, line 19: Why are you changing some of the optional arguments of the form `--x n` to the form `--x=n` but not others? E.g., `--use_statement "%use_statement%"` lacks the `=`. Is there a reason?

src/dashboard/src/main/migrations/0053_quote_args.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def data_migration_put_quotes_in_compress_aip_args(apps, schema_editor):
```
jrwdunham, line 7: Does the function name here really reflect what it's doing? Relatedly, why are you targeting these and only these 6 `StandardTaskConfig`s? They seem to all contain values for `SIPName`. But there are 18 others:

    >>> len(models.StandardTaskConfig.objects.filter(arguments__contains='SIPName'))
    24

Why do the others not need mods to their arguments value also?

Maybe a docstring in this migration func would be helpful.

jrwdunham (commented):
I have some questions ...

sevein:
Hey @jrwdunham, thanks for your comments. I think you're right, this commit may need some more work as it doesn't seem to cover all the cases. Thanks.

Issue #1133, opened 2018-06-13
Problem: SS client timeout setting for async API endpoints is missing


sevein:
Storage Service 0.12 brings asynchronous API endpoints. This means that certain API requests that we make are now going to be responded by SS immediately instead of holding until the work is done. With this in place now we need to update our client so it can have different timeouts when it comes to synchronous vs asynchronous endpoints. For the former you may still want to wait for hours while for the latter case you wouldn't wait too long, perhaps no more than five seconds.

Pull request @ JiscRDSS: https://github.com/JiscRDSS/archivematica/pull/58.

jraddaoui:
Code changes look great and I've been notified that this has been tested for a while.

PR #1170, opened 2018-06-28
storageService: add new query timeout

sevein:
This is in addition to the existing client timeout that already exists in the
storageService module. This new timeout is used in requests that we know they
should be processed immediately.

This is connected to #1133.

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -57,7 +57,7 @@ def _storage_service_url():
     return storage_service_url
 
 
-def _storage_api_session(timeout=django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT):
+def _storage_api_session(timeout=django_settings.STORAGE_SERVICE_CLIENT_QUICK_TIMEOUT):
     """Return a requests.Session with a customized adapter with timeout support."""
     class HTTPAdapterWithTimeout(requests.adapters.HTTPAdapter):
         def __init__(self, timeout=None, *args, **kwargs):
@@ -73,6 +73,11 @@ def _storage_api_session(timeout=django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT)
     session.mount('http://', HTTPAdapterWithTimeout(timeout=timeout))
     session.mount('https://', HTTPAdapterWithTimeout(timeout=timeout))
     return session
+
+
+def _storage_api_slow_session():
+    """Return a requests.Session with a higher configurable timeout."""
+    return _storage_api_session(django_settings.STORAGE_SERVICE_CLIENT_TIMEOUT)
 
 
 def _storage_api_params():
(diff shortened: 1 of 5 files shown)


sevein:
I think I need to review the client module again because I'm not entirely sure that the new timeout setting is used in all places where we could be taking advantage of it.

src/archivematicaCommon/lib/storageService.py
```diff
@@ -249,7 +253,7 @@ def copy_files(source_location, destination_location, files):
 
     url = _storage_service_url() + 'location/' + destination_location['uuid'] + '/async/'
     try:
-        response = _storage_api_session().post(url, json=move_files)
```
jrwdunham: Unable to put the comment at line 209, but shouldn't `_storage_api_session` in `wait_for_async` be `_storage_api_quick_session`? That should be a quick response too, no?

src/archivematicaCommon/lib/storageService.py
```diff
@@ -441,7 +446,7 @@ def request_file_deletion(uuid, user_id, user_email, reason_for_deletion):
         'user_id': user_id,
```
jrwdunham: Again, unable to comment at actual line, but is there a reason not to use `_storage_api_quick_session` in `get_file_info` above?

jrwdunham: As Rich Hickey says, "perception should be free".

jrwdunham: Aw man, `get_file_metadata` can take a long time because it queries all of the files of a package into memory and json-serializes them. That endpoint needs some pagination desperately. Created https://github.com/artefactual/archivematica-storage-service/issues/389

jrwdunham: My general comment is that `_storage_api_session` should be the exceptional case and all instances of it should be accompanied by a comment indicating why it is necessary. I think the above comment could be one such code comment.

To take the next nitpicky step given the above, I think the two func names should instead be `_storage_api_session` and `_storage_api_slow_session`.

jrwdunham: Ok, maybe all instances where the slow session is used do **not** need to be commented. For some it may be obvious.

jrwdunham (commented):
I have some comments, questions and ideas about the function naming and where the quick timeout could be used. 

I think maybe the quick timeout session could be used in `session = storage_service._storage_api_session()` of `failedTransferCleanup.py` (and `failedSIPCleanup.py`) because those PATCH requests just update a Package instance in the db.

sevein:
@jrwdunham ok done! `_storage_api_slow_session` is used in:

- `storageService.create_file`
- `storageService.get_file_info`
- `storageService.extract_file`
- `storageService.request_reingest`
- `storageService.post_store_aip_callback`
- `storageService.get_file_metadata`
- `storageService.remove_files_from_transfer`
- `storageService.index_backlogged_transfer_contents`
- `storageService.reindex_file`

`_storage_api_session` (the new "quick") is used in:

- `failedSIPCleanup.py`
- `failedTransferCleanup.py`
- `storageService.create_pipeline`
- `storageService.get_pipeline`
- `storageService.get_location`
- `storageService.browse_location`
- `storageService.wait_for_async`
- `storageService.copy_files` (async)
- `storageService.create_file` (non-update branch, async)
- `storageService.request_file_deletion`

jrwdunham (approved):

Issue #1022, opened 2018-03-27
Problem: the "normalize for thumbnail" micro-service has inefficiencies and redundancies


jrwdunham:
Proposed solution: reduce and/or remove some "normalize for thumbnail" tasks. Some normalization can be avoided altogether, e.g., the creation of thumbnails for XML files. See also:

- https://github.com/artefactual-labs/archivematica-acceptance-tests/issues/62
- https://docs.google.com/document/d/1Fv8SthGKndeIg_TzVNNXXJUg89Rm0PYC80ebf50gIuI/edit

jhsimpson:
Some additional detail:

The Normalization Microservice includes a job that creates thumbnail images for every file. These thumbnails are used in the Archival Storage tab when the 'show files' checkbox is clicked.  

If search indexing is disabled (a new feature in v1.7.0) then the thumbnails can't be accessed, but they are still created.

Thumbnails often aren't unique. A thumbnail for an xml file just tells you it is an xml file. The current implementation physically copies a placeholder image once for every file.  The same thumbnail image could be use for every file.

This job within the Microservice should be configurable so that users can determine not to run it. This will save processing time and reduce the size of preserved datasets.

Some analysis is required to determine best model for configuring this (i.e. as part of FPR rule, or as part of processing configuration?)

Acceptance Criteria
produce a decision matrix showing options for how to make this microservice configurable
implement a first iteration of the selection solution
test to demonstrate impact of the change (performance and preservation quality).

This issue has been filed in the JiscRDSS fork of Archivematica as well: https://github.com/JiscRDSS/archivematica/issues/40

jambun:
We looked at this today and walked through how we might implement it using the processing configuration to set pass a purpose to the normalize script to tell it not to generate thumbnails. It seems like the right approach, but maybe we're not seeing and alternative. It seems to us that using an FP rule would not help in this case. But again, we might be missing a subtlety here.

Should we go ahead and implement this?

jambun:
Thanks @sromkey - that sounds like a good plan.

@jhsimpson - shall we go ahead with the first part of Sarah's suggested approach?

jhsimpson:
I like the idea of doing a couple or three iterations.  How does this sound?

Iteration 1:
add a new question to the processing configuration page called 'generate thumbnails' with three options (Yes, No, None).  None allows the user to choose during processing.

Iteration 2:
[...]

PR #1176, opened 2018-07-02
Thumbnail mode

marktriggs:
We've been working on making thumbnail generation configurable, and think we have a set of changes that work.  At the point of normalisation, the user can choose either:

  * Yes

  * Yes, without default

  * No

The behaviour of these corresponds to the first two iterations Justin described in issue https://github.com/artefactual/archivematica/issues/1022#issuecomment-398960933.

Digging more into the technical details, the simplified version is that the user's answer to this question sets a replacement dict value (`%ThumbnailMode%`) that gets passed to the normalise client script, and it changes thumbnail behaviour based on this value.

The fact that the thumbnails might be absent also required two changes further down the line: "Copy thumbnails to DIP" assumed that the thumbnails directory would exist (it doesn't, because Create METS V2 removes empty directories), and "Remove bagged files" would attempt to delete the thumbnails directory and fail when it wasn't there.  We've changed both of these cases to use new client scripts that let it slide if the thumbnails directory is gone.

That's the high-level technical summary.  Now, diving head-first into the weeds, the changes to the microservice links required a few iterations to get right.

Our first instinct was to prefix the normalise chain with a new link that would ask our new question, set a replacement dict value, pass it through to the normalise client script, and off to the races.  This didn't work because there are actually 6 different normalisation paths, each one represented as its own chain.  Placing our question at the point that chooses between these normalisation chains didn't work because replacements dicts don't get passed between chains (so our variable would get set and then immediately cleared).

In attempt 2 we duplicated the question to sit at the front of each normalisation chain (creating 6 new replacement dict links in the DB).  This worked when running an ingest through the dashboard, but when we went to add the option to the Administration section, we realised it was no good: with the question duplicated 6 times, setting a default to cover all of them wasn't possible.

Our final attempt used a pattern that already gets used to return to normalisation after the file identification command has run.  What we ended up doing:

  1. Each normalisation chain now starts with a link that uses a SetUnitVariable to store the location of the next link (the one that will actually do the normalise thumbnail)

  2. It then jumps to a shared link that asks our question, setting the replacement dict value as before.

  3. Once the value is set, we jump to a second shared link that does a UnitVariableLinkPull on the variable added in step 1.  This sends us back to the normalisation link with the replacement dict now set.

  4. As before, we pass that `%ThumbnailMode%` value to the client script.

I hope I'm making sense :)

To aid in understanding all of this, I wrote a program (https://github.com/hudmol/am-generate-microservice-chains/blob/master/generate_microservice_chains.rb) that reads the definitions out of the MCP DB and spits out an equivalent Ruby/Crystal program (largely stealing most of the logic from Jesús' SVG generation program ;) The before and after versions are here:

  * Before: http://tsp.nz/d/e2ffd50fb43d03b61ff3f26324b5ecd28ba6a36b.rb#/pre-change.rb

  * After: http://tsp.nz/d/663d743eb117520aec41ba913cdcfa7c59a97b86.rb#/post-change.rb

and diffing the two looked about like what we expected.

connects to #1022

### DIFF
--- a/src/MCPClient/lib/archivematicaClientModules
+++ b/src/MCPClient/lib/archivematicaClientModules
@@ -133,5 +133,5 @@ elasticsearchindex_v0.0 = elastic_search_index_process_transfer
 restructuredipforcontentdmupload_v0.0 = restructure_dip_for_content_dm_upload
 upload-qubit_v0.0 = upload_qubit
 upload-archivesspace_v0.0 = upload_archivesspace
-copyThumbnailsToDIPDirectory_v0.0 = copyThumbnailsToDIPDirectory
-removeDirectories_v0.0 = removeDirectories
+copyThumbnailsToDIPDirectory_v0.0 = copy_thumbnails_to_dip_directory
+removeDirectories_v0.0 = remove_directories
--- a/src/MCPClient/lib/archivematicaClientModules
+++ b/src/MCPClient/lib/archivematicaClientModules
@@ -133,3 +133,4 @@ elasticsearchindex_v0.0 = elastic_search_index_process_transfer
 restructuredipforcontentdmupload_v0.0 = restructure_dip_for_content_dm_upload
 upload-qubit_v0.0 = upload_qubit
 upload-archivesspace_v0.0 = upload_archivesspace
+copyThumbnailsToDIPDirectory_v0.0 = copyThumbnailsToDIPDirectory
--- a/src/MCPClient/lib/archivematicaClientModules
+++ b/src/MCPClient/lib/archivematicaClientModules
@@ -134,3 +134,4 @@ restructuredipforcontentdmupload_v0.0 = restructure_dip_for_content_dm_upload
 upload-qubit_v0.0 = upload_qubit
 upload-archivesspace_v0.0 = upload_archivesspace
 copyThumbnailsToDIPDirectory_v0.0 = copyThumbnailsToDIPDirectory
+removeDirectories_v0.0 = removeDirectories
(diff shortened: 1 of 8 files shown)


jhsimpson:
The Batching work got merged (#1105 ) so I think this should be rebased, but we can continue reviewing in the meantime.

payten:
Thanks @jhsimpson, I've just rebased this with the latest and have migrated the new client scripts to be batch-friendly.

jhsimpson:
I have tested this, and it works correctly.  I didn't test all the permutations yet, but I set the 'generate thumbnails' config choice to 'do not generate', and there were no thumbnails generated.

I have 2 comments/questions, neither are blocking concerns at all:

1) There is a job that runs inside the Normalize MicroServiceGroup called `Create thumbnails directory`.  This is not really a problem, but I find myself trying to think of a way to prevent that from running at all, when do not generate is chosen.  a minor optmization probably.

2) The `Normalize for thumbnails` job executes for every file and then doesn't do any work.  Again an optimzation, not necessarily easy or in scope for this PR, but is there a way to not run all those tasks in the first place, given that the system knows before the Job that 'do not generate' was selected.  In the case where 'yes, without default' and 'yes this comment is not relevant, so maybe trying to change the behaviour for one user choice, but not  the others, is not really possible.

The archival storage tab search results look a bit funny, when there are no thumbnails, but some aips might have thumbnails and some might not, so it is not a straightforward change to improve that layout.  That is not a question, just a random observation.

sevein (commented):
This is great. We must have the most unnecessarily complex way of describing workflows but you still figured it out nicely.

I've also found useful to compare the graphs before and after your migration. I've uploaded a gist with both graphs before and after: https://gist.github.com/sevein/2a2661ec586e0b35db321b120bcd2ebe. 

I'm planning to submit a new pull request bumping the migration to 0055 because 0053 and 0054 are already taken!

sevein:
Replaced by https://github.com/artefactual/archivematica/pull/1216.

marktriggs:
Thanks @sevein!  Yeah, it took some figuring out :)  Having  `graph_links.py` (both the source and its SVG output) was a huge help, thanks!

PR #1177, opened 2018-07-03
Test to try and add a new user choice in the ArchivesSpace

### LINKED ISSUE #1112: Problem: upload DIP to ArchivesSpace not working

jrwdunham:
This PR adds a new link in the workflow between "Choose config for ArchivesSpace DIP Upload" and "Upload to ArchivesSpace":

    Choose config for ArchivesSpace DIP Upload (get replacement dic from user choice)
    |
    v
    Continue DIP Upload to ArchivesSpace? (get user choice to proceed with)
    |
    v
    Upload to ArchivesSpace (one instance)

This is needed because when the user chooses to upload to AS they are presented with an interface for matching AS resources and AM entities. However, given the workflow as it is now, the upload requests to AS begin immediately so that by the time the user has finished the AS/AM pairing they are irrelevant (the request has been sent).

The fix being explored here tries to introduce another choice point between the user choice and the AS upload so that the user-performed pairing can terminate before upload starts. There is at least one problem with this approach, at least as it is currently implemented: the replacement dic from the first link only affects the next link. This means that the `Upload to ArchivesSpace (one instance)` link lacks the AS config and will make a failed request.

This is connected to #1112.

### DIFF
--- a/src/MCPServer.Dockerfile
+++ b/src/MCPServer.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7
+FROM python:2.7-jessie
 
 ENV DEBIAN_FRONTEND noninteractive
 ENV DJANGO_SETTINGS_MODULE settings.common
--- a/src/dashboard.Dockerfile
+++ b/src/dashboard.Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7
+FROM python:2.7-jessie
 
 ENV DEBIAN_FRONTEND noninteractive
 ENV DJANGO_SETTINGS_MODULE settings.production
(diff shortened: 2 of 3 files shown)


sevein:
Reminder: ideally we wouldn't be introducing database migrations to stable/1.7.x. If we do, qa/1.x migrations will need to be reordered. Related: https://github.com/archivematica/Issues/issues/54.
Is it possible to solve this problem without migrations?

sevein:
Replaced by #1233. I'm going to leave the branch just in case we realize that we need to go back to this solution.

sevein (commented):
Replaced by #1233.

PR #1178, opened 2018-07-03
Correct PREMIS:originalName for packaged objects

ross-spencer:
This commit does a number of things. First, the logic for extracting
zipped packages is reversed so the package is temporarily renamed so
they can be extracted into a folder that reflects their original name.

Second the commit adds behavior that allows the originalLocation to be
set verbosely inside the code, where previously, we were not exploiting
the full capabilities of the Directory model type in the ORM.

--

**For discussion** should I replace the use of a tuple in all instances where we use the Directory create_many member function? https://github.com/artefactual/archivematica/search?q=create_many&unscoped_q=create_many There doesn't seem that many, and I think I prefer the verbosity.

**To test** the zip attached contains another zip. There are accents in the directory names, spaces, and an accent in the file names. All should be preserved in all the `premis:originalName` entries when reviewed, e.g. 

`<premis:originalName>%transferDirectory%objects/space space.zip/space space/ÉtoileBrûler.zip/ÉtoileBrûler/filè.txt</premis:originalName>`

**NB.** From an analyst's perspective, do we think this is accurate? Do we want to have the `.zip` extension?

[space space.zip](https://github.com/artefactual/archivematica/files/2161189/space.space.zip)

Connects to #1094

### DIFF
--- a/src/MCPClient/lib/clientScripts/extract_contents.py
+++ b/src/MCPClient/lib/clientScripts/extract_contents.py
@@ -15,7 +15,7 @@ from main.models import Directory, FileFormatVersion, File, Transfer
 from custom_handlers import get_script_logger
 from executeOrRunSubProcess import executeOrRun
 from databaseFunctions import fileWasRemoved
-from fileOperations import addFileToTransfer, updateSizeAndChecksum
+from fileOperations import addFileToTransfer, updateSizeAndChecksum, rename
 from archivematicaFunctions import get_dir_uuids, format_subdir_path
 
 # clientScripts
@@ -23,11 +23,12 @@ from has_packages import already_extracted
 
 file_path_cache = {}
 
-
 logger = get_script_logger("archivematica.mcp.client.extractContents")
 
+TRANSFER_DIRECTORY = "%transferDirectory%"
 
-def output_directory(file_path, date):
+
+def temporary_directory(file_path, date):
     if file_path_cache.get(file_path):
         return file_path_cache[file_path]
     else:
@@ -42,26 +43,60 @@ def tree(root):
             yield os.path.join(dirpath, file)
 
 
-def assign_uuid(job, filename, package_uuid, transfer_uuid, date, task_uuid, sip_directory, package_filename):
+def assign_uuid(job, filename, package_uuid, transfer_uuid, date, task_uuid,
+                sip_directory, package_filename, original_location,
+                sanitized_original_location):
+    """Assign a uuid to each file in the extracted package."""
     file_uuid = uuid.uuid4().__str__()
-    relative_path = filename.replace(sip_directory, "%transferDirectory%", 1)
-    relative_package_path = package_filename.replace(sip_directory, "%transferDirectory%", 1)
+    location_to_replace = format_subdir_path(sanitized_original_location,
+                                             sip_directory)
+    # Correct the information in the path strings sent to this function. First
+    # remove the SIP directory from the string. Second, make sure that file
+    # paths have not been modified for processing purpose, i.e. in
+    # Archivematica current terminology, sanitized.
+    relative_path = filename.replace(sip_directory, TRANSFER_DIRECTORY, 1)
+    transfer_path = relative_path.replace(location_to_replace,
+                                          os.path.join(original_location, ''))
+    relative_package_path = package_filename.replace(sip_directory,
+                                                     TRANSFER_DIRECTORY, 1)
     package_detail = "{} ({})".format(relative_package_path, package_uuid)
     event_detail = "Unpacked from: " + package_detail
     addFileToTransfer(relative_path, file_uuid, transfer_uuid, task_uuid, date,
-                      sourceType="unpacking", eventDetail=event_detail)
+                      sourceType="unpacking", eventDetail=event_detail,
+                      originalLocation=transfer_path)
     updateSizeAndChecksum(file_uuid, filename, date, uuid.uuid4().__str__())
-
     job.pyprint('Assigning new file UUID:', file_uuid, 'to file', filename)
 
 
-def _get_subdir_paths(root_path, path_prefix_to_repl):
+def _get_subdir_paths(job, root_path,
+                      path_prefix_to_repl,
+                      original_location):
     """Return a generator of subdirectory paths in ``root_path`` with
     the ancestor path ``path_prefix_to_repl`` replaced by a placeholder
-    string.
+    string. ``original_location`` should be the zip container that the content
+    was extracted from as it was transferred. Between then and now the path
+    will have been changed to remove characters that are difficult for tools
+    to handle. PREMIS describes the original location as how it was received
+    not what it was changed to, so we make these changes below to reflect the
+    original transfer material.
     """
-    return (format_subdir_path(dir_path, path_prefix_to_repl) for
-            dir_path, __, ___ in os.walk(root_path))
+    # Make sure that the original location is suffixed with a trailing slash
+    # as it is a directory.
+    original_location = os.path.join(original_location, '')
+    # Transform the processing directory string to reflect where it was when
+    # transferred, that is, in the ``%transferDirectory%``.
+    location_to_replace = format_subdir_path(root_path,
+                                             path_prefix_to_repl)
+
+    # Return a generator here that contains information about the current path
+    # and the original path for the PREMIS information in the METS file.
+    for dir_path, __, ___ in os.walk(root_path):
+        formatted_path = format_subdir_path(dir_path, path_prefix_to_repl)
+        for dir_uuid in get_dir_uuids(
+                [formatted_path], logger, printfn=job.pyprint):
+            dir_uuid["originalLocation"] = formatted_path\
+                .replace(location_to_replace, original_location)
+            yield dir_uuid
 
 
 def delete_and_record_package_file(job, file_path, file_uuid, current_location):
@@ -72,7 +107,8 @@ def delete_and_record_package_file(job, file_path, file_uuid, current_location):
 
 
 def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
-    files = File.objects.filter(transfer=transfer_uuid, removedtime__isnull=True)
+    files = File.objects.filter(transfer=transfer_uuid,
+                                removedtime__isnull=True)
     if not files:
         job.pyprint('No files found for transfer: ', transfer_uuid)
 
@@ -123,18 +159,28 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
                         file=sys.stderr)
             continue
 
-        file_path = file_.currentlocation.replace('%transferDirectory%', sip_directory)
+        output_file_path = file_.currentlocation.replace(TRANSFER_DIRECTORY,
+                                                         sip_directory)
 
+        # Temporarily rename the input package so that when we extract the
+        # contents we don't extract it to a directory that will conflict with
+        # the names we want to preserve in our PREMIS:originalLocation.
+        temp_dir = temporary_directory(output_file_path, date)
+        rename(output_file_path, temp_dir)
+
+        # Create the extract packages command.
         if command.script_type == 'command' or command.script_type == 'bashScript':
             args = []
             command_to_execute = command.command.replace('%inputFile%',
-                                                         file_path)
+                                                         temp_dir)
             command_to_execute = command_to_execute.replace('%outputDirectory%',
-                                                            output_directory(file_path, date))
+                                                            output_file_path)
         else:
             command_to_execute = command.command
-            args = [file_path, output_directory(file_path, date)]
+            args = [temp_dir, output_file_path]
 
+        # Make the command clear to users when inspecting stdin/stdout.
+        logger.info("Command to execute is: %s", command_to_execute)
         exitstatus, stdout, stderr = executeOrRun(command.script_type,
                                                   command_to_execute,
                                                   arguments=args,
@@ -145,37 +191,48 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
 
         if not exitstatus == 0:
             # Dang, looks like the extraction failed
-            job.pyprint('Command', command.description, 'failed!', file=sys.stderr)
+            job.pyprint('Command', command.description, 'failed!',
+                        file=sys.stderr)
         else:
             extracted = True
-            job.pyprint('Extracted contents from', os.path.basename(file_path))
+            job.pyprint('Extracted contents from',
+                        os.path.basename(output_file_path))
 
             # Assign UUIDs and insert them into the database, so the newly-
             # extracted files are properly tracked by Archivematica
-            extracted_path = output_directory(file_path, date)
-            for extracted_file in tree(extracted_path):
-                assign_uuid(job,
-                            extracted_file, file_.uuid, transfer_uuid, date,
-                            task_uuid, sip_directory, file_.currentlocation)
+            for extracted_file in tree(output_file_path):
+                assign_uuid(
+                    job, extracted_file, file_.uuid, transfer_uuid, date,
+                    task_uuid, sip_directory, file_.currentlocation,
+                    file_.originallocation, output_file_path)
 
-            # Assign UUIDs to directories via ``Directory`` objects in the
-            # database.
             if transfer_mdl.diruuids:
-                Directory.create_many(
-                    get_dir_uuids(
-                        _get_subdir_paths(extracted_path, sip_directory),
-                        logger,
-                        printfn=job.pyprint),
-                    transfer_mdl)
+                create_extracted_dir_uuids(
+                    job, transfer_mdl, output_file_path, sip_directory, file_)
 
-            # We may want to remove the original package file after extracting its contents
+            # We may want to remove the original package file after extracting
+            # its contents
             if delete:
-                delete_and_record_package_file(job, file_path, file_.uuid, file_.currentlocation)
+                delete_and_record_package_file(
+                    job, temp_dir, file_.uuid, file_.currentlocation)
 
     if extracted:
         return 0
     else:
         return 255
+
+
+def create_extracted_dir_uuids(
+        job, transfer_mdl, output_file_path, sip_directory, file_):
+    """Assign UUIDs to directories via ``Directory`` objects in the database.
+    """
+    Directory.create_many(
+        _get_subdir_paths(
+            job,
+            output_file_path,
+            sip_directory,
+            file_.originallocation),
+        transfer_mdl)
 
 
 def call(jobs):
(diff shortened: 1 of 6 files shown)


sallain:
Related to #1094

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -145,32 +189,38 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
...
 
             # Assign UUIDs and insert them into the database, so the newly-
             # extracted files are properly tracked by Archivematica
-            extracted_path = output_directory(file_path, date)
-            for extracted_file in tree(extracted_path):
-                assign_uuid(job,
-                            extracted_file, file_.uuid, transfer_uuid, date,
-                            task_uuid, sip_directory, file_.currentlocation)
+            for extracted_file in tree(output_file_path):
+                assign_uuid(extracted_file, file_.uuid, transfer_uuid, date,
```
sevein: You need to pass the `job` too.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -80,9 +81,12 @@ def insertIntoFiles(fileUUID, filePath, enteredSystem=None, transferUUID="", sip
     if enteredSystem is None:
         enteredSystem = getUTCDate()
 
+    if not originalLocation:
+        originalLocation = filePath
```
sevein, line 86: You could test this in `archivematicaCommon/tests/test_database_functions.py` where there's already some coverage for this function.

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -64,7 +64,8 @@ def getDeciDate(date):
     return str("{:10.10f}".format(float(ret)))
 
 
-def insertIntoFiles(fileUUID, filePath, enteredSystem=None, transferUUID="", sipUUID="", use="original"):
+def insertIntoFiles(fileUUID, filePath, enteredSystem=None, transferUUID="",
+                    sipUUID="", use="original", originalLocation=None):
```
sevein, line 68: `originalLocation` needs to be added to the docstring.

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -145,32 +189,38 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
...
+                    get_dir_uuids(subdir_paths,
+                                  logger,
+                                  printfn=job.pyprint), transfer_mdl)
 
-            # We may want to remove the original package file after extracting its contents
+            # We may want to remove the original package file after extracting
+            # its contents
             if delete:
-                delete_and_record_package_file(job, file_path, file_.uuid, file_.currentlocation)
+                delete_and_record_package_file(temp_dir, file_.uuid,
```
sevein: Here also, you need to pass the `job` too.

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -15,19 +15,20 @@
 from custom_handlers import get_script_logger
 from executeOrRunSubProcess import executeOrRun
 from databaseFunctions import fileWasRemoved
-from fileOperations import addFileToTransfer, updateSizeAndChecksum
+from fileOperations import addFileToTransfer, updateSizeAndChecksum, rename
 from archivematicaFunctions import get_dir_uuids, format_subdir_path
 
 # clientScripts
 from hasPackages import already_extracted
```
sevein: This import may not work, #1105 renamed the  `hasPackages` module to `has_packages`.

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -145,32 +189,38 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
...
+                                                 file_.originallocation)
                 Directory.create_many(
-                    get_dir_uuids(
-                        _get_subdir_paths(extracted_path, sip_directory),
-                        logger,
-                        printfn=job.pyprint),
-                    transfer_mdl)
+                    get_dir_uuids(subdir_paths,
+                                  logger,
+                                  printfn=job.pyprint), transfer_mdl)
```
sevein: Nitpick: you don't have to take every chance to alter the code style. When you do there's more noise in the diff which makes harder for the reader to understand the changes.

Here's the original block just with the functional change you wanted to make:

```python
            if transfer_mdl.diruuids:
                Directory.create_many(
                    get_dir_uuids(
                        _get_subdir_paths(
                            output_file_path, sip_directory,
                            file_.originallocation),
                        logger,
                        printfn=job.pyprint),
                    transfer_mdl)
```

The diff will look like:

```diff
4c4,6
<                         _get_subdir_paths(extracted_path, sip_directory),
---
>                         _get_subdir_paths(
>                             output_file_path, sip_directory,
>                             file_.originallocation),
```

Instead of:

```diff
1a2,4
>                 subdir_paths = _get_subdir_paths(output_file_path,
>                                                  sip_directory,
>                                                  file_.originallocation)
3,7c6,8
<                     get_dir_uuids(
<                         _get_subdir_paths(extracted_path, sip_directory),
<                         logger,
<                         printfn=job.pyprint),
<                     transfer_mdl)
---
>                     get_dir_uuids(subdir_paths,
>                                   logger,
>                                   printfn=job.pyprint), transfer_mdl)
```

src/MCPClient/lib/clientScripts/extract_contents.py
sevein: Sad that the tests didn't find this.

sevein (changes requested):
Looking good! I've added some comments. I think that the PR may be broken as it is right now, probably a result of a recent rebase?

> For discussion should I replace the use of a tuple in all instances where we use the Directory create_many member function? https://github.com/artefactual/archivematica/search?q=create_many&unscoped_q=create_many. There doesn't seem that many, and I think I prefer the verbosity.

I'm personally struggling trying to make a recommendation here. Both `get_dir_uuids` and `create_many` were really simple which I like, just working with a 2-tuple, easy to read and test. I'd encourage you to keep it that way if possible. You could implement your own local generator wrapping `get_dir_uuids` but to be honest I'm not certain wether that's going to be any better. I'd like to hear @jrwdunham's opinion as he wrote the original functions.

src/MCPClient/lib/clientScripts/extract_contents.py
ross-spencer: I don't think this was me taking an opportunity to changing the code style as much as me wrestling with how to make my changes work with PEP8 recommendations as much as anything. I'm learning how to do this more elegantly each day. I do recognise your suggestion is a nicer approach to style and will make this a change in the review changes. :+1:

src/MCPClient/lib/clientScripts/extract_contents.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/extract_contents.py
sevein: Cool that makes sense, thank you!

src/MCPClient/lib/clientScripts/extract_contents.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/extract_contents.py
ross-spencer: @sevein given this isn't showing up when rebasing, this needs to be changed in the main `qa/1.x` branch? https://github.com/artefactual/archivematica/blob/f996d47073a26f98ba5d0504686c0de9fae13d4d/src/MCPClient/lib/clientScripts/extract_contents.py

sevein: That'd be ideal but you can just fix here if you don't want to spend time submitting another PR.

sevein: See https://github.com/artefactual/archivematica/pull/1219.

src/archivematicaCommon/lib/databaseFunctions.py
ross-spencer: :+1:

ross-spencer:
@sevein this took some time (docker woes, and logic fails!) but I think I have a solution, do you think you have time to provide feedback on my changes for Monday so that we can fix the issues for IISH?

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -74,15 +75,19 @@ def insertIntoFiles(fileUUID, filePath, enteredSystem=None, transferUUID="", sip
     :param str transferUUID: UUID for the transfer containing this file. Can be empty. At least one of transferUUID or sipUUID must be defined. Mutually exclusive with sipUUID.
     :param str sipUUID: UUID for the SIP containing this file. Can be empty. At least one of transferUUID or sipUUID must be defined. Mutually exclusive with transferUUID.
     :param str use: A category used to group the file with others of the same kind. Will be included in the AIP's METS document in the USE attribute. Defaults to "original".
+    :param str oriignalLocation: where the original location of the file needs to be recorded, such as premis:OriginalName fields, it can be set using this parameter here.
```
sevein: Typo: s/oriignalLocation/originalLocation

src/archivematicaCommon/lib/fileOperations.py
```diff
@@ -60,9 +60,17 @@ def updateSizeAndChecksum(fileUUID, filePath, date, eventIdentifierUUID, fileSiz
...
+                      date, sourceType="ingestion", eventDetail="",
+                      use="original", originalLocation=None):
+    if not originalLocation:
+        originalLocation = filePathRelativeToSIP
+    insertIntoFiles(fileUUID,
+                    filePathRelativeToSIP,
+                    date,
+                    transferUUID=transferUUID,
+                    use=use,
+                    originalLocation=originalLocation)
```
sevein: Since you mentioned that you're still exploring better ways to style the code, I'm going to make a suggestion about the way of writing this function call. If you replaced the unused space with underscores you can see that there's a lot of real estate wasted:

```
insertIntoFiles(fileUUID,
________________filePathRelativeToSIP,_________________________________________
________________date,__________________________________________________________
________________transferUUID=transferUUID,_____________________________________
________________use=use,_______________________________________________________
________________originalLocation=originalLocation)_____________________________
```

Two alternatives that are far more common in our codebase:

```python
insertIntoFiles(
    fileUUID, filePathRelativeToSIP, date,
    transferUUID=transferUUID, use=use, originalLocation=originalLocation)
```

```python
insertIntoFiles(fileUUID, filePathRelativeToSIP, date,
                transferUUID=transferUUID, use=use,
                originalLocation=originalLocation)
```

Both alternatives fit in three lines instead of six. I think they also align better with Python's conciseness where you usually have a lot of things happening within one single line.

src/archivematicaCommon/tests/test_database_functions.py
```diff
@@ -35,8 +34,42 @@ def test_insert_into_files_raises_if_both_sip_and_transfer_provided(self):
...
+        # inaccurately.
+        databaseFunctions.insertIntoFiles(
+            fileUUID="e0a1fdc4-605a-4104-bf59-039859ee8238",
+            filePath=file_path,
+            sipUUID="0049fa6c-152f-44a0-93b0-c5e856a02292",
+            originalLocation=original_location)
+        assert File.objects.filter(originallocation=original_location)\
+            .count() == 1
+        assert File.objects.filter(originallocation=original_location)[0]\
+            .currentlocation != original_location
```
sevein: I think that the following is equivalent. With `get()` you're already checking that there's only one because otherwise a `MultipleObjectsReturned` exception is raised.

```python
assert File.objects.get(
    originallocation=original_location).currentlocation != original_location
```

src/dashboard/src/main/models.py
```diff
@@ -481,18 +481,28 @@ def __unicode__(self):
         }
 
     @classmethod
-    def create_many(cls, dir_paths_uuids, unit_mdl, unit_type='transfer'):
+    def create_many(cls, dir_paths_uuids, unit_mdl, unit_type='transfer',
+                    original_location=None):
```
sevein: So `original_location` is not used anymore right?

src/dashboard/src/main/models.py
```diff
@@ -481,18 +481,28 @@ def __unicode__(self):
...
-        directory path/UUID pair in ``dir_paths_uuids`` and the ``Transfer`` model
-        that the directories are a part of.
+        directory path/UUID entry in ``dir_paths_uuids`` and the ``Transfer``
+        model that the directories are a part of. ``dir_paths_uuids`` is
+        expected to be a dict instance where the ``originallocation`` field is
+        optional. The ``originallocation`` but can be set according to the
+        requirements of the PREMIS record that eventually needs to be created
+        using this model.
         """
-        unit_type = {'transfer': 'transfer'}.get(unit_type, 'sip')
```
sevein: This line should not be removed. It's a filter that replaces unknown values with `sip`.

| Input | Output | Comment |
| - | - | - |
| `None` | `"transfer"` | Default value in function argument kicks in. |
| `"transfer"` | `"transfer"` | The item is found in the dictionary and its value is returned. |
| `"sip"` | `"sip"` | The item is not found in the dictionary, default provided is used instead. |
| `"unknown"` | `"sip"` | The item is not found in the dictionary, default provided is used instead. |

With that filter removed:

| Input | Output |
| - | - |
| `None` | `"transfer"` |
| `"transfer"` | `"transfer"` |
| `"sip"` | `"sip"` |
| `"unknown"` | `"unknown"` |

src/dashboard/src/main/models.py
```diff
@@ -481,18 +481,28 @@ def __unicode__(self):
...
-        return cls.objects.bulk_create([
-            cls(**{'uuid': dir_uuid,
-                   unit_type: unit_mdl,
-                   'originallocation': dir_path,
-                   'currentlocation': dir_path})
-            for dir_path, dir_uuid in dir_paths_uuids])
+        paths = []
+        for dir_ in dir_paths_uuids:
+            orig_path = None
```
sevein: Not needed?

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -145,39 +191,52 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
...
     if extracted:
         return 0
     else:
         return 255
 
 
+def create_extracted_dir_uuids(
+        job, transfer_mdl, output_file_path, sip_directory, file_):
+    # Assign UUIDs to directories via ``Directory`` objects in the
+    # database.
```
sevein: Because this is now a function, the comment should be a proper docstring.

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -145,39 +191,52 @@ def main(job, transfer_uuid, sip_directory, date, task_uuid, delete=False):
 
         if not exitstatus == 0:
             # Dang, looks like the extraction failed
-            job.pyprint('Command', command.description, 'failed!', file=sys.stderr)
+            job.pyprint('Command',
+                        command.description,
+                        'failed!',
+                        file=sys.stderr)
```
sevein: Perhaps this is better? We don't have to go all vertical, just enough so we don't exceed 79 columns.

```python
            job.pyprint('Command', command.description, 'failed!',
                        file=sys.stderr)
```

Fits in two lines versus five.

sevein (changes requested):
Looking good. I've added some smaller observations.

src/archivematicaCommon/tests/test_database_functions.py
sevein: Feel free to ignore, yours is also okay - I appreciate being explicit plus you'd see the count when it's not just one, which is something you wouldn't see using my suggested alternative.

src/archivematicaCommon/lib/fileOperations.py
ross-spencer: Hi @sevein I think this is a good example of where our discussion about using `black` or alternative would save a lot of CR time. 

`    fileUUID, filePathRelativeToSIP, date, transferUUID=transferUUID, `

Still fits within 79 characters, so the are we just electing to split the line based on named vs. unnamed arguments? 

if you look at our migrations, then we use the vertical approach pretty consistently. 

I haven't been maintaining the code-base as long, so I guess I'm pretty easy going on either approach to PEP8 conformance. Once I had to make a decision about what does and doesn't fit inside 80 characters when there were different configurations it was just easier to make the function call use the vertical method. 

We could adopt a formatting tool, or provide some additional nuance to the CR docs? I.e. what real estate means to the AM code-base.

src/dashboard/src/main/models.py
ross-spencer: You're right. If it is provided in the `dict` then it is used, else it shouldn't be an argument provided to this function. I should have caught this.

src/dashboard/src/main/models.py
ross-spencer: This is a good example why we have code-review. It wasn't removed intentionally. I must have deleted it when clearing up the code. I should have caught this when I read through my diffs at the end.

src/archivematicaCommon/lib/fileOperations.py
sevein: Ok, agree. I'll avoid making more suggestions like this again so we save CR time. UPDATE: and make sure it's captured in the docs.

src/dashboard/src/main/models.py
ross-spencer: This function can create many new directory entries. If `originalLocation` is passed via the iterator that can be passed to this function then it overrides the `currentLocation` so I believe it is needed. Is there a reason you believe it shouldn't be here? 

The named argument `original_location=None` will be removed, so maybe this was the problem?

src/MCPClient/lib/clientScripts/extract_contents.py
ross-spencer: :+1:

src/dashboard/src/main/models.py
sevein: > Is there a reason you believe it shouldn't be here?

I only thought it was not needed because `orig_path` is assigned a new value in L499 anyways.

src/archivematicaCommon/lib/databaseFunctions.py
ross-spencer: :+1:

src/dashboard/src/main/models.py
ross-spencer: No worries :)

ross-spencer:
Thanks @sevein - I think this is okay now. If you've time can you look at the changes and see if they are appropriate to merge?

sevein (approved):

PR #1185, opened 2018-07-05
MCP batching followups

sevein:
Smaller fixes in addition to https://github.com/artefactual/archivematica/pull/1105. See the commits individually for more details.

This is connected to #938.

CC @marktriggs @payten @jambun


marktriggs:
That looks good to me, thanks @sevein!

I think I'll have a follow-on patch of my own soon, but I'm just doing some more testing.  We've been testing how things go when running multiple concurrent transfers, and I hit an interesting case where a transaction inserting rows into `Tasks` collided with a transaction updating (different!) rows in Tasks, and the second transaction got aborted.

I'll send more detail in a separate PR soon... I don't think it's a new bug, and it's quite hard to reproduce (even when I'm doing it deliberately), but might save some painful troubleshooting down the line if I fix it now :)

Issue #1136, opened 2018-06-13
Problem: processing is aborted after failure in Index AIP


sevein:
Pull request @ JiscRDSS: https://github.com/JiscRDSS/archivematica/pull/61.

joel-simpson:
Adding some background to this problem (JiscRDSS#61 points to documentation that is not available publicly: 

During our scalability testing on the Jisc project, we saw datasets fail the "Index AIP" microservice, particularly where they included a very large number of files. This microservice creates an index using ElasticSearch based on the contexts of the METS file.   If the METS is very large, then the document that ElasticSearch tries to index is too big and ElasticSearch throws an error. 
JiscRDSS#61 fixes this issue by ensuring that the workflow proceeds even if the Index AIP microservice fails. When indexing fails, the AIP can't be found using the Dashboard search functionality, but it is possible (as a workaround) to retrieve the AIP from the storage service.

joel-simpson:
This is working as expected. If the Indexing AIP fails, processing still completes. 
![image](https://user-images.githubusercontent.com/17187309/43232536-6ad40c78-903f-11e8-8f73-709559a4d59b.png)
(this was demonstrated in https://github.com/artefactual/archivematica/issues/1194 as well)

PR #1190, opened 2018-07-06
MCPClient: add indexAIP error handling

sevein:
This commit adds a new exit code branch (`179`) to the indexAIP chain link.
The client script reports this code when the desire is to set the job status as
failing but the processing should continue.

This is connected to #1136.

### DIFF
--- a/src/MCPClient/install/README.md
+++ b/src/MCPClient/install/README.md
@@ -146,6 +146,12 @@ This is the full list of variables supported by MCPClient:
     - **Type:** `boolean`
     - **Default:** `true`
 
+- **`ARCHIVEMATICA_MCPCLIENT_MCPCLIENT_INDEX_AIP_CONTINUE_ON_ERROR`**:
+    - **Description:** controls whether Archivematica continues processing the package when an error occurs in `indexAIP_v0.0`.
+    - **Config file example:** `MCPClient.index_aip_continue_on_error`
+    - **Type:** `boolean`
+    - **Default:** `false`
+
 - **`ARCHIVEMATICA_MCPCLIENT_MCPCLIENT_REMOVABLEFILES`**:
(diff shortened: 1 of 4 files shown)


jhsimpson (approved):
This looks good to me.  I tested by setting the `ARCHIVEMATICA_MCPCLIENT_MCPCLIENT_INDEX_AIP_CONTINUE_ON_ERROR` variable to yes in my docker-compose.yml, and rebulding the mcp client container.  After starting a transfer, I stopped the elasticsearch container.  The resulting SIP shows the 'index aip' job in red but continues on and stores the aip.

The AIP can't be found in the archival storage tab, but I was able to locate it in the storage service.

One comment, I noticed that the 'index aip' job has no stdout or stderr in the task details page.  The next job, 'Clean up after storing AIP' is also red, has stderr containing a stacktrace from the elasticsearch connection error.  

I didn't simulate the scenario that this feature is meant to address (when indexing starts but fails for some reason, such as a timeout or ES returning an error).  I tested the scenario when ES is not available.  But it is a bit confusing, the stderr should be in the 'index aip' job, shouldn't it, and not the one after?

:+1:

sevein:
> One comment, I noticed that the 'index aip' job has no stdout or stderr in the task details page. The next job, 'Clean up after storing AIP' is also red, has stderr containing a stacktrace from the elasticsearch connection error.

> I didn't simulate the scenario that this feature is meant to address (when indexing starts but fails for some reason, such as a timeout or ES returning an error). I tested the scenario when ES is not available. But it is a bit confusing, the stderr should be in the 'index aip' job, shouldn't it, and not the one after?

I will investigate, thanks.

sevein:
You're right @jhsimpson.

I think that this is a side effect introduced in #1105, I'll file an issue.
Update: issue filed - https://github.com/artefactual/archivematica/issues/1194.

Issue #1139, opened 2018-06-13
Problem: start transfer API endpoint is blocking


sevein:
In RDSS we started the development of a new`/api/v2beta/package/` API endpoint.

Pull request @ JiscRDSS: https://github.com/JiscRDSS/archivematica/pull/66. Also, https://github.com/JiscRDSS/archivematica/pull/73 - where we reverted the changes in the transfer browser widget due to an incompatibility with Shibboleth that needs to be investigated further.

joel-simpson:
I think the new beta package endpoint is a great addition for performance and reliability. While it had to be reverted in the Jisc production environment due to shibboleth issues, I did test it quite a bit in the Jisc QA environment. 

I also think that the asynchronous endpoint to create transfers may be necessary / useful for the work on rate limiting that we are currently looking at for the Jisc project (which is currently only documented on the wiki, but if we decide to proceed with it, I will create a public issue to document the concerns and rationale for that work).

sevein:
There was an early attempt in #936 but I'm going to be submitting a new PR.

sromkey:
@sevein Could you please restate the problem in a way that describes how this effects users? My understanding is that the connected PR removes the user approval step, non-optionally? I think this warrants discussion. (For the record I'm not opposed to the change, but it is a pretty big one to go through without further user discussion).

joel-simpson:
This problem arose during the Jisc project (so the description of the problem and the initial analysis was done elsewhere)... I will attempt to describe the problem better:

**Problem: Very large datasets can't be successfully transferred into Archivematica.** 
When selecting a very large directory (> 100s of GBs) using the Transfer tab of the Dashboard, the Transfers were not successfully started, and the user receives no feedback that the operation has failed. 
We encountered this using a fork of Archivematica 1.6 deployed using Docker containers in an AWS environment.

**Analysis of root cause:** When a directory is selected in the Transfer tab and the user selects "Start Transfer" the files from the Transfer Source Location selected are copied to the 'currently processing' location. This process relies on calls to the Archivematica and Storage Service APIs, which are synchronous. This means that those APIs "block" further work until their operations are complete. 

In the Jisc deployment there is a constraint that all API calls must complete within 1 hour. (In AWS, when an ALB (Application Load Balancer) is used, the maximum timeout setting is 1 hour.)
[...]

PR #1191, opened 2018-07-07
Add /api/v2beta/package endpoint

sevein:
This pull request adds a new API endpoint `/api/v2beta/package`.

The main differences with  `/api/transfer/start_transfer` are:

- Once the transfer arrives to Archivematica we start processing right away. We don't request the user to approve it.
- The server returns immediately. It does not block while the transfer is being copied which fits better with certain deployments, e.g. when Gunicorn uses sync workers or a load balancer can't hold a connection for too long.
- The response contains the UUID of the package - further development should make possible to request the status of the package.
- It does not accept multiple paths - you can still use the old API endpoint or send multiple requests.
- All the copying/processing happens in MCPServer. The Dashboard performs only as an intermediate. 

Other considerations:
- **The interface is not stable and it will likely change.**
- We're aware that the transfer browser fails to authenticate when Shibboleth is set up. This will need to be fixed. Before this PR, the transfer browser was using endpoints that are not part of the API namespace. The latter uses tastypie's authentication classes and that seems to be where the problem is.
- The `/api/transfer/start_transfer` endpoint needs to be updated to make use of the new `packageCreate` RPC method. Once this is ready we'll be able to remove some duplicate code that exists in both Dashboard and MCPServer.

Usage example:

```bash
#!/usr/bin/env bash

TRANSFER_UUID="de115557-8eae-4795-95c0-bec173664f22"
TRANSFER_PATH="${TRANSFER_UUID}:/home/interactive/test"
TRANSFER_PATH=$(echo -n ${TRANSFER_PATH} | base64 | tr -d '\n')
HOST="127.0.0.1:32789"
NUM=$(shuf -i 1-100000 -n 1)

curl -v \
  -X POST \
  -H "Content-Type: application/json" \
  -H "Authorization: ApiKey test:test" \
        -d "{\"name\": \"Test${NUM}\", \"type\": \"standard\", \"path\": \"${TRANSFER_PATH}\"}" \
    http://${HOST}/api/v2beta/package/
```

Compare with the old method: https://gist.github.com/sevein/4cf84be9400a696be11d3808f2bd6318.

This endpoint is going to be adopted in:
- The transfer browser widget (see 4600f25),
- Automation tools (see [transfer_async](https://github.com/artefactual/automation-tools/blob/master/transfers/transfer_async.py)).
- Channel adapter (work in progress).

This is connected to #1139.

### DIFF
--- a/src/MCPServer/install/README.md
+++ b/src/MCPServer/install/README.md
@@ -122,6 +122,18 @@ This is the full list of variables supported by MCPServer:
     - **Type:** `boolean`
     - **Default:** `true`
 
+- **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_STORAGE_SERVICE_CLIENT_TIMEOUT`**:
+    - **Description:** configures the Storage Service client to stop waiting for a response after a given number of seconds.
+    - **Config file example:** `MCPClient.storage_service_client_timeout`
+    - **Type:** `float`
+    - **Default:** `86400`
+
+- **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_STORAGE_SERVICE_CLIENT_QUICK_TIMEOUT`**:
+    - **Description:** configures the Storage Service client to stop waiting for a response after a given number of seconds. This applies only to requests that are supposed to return immediately.
+    - **Config file example:** `MCPClient.storage_service_client_timeout`
+    - **Type:** `float`
+    - **Default:** `5`
+
 - **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_WATCHDIRECTORIESPOLLINTERVAL`**:
     - **Description:** time in seconds between filesystem poll intervals.
     - **Config file example:** `MCPServer.watchDirectoriesPollInterval`
--- a/src/MCPServer/lib/RPCServer.py
+++ b/src/MCPServer/lib/RPCServer.py
@@ -1,6 +1,8 @@
+"""Archivematica MCPServer RPC."""
+
 # This file is part of Archivematica.
 #
-# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
+# Copyright 2010-2018 Artefactual Systems Inc. <http://artefactual.com>
 #
 # Archivematica is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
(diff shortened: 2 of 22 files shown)


sevein:
CC @marktriggs hey you may want to take a look at this pull request. It's a new way to start transfers that we were already using in our RDSS fork. I've updated it so it uses your `taskThreadPool`. Namely, It uses the `apply_async` method but I've left some code that allows us to use the synchronous `apply` as well so we can reuse the same code dealing to perform transfers the old way (`/api/transfer/start_transfer`) - just that I haven't updated the old API endpoint.

src/MCPServer/install/README.md
```diff
@@ -122,6 +122,18 @@ This is the full list of variables supported by MCPServer:
...
+    - **Description:** configures the Storage Service client to stop waiting for a response after a given number of seconds.
+    - **Config file example:** `MCPClient.storage_service_client_timeout`
+    - **Type:** `float`
+    - **Default:** `86400`
+
+- **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_STORAGE_SERVICE_CLIENT_QUICK_TIMEOUT`**:
+    - **Description:** configures the Storage Service client to stop waiting for a response after a given number of seconds. This applies only to requests that are supposed to return immediately.
+    - **Config file example:** `MCPClient.storage_service_client_timeout`
+    - **Type:** `float`
+    - **Default:** `5`
```
sevein, line 135: The new `package` module in MCPServer responsible for starting transfers/packages talks to Storage Service. As we had never talked to Storage Service from MCPServer before we did not have settings. The `storageService` module does not provide defaults so we have to provide them instead.

src/MCPServer/lib/RPCServer.py
```diff
@@ -15,74 +17,116 @@
...
     gm_worker = gearman.GearmanWorker([django_settings.GEARMAN_SERVER])
     hostID = gethostname() + "_MCPServer"
     gm_worker.set_client_id(hostID)
-    gm_worker.register_task("approveJob", gearmanApproveJob)
-    gm_worker.register_task("getJobsAwaitingApproval", gearmanGetJobsAwaitingApproval)
+
+    # The tasks registered in this worker should not block.
+    gm_worker.register_task("approveJob", job_approve_handler)
+    gm_worker.register_task("getJobsAwaitingApproval", job_awaiting_approval_handler)
+    gm_worker.register_task("packageCreate", package_create_handler)
```
sevein, line 128: `packageCreate` is really the only endpoint I'm adding this time. But I took the chance to introduce some decorators that I think can be helpful.

src/MCPServer/lib/package.py
```diff
@@ -0,0 +1,380 @@
+"""Package management."""
```
sevein, line 1: A lot of the code here already exists in the Dashboard. We can get rid of the latter once we update the old start_transfer endpoint so it uses the new RPC method.

src/dashboard/frontend/transfer-browser/app/services/transfer.service.js
```diff
@@ -39,24 +39,38 @@ class Transfer {
...
-    let params = {
-      name: name,
-      type: this.type,
-      accession: this.accession,
-      'paths[]': this.components.map(component => Base64.encode(`${component.location}:${component.path}`)),
-      'row_ids[]': this.components.map(component => component.id || ''),
-    };
-
-    // Cleanup object state on success or failure
-    let promise = jQuery.post('/filesystem/transfer/', params);
```
sevein, line 57: Notice that we were not using the `/api` namespace before. Now that we do we've found that the request sent from this file won't authenticate when the deployment uses Shibboleth. I suspect the problem is here:

https://github.com/artefactual/archivematica/blob/0894906bf6f0401bd766285f7f14606a70f0ea70/src/dashboard/src/components/api/views.py#L99-L104

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -201,31 +201,6 @@ def delete_arrange(request, filepath=None):
     return helpers.json_response({'message': _('Delete successful.')})
 
 
-def start_transfer_logged_in(request):
```
sevein, line 204: This was the old endpoint used by the transfer browser widget.

src/dashboard/src/contrib/mcp/client.py
```diff
@@ -56,3 +63,25 @@ def notifications(self):
...
+                       auto_approve=True, wait_until_complete=False):
+        gm_client = gearman.GearmanClient([self.server])
+        data = cPickle.dumps({
+            'name': name,
+            'type': type_,
+            'accession': accession,
+            'path': path,
+            'metadata_set_id': metadata_set_id,
+            'auto_approve': auto_approve,
+            'wait_until_complete': wait_until_complete,
```
sevein, line 77: We're not using `auto_approve` and `wait_until_complete` at the moment but I'm hoping they'll do the trick when we have time to refactor the old `start_transfer` endpoint to use this RPC method.

jhsimpson:
I tested this, and once I correctly set the transfer source location uuid, I got back a new uuid for the package and the transfer appeared in the dashboard and processed.

When  I set the transfer source location uuid incorrectly, there was an error in the mcp server log, but there was no other indication.  The http response from the api call was still 202 (accepted). This is not the behaviour I expected, but maybe it is correct.  

Either way, adding a method for returning package status info is out of scope for this PR.  So it looks great to me.

:+1:

jhsimpson (approved):

sevein:
I'm going to merge. I'm sorry if this is happening too soon. Code freeze is around the corner. If you have feedback feel free to file new issues or submit a PR against the original issue.

marktriggs:
Cool, thanks for the heads up @sevein.  Looks good, and allowed me to cross "Figure out what subjobof means" off my todo list :)

joel-simpson:
In the issue connected to this PR (#1139), @sromkey asked about the change included here that removes the "Approve Transfer" step. 

My understanding is that the "Approve Transfer" step exists to allow users to review the contents of the Transfer to confirm that it is complete, before any processing (microservices) begin. 

This is important in the current (i.e. before this PR) design, because processing of the transfer is triggered using a watched directory. Selecting "start transfer" instructs the Dashboard to copy the files from the Transfer Source location selected to the 'currently processing' location. Archivematica 'watches' directories in this location and when it detects a new directory it will trigger processing of that directory. The problem with watched directories is that Archivematica doesn't know for sure when copying has completed. It is possible, particularly for large transfers, that not all of the files have been successfully copied over. 
The "Approve Transfer" allows users to inspect that directory and confirm it is complete before any processing takes place. (whether this is commonly done is another question - it isn't actually possible to do from the Archivematica dashboard, so users need access to the file storage, directly or through some other client, to do this)

This PR removes the need for the watched directory. Moving files is under the control of the API which is able to confirm when the copying is complete, and only trigger processing from that stage onward. That makes the "Approve Transfer" step redundant. At least that is true if our understanding of the purpose of that step is correct... I couldn't find any documentation on this so there may be other reasons for it?

Issue #1192, opened 2018-07-07
Problem: can't set custom premis agent version number in qa or dev environments


jhsimpson:
When deploying Archivematica from the qa/1.x branch, the version number is currently set to `1.7.0` - which is not accurate.  No version number is technically accurate, since the code in qa/1.x has not been officially release yet.

The version number is set in more than one place - in the UI it is populated from version.py in archivematicaCommon, in the METS files inside AIP's, the premis agent for archivematica has an identifiervalue that is read out of the database.

At a minimum, there should be a way to override the database value for the archivematica premis agent, so that aip's produced by a qa system are mistakenly marked as being made by production code.   It would be better to be able to set this value at buildtime/deploytime, rather than have to add a new django migration for each change.

Changing the version number in the UI would be nice, but is much less important,.

This is important for users like JiscRDSS - who are planning on deploying a production system from a qa branch of archivematica.

sevein:
Related:
- https://github.com/artefactual/archivematica/issues/658
- https://github.com/artefactual/archivematica/issues/1183

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Jisc RDSS](https://trello.com/c/onaPPfEg/3-jisc-rdss)

peterVG:
I'm not clear on how to test for this?

jraddaoui:
Executed:
```
docker-compose run \
        --rm \
        --entrypoint /src/dashboard/src/manage.py \
            archivematica-dashboard \
                set_agent_code v1.8.0-rc1
```

Created an AIP and found the following PREMIS agent in the METS and pointer files:
```
<premis:agent xmlns:premis="info:lc/xmlns/premis-v2" xsi:schemaLocation="info:lc/xmlns/premis-v2 http://www.loc.gov/standards/premis/v2/premis-v2-2.xsd" version="2.2">
    <premis:agentIdentifier>
        <premis:agentIdentifierType>preservation system</premis:agentIdentifierType>
        <premis:agentIdentifierValue>v1.8.0-rc1</premis:agentIdentifierValue>
    </premis:agentIdentifier>
    <premis:agentName>Archivematica</premis:agentName>
    <premis:agentType>software</premis:agentType>
</premis:agent>
```

It also appears as the agent code in `.../administration/version/`.

PR #1193, opened 2018-07-07
Dashboard: set_agent_code command

sevein:
This command enables users to override the Archivematica identifier
(`Agent.pk=1`), also known as the Archivematica PREMIS agent.

Usage example seen in the RDSS project:
```
# Set agent code in Dashboard
docker-compose run \
	--rm \
	--entrypoint /src/dashboard/src/manage.py \
		archivematica-dashboard \
			set_agent_code v0.6.0-14-g89693bb
```

Optionally you can use the `--pk=` argument to pass a different primary key. You can only update existing agents.

Connects to https://github.com/artefactual/archivematica/issues/1192.
Based on https://github.com/JiscRDSS/archivematica/pull/48.

### DIFF
--- a/src/dashboard/src/components/administration/management/commands/set_agent_code.py
+++ b/src/dashboard/src/components/administration/management/commands/set_agent_code.py
@@ -1,0 +1,22 @@
+from django.core.management.base import BaseCommand
+
+from installer.steps import set_agent_code
+from main.models import Agent
+
+
+class Command(BaseCommand):
+    help = """Set the code of an agent.
+
+    By default it updates the agent code of the Archivematica preservation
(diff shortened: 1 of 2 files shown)


jhsimpson (approved):
I reviewed the code and it looks good, I have no suggested changes.

I tested this and it works as advertised.  I set a new agent code with the management command and made an aip. The aip contains premis agent xml like so:

```xml
<premis:linkingAgentIdentifier>
  <premis:linkingAgentIdentifierType>preservation system</premis:linkingAgentIdentifierType>
  <premis:linkingAgentIdentifierValue>its-coming-home</premis:linkingAgentIdentifierValue>
</premis:linkingAgentIdentifier>
```

Issue #1187, opened 2018-07-05
Problem: MCPServer internals can't be monitored


sevein:
Currently we don't have the ability to monitor the internals of MCPServer.
This makes really hard to reason about errors.

This is a good start: https://prometheus.io/docs/practices/instrumentation/.

ross-spencer:
Hi @sevein, outside of logging which we can control and monitor quite easily, I'm not sure what internals means in this issue?

sevein:
We're thinking on details that are not necessarily actionable, e.g. we have pools of threads so how long do threads take to complete work (histogram), how often they error (counter), how many items are queued (gauge)...

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Jisc RDSS](https://trello.com/c/onaPPfEg/3-jisc-rdss)

sevein:
#1206 added some basic instrumentation exposed via a HTTP server. The server is enabled only when `ARCHIVEMATICA_MCPSERVER_MCPSERVER_PROMETHEUS_HTTP_SERVER` is defined. cURL can be used to query the server and retrieve the samples. In a normal setup, the user would run Prometheus to collect the samples and some other tools for graphing, alerting, etc...

jraddaoui:
Hi @mamedin, passing this to you as you have some experience with Prometheus/Grafana.

mamedin:
The #1206 has been used and tested with a Prometheus-Grafana instrumentation solution in JISC/RDSS 3 months ago: https://github.com/JiscRDSS/rdss-archivematica/commit/5f685ccd64645c089e103adc5c781006a20fcb8c

I tested again adding the line to the /etc/default/archivematica-mcp-server file:
```
ARCHIVEMATICA_MCPSERVER_MCPSERVER_PROMETHEUS_HTTP_SERVER=127.0.0.1:12345
```
Restarting mcp-server and checking the new prometheus http server:

```
wget 51.38.224.179:12345
```
I got the file:
```ubuntu@mamedin-bionic:~$ cat index.html 
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 2752106496.0
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 59232256.0
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
[...]

PR #1206, opened 2018-07-11
MCPServer: add active unit gauge with prometheus_client

sevein:
Connects to https://github.com/artefactual/archivematica/issues/1187.

### DIFF
--- a/src/MCPServer/lib/archivematicaMCP.py
+++ b/src/MCPServer/lib/archivematicaMCP.py
@@ -36,6 +36,7 @@ import logging
 import logging.config
 import getpass
 import os
+import re
 import signal
 import sys
 import threading
@@ -44,10 +45,10 @@ import uuid
 
 import django
 django.setup()
-
 from django.conf import settings as django_settings
 from django.db.models import Q
 from django.utils import six
+from prometheus_client import start_http_server
 
 # This project, alphabetical by import source
 import watchDirectory
@@ -68,6 +69,7 @@ import dicts
 
 from main.models import Job, SIP, Task, WatchedDirectory
 from multiprocessing.pool import ThreadPool
+
 
 taskThreadPool = ThreadPool(django_settings.LIMIT_TASK_THREADS)
 
@@ -317,6 +319,35 @@ def created_shared_directory_structure():
         processing.install_builtin_config(config)
 
 
+naiveip_re = re.compile(r"""^(?:
+(?P<addr>
+    (?P<ipv4>\d{1,3}(?:\.\d{1,3}){3}) |         # IPv4 address
+    (?P<ipv6>\[[a-fA-F0-9:]+\]) |               # IPv6 address
+    (?P<fqdn>[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*) # FQDN
+):)?(?P<port>\d+)$""", re.X)
+
+
+def start_prometheus_http_server(addrport):
+    if not addrport:
+        return
+    m = re.match(naiveip_re, addrport)
+    if m is None:
+        logger.error('[prometheus_http_server]'
+                     ' "%s" is not a valid port number or address:port pair.',
+                     addrport)
+        return
+    addr, _ipv4, _ipv6, _fqdn, port = m.groups()
+    try:
+        port = int(port)
+    except ValueError:
+        logger.error('[prometheus_http_server]'
+                     ' "%r" is not a valid port number.', port)
+        return
+    if addr is None:
+        addr = '127.0.0.1'
+    start_http_server(*(port, addr))
+
+
 def _except_hook_log_everything(exc_type, exc_value, exc_traceback):
     """
     Replacement for default exception handler that logs exceptions.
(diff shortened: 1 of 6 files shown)


sevein:
@marktriggs can you take a look? The Prometheus client library has a global `CollectorRegistry`, that's where the samples are kept until a Prometheus server collects them and puts them in the time series database.

I suspect we still want to do the sampling + max() that you implemented.

sevein:
Example of document generated by the prometheus http interface:

```
# HELP task_group_runner_active_units Number of units currently being processed
# TYPE task_group_runner_active_units gauge
task_group_runner_active_units 3.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="2",minor="7",patchlevel="15",version="2.7.15"} 1.0
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 3427557376.0
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 63528960.0
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1531346625.42
# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 19.240000000000002
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 14.0
# HELP process_max_fds Maximum number of open file descriptors.
# TYPE process_max_fds gauge
process_max_fds 1048576.0
# HELP rpc_job_package_create_request_time Time spent processing request
# TYPE rpc_job_package_create_request_time summary
rpc_job_package_create_request_time_count 3.0
rpc_job_package_create_request_time_sum 0.026831865310668945
# HELP request_processing_seconds Time spent processing request
# TYPE request_processing_seconds summary
request_processing_seconds_count 0.0
request_processing_seconds_sum 0.0
# HELP rpc_job_awaiting_approval_request_time Time spent processing request
# TYPE rpc_job_awaiting_approval_request_time summary
rpc_job_awaiting_approval_request_time_count 108.0
rpc_job_awaiting_approval_request_time_sum 0.040329933166503906
# HELP rpc_job_approve_request_time Time spent processing request
# TYPE rpc_job_approve_request_time summary
rpc_job_approve_request_time_count 0.0
rpc_job_approve_request_time_sum 0.0
```

marktriggs:
@sevein Cool, that looks good.  And yeah, I think the sampling is still needed.

It's going to update that gauge value 5 times per second but it looks like that won't cause problems... it seems like each update just replaces the previous value, so not chewing up memory or anything like that.

sevein:
Here's the gauge in action using Prometheus's graphing tool. It's showing a period where ten transfers were started using the new Package endpoint. The samples captured are higher than the number of expected units at some point. I suspect this is because we have a few entities (Transfer, SIP, DIP...) related to the same package, not something that we can fix now.

image

src/MCPServer/lib/taskGroupRunner.py
```diff
@@ -218,6 +241,12 @@ def _monitor_running_jobs(self, gm_client):
                         len(self.task_group_jobs_by_uuid))
             self.last_notification_time = now
 
+        active_count = len(set([task_group.task_group.unit_uuid
```
marktriggs: Oops, this is a bug from me.  I meant to write:

```
 active_count = len(set([task_group.task_group.unit_uuid()
```

(i.e. call `unit_uuid` as a method--missed the trailing parens!)

It works by accident at the moment.

marktriggs:
@sevein The inflated numbers are interesting... with a single transfer, I can see the unit UUID switching as it moves from Transfer to SIP, but I was surprised it got counted twice since I thought those stages wouldn't overlap.

I think it might just be that the move_transfer client script has finished moving the transfer into the watched directory but hasn't yet returned its result to Gearman.  Then, the watched directory picks up the new SIP and sends the task to the task runner, which counts it.  So yeah.  Tricky :)

sevein:
Ok, fixed. Rewrote your commit.

marktriggs:
Much obliged, thanks :)

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -338,6 +339,8 @@ def _except_hook_log_everything(exc_type, exc_value, exc_traceback):
     logger.info('This PID: %s', os.getpid())
     logger.info('User: %s', getpass.getuser())
 
+    start_http_server(7999)
```
sevein: I need to make this configurable.

sevein:
I've added a new `PROMETHEUS_HTTP_SERVER` setting that defaults to empty string. The Prometheus server is only created when the setting carries a non-empty string, e.g. forms `12345` or `127.0.0.1:12345` are accepted - ignored otherwise.

PR #1208, opened 2018-07-12
Fix JSON->CSV metadata conversion

jrwdunham:
Fixes logic and adds tests to confirm that the following are true.

- The headers of the CSV are the union of all attributes of all objects in the JSON array: `[{'a': ...}, {'b': ...}]` yields `'a,b\r\n'`.
- JSON object attribute values can be integers or floats.
- The order of the JSON object attributes does not matter.
- A JSON object value may be an array or a string: `[{'a': ['x', 'y']}, {'a': 'z'}]` yields `'a,a\r\nx,y\r\nz,\r\n'`.

Connected to archivematica/Issues#7


sevein (approved):
Nicely done!

PR #1209, opened 2018-07-17
MCPClient: job.py allow Unicode printing

jrwdunham:
Adds exception handling to the IO-related methods in MCPClient's job.py module so that Unicode objects can be "written" and "printed".

Connected to archivematica/Issues#16

### DIFF
--- a/src/MCPClient/tests/test_job.py
+++ b/src/MCPClient/tests/test_job.py
@@ -1,0 +1,23 @@
+import os
+import sys
+from uuid import uuid4
+
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.append(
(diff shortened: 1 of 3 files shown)


src/MCPClient/tests/test_job.py
```diff
@@ -0,0 +1,24 @@
...
+from uuid import uuid4
+
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+sys.path.append(
+    os.path.abspath(os.path.join(THIS_DIR, '../lib/clientScripts')))
+from job import Job
+
+
+def test_job(tmpdir):
+    assert True
```
sevein: !

sevein (approved):

PR #1210, opened 2018-07-18
Unicode issues

marktriggs:
Fix issues with multibyte and shell characters in filenames.  Discussion here: https://github.com/archivematica/Issues/issues/16.

Connects to https://github.com/archivematica/Issues/issues/16.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -120,7 +120,7 @@ def handle_batch_task(gearman_job):
 
         job = Job(gearman_job.task,
                   task_data['uuid'],
-                  shlex.split(arguments),
+                  _parse_command_line(arguments),
                   caller_wants_output=task_data['wants_output'])
         jobs.append(job)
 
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/archivematicaClient.py
```diff
@@ -143,6 +143,29 @@ def set_start_times():
...
+            # character.  Shlex doesn't do this but bash unescaping does, and we
+            # want to remain compatible.
+            result += s[i + 1]
+            i += 1
+        else:
+            result += s[i]
+
+        i += 1
+
+    return result
```
jrwdunham: @marktriggs this func could be replaced with something more concise like the following, though depending on your taste it may reduce readability (in which case feel free to ignore):

    def _shlex_unescape(s):
        return ''.join(c1 for c1, c2 in zip(s, s[1:] + '.') if (c1, c2) != ('\\', '`'))

jrwdunham (approved):
lgtm

src/MCPClient/lib/archivematicaClient.py
marktriggs: Nice!  I'll take it :)  I've just updated the PR branch, thanks.

sevein (approved):

PR #1211, opened 2018-07-19
Allow package create without transfer source details

sevein:
This commit makes possible to take `path` values in the Package create API
where the UUID of the transfer source location is not included, e.g. `<path>`
instead of `<uuid>:<path>`.

When this is the case, we attempt to determine the default transfer source
location asking the Storage Service.

Connects to https://github.com/archivematica/Issues/issues/21.

### DIFF
--- a/src/MCPServer/lib/package.py
+++ b/src/MCPServer/lib/package.py
@@ -139,6 +139,18 @@ def _check_filepath_exists(filepath):
     return None
 
 
+_default_location_uuid = None
+
+
+def _default_transfer_source_location_uuid():
+    if _default_location_uuid is not None:
+        return _default_location_uuid
+    location = storage_service.get_default_location('TS')
+    global _default_location_uuid
+    _default_location_uuid = location['uuid']
+    return _default_location_uuid
+
+
 def _copy_from_transfer_sources(paths, relative_destination):
     """Copy files from source locations to the currently processing location.
 
@@ -154,11 +166,9 @@ def _copy_from_transfer_sources(paths, relative_destination):
     files = {l['uuid']: {'location': l, 'files': []} for l in transfer_sources}
 
     for item in paths:
-        try:
-            location, path = item.split(':', 1)
-        except ValueError:
-            raise Exception(
-                'Path {} cannot be split into location:path'.format(item))
+        location, path = Path(item).parts()
+        if location is None:
+            location = _default_transfer_source_location_uuid()
         if location not in files:
             raise Exception('Location %(location)s is not associated'
                             ' with this pipeline' % {'location': location})
@@ -316,7 +326,7 @@ def _capture_transfer_failure(fn):
 def _determine_transfer_paths(name, path, tmpdir):
     if _file_is_an_archive(path):
         transfer_dir = tmpdir
-        p = path.split(':', 1)[1]
+        p = Path(path).path
         filepath = os.path.join(tmpdir, os.path.basename(p))
     else:
         path = os.path.join(path, '.')  # Copy contents of dir but not dir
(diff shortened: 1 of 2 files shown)


jhsimpson:
I tested this and it is working as described.

There needs to be some documentation of the packages endpoint, and currently it is difficult to troubleshoot when something goes wrong, but I think the PR should be merged as is and those issues worked out separately.  

Some notes about using this endpoint:

the url for the endpoint is api/v2beta/package/
I set two http headers, `Authorization: ApiKey test:test` and `Content-Type: application/json`.
I POST'ed a json document like so:
`{"name": "Justin", "type": "standard", "path": "YXJjaGl2ZW1hdGljYS9hcmNoaXZlbWF0aWNhLXNhbXBsZWRhdGEvU2FtcGxlVHJhbnNmZXJzL0ltYWdlcw=="}`

where name = the name of the transfer as I want it to appear in the dashboard, type = transfer type, and path = the base64 encoded path of the directory I want to use as my transfer.  In my test, I had the archivematica-sampledata checked out at /home/archivematica/archivematica-sampledata/.  The transfer source location in my storage service was setup with a path of '/home'.  The path that I needed to base64 encode was 'archivematica/archivematica-sampledata/SampleTransfers/Images' - the path has to be relative to the transfer source locations path.

So you don't need to know the uuid of the transfer source location, but you do need to know its path and exclude that from the path you POST.

In my first attempt, I didn't do this and the POST returned a 202 accepted, but no transfer appeared in the dashboard. I was able to determine what went wrong by going to the archivmatica-mcp-server logs and finding this error:

`rsync: change_dir "/home/home/archivematica/archivematica-sampledata/SampleTransfers/Images" failed: No such file or directory (2)`

The api/v2beta/package/ endpoint will need to have a status endpoint added sooner than later (say api/v2beta/package/{uuid}/status or something like that), in order to find out if there was an error.  But in the meantime, it is possible to find out about errors from the logs. Not great, but a work around.

Here is the session from when I was able to successfully start a transfer without supplying the transfer source locations uuid:

```
* Preparing request to http://127.0.0.1:62080/api/v2beta/package/
* Using libcurl/7.59.0 OpenSSL/1.1.0h zlib/1.2.11 nghttp2/1.31.1
* Enable automatic URL encoding
* Enable SSL validation
* Enable cookie sending with jar of 0 cookies
* Connection 2 seems to be dead!
* Closing connection 2
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 62080 (#3)
> POST /api/v2beta/package/ HTTP/1.1
> Host: 127.0.0.1:62080
> User-Agent: insomnia/5.16.6
> Authorization: ApiKey test:test
> Content-Type: application/json
> Accept: */*
> Content-Length: 134
| {"name": "Justin", "type": "standard", "path": "YXJjaGl2ZW1hdGljYS9hcmNoaXZlbWF0aWNhLXNhbXBsZWRhdGEvU2FtcGxlVHJhbnNmZXJzL0ltYWdlcw=="}
* upload completely sent off: 134 out of 134 bytes
< HTTP/1.1 202 ACCEPTED
< Server: nginx/1.12.2
< Date: Fri, 20 Jul 2018 16:40:23 GMT
< Content-Type: application/json
< Transfer-Encoding: chunked
< Connection: keep-alive
< Vary: Accept-Language, Cookie
< Content-Language: en

* Received 57 B chunk
* Connection #3 to host 127.0.0.1 left intact
```

jhsimpson (approved):
I have no complaints or comments about the code and when testing this works for me.

:+1:

PR #1212, opened 2018-07-19
Feature: allow DIP upload to Binder

jrwdunham:
- Adds `access_system_id` attribute to the `Transfer` model (including migration).
- Adds "Access System ID" input to transfer browser GUI.
- Adds `access_system_id` param to payload of POST v2beta/package endpoint.
- Fixes dependency of access_system_id migration.
- Updates code comments and HTML help text to explain how to use the "Access System ID" at the start of transfer or the AtoM/Binder Identifier in the modal dialog to specify the identifier required by the AtoM or Binder SWORD V1 deposit endpoint during DIP upload to AtoM or Binder.

Connected to archivematica/Issues#23

**Note 1**: this PR was built by first cherry-picking the following four commits from the `dev/1.6.x-moma` branch and then adding a new commit:

- fa54b3014cb63f0a656d37624f31d93137a8a43f
- 3a7e56cb8d57a828233ac7c1c486e1f9300e3b06
- 15cb8d1eacdf821850517d18959e380b1075e290
- 0fa2fb4392f8e8ac54a8939a83d875db96bfbbe9

**Note 2**: this work has NOT yet been tested with DIP upload to a real AtoM or Binder instance.

The transfer browser should now look like the following:

screen shot 2018-07-20 at 9 40 00 am

The modal dialog in ingest at the "Choose config for AtoM DIP upload" micro-service should now look like the following:

screen shot 2018-07-20 at 2 28 41 pm


src/dashboard/frontend/transfer-browser/app/front_page/content.html
```diff
@@ -22,6 +22,11 @@
   <div class="help-block">{{ "Accession no." | translate }}</div>
   </div>
 
+  <div class="col-xs-2">
+  <input class="form-control" ng-model="vm.transfer.access_system_id"/>
+  <div class="help-block">{{ "Access System ID" | translate }}</div>
```
sevein: Other labels are only sentence cased. Should be `Access system ID` instead?

src/dashboard/src/templates/ingest/grid.html
```diff
@@ -246,13 +246,16 @@ <h3>{% trans "Edit metadata" %}</h3>
...
         </div>
         <div class="modal-body">
-          <p>{% trans "Upload the generated DIP to AtoM using the permalink of the target description." %}</p>
+          <p>{% trans "Please provide the identifier required by the AtoM or Binder SWORD V1 deposit endpoint." %}</p>
+          <ul>
+              <li><strong>AtoM:</strong>&nbsp;{% trans "In AtoM, this is the permalink or slug of the target archival description where the DIP will be attached as a child. For example, if the URL of the archival description is 'http://myAtoM.ca/newsletters-2;rad', then you would enter 'newsletters-2'." %}</li>
+              <li><strong>Binder:</strong>&nbsp;{% trans "In Binder, this is an identifier that indicates which artwork or technology record the DIP will be attached to. For example, if the DIP is being uploaded to an artwork record, and that artwork has a TMS Object ID of 123, you would enter 'ar:123'. If the DIP is being uploaded to a technology record, and that technology record has a URL in Binder of 'http://myBinder.org/technology-record/456' you would enter 'tr:456'." %}</li>
```
sevein: Two observations:
- Our templates use two spaces for indentation.
- I think that it's preferably to annotate the whole sentence, e.g. `AtoM: ...` even if it needs to contains HTML elements that way we don't enforce a grammar structure.

src/dashboard/src/main/migrations/0055_transfer_access_system_id.py
```diff
@@ -7,7 +7,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('main', '0029_clean_transfers'),
+        ('main', '0054_index_aip_silent_error_branch'),
```
sevein: I've just added 0055 (#1216).

sevein (approved):
LGTM.

sromkey:
I have a question and a suggestion:
Question: in the user interface at DIP upload, has that text changed? If Binder isn't split out into its own DIP upload selection in the drop-down menu we should ammend the text to read "Upload DIP to AtoM/Binder."
Suggestion: In the modal dialog, to make the Binder text a little less art-repository-specific and to generalize away the use of TMS (it's a longer-term goal that Binder shouldn't rely on a connected TMS instance to work). Suggested text:

`Binder: In Binder, this is an identifier that indicates which resource or technology record the DIP will be attached to. Resources are preceded by 'ar:' (e.g. 'ar:123') and technology records are preceded by 'tr:' (e.g. 'tr:456')`

Further explanation about how to find the appropriate id numbers for resource and technology records should be covered in documentation.

src/dashboard/frontend/transfer-browser/app/front_page/content.html
jrwdunham: k.

jrwdunham:
@sevein @sromkey I have made the changes you suggested.

@sromkey here's what the UI currently looks like (in this PR) wrt AtoM/Binder:

- On the processing config page (e.g., /administration/processing/edit/default/), the decision says "Upload DIP to AtoM".
- On the ingest page under "Upload DIP", the relevant micro-service name is "Choose config for AtoM DIP upload" with the only option in the select box being "- upload-qubit_v0.0".

Should these ^ be changed to reference Binder as well?

sromkey:
@jrwdunham-

- Yeah, I think the Processing Config page should also refer to Binder, good catch
- The Choose Config for AtoM DIP upload is not actually something a user ever sees, I'm talking about one step before that- choosing between DIP upload methods (e.g. ArchivesSpace, CONTENTdm, AtoM, etc)

sevein (approved):

ablwr:
+1 to @sromkey's suggestions of directing the user to upload DIP to AtoM/Binder and the non-TMS-specific language. Which you've already done soooo.... LGTM!!! 👍

sromkey (approved):

PR #1213, opened 2018-07-19
Dashboard: disable CSRF in /api (shib only)

sevein:
Dashboard's `/api` expects a `csrftoken` when you're authenticating using your session cookie, i.e. not your API keys. This is the method used by the transfer browser widget now that we've updated it so it uses `/api/v2beta/package`. The `csrftoken` is emitted by Django in the login page. Shibboleth-enabled deployments have a different authentication flow where the login page is never visited so the `csrftoken` wasn't made available to the client.

The solution here is to disable the CSRF check in `/api` when Shibboleth is enabled.

This is connected to https://github.com/archivematica/Issues/issues/18.


jhsimpson (approved):
I tested and works for me.  

:+1:

PR #1214, opened 2018-07-20
Capture any exception in index_aip to job's stderr

### LINKED ISSUE #1194: Problem: task stderr is not captured after client script raises an exception
How to reproduce:
1. Update `index_aip.index_aip()` so it raises an exception.
2. Start a transfer, store the AIP.
3. You should see something like the following: image
4. Click on the gear icon.

The exit code is 179, this is expected (see #1190).
**However the stderr stream has not been recorded. It's not in the database.**

image

I would expect to see the message of my exception.

This was originally found by @jhsimpson - see https://github.com/artefactual/archivematica/pull/1190#pullrequestreview-135179192.

payten:
Fixes #1194 whereby exceptions were being caught but not logged.  Patch ensures any exceptions are logged to the job's stderr.

This is connected to #1194.

### DIFF
--- a/src/MCPClient/lib/clientScripts/index_aip.py
+++ b/src/MCPClient/lib/clientScripts/index_aip.py
@@ -3,6 +3,7 @@
 from glob import glob
 import os
 import sys
+import traceback
 
 # dashboard
 from main.models import UnitVariable
@@ -133,8 +134,11 @@ def call(jobs):
         with job.JobContext(logger=logger):
             try:
                 status_code = index_aip(job)
-            except Exception:
+            except Exception as err:
                 # We want to capture any exception so ``filter_status_code``
                 # makes the last call on what is the status returned.
                 status_code = 1
+                job.print_error(repr(err))
+                job.print_error(traceback.format_exc())
+
             job.set_status(filter_status_code(status_code))


sevein:
@payten I'm wondering if we should approach this in a way that it applies to all client scripts instead of dealing with them individually.

`fork_runner.call` seems to have a mechanism in place to have uncaught exceptions recorded in the job but when the module is executed inline we seem to rely on what the client script is capable to record.

https://github.com/artefactual/archivematica/blob/027785feb99229ea5b80ef52a745a0d91f61f941/src/MCPClient/lib/archivematicaClient.py#L138-L141

Perhaps #L141 should be wrapped with a try-except for consistency with `fork_runner`.

Does that make sense? I'm not entirely sure that I'm understanding the existing solution correctly so take my suggestion less seriously.

payten:
Hi @sevein, I believe what your describing is provided by the `JobContext`, whereby any uncaught exceptions are logged and an error status of `1` is set on the job: https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPClient/lib/job.py#L103. 
 
Looking at `index_aip`, it looks like we want to handle this exception case in a slightly different way by setting the error code based on the result of `filter_status_code`.

sevein:
Oh I get it, thank you that's very useful.
Maybe we can raise the exception anyways, e.g.:

```python
# [...]
job.set_status(filter_status_code(status_code))
try:
    raise err
except NameError:
    pass
```

Would that work?

payten:
Hm.. currently the `JobContext` code sets the exit code to `1` when it encounters an exception: https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPClient/lib/job.py#L106.  When this happens we could change it so we only `set_status(1)` when there isn't already a job status set, but that may result in false positives.. maybe we only  `set_status(1)` if `job.status == 0`?

Our thinking was that the client scripts would do their best to catch and handle exceptions and the `JobContext` would then catch any that slipped through, allowing the batch to continue.  As part of the batching work you'll see a few client scripts where we've added exception handling where previously there was none.  Happy to rethink things to make cases like this more manageable!

sevein:
Gotcha! It's also worth saying that I broke this in ccf6464e50515f84cb25f212116ace9611e837c2 because I didn't know then what `job.JobContext` was doing. Thanks.

sevein (approved):

PR #1217, opened 2018-07-24
Transfer Browser: add auto_approve checkbox

sevein:
The new Package API endpoint starts transfers automatically. In this commit,
the endpoint learns a new optional parameter `auto_approve` which defaults to
`True`. In the transfer browser, a new checkbox `Approve automatically` has
been added and it is disabled by default.

image

This is connected to #1139.

### DIFF
--- a/src/MCPServer/lib/RPCServer.py
+++ b/src/MCPServer/lib/RPCServer.py
@@ -114,6 +114,8 @@ def package_create_handler(*args, **kwargs):
         payload.get('access_system_id'),
         payload.get('path'),
         payload.get('metadata_set_id'),
+        payload.get('auto_approve'),
+        payload.get('wait_until_complete'),
     )
     return create_package(*create_package_args).pk
 
--- a/src/dashboard/frontend/transfer-browser/app/front_page/content.html
+++ b/src/dashboard/frontend/transfer-browser/app/front_page/content.html
@@ -12,7 +12,7 @@
   <div class="help-block">{{ "Transfer type" | translate }}</div>
   </div>
 
-  <div class="col-xs-3" ng-if="vm.transfer.type !== 'zipped bag'">
+  <div class="col-xs-2" ng-if="vm.transfer.type !== 'zipped bag'">
   <input class="form-control" ng-model="vm.transfer.name"/>
   <div class="help-block">{{ "Transfer name" | translate }}</div>
   </div>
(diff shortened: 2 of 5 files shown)


src/MCPServer/lib/RPCServer.py
```diff
@@ -113,6 +113,8 @@ def package_create_handler(*args, **kwargs):
         payload.get('accession'),
         payload.get('path'),
         payload.get('metadata_set_id'),
+        payload.get('auto_approve'),
+        payload.get('wait_until_complete'),
```
sevein: When `auto_approve=False` and `wait_until_complete=True` we would be effectively implementing the same behaviour of the old `start_transfer` endpoint from the user point of view (transfer prompts for auto approval, request holds until the files are copied). So I'm hoping with some further development we can have the `start_transfer` view updated so it delegates on MCPServer's new package.py module instead of the old `filesystem_ajax` Django app.

src/dashboard/src/components/api/views.py
```diff
@@ -782,6 +782,7 @@ def _package_create(request):
         payload.get('accession'),
         path,
         payload.get('metadata_set_id'),
+        payload.get('auto_approve', True)
```
sevein: With a default value so the parameter does not become mandatory. `True` because we're finding this behaviour a better fit but that may change - remember that this is still in beta.

src/dashboard/frontend/transfer-browser/app/services/transfer.service.js
```diff
@@ -17,6 +17,7 @@ class Transfer {
     this.type = 'standard';
     this.accession = '';
     this.components = [];
+    this.auto_approve = false;
```
sevein: The transfer browser widget prefers the behaviour to be set to false by default because otherwise we'd be introducing a backward incompatible change.

src/dashboard/frontend/transfer-browser/app/front_page/content.html
```diff
@@ -22,22 +22,20 @@
...
           class="btn btn-default"
           data-toggle="collapse"
           data-target="#transfer_browse_tree"
           aria-expanded="false"
-          aria-controls="transfer_browse_tree">
-    {{ "Browse" | translate }}
-  </button>
-  </div>
-
-  <div class="col-xs-2">
```
sevein, line 41: I've put the "Browse" and "Start transfer" buttons together in the same column so when the "Approve automatically" checkbox overflows there's more real estate for the text to be presented.

To be clear:

image

Instead of:

image

sromkey (approved):

Issue #1148, opened 2018-06-20
Problem: Parse Dataverse Mets fails for some datasets


joel-simpson:
Testing the new "Parse Dataverse  Mets" job within the "Parse External Files" microservice. 

When testing with datasets I added to Dataverse, this job completes successfully. 

When testing datasests created by the Scholar's Portal team, this job is failing. 
![screen shot 2018-06-20 at 10 07 14 am](https://user-images.githubusercontent.com/17187309/41663623-d9a5d85a-7471-11e8-9af3-7c4161ec2850.png)

The error message from the task that doesn't complete is: 
![screen shot 2018-06-20 at 10 07 41 am](https://user-images.githubusercontent.com/17187309/41663649-ea0b5a08-7471-11e8-984b-3e8cc716e474.png)

I am not sure if it is relevant, but to get to access this dataset, I used "data&subtree=archivematica" in the relative path of the Dataverse location. For the datasets that have passed this job, that field is only set to "archivematica".

joel-simpson:
The following dataset from the Archivematica Test Dataverse fails at the Parse Dataverse Mets microservice: 
[3D Laser Images of a road cut at Ivy Lea, Ontario (2007), underground in Sudbury, Ontario (2007), underground in Thompson, Manitoba (2009) [test]](https://demodv.scholarsportal.info/dataset.xhtml?persistentId=doi:10.5072/FK2/HAQQUC)

ross-spencer:
Discussing this last-week with a Kaitlin at OCUL, it was noted that to include a ZIP file in a Dataverse is a deliberate decision by the researcher because Dataverse will automatically extract the first level of zips that it comes across.

From: http://guides.dataverse.org/en/latest/user/dataset-management.html
> Compressed Files
Compressed files in zip format are unpacked automatically. If it fails to unpack, for whatever reason, it will upload as is. If the number of files inside are more than a set limit (1,000), you will get an error message and the file will uploads as is.

Support for unpacking tar files will be added when this ticket is closed: https://github.com/IQSS/dataverse/issues/2195.

[...]

Issue #1147, opened 2018-06-19
Problem: Dataverse requires its own transfer type


ross-spencer:
For Dataverse transfers to be picked up in Archivematica 1.7 proper, i.e. not via Automation Tools, we require a Dataverse transfer-type. 

In future, other mechanisms for different 'types' of transfer could also be investigated.

ross-spencer:
**NB.** From the previous Dataverse branch we had this comment: 
```
    # TEMPORARY
    # TODO Update this so it
    # A) autodetects transfer type OR
    # B) add a Dataverse type that sets postExtractSpecializedProcessing
```
**To lookup:** `postExtractSpecializedProcessing... `

ross-spencer:
Related: https://github.com/artefactual/archivematica/pull/347

ross-spencer:
Related: https://github.com/artefactual/archivematica/issues/1121#issuecomment-413954754

PR #1218, opened 2018-07-24
Creates a Dataverse transfer type

ross-spencer:
This PR introduces a new module `processworkflowjson.py` which allows a workflow inside Archivematica to be defined using a JSON configuration.

The module is used to establish a completely new workflow that runs in parallel to the current Archivematica workflow. 

Initially, as the Dataverse POC is evaluation this workflow remains completely in parallel. Some links may need to be removed, some might need to happen earlier in the process and some might need to be added. 

It is recognised talking to @sevein that using the `variablelinkpull` technique to link the chain together in places will be more appropriate. 

Weaknesses when not using that approach include:

* Not being able to use the processing config to automate decisions such as format identification because they are completely different chain links now. 

Compromises made in this code to get up and running:

* AIP METS 2 still needs to be merged to incorporate Dataverse specific changes instead of forking out to a Dataverse module (this should happen post POC once the Dataverse METS has been analysed). 
* No quarantine links
* No create tree link (in some cases for Dataverse this happens too early in the process because some Dataverse datasets are zipped first)
* No policy check microservices on transfer
* No examine contents microservices on transfer 

A benefit:

* Extract contents is no longer recursive so this is good for zipped Dataverses: https://github.com/artefactual/archivematica/issues/1148

In my mind, to complete this work:

[ ] The new chainlink migration approach needs to be signed-off on.
[ ] We need to decide if this is an appropriate Dataverse workflow, e.g. no quarantine, etc. 
[ ] The storage service PR needs review: https://github.com/artefactual/archivematica-storage-service/pull/386
[ ] The storage service needs to be changed to improve transfer browsing, i.e. restrictions removed from the Dataverse query string
[ ] I need to learn how to use variable link pull fields in the database to improve chain linking
[ ] Some fields need removing from the new JSON workflow files, they are either redundant, or have been implemented in a different way, e.g. the chain sections, vs. the chain attributes.
[ ] **I will come back to this list and refine it with the help of the team**

Connects to https://github.com/artefactual/archivematica/issues/1147.

Out of scope:

[ ] Creating a microservice to extract packages earlier in the workflow (this could be done much sooner rather than relying on extract packages.
[ ] Anything else?


ross-spencer:
Once merged, we can close this too: https://github.com/artefactual/archivematica/pull/347 (Let's verify we have captured everything in the PR)

src/dashboard/src/main/processworkflowjson.py
```diff
@@ -0,0 +1,470 @@
+# -*- coding: utf-8 -*-
```
sevein, line 1: The problem with a helper like this is that it's designed against a specific version of the data model. Perhaps the migrations in this pull request can leverage this work but this module will become obsolete as soon as we introduce backward incompatible in the schema, which we will for sure, we do all the time.

I think this is why we haven't implemented something like this before. Each migration is targeting a specific snapshot of the schema, e.g. migration 0056 was developed with 0055 in mind. If we had your `processworkflowjson` module up to date according to the schema changes then the old migrations would potentially break.

sevein: I think this problem to some extend also affects your idea of having JSON documents describing the changes. Who can read these files? `processworkflowjson`? Then we'd have the same problem.

You could have a particular migration being fed from that JSON document(s) but what would be the point then? Can't you just achieve the same with simple Python data structures like lists, sets, dicts, ordered dicts?

Also I suspect that you wouldn't feel the urge for this solution if the set of workflow changes you needed were smaller. Now it's pretty big because you're wanting to have a whole different workflow. My recommendation would be to step back, identify the minimal set of changes and implement those. Potentially your migration would look much simpler, e.g. see `0041_bind_pids.py`.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -57,7 +57,8 @@
     'zipped bag': 'baggitZippedDirectory',
     'dspace': 'Dspace',
     'maildir': 'maildir',
-    'TRIM': 'TRIM'
+    'TRIM': 'TRIM',
+    'dataverse': 'dataverseTransfer'
```
sevein, line 61: Now in qa/1.x you also have to make similar update in `package.py`.

src/dashboard/src/main/migrations/0052_dataverse_foreign_keys.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import unicode_literals
```
sevein, line 1: Something to remember: migrations are very slow! A changeset like the one in this PR should only bring one migration so it's important to jointhem before merging.

src/MCPClient/lib/clientScripts/parse_dataverse.py
```diff
@@ -0,0 +1,187 @@
...
+    update_file_use(mapping)
+    agent = add_external_agents(unit_path)
+    create_derivatives(mapping, agent)
+    validate_checksums(mapping, unit_path)
+    return 0
+
+if __name__ == '__main__':
+    print('Parse Dataverse METS')
+    parser = argparse.ArgumentParser(description='Parse Dataverse METS file')
+    parser.add_argument('transfer_dir', action='store', type=str, help="%SIPDirectory%")
```
sevein: FYI: `store` is the default action, `str` is the default type.

sevein (changes requested):

ross-spencer:
Closing as this will be superseded by the work here: https://github.com/artefactual/archivematica/tree/dev/issue-1147-minimal-dataverse-transfer-type

ross-spencer:
Replaced by: https://github.com/artefactual/archivematica/pull/1242

PR #1219, opened 2018-07-24
Small fixes after MCP batching changes

sevein:
This is connected to #938.

### DIFF
--- a/src/MCPClient/lib/clientScripts/extract_contents.py
+++ b/src/MCPClient/lib/clientScripts/extract_contents.py
@@ -19,7 +19,7 @@ from fileOperations import addFileToTransfer, updateSizeAndChecksum
 from archivematicaFunctions import get_dir_uuids, format_subdir_path
 
 # clientScripts
-from hasPackages import already_extracted
+from has_packages import already_extracted
 
 file_path_cache = {}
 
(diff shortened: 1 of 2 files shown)


src/MCPServer/lib/taskGroupRunner.py
```diff
@@ -235,7 +235,7 @@ def _monitor_running_jobs(self, gm_client):
 
         now = time.time()
         if (now - self.last_notification_time) > TaskGroupRunner.NOTIFICATION_INTERVAL_SECONDS:
-            LOGGER.info("%d jobs pending; %d jobs running; %d known task groups",
+            LOGGER.debug("%d jobs pending; %d jobs running; %d known task groups",
```
ross-spencer, line 238: Do you think this will generate a lot of noise in the logs? I'm not sure the best practice or the rationale to change this from info so really just an open question.

src/MCPClient/lib/clientScripts/extract_contents.py
```diff
@@ -19,7 +19,7 @@
 from archivematicaFunctions import get_dir_uuids, format_subdir_path
 
 # clientScripts
-from hasPackages import already_extracted
+from has_packages import already_extracted
```
ross-spencer, line 22: :+1:

ross-spencer (approved):
I left an open question about the logging, but otherwise, LGTM.

src/MCPServer/lib/taskGroupRunner.py
sevein: Yeah I think we're printing that on every poll which I think it defaults to 200ms.

We should just be logging human-actionable events, as from what I've been reading as a good practice in the industry. This code could be instrumented if we really wanted to have gauges with number of errors, sum of jobs, etc...

sevein:
Ops I totally missed the build complaining about that issue reported by flake8. Hopefully we can fix in the next PR.

Issue #800, opened 2017-11-19
Problem: user can't choose processing configuration


sevein:
The Dashboard allows the user to create multiple configurations and they can be chosen during re-ingest but not in new transfers.

sromkey:
Are we considering this issue to be fully addressed by #799? With that PR you can download the MCP file but AFAIK there's not a way in the dashboard for users to choose a processing config.

sevein:
Yes, 799 is not enough.

sevein:
@sromkey when you have a chance can you please take a look at the solution suggested in #1222? It does two things: 1) it makes possible via the new Package API to choose the processing configuration by adding an additional attribute `processing_config`, 2) it makes a small change in the transfer browser widget so developers can choose define a value for `processing_config` - as a temporary solution so we can start testing while we figure out the user interface.

PR #1222, opened 2018-07-25
Package API: add processing_config parameter

sevein:
This pull request includes two commits:

#### Package API: add processing_config parameter

In this commit, the new `/api/v2beta/package/` API endpoint learns a new `processing_config` parameter where the user can define the name of the processing configuration that wants to be used, e.g. `automated` or any other configuration added to the system.

#### Transfer Browser: add basic support for processing config

The transfer browser widget is updated so the user can set the processing configuration by opening the web console and creating a new property in the `window` object named `processing_config`. Usage:

1. Open JavaScript console.
2. Type:

        window.processing_config = 'automated';

This meant to be a temporary solution until we have time to explore ways to integrate this in the user interface.

Connects to https://github.com/artefactual/archivematica/issues/800.

### DIFF
--- a/src/MCPServer/lib/RPCServer.py
+++ b/src/MCPServer/lib/RPCServer.py
@@ -107,17 +107,22 @@ def job_awaiting_approval_handler(*args):
 def package_create_handler(*args, **kwargs):
     """Handle create package request."""
     payload = kwargs['payload']
-    create_package_args = (
+    args = (
         payload.get('name'),
         payload.get('type'),
         payload.get('accession'),
         payload.get('access_system_id'),
         payload.get('path'),
         payload.get('metadata_set_id'),
-        payload.get('auto_approve'),
-        payload.get('wait_until_complete'),
     )
-    return create_package(*create_package_args).pk
+    kwargs = {
+        'auto_approve': payload.get('auto_approve'),
+        'wait_until_complete': payload.get('wait_until_complete'),
+    }
+    processing_config = payload.get('processing_config')
+    if processing_config is not None:
+        kwargs['processing_config'] = processing_config
+    return create_package(*args, **kwargs).pk
 
 
 def startRPCServer():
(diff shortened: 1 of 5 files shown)


jrwdunham (commented):
Code looks good and I confirmed it does what it claims to. I'm wondering what is the motivation behind merging a partial solution like this? Why not wait until we know what we want the GUI to look like and implement it instead of doing this JS console stuff?

Relatedly, in exploring this, I noticed that you can overwrite an existing processing config by creating a new one with the same name. I wonder if this is a bug. I noticed that there is some validation of the pcfg name (e.g., no spaces in it) so why not validate against duplicates also? Maybe something for a separate issue but bringing it up here anyway :)

... another potential gotcha with this work is that if you navigate to a new page (or refresh the page), then obviously your console-specified processing config name will be lost.

sevein:
> Code looks good and I confirmed it does what it claims to. I'm wondering what is the motivation behind merging a partial solution like this? Why not wait until we know what we want the GUI to look like and implement it instead of doing this JS console stuff?

We can definitely wait. I wanted to have support in the API and some kind of preliminary support in the transfer browser so we can start testing from the browser.

> ... another potential gotcha with this work is that if you navigate to a new page (or refresh the page), then obviously your console-specified processing config name will be lost.

Yes, that's definitely a limitation. My idea is to introduce it as an interim solution that could potentially be used by developers while we figure out the user interface. The intention was to make it as simple as possible.

> Relatedly, in exploring this, I noticed that you can overwrite an existing processing config by creating a new one with the same name. I wonder if this is a bug. I noticed that there is some validation of the pcfg name (e.g., no spaces in it) so why not validate against duplicates also? Maybe something for a separate issue but bringing it up here anyway :)

Yes, I'm planning to work on that soon or at the very least I will file an issue. Thanks.

jrwdunham (approved):

Issue #1174, opened 2018-06-29
Problem: disabling ES search appears to break post_store_aip_callback and dspace_handle_to_archivesspace


jrwdunham:
See https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPClient/lib/clientScripts/post_store_aip_hook.py#L88-L94. 

It appears that when ES is disabled, then `dspace_handle_to_archivesspace` and `post_store_aip_callback` will never get called. This indicates to me that if ES is disabled, then some workflows will be broken that should not be.

jraddaoui:
Agree. I may be able to tackle this in #1172 if it has not been fixed sooner.

jraddaoui:
To test this, in with the `search_enabled` config set to `false`, after an AIP has been successfully stored, the following workflows should be executed:

- SS file deletion request (when all files have been arranged).
- DuraSpace handle to ArchivesSpace.
- SS post store AIP callback.

ablwr:
I think I have tested this successfully but @jraddaoui we can talk more about it tomorrow :)

ablwr:
@mamedin and I both tested this and it is VERIFIED FIXED!

PR #1223, opened 2018-07-26
MCPClient: fix post_store_aip_hook (indexless)

jraddaoui:
When the search functionality is disabled, the post_store_aip_hook will
only avoid making the changes related to the transfers ES index. The
following workflows will be restored in that case:

- SS file deletion request (when all files have been arranged).
- DuraSpace handle to ArchivesSpace.
- SS post store AIP callback.

Connects to #1174.


sevein (approved):

jraddaoui:
Thanks @sevein!

40c11bb in stable/1.7.x.

PR #1225, opened 2018-07-27
Standardise global state in create_mets_v2

marktriggs:
The `create_mets_v2` client script uses a bunch of global state for things like counters and file IDs.  The original implementation could rely on these globals being reset between runs (by virtue of running the script in a new subprocess each time) but, after the batching changes, this is no longer the case.

To make things more manageable, we pull all of this global stuff into a single top-level object which gets replaced at the beginning of each job run.  Where other modules call the functions of create_mets_v2, they reset the global state as appropriate as well.

A secondary piece of shared state is the `sharedVariablesAcrossModules` module, used by `create_mets_v2` and a handful of other scripts.  Handling for this needed to be updated for the batching changes in a small number of places.

This is connected to https://github.com/archivematica/Issues/issues/42.
This is connected to https://github.com/archivematica/Issues/issues/43.
This is connected to https://github.com/artefactual/archivematica/issues/938.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
@@ -300,13 +300,14 @@ def add_new_files(job, mets, sip_uuid, sip_dir):
         return mets
 
     # Set global counters so getAMDSec will work
-    createmets2.globalAmdSecCounter = int(
-        mets.tree.xpath('count(mets:amdSec)', namespaces=ns.NSMAP))
-    createmets2.globalTechMDCounter = int(
-        mets.tree.xpath('count(mets:amdSec/mets:techMD)', namespaces=ns.NSMAP))
-    createmets2.globalDigiprovMDCounter = int(
-        mets.tree.xpath('count(mets:amdSec/mets:digiprovMD)',
-                        namespaces=ns.NSMAP))
+    createmets2.initGlobalState(
+        globalAmdSecCounter=int(mets.tree.xpath('count(mets:amdSec)',
+                                                namespaces=ns.NSMAP)),
+        globalTechMDCounter=int(mets.tree.xpath('count(mets:amdSec/mets:techMD)',
+                                                namespaces=ns.NSMAP)),
+        globalDigiprovMDCounter=int(mets.tree.xpath('count(mets:amdSec/mets:digiprovMD)',
+                                    namespaces=ns.NSMAP))
+    )
 
     objects_fsentry = mets.get_file(label='objects', type='Directory')
 
--- a/src/MCPClient/lib/clientScripts/create_mets_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_v2.py
@@ -55,53 +55,51 @@ from archivematicaFunctions import normalizeNonDcElementName
 from custom_handlers import get_script_logger
 import namespaces as ns
 from sharedVariablesAcrossModules import sharedVariablesAcrossModules
-sharedVariablesAcrossModules.globalErrorCount = 0
 
 from bagit import Bag, BagError
 
-# Global Variables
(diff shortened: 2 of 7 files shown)


marktriggs:
Oops.  Just investigating the failing test here.

marktriggs:
OK!  Tests are happy again.  Ready for review.

ross-spencer:
> The create_mets_v2 client script uses a bunch of global state for things like counters and file IDs. The original implementation could rely on these globals being reset between runs (by virtue of running the script in a new subprocess each time) but, after the batching changes, this is no longer the case.

This is useful to understand the behaviour. Do you have an instinct about other MCP Client scripts with global state that this might be a gotcha for?

src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
```diff
@@ -300,13 +300,14 @@ def add_new_files(job, mets, sip_uuid, sip_dir):
...
-        mets.tree.xpath('count(mets:amdSec/mets:digiprovMD)',
-                        namespaces=ns.NSMAP))
+    createmets2.initGlobalState(
+        globalAmdSecCounter=int(mets.tree.xpath('count(mets:amdSec)',
+                                                namespaces=ns.NSMAP)),
+        globalTechMDCounter=int(mets.tree.xpath('count(mets:amdSec/mets:techMD)',
+                                                namespaces=ns.NSMAP)),
+        globalDigiprovMDCounter=int(mets.tree.xpath('count(mets:amdSec/mets:digiprovMD)',
+                                    namespaces=ns.NSMAP))
+    )
```
sevein, line 310: If you don't mind me offering a style suggestion so you know next time, here's an alternative that does not exceed the 79 columns limit suggested by pep8.

```python
    createmets2.initGlobalState(
        globalAmdSecCounter=int(mets.tree.xpath(
            'count(mets:amdSec)', namespaces=ns.NSMAP)),
        globalTechMDCounter=int(mets.tree.xpath(
            'count(mets:amdSec/mets:techMD)', namespaces=ns.NSMAP)),
        globalDigiprovMDCounter=int(mets.tree.xpath(
            'count(mets:amdSec/mets:digiprovMD)', namespaces=ns.NSMAP)),
    )
```

E501 is one of the few errors emitted by flake8 that we're ignoring.
But we're trying that new code is formatted that way.

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -55,53 +55,51 @@
...
+# The collection of global state for the currently executing job.  Ultimately it
+# would probably be best to pass this around, but that's a pretty big
+# refactoring.
+#
+# We'll reinitialise this as each job runs.
+globalState = None
 
 
-logger = get_script_logger("archivematica.mcp.client.createMETS2")
+class MetsGlobalState():
```
sevein, line 69: This is a good compromise. We'll have time to clean up this script another time.

sevein (approved):

src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
marktriggs: Thanks!

marktriggs:
@ross-spencer as far as I can see, `create_mets_v2` is the only one using global variables in this way.  A few other scripts use the `sharedVariablesAcrossModules` module to similar ends (storing things like counters), so care needs to be taken to reset those counters at the right time.  I think those cases are now covered too.

I'll do some more testing next week to confirm that things are looking good.  Unfortunately my previous testing didn't turn this up because, although I was comparing outputs of the batching and pre-batching versions of the code, I was restarting my development environment between tests (and therefore clearing that global state).  I'll retest by running several successive ingests against each version of the code and check that the outputs are comparable.

Issue #1090, opened 2018-05-15
Problem: job resume after normalization file identification tool selected fails


sevein:
This is reproducible in qa/1.x.

The task page is blank so no more details are reported here.

This was probably caused by https://github.com/artefactual/archivematica/pull/948 or one of the follow-ups (the fix for #980 maybe?).

sevein:
This could be fixed I believe by adding a new ExitCode in the workflow data associating the `0` value to the success job status. This is probably what's missing in the original migration done in #948.

An alternative is to do some changes in the code so the task can suggest the status when the exit code entry is missing in the workflow data.

joel-simpson:
Doing some testing in qa/1.x today and seeing this fail every time. It does appear to only be a display issue -- the correct workflow decision is being made, it is showing a failed status in the dashboard.

PR #1226, opened 2018-07-30
Migration: fix unit variable pull link

sevein:
Set `defaultExitMessage` to `2` (`Job.STATUS_COMPLETED_SUCCESSFULLY`) so the
job is marked as successful when `linkTaskManagerUnitVariableLinkPull`
completes.

This is connected to #1090.


ross-spencer (approved):
:+1:

PR #1227, opened 2018-07-30
MCPServer: use auto_close_db

sevein:
Ensure that threads close the database connections, otherwise they remain open
which can cause "MySQL server has gone away" errors when MySQL recycles its
thread associated after a long inactivity period.

CC @marktriggs @payten @jambun.

This is connected to #938.

### DIFF
--- a/src/MCPClient/lib/fork_runner.py
+++ b/src/MCPClient/lib/fork_runner.py
@@ -22,6 +22,7 @@ import sys
 import tempfile
 import traceback
 
+from databaseFunctions import auto_close_db
 from executeOrRunSubProcess import launchSubProcess
 
 logger = logging.getLogger('archivematica.mcp.client')
@@ -75,6 +76,7 @@ def _split_jobs(jobs, n):
     return result
 
 
+@auto_close_db
 def _run_jobs(module_name, jobs):
     (fd, output_file) = tempfile.mkstemp()
 
--- a/src/MCPServer/lib/RPCServer.py
+++ b/src/MCPServer/lib/RPCServer.py
@@ -26,6 +26,7 @@ import time
 from django.conf import settings as django_settings
 import gearman
 
+from databaseFunctions import auto_close_db
 from linkTaskManagerChoice import choicesAvailableForUnits
 from package import create_package
 
@@ -78,6 +79,7 @@ def pickle_result(fn):
     return wrap
 
 
+@auto_close_db
 @pickle_result
 @unpickle_payload
 @log_exceptions(LOGGER)
@@ -92,6 +94,7 @@ def job_approve_handler(*args, **kwargs):
     return "approving: ", job_id, chain
 
 
+@auto_close_db
 @pickle_result
 @log_exceptions(LOGGER)
 def job_awaiting_approval_handler(*args):
@@ -101,6 +104,7 @@ def job_awaiting_approval_handler(*args):
     return etree.tostring(ret, pretty_print=True)
 
 
+@auto_close_db
 @pickle_result
 @unpickle_payload
 @log_exceptions(LOGGER)
--- a/src/MCPServer/lib/package.py
+++ b/src/MCPServer/lib/package.py
@@ -28,6 +28,7 @@ from django.conf import settings as django_settings
 
 from archivematicaMCP import taskThreadPool
 from archivematicaFunctions import unicodeToStr
+from databaseFunctions import auto_close_db
 from jobChain import jobChain
 from main.models import Transfer, TransferMetadataSet
 import storageService as storage_service
@@ -270,6 +271,7 @@ def create_package(name, type_, accession, access_system_id, path,
     transfer = Transfer.objects.create(**kwargs)
     logger.debug('Transfer object created: %s', transfer.pk)
 
+    @auto_close_db
     def _start(transfer, name, type_, path):
         # TODO: use tempfile.TemporaryDirectory as a context manager in Py3.
         tmpdir = mkdtemp(


sevein:
```
ERROR     2018-07-30 01:50:47  archivematica.mcp.server.rpcserver:RPCServer:wrap:53:  Exception raised by handler package_create_handler: (2006, 'MySQL server has gone away')

  File "/src/MCPServer/lib/RPCServer.py", line 49, in wrap
    return func(*args, **kwargs)
  File "/src/MCPServer/lib/RPCServer.py", line 125, in package_create_handler
    return create_package(*args, **kwargs).pk
  File "/src/MCPServer/lib/package.py", line 267, in create_package
    TransferMetadataSet.objects.get(id=metadata_set_id)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 238, in iterator
    results = compiler.execute_sql()
  File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 840, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 124, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 226, in execute
    self.errorhandler(self, exc, value)
  File "/usr/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorvalue
OperationalError: (2006, 'MySQL server has gone away')
```

sevein:
FYI `package.py` uses `archivematicaMCP.taskThreadPool` introduced by #1105.

marktriggs:
Hi @sevein,

Ah!  I hit this issue too and had dug around a little bit.  I also noticed that if I connected to mysql, ran a `show processlist`, identified the MCP Server's connections and ran `kill` on them one by one, I got the same error as above.  I was wondering whether it was holding onto that connection object but failing to reconnect.

The `wait_timeout` mysql setting (https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) seems to default to 8 hours, so I was wondering whether maybe MySQL had killed the connection after it sat idle for a while.  Maybe we also need to make sure MCP Server reconnects to MySQL automatically?

(edited to add...)
Test some more, it seems like I can survive losing all my MySQL connections if I call `close_old_connections()` before running my queries, rather than afterwards (the way `auto_close_db`) does.  Do you think it would be worth having `auto_close_db` call `close_old_connections()` before invoking the wrapped function as well as after?  That might stop the first query fired being sacrificial...

sevein:
On Sun, Jul 29, 2018, 20:10 Mark Triggs <notifications@github.com> wrote:

> The wait_timeout mysql setting (https://dev.mysql.com/doc/ref
> man/8.0/en/server-system-variables.html#sysvar_wait_timeout) seems to
> default to 8 hours, so I was wondering whether maybe MySQL had killed the
> connection after it sat idle for a while.
>
Yes, I think that's the issue. I was able to reproduce today coincidentally
because I had my env running for a couple of days without using it.

> Maybe we also need to make sure MCP Server reconnects to MySQL
> automatically?
>
Django is doing that for us AFAIK.

I believe that we need to close the connections because Django will only
close them after a request is processed which is not a thing outside the
web context. See https://docs.djangoproject.com/en/1.8/ref/
databases/#connection-management. Django won't let you share connections
across threads. It'd be optimal to have a connection pool (internal or
maybe external like vitess) so we can recycle connections.

sevein:
> Test some more, it seems like I can survive losing all my MySQL connections if I call close_old_connections() before running my queries, rather than afterwards (the way auto_close_db) does. Do you think it would be worth having auto_close_db call close_old_connections() before invoking the wrapped function as well as after? That might stop the first query fired being sacrificial...

I'm not sure if I understand why do you think calling it twice would be an improvement?

I realize now that if the allocated threads in our new thread pools are being reused it can totally make sense to run `close_old_connections()` before and after the wrapped function is called because that way we could have it use `CONN_MAX_AGE` as an attempt to recycle the connections associated to that thread as long as they're not obsolete. I'm wondering if this is what you meant above.

Here's how Django does the same in the web context: https://github.com/django/django/blob/master/django/db/__init__.py#L57-L65. Also, its [`ConnectionHandler`](https://github.com/django/django/blob/cebbcaa1baa6eb7f9ec3e223abdc496a983ea482/django/db/utils.py#L135-L142) - see how it uses `threading.local`, I think this is why the connections are never going to be shared between threads.

marktriggs:
Sorry, yeah, I wasn't being clear at all :)

From the testing I did yesterday, it seemed like Django wasn't automatically reconnecting to MySQL if its underlying connection had been closed.  In my test I would run an ingest to completion, then manually kill all of the MCP Server's connections (from the MySQL side), and then try running another ingest.  At that point, I'd consistently get the "MySQL server has gone away" error until I restarted MCP Server.

So we need to make sure that when a thread asks for a DB connection, it doesn't get back a connection that's stale.  I'd been assuming that `close_old_connections` wouldn't necessarily kill off an active DB connection, so if the next activity didn't happen for another 10 hours, MySQL would have killed the connection by that time.  Calling `close_old_connections` at the beginning of the wrapped function seemed like a way of ensuring that the stale connection was detected and replaced at the point the connection was needed.

But, I'd missed the fact that `CONN_MAX_AGE` was set to 0, which I guess means `close_old_connections` really just closes all connections?  So as long as everywhere that takes a database connection does this when it's finished, the next activity that runs shouldn't receive a stale connection?

If that's right, I wonder whether `RPCServer.py` should have `@auto_close_db` on `job_approve_handler` and `job_awaiting_approval_handler` too... it seems like those will also grab a DB connection.  What do you think?

Thanks for bearing with me :)

sevein:
Cool, thanks!

You're right about `job_approve_handler` and `job_awaiting_approval_handler`, good catch. I should also add the decorator to `package_create_handler` perhaps? It uses the database before sending work to the thread pool in `create_package`. They run in the main application thread thought, but I don't think that makes a difference?

Generally I feel it's ugly. I think that once we pull out the workflow data from the database we're just going to be touching a few tables from MCPServer. Perhaps at that point we can easily afford moving to sqlalchemy, it has a pool and sessions are managed explicitly.

marktriggs:
Ah yeah, I think you're right.  I don't think running on the main thread makes a difference, because there's nothing there keeping that connection alive either, so it could die just like the others.  Sad, really.

PR #1229, opened 2018-07-31
Add 7zip copy mode

ross-spencer:
Adds the necessary migration to the database which creates a 7-zip
compression option using file-level storage, that is, a 'copy' mode
which creates a 7-zip wrapper around objects without applying a
compression algorithm, e.g. without lzma, or bzip.

This is connected to https://github.com/archivematica/Issues/issues/46.

### DIFF
--- a/src/dashboard/src/main/migrations/0057_7zip_no_compression.py
+++ b/src/dashboard/src/main/migrations/0057_7zip_no_compression.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""Migration to add a 7-zip method to the list of Archivematica AIP 7-zip
+methods that does not use compression. 7-zip copy mode.
+"""
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+choice_no_compression_uuid = "e3906da0-41fb-4cb8-a598-6c73981b63e9"
+
+
+def get_chain_choice_dict(apps):
+    """Retrieve Chain Choice Replacement Dict model."""
+    return apps.get_model('main', 'MicroServiceChoiceReplacementDic')
+
+
+def get_ms_chain_link_instance(apps, ms_uuid):
+    """Get chainlink instance from the database."""
+    return apps\
+        .get_model("main", "MicroServiceChainLink")\
+        .objects.get(id=ms_uuid)
+
+
+def data_migration_down(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


ross-spencer:
Connected to https://github.com/archivematica/Issues/issues/46

src/dashboard/src/main/migrations/0057_7zip_no_compression.py
```diff
@@ -0,0 +1,45 @@
...
+
+def data_migration_up(apps, schema_editor):
+    ms_prepare_aip = get_ms_chain_link_instance(
+        apps, "01d64f58-8295-4b7b-9cab-8f1b153a504f")
+    get_chain_choice_dict(apps).objects.create(
+        id=choice_no_compression_uuid,
+        description="7z without compression (Copy)",
+        replacementdic="{\"%AIPCompressionAlgorithm%\":\"7z-copy\"}",
+        choiceavailableatlink=ms_prepare_aip,
+    )
```
sevein, line 37: Tip: you don't have to hydrate the related object if you use `choiceavailableatlink_id`, e.g.:

```python
def data_migration_up(apps, schema_editor):
    MicroServiceChoiceReplacementDic = apps.get_model(
        'main', 'MicroServiceChoiceReplacementDic')
    MicroServiceChoiceReplacementDic.objects.create(
        id=choice_no_compression_uuid,
        description="7z without compression (Copy)",
        replacementdic="{\"%AIPCompressionAlgorithm%\":\"7z-copy\"}",
        choiceavailableatlink_id="01d64f58-8295-4b7b-9cab-8f1b153a504f",
    )
```

sevein (approved):
I don't see the extra functions as an improvement necessarily, rather an unnecessary level of indirection for the reader. It's also inconsistent with the way other migrations are implemented.

However the results are the same and it's working.

sevein (commented):
> Connected to archivematica/Issues#46

AFAIK this is not going to work unless you add it to the PR description field.

ross-spencer:
Hmm, yeah, that connected business. I've added it to the description now.

src/dashboard/src/main/migrations/0057_7zip_no_compression.py
ross-spencer: This is useful to know. Thanks @sevein.

PR #1233, opened 2018-08-03
Delay submission of Upload DIP to ArchivesSpace after pair reviewing

sevein:
This pull request is an alternative to #1177 that does not require workflow modifications by making changes in the user interface so the job "Upload DIP" is not executed until the user is presented with the pair reviewing page and submits the form.

Once submitted, the user is forwarded to the ingest grid page and it's prompted with the "Choose Config for ArchivesSpace DIP Upload". Further work could be done to remove this step but that would require a database migration which is not ideal in this work because it's targeting a point release too.

More details have been included in the description of the commits.

This is connected to #1112.

### DIFF
--- a/src/dashboard/src/components/ingest/views_as.py
+++ b/src/dashboard/src/components/ingest/views_as.py
@@ -3,9 +3,12 @@ from functools import wraps
 import json
 import logging
 
+from django.contrib import messages
 from django.core.urlresolvers import reverse
 from django.http import HttpResponse, HttpResponseBadRequest, HttpResponseRedirect, HttpResponseServerError
+from django.utils.translation import ugettext as _
 
+from contrib.mcp.client import MCPClient
 from components import advanced_search
 from main import models
 
@@ -209,3 +212,33 @@ def ingest_upload_as_match(request, uuid):
         return HttpResponse(status=204)
     else:
         return HttpResponse(status=405)
+
+
+def complete_matching(request, uuid):
+    """Complete the process by moving to the ArchivesSpace upload chain.
+
+    This is the final step that we execute when the user has reviewed the
+    matchings and submits the form. It only accepts POST because the action
+    performed is not idempotent.
+    """
+    if request.method != 'POST':
+        return HttpResponse(status=405)
+    try:
+        client = MCPClient()
+        client.execute_unit(
+            uuid,
+            uid=request.user.id,
+            # Microservice: Upload DIP
+            mscl_id="92879a29-45bf-4f0b-ac43-e64474f0f2f9",
+            # Chain: Upload DIP to ArchivesSpace
+            choice="3572f844-5e69-4000-a24b-4e32d3487f82")
+    except Exception as err:
+        messages.error(request, _("Operation failed, please try again later"
+                                  " or contact your administrator."))
+        logger.error("Upload DIP to ArchivesSpace failed: %s", err)
+        return HttpResponseRedirect(
+            reverse(
+                "components.ingest.views_as.ingest_upload_as_review_matches",
+                args=[uuid]))
+    return HttpResponseRedirect(
+        reverse("components.ingest.views.ingest_grid"))
--- a/src/dashboard/src/contrib/mcp/client.py
+++ b/src/dashboard/src/contrib/mcp/client.py
@@ -21,12 +21,23 @@ import logging
 from django.conf import settings
 import gearman
 
+from main.models import Job
+
 
 LOGGER = logging.getLogger('archivematica.dashboard.mcp.client')
 
 
 class RPCError(Exception):
     pass
+
+
+class NoJobFoundError(Exception):
+    def __init__(self, *args, **kwargs):
+        try:
+            message = args[0]
+        except IndexError:
+            message = "No job was found"
+        super(NoJobFoundError, self).__init__(message)
 
 
 INFLIGHT_POLL_TIMEOUT = 5.0
@@ -46,6 +57,24 @@ class MCPClient:
         gm_client.submit_job("approveJob", cPickle.dumps(data), None)
         gm_client.shutdown()
         return
+
+    def execute_unit(self, unit_id, choice, mscl_id=None, uid=None):
+        """Execute the jobs awaiting for approval associated to a given unit.
+
+        Use ``mscl_id`` to pass the ID of the chain link to restrict the
+        execution to a single microservice.
+        """
+        kwargs = {
+            'currentstep': Job.STATUS_AWAITING_DECISION,
+            'sipuuid': unit_id,
+        }
+        if mscl_id is not None:
+            kwargs['microservicechainlink_id'] = mscl_id
+        jobs = Job.objects.filter(**kwargs)
+        if len(jobs) < 1:
+            raise NoJobFoundError()
+        for item in jobs:
+            self.execute(item.pk, choice, uid)
 
     def list(self):
         gm_client = gearman.GearmanClient([self.server])
--- a/src/dashboard/src/media/js/ingest.js
+++ b/src/dashboard/src/media/js/ingest.js
@@ -294,22 +294,28 @@ $(function()
 
           var chainId = $select.find('option:selected').val();
           var unitId = this.model.sip.get('uuid');
-          var loadingMessage = gettext('Loading...');
 
-          // "Upload DIP to Archivists Toolkit" chain matched by its UUID.
-          // Redirect to object/resource mapping pages.
-          if (chainId == 'f11409ad-cf3c-4e7f-b0d5-4be32d98229b')
-          {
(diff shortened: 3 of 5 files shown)


src/dashboard/src/contrib/mcp/client.py
```diff
@@ -47,6 +49,24 @@ def execute(self, uuid, choice, uid=None):
...
+        """
+        kwargs = {
+            'currentstep': Job.STATUS_AWAITING_DECISION,
+            'sipuuid': unit_id,
+        }
+        if mscl_id is not None:
+            kwargs['microservicechainlink_id'] = mscl_id
+        jobs = Job.objects.filter(**kwargs)
+        if len(jobs) < 1:
+            raise Exception('No job was found.')
```
ross-spencer: We can define this as a custom exception, `MCPNoJobFoundException` maybe? This would mean we can be more explicit about the exception we want to handle in calling functions. This reduces the risk of hiding other exceptions/bugs that might be in calling functions.

src/dashboard/src/components/ingest/views_as.py
```diff
@@ -209,3 +212,33 @@ def ingest_upload_as_match(request, uuid):
...
+        client.execute_unit(
+            uuid,
+            uid=request.user.id,
+            # Microservice: Upload DIP
+            mscl_id="92879a29-45bf-4f0b-ac43-e64474f0f2f9",
+            # Chain: Upload DIP to ArchivesSpace
+            choice="3572f844-5e69-4000-a24b-4e32d3487f82")
+    except Exception as err:
+        messages.error(request, _("Operation failed, please try again later"
+                                  "or contact your administrator."))
```
ross-spencer: I think we want to try to avoid catching generic exceptions as much as possible. See below.

src/dashboard/src/media/js/ingest.js
```diff
@@ -295,25 +295,36 @@ $(function()
...
+          // we're going to collect the data required and continue the work.
+          //
+          // In other words, we're not going to execute the next job from
+          // JavaScript as we expect the pairing page to do it once the user
+          // has paired the items.
+          var dipUploadWithMappingPage = {
+            '- Upload DIP to ArchivesSpace':
+              '/ingest/' + this.model.sip.get('uuid') + '/upload/as/',
+            '- Upload DIP to Archivists Toolkit':
+              '/ingest/' + this.model.sip.get('uuid') + '/upload/atk/',
```
ross-spencer: I don't know how likely it is we need to handle a situation where the sip uuid isn't found? i.e. `this.model.sip.get('uuid')` but it seems we can avoid this repetition in calling `get(...)` twice, and increase the potential for handling that by assigning it to a variable first before assigning it in this object.

ross-spencer (changes requested):
Some small changes requested here @sevein. It looks like a nice solution compared to having to introduce a migration for a point release.

src/dashboard/src/components/ingest/views_as.py
sevein: I'm not sure. What I'm doing is a common pattern in model-view-controller. This is a view and the main goal is to output information that can be understood by the user. If I don't handle a particular exception then it would bubble up and be caught by the outer middleware which would present a generic error page - I'm trying to avoid that.

src/dashboard/src/contrib/mcp/client.py
sevein: I didn't do it because I thought it didn't have an utility, do you think it's still needed after reading my response above? I can add it if you wish. The exception message would still be logged but at the moment we don't want to do anything specific about this error.

src/dashboard/src/media/js/ingest.js
sevein: I've declared the variable (`unitId`) as suggested but I don't know how I could handle and report the error. On one hand I don't think a backbone model would make it this far without an ID. But also I don't think there's logic in place at the moment to report the error to the user. What would you do?

```javascript
if (!this.model.sip.get('uuid')) {
    // Would you return?
    // Log to the console?
    // ...?
}
```

sevein:
Ready for review again!

sevein:
Reminder: once merged, I need to port this to stable/1.7.x. It'll conflict in `ingest.js`.

src/dashboard/src/contrib/mcp/client.py
ross-spencer: I can't see the original message in this thread but I remember the gist was that we want to return something simple that the user can understand. I had imagined that you could make the exception specific, because at the end of the day, it is, and then from the dashboard still return a generic message to the user, with a specific message to the logger. I wonder would that work? Or is it likely not robust enough for what a user might experience in the UI?

src/dashboard/src/media/js/ingest.js
```diff
@@ -294,22 +294,28 @@ $(function()
 
           var chainId = $select.find('option:selected').val();
           var unitId = this.model.sip.get('uuid');
```
ross-spencer, line 296: Thanks @sevein. In response to your other message about how you'd handle a potential exception here, I don't think we need to do that at this point. You're right that it seems highly unlikely.

src/dashboard/src/media/js/ingest.js
```diff
@@ -294,22 +294,28 @@ $(function()
...
+            $('body').html('<h1 style="text-align: center;">' + gettext('Loading...') + '</h1>');
+            window.location.href = dipUploadWithMappingPage[chainId];
 
-          // "Upload DIP to ArchivesSpace" chain matched by its UUID.
-          // Redirect to object/resource mapping pages.
-          if (chainId == '3572f844-5e69-4000-a24b-4e32d3487f82')
-          {
-            $('body').html('<h1>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;' + loadingMessage + '</h1>');
-            window.location.href = '/ingest/' + unitId + '/upload/as/';
+            // TODO: needs to be tested with Archivists' Toolkit.
```
ross-spencer, line 315: What's the Artefactual best practice for leaving `TODO` in code? A suggestion I've made in the past is to open a new GitHub issue where this type of test might be out of scope? It could also be left as a note in the acceptance criteria?

ross-spencer (commented):
Feel free to make these changes, or leave as is @sevein. A few discussion points we can pick up on another time. 

I'm away until Monday after today.  If you make any changes you need approval on and get them in later today I can make sure I loop before COP. Otherwise, can pick up Monday if needed. :+1:

src/dashboard/src/media/js/ingest.js
sevein: I'll add a note, good idea, thanks.

src/dashboard/src/contrib/mcp/client.py
sevein: If I understand correctly what you mean, yes it would work, you can definitely create as many custom exceptions as you want and raise them and still result in the same behaviour where I have a catch-all up in the stack as I'm doing.

However I don't know if I see the benefit, to me it's like a `if False: ...` ([we used to have some](https://github.com/artefactual/archivematica-history/blob/407bcb082f382dcfc54c5ec9b22a819ee6903402/src/MCPServer/lib/archivematicaMCP.py#L241)) which I'd remove it right away because it's dead code. Wouldn't this be similar? Excepting that it's harder to reason about because you also have to check its usage outside the function? It's like a fancy goto.

In any case, I can see the value of making this available for a potential use case in the future. Let me know if this would work for you:

```python

# In contrib.mcp.client

class NoJobFoundError(Exception):
    def __init__(self, *args, **kwargs):
        try:
            message = args[0]
        except IndexError:
            message = "No job was found"
        super(NoJobFoundError, self).__init__(message)


# Usage:
raise NoJobFoundError()
raise NoJobFoundError("custom error message")
```

ross-spencer: This looks good @sevein :)

src/dashboard/src/components/ingest/views_as.py
```diff
@@ -209,3 +212,33 @@ def ingest_upload_as_match(request, uuid):
...
+            # Chain: Upload DIP to ArchivesSpace
+            choice="3572f844-5e69-4000-a24b-4e32d3487f82")
+    except Exception as err:
+        messages.error(request, _("Operation failed, please try again later"
+                                  " or contact your administrator."))
+        logger.error("Upload DIP to ArchivesSpace failed: %s", err)
+        return HttpResponseRedirect(
+            reverse(
+                "components.ingest.views_as.ingest_upload_as_review_matches",
+                args=[uuid]))
```
sevein, line 242: @ross-spencer thanks I've added `NoJobFoundError`.

To double check, should the snippet above remain the same?

ross-spencer: :+1:

sevein:
e209c8eafb0855964d542206079b36eac5e66cc3 and bc4c887db40bb3d9cb723eac4580b8bea72c25c3 in `stable/1.7.x`.

My mistake was to target qa/1.x first, I should have started with stable/1.7.x.

Issue #1173, opened 2018-06-28
Enhancement: allow aip reindexing script to be run post-ES upgrade


sromkey:
Related to #1171, once Elasticsearch is upgraded to 6.x users will need to be able to rebuild their AIP index using the new version.

jraddaoui:
There are three commands now available in the Dashboard to rebuild the Elasticsearch indexes:

Rebuild AIPs and AIP files indexes (requires access to the AIPs in the local filesystem):
```
./manage.py rebuild_elasticsearch_aip_index_from_files /var/archivematica/sharedDirectory/www/AIPsStore"
```

Rebuild Transfers and Transfer files indexes (from the database):
```
./manage.py rebuild_transfer_backlog"
```

Rebuild all indexes (from another ES cluster/server using the 1.x version):
```
./manage.py reindex_from_remote_cluster http://192.168.168.196:9200"
```

Check each command help (`-h`) or [their code](https://github.com/artefactual/archivematica/tree/qa/1.x/src/dashboard/src/main/management/commands) for more information on how to run those commands.

sromkey:
We'll need to test this as part of upgrade testing. We should make sure AIPs are in storage and transfers are in backlog pre-update and then rebuild the AIP index/Transfer index after upgrading. Adding @scollazo as an assignee as he can help us coordinate.

sallain:
To do items:
* @jraddaoui to wrap up docs PR 253
* @scollazo to test upgrading a pipeline that has packages in backlog and AIP storage

jraddaoui:
Docs PR is ready for review.

Issue #1172, opened 2018-06-28
Enhancement: make indexless more configurable


sromkey:
Users may want to use Archivematica in indexless mode, but may still want to access transfer backlog (through the Backlog and Appraisal tabs) or archival storage (through the archival storage tab). This would give some increased performance that is associated with running Archivematica in indexless mode but allows me more flexibility for running different digital preservation workflows.

sromkey:
Specifically:
Disabling the search index currently disables three tabs (Backlog, Appraisal and Archival Storage).
The proposal is to be able to disable only those related to the transfers index (Backlog, Appraisal, and the associated micro-services) or only those related to the AIP index (Archival Storage and associated micro-services).

jraddaoui:
Hi @jhsimpson,

First I was thinking to change the current `SEARCH_ENABLED` environment/config variable to accept a list of "indexes" (e.g.: `SEARCH_ENABLED=transfers, aips`), but later I've realized that we could maintain that variable, to avoid changes in the instances using it, and add a couple of variables more (all default to `true`):

- `SEARCH_ENABLED`: takes precedence over the next two if set to `false`.
- `SEARCH_TRANSFERS_ENABLED`: if `SEARCH_ENABLED=true`, enable/disable Transfer related index(es).
- `SEARCH_AIPS_ENABLED`: if `SEARCH_ENABLED=true`, enable/disable AIP related index(es).

This is just to maintain the current var., otherwise we could just use the last two.

Thoughts?

jraddaoui:
Also, we could have this ready before the 1.8 code freeze by the end of this month if we take a decision soon enough. **At least we should think about including #1174 in 1.8 if not this one.**

sevein:
Can't we extend the purpose of SEARCH_ENABLED while maintaining backward compatibility? That would be my choice only because I think we're adding too many settings which is inconvenient for the user particularly because in some cases like in this one you have to define them across more than one service.

A possible approach...

```
def _search_enabled(value):
    known_indices = ("transfers", "aips",)
    boolean = _fancy_booleaner(value) #  e.g. value in ("yes", "y", "True", "true", "1",)
    if boolean is not None:
        return boolean
    else
        return set([item for item in var.split(",") if item in known_indices])

SEARCH_ENABLED = _search_enabled(config.get('search_enabled'))
```

jraddaoui:
I like that idea, thanks @sevein!

[...]

Issue #1171, opened 2018-06-28
Problem: Elasticsearch version is outdated


sromkey:
Archivematica is using an outdated version of Elasticsearch which is known to have security vulnerabilities. It should be upgraded to version 6.x to take advantage of recent updates.

PR #1234, opened 2018-08-04
Enhance search enabled configuration

jraddaoui:
Allow to partially disable/enable Elasticsearch indexing. Extend the
current `search_enabled` config. var. behavior in all the components
to allow a string with a list of enabled search indexes separated by
comma but maintain the current boolean functionality. Available options
for the list are `transfers`, for the Transfer related indexes and
`aips` for the AIPs related indexes. Check the install README files
from each component for more info.

Enhance config. mapping definition to allow a 'process_function' field
where a callback can be passed to allow a more complex parsing process
for specific attributes. Those callbacks must accept the current 
'appconfig.Config' object and the attribute section string.

**Questions:**
This is currently only extending the existing functionality, but I wonder:

- Should we add a note somewhere indicating that the `search_enabled` configuration should be consistent in the three components?
- Should we try to change this multiple declaration at some point and use some global settings for the three components?
- Should this setting be checked somewhere else? For example, showing a warning in the [rebuild commands](https://github.com/artefactual/archivematica/tree/stable/1.7.x/src/dashboard/src/main/management/commands). I can double check this in #1171 and #1173.
- Is this config/env var documented somewhere else besides the install readme files?
- Since we are maintaining backwards compatibility, I guess we don't need to change the declaration of this config/env var in other repositories, but let me know if I'm wrong.

Connects to #1172.

### DIFF
--- a/src/MCPClient/lib/settings/common.py
+++ b/src/MCPClient/lib/settings/common.py
@@ -21,9 +21,8 @@ import logging
 import logging.config
 import os
 
-from appconfig import Config
+from appconfig import Config, process_search_enabled
 import email_settings
-
 
 CONFIG_MAPPING = {
     # [MCPClient]
@@ -37,10 +36,7 @@ CONFIG_MAPPING = {
     'client_modules_file': {'section': 'MCPClient', 'option': 'archivematicaClientModules', 'type': 'string'},
     'elasticsearch_server': {'section': 'MCPClient', 'option': 'elasticsearchServer', 'type': 'string'},
     'elasticsearch_timeout': {'section': 'MCPClient', 'option': 'elasticsearchTimeout', 'type': 'float'},
-    'search_enabled': [
-        {'section': 'MCPClient', 'option': 'disableElasticsearchIndexing', 'type': 'iboolean'},
-        {'section': 'MCPClient', 'option': 'search_enabled', 'type': 'boolean'},
-    ],
+    'search_enabled': {'section': 'MCPClient', 'process_function': process_search_enabled},
     'index_aip_continue_on_error': {'section': 'MCPClient', 'option': 'index_aip_continue_on_error', 'type': 'boolean'},
     'capture_client_script_output': {'section': 'MCPClient', 'option': 'capture_client_script_output', 'type': 'boolean'},
     'removable_files': {'section': 'MCPClient', 'option': 'removableFiles', 'type': 'string'},
--- a/src/archivematicaCommon/lib/appconfig.py
+++ b/src/archivematicaCommon/lib/appconfig.py
@@ -1,3 +1,18 @@
+"""
+Configuration helper for MCPServer, MCPClient and Dashboard, used in:
+
+    - MCPClient/lib/settings/common.py
+    - MCPServer/lib/settings/common.py
+    - Dashboard/src/settings/base.py
+
+Config. attributes are declared on those settings files and they can be defined
+by a dictionary indicating the 'section', 'option' and 'type' to be parsed by
+the Config class. They can also be defined by a list of the same type of
+dictionary and, in that case, the first obtained value will be the one returned.
+Alternatively, they can include the 'section' and a 'process_function' callback
+where a specific parsing process can be defined. Those callbacks must accept the
+current appconfig Config object and the section.
+"""
 import ConfigParser
 
 from django.core.exceptions import ImproperlyConfigured
@@ -6,14 +21,7 @@ from env_configparser import EnvConfigParser
 
 
 class Config(object):
-    """Configuration helper for MCPServer, MCPClient and Dashboard.
-
-    It wraps EnvConfigParser. It is used in the following files:
-
-        - MCPClient/lib/settings/common.py
-        - MCPServer/lib/settings/common.py
-        - Dashboard/src/settings/base.py
-    """
+    """EnvConfigParser wrapper"""
     def __init__(self, env_prefix, attrs):
         self.config = EnvConfigParser(prefix=env_prefix)
         self.attrs = attrs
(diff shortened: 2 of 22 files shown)


src/MCPClient/lib/settings/common.py
```diff
@@ -38,8 +38,9 @@
     'elasticsearch_server': {'section': 'MCPClient', 'option': 'elasticsearchServer', 'type': 'string'},
     'elasticsearch_timeout': {'section': 'MCPClient', 'option': 'elasticsearchTimeout', 'type': 'float'},
     'search_enabled': [
-        {'section': 'MCPClient', 'option': 'disableElasticsearchIndexing', 'type': 'iboolean'},
```
sevein: This was on purpose, now I remember. For backward compatibility reasons. We made a release at some point where `disableElasticsearchIndexing` was the name of the attribute so now we're committed to support it. This was only true in MCPClient, that's why you don't see it in other components.

sevein:
> Should we add a note somewhere indicating that the search_enabled configuration should be consistent in the three components?

That's a good point. I think we should. There is a section in archivematica-docs dedicated to searches (see related PR: https://github.com/artefactual/archivematica-docs/pull/68).

> Should we try to change this multiple declaration at some point and use some global settings for the three components?

Yes, that's something we should do. We need some kind of global application configuration. We're aware of the problem but I don't think we've filed a bug yet!

> Should this setting be checked somewhere else? For example, showing a warning in the rebuild commands. I can double check this in #1171 and #1173.

From the commands you could tell if the config is disabled in the Dashboard so you could show a warning if you have time, I agree that'd be nice.

> Is this config/env var documented somewhere else besides the install readme files? Since we are maintaining backwards compatibility, I guess we don't need to change the declaration of this config/env var in other repositories, but let me know if I'm wrong.

I think you're right. We should be safe without further changes. Good docs in archivematica-docs extending that PR I've linked above should be enough.

src/archivematicaCommon/lib/appconfig.py
```diff
@@ -39,7 +41,10 @@ def get(self, attr, default=None):
 
         attr_opts = self.attrs[attr]
         if isinstance(attr_opts, list):
-            return self.get_from_opts_list(attr, attr_opts, default=default)
+            value = self._get_from_opts_list(attr, attr_opts, default=default)
+            if attr == 'search_enabled':
+                value = self._normalize_search_enabled(value)
```
sevein: Would it be possible if you have time to update this solution so it's not specific to given attributes? E.g. we could have a new `list` type along with `allowed_values`:

    {'section': 'MCPServer', 'option': 'search_enabled', 'type': 'list', 'allowed_values': ['aips', 'transfers']},

I'd like `appconfig.Config` to be agnostic to specific configurations as much as possible. Would that be possible? Am I asking too much? I'm not sure how much extra work would that require at this point.

sevein (commented):
Looking good, thanks @jraddaoui. Only one comment about the changes in `Config`.

src/archivematicaCommon/lib/appconfig.py
jraddaoui: Thanks @sevein!

The problem with this particular attribute is that it should allow boolean values alongside the list and always return a set. I could try to define the normalization rules in the config mapping at the attribute level alongside the options, but that will add some complexity to the mapping and normalization code. Considering that all this (the multi-option and normalization) looks to be done only for this attribute at this point I'm not sure if it worths the effort, but I'll give it a quick try ;)

jraddaoui: How about something like this for the config mapping:

```python
CONFIG_MAPPING = {
    'search_enabled': {
        'section': 'Dashboard',
        'options': [
            {'name': 'disable_search_indexing', 'type': 'iboolean'},
            {'name': 'search_enabled', 'type': 'boolean'},
            {'name': 'search_enabled', 'type': 'list', 'allowed_values': set(['aips', 'transfers'])},
        ],
        'normalize': [
            {
                'type': bool,
                'rules': [
                    {'value': True, 'output': set(['aips', 'transfers'])},
                    {'value': False, 'output': set()},
                ],
            },
        ],
    },
}
```

sevein: I like it but I recognise it's becoming too complicated as you anticipated.

I wonder if we should step back and return to your original suggestion. Perhaps we could teach `Config` how to deal with configs that have a `filter` callable. It'd be like what you did excepting that it's up to the user to pass a callable where the logic specific to a particular configuration attribute is kept, e.g.:

```python
from app_config import Config, process_search_enabled

CONFIG_MAPPING = {
    'search_enabled': {
        'section': 'Dashboard',
        'filter': process_search_enabled,
   }
}
```

`Config` would call the callable in `filter` when available otherwise fallback to the default behaviour (i.e. `options`). The callable would receive as a parameter the instance of `Config` so it can do with it whatever it wants.

src/archivematicaCommon/lib/appconfig.py
```diff
@@ -40,6 +40,8 @@ def get(self, attr, default=None):
         attr_opts = self.attrs[attr]
         if isinstance(attr_opts, list):
             return self.get_from_opts_list(attr, attr_opts, default=default)
+        if all(k in attr_opts for k in ('section', 'process_function')):
+            return attr_opts['process_function'](self, attr_opts['section'])
```
sevein, line 52: :clap:

sevein (approved):
Thank you so much!

sallain:
@sevein does this resolve #1173?

jraddaoui:
Hi @sara, it does not. I'll work on that one alongside #1171 in the near future.

jraddaoui:
Sorry, @sallain.

Issue #1101, opened 2018-05-21
Problem: workflow labels can't be translated


sevein:
This feature is not being included in Archivematica 1.8. But it may still show up in the AM18 waffle board because some of the refactoring needed in advance was included in AM18: see #1235, #1238 and #1239.

---

Archivematica 1.7 included user interface translation support (see #506). This work did not include the interface strings found in [Archivematica's workflows](https://wiki.archivematica.org/MCP) because they are in the application database making very difficult to model a translatable schema and push changes via database migrations.

It was decided that the workflow data should be removed from the application database. There was an early attempt in #576 that we are planning to revive.

Action items:

- [x] (PR https://github.com/artefactual/archivematica/pull/1235) **Pass `link_id` to front-end.** This is so our JavaScript code can continue doing conditional representation of micro-services when the labels are translated by using their UUIDs instead.
- [x] (PR https://github.com/artefactual/archivematica/pull/1239) **Refactor processing configuration page.** We want to have the Django form behind this page updated so it does not interact with the workflow models and the processing configuration - this functionality should be moved to MCPServer and exposed via the RPC interface where further work is going to manage workflows using the JSON-encoded document.
- [x] (PR https://github.com/artefactual/archivematica/pull/1238) **Update API's approve_transfer**. Its view uses workflow models, it should be updated and talk to MCPServer's RPC interface as needed.
- [x] (PR https://github.com/artefactual/archivematica/pull/1251) **Creation of workflow module**. So we can load our JSON-encoded workflow document. Update MCPServer to make use of it. Use translated labels in RPC's job_awaiting_for_approval method (used in `ingest_status` and `transfer_status` views). Also: remove workflow models. This will require a new migration that removes the tables and the removal of our fixture imported using `loaddata`. Processing models `UnitVariable` or `Job` need to be updated so foreign keys are updated by `UUIDField`. `json_links` and `graph_links` will stop working therefore they need to be removed.

For future releases:
- **Workflow schema improvements**. See https://gist.github.com/sevein/63e751acd37bde5590d680344ad8b3d5 for more details.
- **New workflow visualization and editor tool**. We need to provide the same functionality offered by `graph_links` until now. We also have to make sure that editing the workflow is still possible as it was before via database migrations. Work in progress: https://github.com/sevein/amflow.

Other improvements that were included in #576 like the gRPC server, removal of the `/mcp` endpoints or the thread-safe `UnitChoices` dict have been excluded this time.

[...]

PR #576, opened 2017-02-28
Dashboard: i18n, refs #8161 (2/2)

sevein:
Continuation of #506 which is already merged!

**https://github.com/artefactual/archivematica/pull/584 needs to be merged first!**

- [x] MCPServer API: server
- [x] MCPServer API: client
- [x] archivematica-workflow

It needs some polishing and testing.

PR #1235, opened 2018-08-07
Dashboard: retrieve config using UUIDs

sevein:
This commit updates our `ingest.js` and `jobs.js` files so they don't
use labels for conditional behaviour. Instead, we're using UUIDs which
are going to be the same regardless the language used in the user
interface.

Also, this commit is removing some dead code:
- Review button in Store DIP (broken and unused).
- Dialog in unexistent micro-service:
  "Normalize submission documentation to preservation format".

This is connected to https://github.com/artefactual/archivematica/issues/1101, https://github.com/archivematica/Issues/issues/62.

### DIFF
--- a/src/dashboard/src/components/transfer/views.py
+++ b/src/dashboard/src/components/transfer/views.py
@@ -156,6 +156,7 @@ def status(request, uuid=None):
                 item['jobs'].append(newJob)
                 newJob['uuid'] = job.jobuuid
                 newJob['type'] = job.jobtype
+                newJob['link_id'] = job.microservicechainlink.pk
                 newJob['microservicegroup'] = job.microservicegroup
                 newJob['subjobof'] = job.subjobof
                 newJob['currentstep'] = job.currentstep
--- a/src/dashboard/src/media/js/ingest.js
+++ b/src/dashboard/src/media/js/ingest.js
@@ -177,15 +177,6 @@ $(function()
                         }
                     }]
                 });
-
-              if (self.model.jobs.detect(function(job)
-                {
-                  return job.get('type') === 'Normalize submission documentation to preservation format';
-                }))
-              {
-                dialog.find('input, select, textarea').prop('disabled', true).addClass('disabled');
-                dialog.dialog('option', 'buttons', dialog.dialog('option', 'buttons').splice(0,1));
-              }
             };
 
           $.ajax({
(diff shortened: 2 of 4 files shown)


sevein:
For QA, after merging, it should be possible to verify that:

- "Approve normalization" includes Review button (only while waiting for approval).
- "Approve normalization" includes Report button.
- Clicking on "Upload DIP to Archivists Toolkit" opens new page as usual.
- Clicking on "Upload DIP to ArchivesSpace" opens new page as usual.
- Clicking on "Upload DIP to AtoM/Binder" opens new modal as usual.
- "Move to the uploadedDIPs directory" includes Review button.
- "Store AIP" includes Review button (while waiting for approval).

src/dashboard/src/media/js/ingest.js
```diff
@@ -177,15 +177,6 @@ $(function()
...
                 });
-
-              if (self.model.jobs.detect(function(job)
-                {
-                  return job.get('type') === 'Normalize submission documentation to preservation format';
-                }))
-              {
-                dialog.find('input, select, textarea').prop('disabled', true).addClass('disabled');
-                dialog.dialog('option', 'buttons', dialog.dialog('option', 'buttons').splice(0,1));
-              }
```
sevein, line 188: Unused.

src/dashboard/src/media/js/jobs.js
```diff
@@ -472,11 +471,13 @@ var MicroserviceGroupView = Backbone.View.extend({
 
         // Add link to browse SIP before it's made into an AIP
         if (
-          (job.get('currentstep') == job.sip.statuses['STATUS_AWAITING_DECISION'] && -1 < jQuery.inArray(job.get('type'), ['Store AIP', 'Approve normalization']))
-          || job.get('type') == 'Move to the uploadedDIPs directory'
-          || job.get('type') == 'Store DIP'
```
sevein: `Store DIP` not included in my change because is unused - see https://github.com/archivematica/Issues/issues/62.

src/dashboard/src/components/transfer/views.py
```diff
@@ -156,6 +156,7 @@ def encoder(obj):
                 item['jobs'].append(newJob)
                 newJob['uuid'] = job.jobuuid
                 newJob['type'] = job.jobtype
+                newJob['link_id'] = job.microservicechainlink.pk
```
ross-spencer, line 159: A question outside of CR, but with these values in `transfer/views/py` and `ingest/views.py` could they be inherited from elsewhere instead of defining them twice?

src/dashboard/src/media/js/ingest.js
```diff
@@ -242,20 +233,24 @@ $(function()
             this.$('.job-detail-actions').append($select);
           }
 
-          if ('Approve normalization' == this.model.get('type'))
+          // Approve normalization
+          if ('de909a42-c5b5-46e1-9985-c031b50e9d30' == this.model.get('link_id'))
```
ross-spencer: I believe that we can make this snippet clearer by adding some verbosity, e.g.

`approve_normalization_mscl = 'de909a42-c5b5-46e1-9985-c031b50e9d30'`

Or a comment highlighting the `de909a4` is the UUID for approve normalization from the MS chainlink table. 

Second, is there a reason for the construct to be:

`if 'constant' == 'value returned from function'` - I would normally expect it to be written:

`if (this.model.get('link_id') == 'value_we_want_to_confirm') `

(This seems to be the same for most of the calls below so this comment applies to those as well. My concern is that UUIDs are in a lot of places in Archivematica. If we can be as explicit as possible about the source of a UUID to future readers or folks new to the code (even if it is obvious to us), then it saves a little time for them.)

src/dashboard/src/media/js/ingest.js
```diff
@@ -295,25 +290,28 @@ $(function()
...
             window.location.href = '/ingest/' + this.model.sip.get('uuid') + '/upload/as/';
           }
 
+          // Upload DIP to AtoM/Binder
           // If no identifier for the AtoM or Binder SWORD V1 deposit endpoint
           // provided at start of transfer, display a modal dialog to request
           // such here.
           if (
-            '- upload-qubit_v0.0' == $select.find('option:selected').text()
+            '0fe9842f-9519-4067-a691-8a363132ae24' == $select.find('option:selected').val()
```
ross-spencer: Related to the above. This UUID is a chain not a chain-link, so it might be useful to elaborate on that.

ross-spencer (changes requested):
Some changes requested to improve the documentation of the code now we're using UUIDs more widely but otherwise LGTM.

src/dashboard/src/components/transfer/views.py
sevein: Yes, they're almost identical.

sevein: During a big refactor that was never merged because of its size I tried to fix that:

https://github.com/artefactual/archivematica/pull/576/files#diff-254823927ae6fbe4584582f13c9b0c18R327

I learned the lesson and now I keep my work laser focused, e.g. see the second take of workflow translations here: https://github.com/artefactual/archivematica/issues/1101.

sevein:
@ross-spencer ok noted and fixed, thanks. Ready to be reviewed again.

ross-spencer (approved):
:+1:

PR #1237, opened 2018-08-12
Fix flake8 builds in Py3

sevein:
As you probably know already, in Py3 `str` is text and `bytes` is binary, `basestring` and `unicode` were removed and many other subtle differences that caused our flake8-py36 build to break. We used to ignore its build status. This PR uses the `six` library so the flake8 errors disappear and updates our Travis config so the outcome of flake8-py36 is not ignored anymore.

Connects to https://github.com/archivematica/Issues/issues/70.

### DIFF
--- a/src/archivematicaCommon/lib/databaseFunctions.py
+++ b/src/archivematicaCommon/lib/databaseFunctions.py
@@ -31,7 +31,7 @@ import uuid
 
 from django.db import close_old_connections
 from django.db.models import Q
-from django.utils import timezone
+from django.utils import six, timezone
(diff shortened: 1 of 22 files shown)


jraddaoui (approved):
Awesome!

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -359,7 +359,7 @@ def deUnicode(str):
     """
     if str is None:
         return None
-    return unicode(str).encode('utf-8')
+    return six.text_type(str).encode('utf-8')
```
sevein: I should change the name of the argument because `str` is a known type.

PR #1238, opened 2018-08-12
API: move workflow model lookups to MCPServer

sevein:
We want MCPServer to be the only component interacting with workflow
models. In this commit, MCPServer learns two new functions in order to
approve transfers and reingests. The Dashboard is updated accordingly to
invoke these functions and have the old code removed. With this, the Dashboard
API stops using workflow models.

Our pseudo-RPC interface based on Gearman has been updated so the server
can return error messages to the client. Dashboard's RPC client has been
updated to identify these errors and raise a `RPCServerError` exception.

This is connected to #1101.

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -85,6 +85,21 @@ def _api_endpoint(expected_methods):
     return decorator
 
 
+def _ok_response(message, **kwargs):
+    """Mixin to return API responses to the user."""
+    payload = {"message": message}
+    status_code = kwargs.pop("status_code", 200)
+    payload.update(kwargs)
+    return helpers.json_response(payload, status_code=status_code)
+
+
+def _error_response(message, status_code=400):
+    """Mixin to return API errors to the user."""
+    return helpers.json_response(
+        {"error": True, "message": message},
+        status_code=status_code)
+
+
 class HttpResponseNotImplemented(django.http.HttpResponse):
     status_code = 501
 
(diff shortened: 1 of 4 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -381,97 +396,59 @@ def unapproved_transfers(request):
...
-        return helpers.json_response(response, status_code=500)
-    else:
-        response['message'] = 'Approval successful.'
-        response['uuid'] = unit_uuid
-        return helpers.json_response(response)
+    The user may find the Package API a better option when the ID of the
+    unit is known in advance.
 
+    The errors returned use the 500 status code for backward-compatibility
+    reasons.
```
jraddaoui, line 405: :cry:

src/dashboard/src/contrib/mcp/client.py
```diff
@@ -27,25 +27,94 @@
...
+    def __init__(self, message=None):
+        if message is None:
             message = "No job was found"
         super(NoJobFoundError, self).__init__(message)
 
 
 INFLIGHT_POLL_TIMEOUT = 5.0
 
 
-class MCPClient:
```
jraddaoui, line 46: :astonished:

src/dashboard/src/contrib/mcp/client.py
```diff
@@ -27,25 +27,94 @@
...
+class MCPClient(object):
+    """MCPServer client (RPC via Gearman)."""
+
     def __init__(self):
         self.server = settings.GEARMAN_SERVER
+        self.timeout = INFLIGHT_POLL_TIMEOUT
```
jraddaoui: Is this really needed? It's passed to the bellow function using the attribute directly and it's not used anywhere else.

src/dashboard/src/contrib/mcp/client.py
```diff
@@ -27,25 +27,94 @@
...
+
+    def _rpc_sync_call(self, ability, data, timeout=INFLIGHT_POLL_TIMEOUT):
```
jraddaoui: Not a big deal but, wouldn't this function be nicer as a decorator?

jraddaoui (approved):
Great job placing functionality where it should be and cleaning things up, thanks @sevein!

I don't know a lot about this workflows yet but the code looks great, just a couple of minor comments. Also, there is quite some inconsistency in the use of single vs double quotes, of course, not a big deal either ;)

src/dashboard/src/contrib/mcp/client.py
sevein: Ops, fixed, thanks!

src/dashboard/src/contrib/mcp/client.py
sevein: I kind of think that we shouldn't abuse decorators because it affects readability. I used them in RPCServer too much and after all I realize they're not idea. There are many discussions out there about them. I'm not sure if I still have a rule of thumb, something that we should explore. For now I'll leave it like this, thanks for the suggestion though, it's worth keeping in mind in further work.

sevein:
> Also, there is quite some inconsistency in the use of single vs double quotes, of course, not a big deal either ;)

I fixed that. I remember Joel mentioning the same thing. I'm trying from now on to stick with double quoting unless the string contains quotes, pep8 recommends not to use the backslash to escape them so I'll just use single quotes only in those cases. It'd be nice to use a flake8 plugin but we'd also need to update the codebase to be consistent and that could be quite a lot of work unless we can automate it, e.g. black does that automatically.

PR #1239, opened 2018-08-13
Populate processing config fields in MCPServer

sevein:
Similar to #1238, we want MCPServer to be the only component interacting
with workflow models. In this commit, the processing configuration form
is updated so all the workflow-related data is pulled from MCPServer via
the RPC interface.

To support this, MCPServer learns a new RPC `getProcessingConfigFields`
and a new `processing_config` module.

The user-facing functionality should not be any different. I've also compared that the XML documents generated are the same. These are only internal implementation changes.

This is connected to #1101.

### DIFF
--- a/src/dashboard/src/components/administration/forms.py
+++ b/src/dashboard/src/components/administration/forms.py
@@ -18,25 +18,23 @@ from __future__ import division
 
 import os
 from lxml import etree
-from collections import OrderedDict
 
 from django import forms
 from django.conf import settings
-from django.db import connection
 from django.forms.widgets import TextInput, Select
 from django.utils.translation import ugettext_lazy as _
 
+from contrib.mcp.client import MCPClient
 from components import helpers
 from installer.forms import site_url_field, load_site_url
-from main import models
+from main.models import Agent, TaxonomyTerm
 
-from abilities import choice_is_available
 import storageService as storage_service
 
 
 class AgentForm(forms.ModelForm):
     class Meta:
-        model = models.Agent
+        model = Agent
         fields = ('identifiervalue', 'name', 'agenttype')
         widgets = {
             'identifiervalue': TextInput(attrs=settings.INPUT_ATTRS),
@@ -221,7 +219,7 @@ class ChecksumSettingsForm(SettingsForm):
 
 class TaxonomyTermForm(forms.ModelForm):
     class Meta:
-        model = models.TaxonomyTerm
+        model = TaxonomyTerm
         fields = ('taxonomy', 'term')
         widgets = {
             "term": TextInput(attrs=settings.INPUT_ATTRS)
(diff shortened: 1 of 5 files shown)


src/dashboard/src/components/administration/views_processing.py
```diff
@@ -47,26 +47,51 @@ def list(request):
...
-            return redirect('components.administration.views_processing.list')
-    else:
-        form = ProcessingConfigurationForm()
-        if name is not None:
-            try:
-                form.load_config(name)
-            except IOError:
-                raise Http404
-    return render(request, 'administration/processing_edit.html', {'form': form})
+
```
sevein, line 51: Functionally this function is very similar but with more exhaustive error handling, e.g. internal server error page is avoided as much as possible and error messages are more explicit, also better logging. The code is also written to avoid nested blocks, i.e. returning early.

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -47,26 +47,51 @@ def list(request):
...
+        return render(
+            request, 'administration/processing_edit.html',
+            {'form': form})
+
+    # Initialize form.
+    try:
+        form = ProcessingConfigurationForm(request.POST or None)
+    except Exception as err:
+        return _report_error(
+            err, "Unable to load processing configuration page.")
```
sevein: It's important to capture potential errors here because the constructor of ProcessingConfigurationForm is going to communicate with MCPServer so we should really expect errors to occur at that point.

src/dashboard/src/components/administration/forms.py
```diff
@@ -519,44 +362,16 @@ def save_config(self):
...
-                    pass
-                else:
-                    for i, item in enumerate(get_duplicated_choices(fprops['label'], choice_name)):
-                        comment = '{} (match {} for "{}")'.format(fprops['label'], i + 1, choice_name)
-                        config.add_choice(item[0], item[1], comment=comment)
+            elif fprops['type'] == 'chain_choice' and 'duplicates' in fprops:
+                desc, matches = fprops['duplicates'][value]
+                for i, match in enumerate(matches):
+                    comment = '{} (match {} for "{}")'.format(fprops['label'], i + 1, desc)
+                    config.add_choice(match[0], match[1], comment=comment)
```
sevein, line 369: This is where we generate config items for multiple links with the same name in our XML document, e.g.:

```xml
        <!-- Normalize (match 1 for "Normalize for preservation") -->
        <preconfiguredChoice>
          <appliesTo>cb8e5706-e73f-472f-ad9b-d1236af8095f</appliesTo>
          <goToChain>612e3609-ce9a-4df6-a9a3-63d634d2d934</goToChain>
        </preconfiguredChoice>
        <!-- Normalize (match 2 for "Normalize for preservation") -->
        <preconfiguredChoice>
          <appliesTo>7509e7dc-1e1b-4dce-8d21-e130515fce73</appliesTo>
          <goToChain>612e3609-ce9a-4df6-a9a3-63d634d2d934</goToChain>
        </preconfiguredChoice>
```

src/dashboard/src/components/administration/forms.py
```diff
@@ -432,49 +243,81 @@ class ProcessingConfigurationForm(forms.Form):
...
+                continue
+
+            choices = opts['choices'] = list(self.EMPTY_CHOICES)
+            if ftype == 'boolean':
+                if 'yes_option' in field:
+                    choices.append((field['yes_option'], _('Yes')))
+                if 'no_option' in field:
+                    choices.append((field['no_option'], _('No')))
+            elif ftype in ('chain_choice', 'replace_dict'):
+                choices.extend(field['options'])
```
sevein, line 305: `field['options']` is now all computed in the server so it gets much simpler here.

src/dashboard/src/components/administration/forms.py
```diff
@@ -432,49 +243,81 @@ class ProcessingConfigurationForm(forms.Form):
...
+        '7079be6d-3a25-41e6-a481-cee5f352fe6e': _('Transcribe files (OCR)'),
+        '087d27be-c719-47d8-9bbb-9a7d8b609c44': _('Select file format identification command (Submission documentation & metadata)'),
+        '01d64f58-8295-4b7b-9cab-8f1b153a504f': _('Select compression algorithm'),
+        '01c651cb-c174-4ba4-b985-1d87a44d6754': _('Select compression level'),
+        '2d32235c-02d4-4686-88a6-96f4d6c7b1c3': _('Store AIP'),
+        'b320ce81-9982-408a-9502-097d0daa48fa': _('Store AIP location'),
+        '92879a29-45bf-4f0b-ac43-e64474f0f2f9': _('Upload DIP'),
+        '5e58066d-e113-4383-b20b-f301ed4d751c': _('Store DIP'),
+        'cd844b6e-ab3c-4bc6-b34f-7103f88715de': _('Store DIP location'),
+    }
```
sevein, line 277: We have to do this until i18n makes it into workflow data. Ugly but just a temporary hack, soon to be fixed as part of #1101.

src/dashboard/src/components/administration/forms.py
```diff
@@ -432,49 +243,81 @@ class ProcessingConfigurationForm(forms.Form):
         'initial': None,
     }
 
+    # This extends the fields defind in the processing_config module.
```
jraddaoui: "defined"

src/dashboard/src/components/administration/forms.py
jraddaoui: :clap:

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -47,26 +47,51 @@ def list(request):
...
-    else:
-        form = ProcessingConfigurationForm()
-        if name is not None:
-            try:
-                form.load_config(name)
-            except IOError:
-                raise Http404
-    return render(request, 'administration/processing_edit.html', {'form': form})
+
+    def _redirect_list(msg=None):
```
jraddaoui: In the end, it's only used once so it may not deserve its own function.

src/dashboard/src/components/administration/views_processing.py
```diff
@@ -47,26 +47,51 @@ def list(request):
...
+                err, "Failed to save processing configuration.")
+        return _redirect_list("Saved!")
+
+    # Process form load.
+    try:
+        form.load_config(name)
+    except IOError:
+        raise Http404
+    except Exception as err:
+        return _report_error(err, "Failed to load processing configuration.")
```
jraddaoui: We should add i18n to the error/success messages in this view.

jraddaoui (changes requested):
This is really nice, thanks @sevein!

Just a couple of comments.

jraddaoui (approved):
Thanks!!

PR #1241, opened 2018-08-15
Docker: multi-stage frontend build

sevein:
This commit changes `dashboard.Dockerfile` so the frontends are built
from a different stage. Intended consequences:

- Reduce the size of the final image (1.42GB -> 1.01GB).
- Avoid issues with shared node_modules in osxfs/docker-for-mac.

It requires Docker 17.05 or newer.

This is connected to https://github.com/artefactual-labs/am/issues/49.
This is connected to https://github.com/artefactual-labs/am/issues/4.

### DIFF
--- a/src/.dockerignore
+++ b/src/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/src/dashboard.Dockerfile
+++ b/src/dashboard.Dockerfile
@@ -1,3 +1,13 @@
+FROM node:10-alpine as webpack-bundle
+COPY dashboard/frontend/transfer-browser/ /src/transfer-browser/
+COPY dashboard/frontend/appraisal-tab/ /src/appraisal-tab/
+RUN apk add --no-cache git && chown -R node:node /src
+USER node
+RUN cd /src/transfer-browser \
+	&& npm install \
+	&& cd /src/appraisal-tab \
+	&& npm install
+
 FROM python:2.7-jessie
 
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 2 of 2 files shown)


sevein:
If this is merged, our `make bootstrap-dashboard-frontend` will stop working. We'd need to update am.git: 1) remove goal from Makefile, 2) add instructions on how to build the frontends from the host during frontend dev work.

ablwr:
Builds great.
Failed when `make bootstrap` as expected.
After locally modifying my Makefile, it ran without fail.

Then I went into transfer-browser to manually build the frontend, but I still get the webpack bug issue and can't build.

Everything loads fine. I can see the transfer browse widget and browse/select files. However, when I select a file, I cannot start a transfer. (The button is available to click but nothing happens).

In the console, the error is `POST http://127.0.0.1:62080/filesystem/transfer/ 404 (NOT FOUND)`

sevein:
> In the console, the error is POST http://127.0.0.1:62080/filesystem/transfer/ 404 (NOT FOUND)

This means that you're using an old version of the transfer browser widget (needs to be built again) that is trying to hit that endpoint which was removed as part of the Package API improvements.

> Then I went into transfer-browser to manually build the frontend, but I still get the webpack bug issue and can't build.

To be clear for others looking a this pull request, the problem that @ablwr is seeing is the following:

```
$ webpack --progress --colors --entry ./app.js
/bin/sh: webpack: command not found
error Command failed with exit code 127.
```

We don't know why this is happening, I can't reproduce.

sevein:
Closing this until we have time to investigate further.

PR #1242, opened 2018-08-21
Create a Dataverse transfer type

### LINKED ISSUE #1119: Problem: Dataverse branch is out of date
The existing Dataverse branch is out of date and needs to be cherry-picked into `qa/1.x`.

Existing: https://github.com/artefactual/archivematica/tree/dev/dataverse-poc (last commits September 2016 and April 2017, looks to have squisked some of the 1_6_0 branch below) 

Additionally, we have two other branches that will all need deleting at the completion of this work. 

* https://github.com/artefactual/archivematica/tree/dev/issue-8693-dataverse-1_6_0 (last commits September 2016 - looks to be based on issue-8693)
* https://github.com/artefactual/archivematica/tree/dev/issue-8693-dataverse (last commits June 2016)

### LINKED ISSUE #1121: Problem: Extract Packages is a mandatory stage for a Dataverse Transfer
The error here: https://github.com/artefactual/archivematica/issues/1120 is a result of `Extract Packages` not running. 

Consider what the best way to approach this is. I.e. do we want extract packages to be mandatory? Does the task call for a different extract packages? Do we want to work with the zip as well as? (i.e. provide the option?)

### LINKED ISSUE #1129: Problem: transfer names can't start with a dash

### LINKED ISSUE #1147: Problem: Dataverse requires its own transfer type

### LINKED ISSUE #1148: Problem: Parse Dataverse Mets fails for some datasets

### LINKED ISSUE #1149: Problem: Generate AIP METS fails for Dataverse SIPs when document empty directories is set in workflow
Testing Dataverse datasets, the Generate AIP METS microservice is failing. This has worked with some datasets, but appears to be failing about half the time. 

screen shot 2018-06-20 at 10 17 53 am

The error message produced is: 
screen shot 2018-06-20 at 10 18 16 am

### LINKED ISSUE #1163: Problem: AIP METS2 in Dataverse Branch needs to be reviewed and merged
Dataverse specific components of the AIP METS2 client script needs to be reviewed, and merged. The merge would see the primary METS work brought up to date with that in `stable/1.7` and the Dataverse METS worked added (or included in a separate module script). 

This is connected to https://github.com/artefactual/archivematica/issues/1119

ross-spencer:
This work introduces a new transfer type for Dataverse, https://dataverse.org/ datasets.

The migration also introduces two new microservices:

* Convert Dataverse Structure
* Parse Dataverse METS

In order to do that, the transfer workflow requires two signals in the form of unit variables to read from when the new microservice tasks need to be performed.

Once the tasks have completed, the workflow is picked up from where it would normally for a standard transfer type.

The migration has a down migration component which does seem to function as anticipated but the code-reviewer may want to test that out as well. 

Further elements you will see in the PR as you look through the commits are the introduction of unit tests for `convert_dataverse_structure.py`. The tests have been introduced to enable refactor of that script over time. It seemed to conform least to AM coding guidelines having been ported from the automation tools. The other script changes have been made as minimal as possible and it is hoped any issues in these scripts can be ironed out over time. 

Comments have been left in code where compromises have been made around scope given the time we have on this work left. 

This work requires storage service `qa/0.x` and (I believe) eventually `stable/0.13` to function. 

Resolves #1147 
Resolves #1163 
Resolves #1149 
Resolves #1121 
Resolves #1119 
Resolves #1148 
Resolves archivematica/issues#59
Resolves archivematica/issues#152
Resolves archivematica/issues#187

Connected to #1147  
Connected to archivematica/issues#104 (Creates a Convert Dataverse Structure set of tests)
Connected to archivematica/issues#57 (Add logging to Dataverse scripts)
Connected to archivematica/issues#131 (AIP Validation)

Replaces https://github.com/artefactual/archivematica/pull/1218.

### DIFF
--- a/src/MCPClient/lib/archivematicaClientModules
+++ b/src/MCPClient/lib/archivematicaClientModules
@@ -1,7 +1,7 @@
 # This file is part of Archivematica.
 #
 # Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
-# 
+#
 # Archivematica is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
@@ -135,3 +135,7 @@ upload-qubit_v0.0 = upload_qubit
 upload-archivesspace_v0.0 = upload_archivesspace
 copyThumbnailsToDIPDirectory_v0.0 = copy_thumbnails_to_dip_directory
 removeDirectories_v0.0 = remove_directories
+
+# Dataverse Scripts
+convertDataverseStructure_v0.0 = convert_dataverse_structure
+parseDataverse_v0.0 = parse_dataverse_mets
--- a/src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
+++ b/src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
@@ -1,0 +1,546 @@
+#!/usr/bin/env python2
+# -*- coding: utf-8 -*-
+
+"""Given a transfer type Dataverse, read the metadata submission object
+``dataset.json`` to generate a Dataverse METS.xml file. This METS file will
+then later be used to create a Transfer METS.xml document as part of the
+Archivematica submission information package (SIP).
+
+The METS.xml will reflect various properties of the ``dataset.json`` file. An
+example of a specific feature of Dataverse is the existence of Bundle objects
+for Tabular data. Bundles contain derivatives of a tabular data file that are
+created by Dataverse to enable the data to be interacted with using as wide a
+range of tools as possible. These Derivatives are transcribed to the Dataverse
+METS.xml.
+
+More information about Dataverse in Archivematica can be found here:
+https://wiki.archivematica.org/Dataverse
+"""
+
+import json
+import os
+import uuid
+
+from lxml import etree
+
+# databaseFunctions requires Django to be set up
+import django
+django.setup()
+
+from custom_handlers import get_script_logger
+import metsrw
+
+
+logger = get_script_logger("archivematica.mcp.client.convert_dataverse_struct")
+
+
+class ConvertDataverseError(Exception):
+    """Exception class for failures that might occur during the execution of
+    this script.
+    """
+
+
+# Mapping from originalFormatLabel in dataset.json to file extension. The
+# values here are associated with Dataverse Bundles, created when Tabular data
+# is ingested, see: http://guides.dataverse.org/en/latest/user/dataset-management.html?highlight=bundle
+# The formats supported for tabluar data ingests are here:
+# http://guides.dataverse.org/en/latest/user/tabulardataingest/supportedformats.html
+EXTENSION_MAPPING = {
+    "Comma Separated Values": ".csv",
+    "MS Excel (XLSX)": ".xlsx",
+    "R Data": ".RData",
+    "SPSS Portable": ".por",
+    "SPSS SAV": ".sav",
+    "Stata Binary": ".dta",
+    "Stata 13 Binary": ".dta",
+    "UNKNOWN": "UNKNOWN",
+}
+
+
+def get_ddi_title_author(dataset_md_latest):
+    """Retrieve the title and the author of the dataset for the DDI XML
+    snippet to be included in the METS file.
+    """
+    title_text = author_text = None
+    citation = dataset_md_latest.get("metadataBlocks", {}).get("citation")
+    fields = citation.get("fields", None)
+    if fields:
+        for field in fields:
+            if field.get("typeName") == "title":
+                title_text = field.get("value")
+            if field.get("typeName") == "author":
+                author_text = field.get("value")[0].get("authorName")\
+                    .get("value")
+        return title_text.strip(), author_text.strip()
+    raise ConvertDataverseError(
+        "Unable to retrieve MD fields from dataset.json")
+
+
+def create_ddi(job, json_metadata, dataset_md_latest):
+    """Create the DDI dmdSec from the JSON metadata."""
+    ddi_elems = {}
+
+    try:
+        ddi_elems["Title"], \
+            ddi_elems["Author"] = get_ddi_title_author(dataset_md_latest)
+    except TypeError as err:
+        logger.error(
+            "Unable to gather citation data from dataset.json: %s", err)
+        return None
+    except ConvertDataverseError as err:
+        logger.error(err)
+        return None
+
+    ddi_elems["PID Type"] = json_metadata.get("protocol", "")
+    ddi_elems["IDNO"] = json_metadata.get("persistentUrl", "")
+    ddi_elems["Version Date"] = dataset_md_latest.get("releaseTime", "")
+    ddi_elems["Version Type"] = dataset_md_latest.get("versionState", "")
+    ddi_elems["Version Number"] = "{}.{}".format(
+        dataset_md_latest.get("versionNumber", ""),
+        dataset_md_latest.get("versionMinorNumber", "")
+    )
+    ddi_elems["Restriction Text"] = dataset_md_latest.get("termsOfUse", "")
+    ddi_elems["Distributor Text"] = json_metadata.get("publisher", "")
+
+    draft = False
+    job.pyprint("Fields retrieved from Dataverse:")
+    for ddi_k, ddi_v in ddi_elems.iteritems():
+        if ddi_k == "Version Type" and ddi_v == "DRAFT":
+            draft = True
+        job.pyprint("{}: {}".format(ddi_k, ddi_v))
+
+    if draft:
+        job.pyprint(
+            "Dataset is in a DRAFT state and may not transfer correctly")
+        logger.error(
+            "Dataset is in a DRAFT state and may not transfer correctly")
+
+    # Create XML.
+    nsmap = {"ddi": "http://www.icpsr.umich.edu/DDI"}
+    ddins = "{" + nsmap["ddi"] + "}"
+    ddi_root = etree.Element(ddins + "codebook", nsmap=nsmap)
+    ddi_root.attrib["version"] = "2.5"
+
+    root_ns = "{http://www.w3.org/2001/XMLSchema-instance}schemaLocation"
+    dv_ns = (
+        "http://www.ddi:codebook:2_5 "
+        "http://www.ddialliance.org/Specification/DDI-Codebook/2.5/"
+        "XMLSchema/codebook.xsd"
+    )
+    ddi_root.attrib[root_ns] = dv_ns
+
+    stdydscr = etree.SubElement(ddi_root, ddins + "stdyDscr", nsmap=nsmap)
+    citation = etree.SubElement(stdydscr, ddins + "citation", nsmap=nsmap)
+
+    titlstmt = etree.SubElement(citation, ddins + "titlStmt", nsmap=nsmap)
+    etree.SubElement(titlstmt, ddins + "titl", nsmap=nsmap).text \
+        = ddi_elems["Title"]
+
+    etree.SubElement(
+        titlstmt, ddins + "IDNo", agency=ddi_elems["PID Type"]).text \
+        = ddi_elems["IDNO"]
+
+    rspstmt = etree.SubElement(citation, ddins + "rspStmt")
+    etree.SubElement(rspstmt, ddins + "AuthEnty").text \
+        = ddi_elems["Author"]
+
+    diststmt = etree.SubElement(citation, ddins + "distStmt")
+    etree.SubElement(diststmt, ddins + "distrbtr").text \
+        = ddi_elems["Distributor Text"]
+
+    verstmt = etree.SubElement(citation, ddins + "verStmt")
+    etree.SubElement(
+        verstmt, ddins + "version", date=ddi_elems["Version Date"],
+        type=ddi_elems["Version Type"]
+    ).text = ddi_elems["Version Number"]
+
+    dataaccs = etree.SubElement(stdydscr, ddins + "dataAccs")
+    usestmt = etree.SubElement(dataaccs, ddins + "useStmt")
+    etree.SubElement(usestmt, ddins + "restrctn").text \
+        = ddi_elems["Restriction Text"]
+
+    return ddi_root
+
+
+def display_checksum_for_user(job, fname, checksum_value, checksum_type="MD5"):
+    """Provide some feedback to the user that enables them to understand what
+    this script is doing in the Dataverse workflow.
+    """
+    job.pyprint(
+        "Checksum for '{}' retrieved from dataset.json: {} ({})"
+        .format(fname, checksum_value, checksum_type))
+
+
+def create_bundle(job, tabfile_json):
+    """Create the FSEntry objects for the various files in a Dataverse bundle
+    identified initially by a ```.tab``` file being requested from the
+    Dataverse API.
+
+    A bundle is a collection of multiple representations of a tabular data
+    file. Bundles are created by Dataverse to allow interaction with as wide a
+    range of tools as possible.
+
+    Documentation on Bundles can be found on the Dataverse pages:
+
+       * http://guides.dataverse.org/en/latest/user/dataset-management.html?highlight=bundle
+    """
+    # Base name is .tab with suffix stripped
+    tabfile_name = tabfile_json.get("label")
+    if tabfile_name is None:
+        return None
+
+    # Else, continue processing.
+    job.pyprint("Creating entries for tabfile bundle {}"
+                .format(tabfile_name))
+    base_name = tabfile_name[:-4]
+    bundle = metsrw.FSEntry(path=base_name, type="Directory")
+    # Find the original file and add it to the METS FS Entries.
+    tabfile_datafile = tabfile_json.get("dataFile")
+    fname = None
+    ext = EXTENSION_MAPPING.get(
+        tabfile_datafile.get("originalFormatLabel", ""), "UNKNOWN")
+    logger.info("Retrieved extension mapping value: %s", ext)
+    logger.info(
+        "Original file format listed as %s",
+        tabfile_datafile.get("originalFileFormat", "None"))
+    if ext == "UNKNOWN":
+        fname = tabfile_datafile.get("filename")
+        logger.info(
+            "Original Format Label is UNKNOWN, using filename: %s",
+            fname)
+    if fname is None:
+        fname = "{}{}".format(base_name, ext)
+    checksum_value = tabfile_datafile.get("md5")
+    if checksum_value is None:
+        return None
+    display_checksum_for_user(job, fname, checksum_value)
+    original_file = metsrw.FSEntry(
+        path="{}/{}".format(base_name, fname),
+        use="original",
+        file_uuid=str(uuid.uuid4()),
+        checksumtype="MD5",
+        checksum=checksum_value,
+    )
+    bundle.add_child(original_file)
+    if tabfile_datafile.get("originalFormatLabel") != "R Data":
+        # RData derivative
+        fsentry = metsrw.FSEntry(
+            path="{}/{}.RData".format(base_name, base_name),
+            use="derivative",
+            derived_from=original_file,
+            file_uuid=str(uuid.uuid4()),
+        )
+        bundle.add_child(fsentry)
+
+    # Bundle contents described earlier in this module are expected to remain
+    # consistent. Among the additional data files in support of a table-based
+    # dataset there are supposed to be three citation files. These are added
+    # as FSEntry objects below.
+
+    # Tabfile
+    fsentry = metsrw.FSEntry(
+        path="{}/{}".format(base_name, tabfile_datafile.get("filename")),
+        use="derivative",
+        derived_from=original_file,
+        file_uuid=str(uuid.uuid4()),
+    )
+    fsentry.add_dmdsec(
+        md="{}/{}-ddi.xml".format(base_name, base_name),
+        mdtype="DDI",
+        mode="mdref",
+        label="{}-ddi.xml".format(base_name),
+        loctype="OTHER",
+        otherloctype="SYSTEM",
+    )
+    bundle.add_child(fsentry)
+    # -ddi.xml
+    fsentry = metsrw.FSEntry(
+        path="{}/{}-ddi.xml".format(base_name, base_name),
+        use="metadata",
+        derived_from=original_file,
+        file_uuid=str(uuid.uuid4()),
+    )
+    bundle.add_child(fsentry)
+    # citation - endnote
+    fsentry = metsrw.FSEntry(
+        path="{}/{}citation-endnote.xml".format(base_name, base_name),
+        use="metadata",
+        derived_from=original_file,
+        file_uuid=str(uuid.uuid4()),
+    )
+    bundle.add_child(fsentry)
+    # citation - ris
+    fsentry = metsrw.FSEntry(
+        path="{}/{}citation-ris.ris".format(base_name, base_name),
+        use="metadata",
+        derived_from=original_file,
+        file_uuid=str(uuid.uuid4()),
+    )
+    bundle.add_child(fsentry)
+    # citation - bib
+    fsentry = metsrw.FSEntry(
+        path="{}/{}citation-bib.bib".format(base_name, base_name),
+        use="metadata",
+        derived_from=original_file,
+        file_uuid=str(uuid.uuid4()),
+    )
+    bundle.add_child(fsentry)
+    return bundle
+
+
+def retrieve_terms_of_access(dataset_md_latest):
+    """Return a tuple that can be used to direct users to information about a
+    dataset if it is restricted.
+    """
+    return dataset_md_latest.get("termsOfAccess")
+
+
+def test_if_zip_in_name(fname):
+    """Check if a file-path ends in a .zip extension. If so, return true. This
+    helps us to log some information about the characteristics of the package
+    as we go.
+    """
+    ext_ = os.path.splitext(fname)[1]
+    if ext_.lower() == '.zip':
+        return True
+    return False
+
+
+def add_ddi_xml(job, sip, json_metadata, dataset_md_latest):
+    """Create a DDI XML data block and add this to the METS."""
+    ddi_root = create_ddi(job, json_metadata, dataset_md_latest)
+    if ddi_root is None:
+        return None
+    sip.add_dmdsec(md=ddi_root, mdtype="DDI")
+    return sip
+
+
+def add_metadata_ref(sip, md_name, md_loc):
+    """Add a single mdref to the METS file."""
+    sip.add_dmdsec(
+        md=md_loc,
+        mdtype="OTHER",
+        mode="mdref",
+        label=md_name,
+        loctype="OTHER",
+        otherloctype="SYSTEM",
+    )
+    return sip
+
+
+def add_md_dir_to_structmap(sip):
+    """Add the metadata directory to the structmap."""
+    md_dir = metsrw.FSEntry(path="metadata", use=None, type="Directory")
+    sip.add_child(md_dir)
+    # Add dataset.json to the fileSec output.
+    fsentry = metsrw.FSEntry(
+        path="metadata/dataset.json", use="metadata",
+        file_uuid=str(uuid.uuid4())
+    )
+    # Add dataset.json to the metadata fileSec group.
+    md_dir.add_child(fsentry)
+    return sip
+
+
+def add_dataset_files_to_md(job, sip, dataset_md_latest, contact_information):
+    """Add file entries to the Dataverse METS document."""
+
+    # Add original files to the METS document.
+    files = dataset_md_latest.get('files')
+    if not files:
+        return None
+
+    # Signal to users the existence of zip files in this transfer.
+    zipped_file = False
+
+    # Signal to users that this transfer might consist of metadata only.
+    if len(files) is 0:
+        logger.info(
+            "Metadata only transfer? There are no file entries in this "
+            "transfer's metadata.")
+
+    for file_json in files:
+        is_restricted = file_json.get("restricted")
+        if is_restricted is True and contact_information:
+            logger.error(
+                "Restricted dataset files may not have transferred "
+                "correctly: %s", contact_information)
+
+        data_file = file_json.get("dataFile", {})
+        if data_file.get("filename", "").endswith(".tab"):
+            # A Tabular Data File from Dataverse will consist of an original
+            # tabular format submitted by the researcher plus multiple
+            # different representations. We need to map that here.
+            bundle = create_bundle(job, file_json)
+            if bundle:
+                sip.add_child(bundle)
+            else:
+                logger.error(
+                    "Create Dataverse transfer METS failed. "
+                    "Bundle returned: %s", bundle)
+                return None
+        else:
+            path_ = None
+            if data_file:
+                path_ = data_file.get("filename")
+            if path_:
+                if test_if_zip_in_name(path_):
+                    # provide some additional logging around the contents of the
+                    # dataset we're processing.
+                    if not zipped_file:
+                        zipped_file = True
+                        logger.info(
+                            "Non-bundle .zip file found in the dataset.")
+                checksum_value = data_file.get("md5")
+                if checksum_value is None:
+                    return None
+                display_checksum_for_user(job, path_, checksum_value)
+                f = metsrw.FSEntry(
+                    path=path_,
+                    use="original",
+                    file_uuid=str(uuid.uuid4()),
+                    checksumtype="MD5",
+                    checksum=checksum_value,
+                )
+                sip.add_child(f)
+            else:
+                logger.error(
+                    "Problem retrieving filename from metadata, returned "
+                    "datafile: %s, path: %s", data_file, path_)
+                return None
+    return sip
+
+
+def write_mets_to_file(sip, unit_path, output_md_path, output_md_name):
+    """Write METS to file."""
+    metadata_path = output_md_path
+    if metadata_path is None:
+        metadata_path = os.path.join(unit_path, "metadata")
+    if not os.path.exists(metadata_path):
+        os.makedirs(metadata_path)
+
+    metadata_name = output_md_name
+    if metadata_name is None:
+        metadata_name = "METS.xml"
+    mets_path = os.path.join(metadata_path, metadata_name)
+
+    # Write the data structure out to a file and ensure that the encoding is
+    # purposely set to UTF-8. This pattern is used in ```create_mets_v2.py```.
+    # Given the opportunity we should add an encoding feature to the metsrw
+    # package.
+    mets_f = metsrw.METSDocument()
+    mets_f.append_file(sip)
+    with open(mets_path, 'w') as xml_file:
+        xml_file.write(etree.tostring(
+            mets_f.serialize(), pretty_print=True, encoding="utf-8",
+            xml_declaration=True))
+
+
+def load_md_and_return_json(unit_path, dataset_md_name):
+    """Load and parse Dataverse metadata from disk. Return JSON to calling
+    function.
+    """
+    json_path = os.path.join(unit_path, "metadata", dataset_md_name)
+    logger.info("Metadata directory exists %s", os.path.exists(json_path))
+    try:
+        with open(json_path, "r") as md_file:
+            return json.load(md_file)
+    except IOError as err:
+        logger.error("Error opening dataset metadata: %s", err)
+        return None
+
+
+def convert_dataverse_to_mets(
+        job, unit_path, dataset_md_name="dataset.json", output_md_path=None,
+        output_md_name=None):
+    """Create a transfer METS file from a Dataverse's dataset.json file"""
+    logger.info(
+        "Convert Dataverse structure called with '%s' unit directory",
+        unit_path)
+
+    json_metadata = load_md_and_return_json(unit_path, dataset_md_name)
+    if json_metadata is None:
+        raise ConvertDataverseError(
+            "Unable to the load Dataverse metadata file")
+    dataset_md_latest = get_latest_version_metadata(json_metadata)
+    if dataset_md_latest is None:
+        raise ConvertDataverseError(
+            "Unable to find the dataset metadata section from dataset.json")
+
+    # If a dataset is restricted we may not have access to all the files. We
+    # may also want to flag this dataset to the users of this service. We
+    # can do this here and below. We do not yet know whether this microservice
+    # should fail because we don't know how all datasets behave when some
+    # restrictions are placed on them.
+    contact_information = retrieve_terms_of_access(dataset_md_latest)
+
+    # Create METS
+    try:
+        sip = metsrw.FSEntry(
+            path="None", label=get_ddi_title_author(dataset_md_latest)[0],
+            use=None, type="Directory"
+        )
+    except TypeError as err:
+        citation_msg = (
+            "Unable to gather citation data from dataset.json: %s", err)
+        logger.error(citation_msg)
+        raise ConvertDataverseError(citation_msg)
+
+    sip = add_ddi_xml(job, sip, json_metadata, dataset_md_latest)
+    if sip is None:
+        raise ConvertDataverseError("Error creating SIP from Dataverse DDI")
+
+    sip = add_metadata_ref(
+        sip, dataset_md_name, "metadata/{}".format(dataset_md_name))
+
+    sip = add_dataset_files_to_md(
+        job, sip, dataset_md_latest, contact_information)
+    if sip is None:
+        raise ConvertDataverseError("Error adding Dataset files to METS")
+
+    # On success of the following two functions, the module will return None
+    # to JobContext which expects non-zero as a failure code only.
+    sip = add_md_dir_to_structmap(sip)
+    write_mets_to_file(sip, unit_path, output_md_path, output_md_name)
+
+
+def get_latest_version_metadata(json_metadata):
+    """If the datatset has been downloaded from the Dataverse web ui then there
+    is a slightly different structure. While the structure is different, the
+    majority of fields should remain the same and work with Archivematica. Just
+    in case, we log the version here and inform the user of potential
+    compatibility issues.
+
+    Ref: https://github.com/IQSS/dataverse/issues/4715
+    """
+    dataset_version = json_metadata.get("datasetVersion")
+    if dataset_version:
+        logger.info(
+            "Dataset seems to have been downloaded from the Dataverse Web UI."
+            "Some features of this method may be incompatible with "
+            "Archivematica at present.")
+        return dataset_version
+    return json_metadata.get("latestVersion")
+
+
+def init_convert_dataverse(job):
+    """Extract the arguments provided to the script and call the primary
+    function concerned with converting the Dataverse metadata JSON.
+    """
+    try:
+        transfer_dir = job.args[1]
+        logger.info("Convert Dataverse Structure with dir: '%s'", transfer_dir)
+        return convert_dataverse_to_mets(job, unit_path=transfer_dir)
+    except IndexError:
+        convert_dv_msg = (
+            "Problem with the supplied arguments to the function len: {}"
+            .format(len(job.args)))
+        logger.error(convert_dv_msg)
+        raise ConvertDataverseError(convert_dv_msg)
+
+
+def call(jobs):
+    """Primary entry point for MCP Client script."""
+    for job in jobs:
+        with job.JobContext(logger=logger):
+            job.set_status(init_convert_dataverse(job))
--- a/src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# This file is part of Archivematica.
+#
+# Copyright 2010-2018 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica.    If not, see <http://www.gnu.org/licenses/>.
+
+"""Maps Dataverse specific elements into the AIP METS file generated on ingest.
+"""
+
+from __future__ import print_function
+import sys
+
+import archivematicaFunctions
+from custom_handlers import get_script_logger
+
+import metsrw
+
+
+# Create a module level logger.
+logger = get_script_logger("archivematica.mcp.client.createMETSDataverse")
+
+
+def create_dataverse_sip_dmdsec(job, sip_path):
+    """
+    Return SIP-level Dataverse dmdSecs for inclusion in the AIP METS.
+
+    :param str sip_path: ...
+    :return: List of dmdSec Elements
+    """
+    logger.info("Create dataverse sip dmdsec %s", sip_path)
+    # Retrieve METS.xml from the file system.
+    metadata_mets_paths = archivematicaFunctions.find_metadata_files(
+        sip_path, 'METS.xml', only_transfers=True)
+    if not metadata_mets_paths:
+        return []
+    ret = []
+    for metadata_path in metadata_mets_paths:
+        try:
+            mets = metsrw.METSDocument.fromfile(metadata_path)
+        except mets.MetsError:
+            job.pyprint('Could not parse external METS (Dataverse)',
+                        metadata_path, file=sys.stderr)
+            continue
+        # Retrieve all directory DMDSecs from the METS.xml.
+        for f in mets.all_files():
+            if f.type == "Directory" and f.dmdsecs:
+                # Serialize
+                ret += [d.serialize() for d in f.dmdsecs]
+    return ret
+
+
+def create_dataverse_tabfile_dmdsec(job, sip_path, tabfile):
+    """
+    Returns dmdSec associated with the given tabfile, if one exists.
+    """
+    logger.info("Create Dataverse tabfile dmdsec %s", sip_path)
+    # Retrieve METS.xml from the file system.
+    metadata_mets_paths = archivematicaFunctions.find_metadata_files(
+        sip_path, 'METS.xml', only_transfers=True)
+    if not metadata_mets_paths:
+        return []
+    ret = []
+    for metadata_path in metadata_mets_paths:
+        try:
+            mets = metsrw.METSDocument.fromfile(metadata_path)
+        except mets.MetsError:
+            job.pyprint('Could not parse external METS (Dataverse)',
+                        metadata_path, file=sys.stderr)
+            continue
+        # Retrieve all Item DMDSecs from the METS.xml.
+        for f in mets.all_files():
+            if f.type == "Item" and f.path.endswith(tabfile):
+                # Found the correct tabfile
+                return [d.serialize() for d in f.dmdsecs]
+    return ret
--- a/src/MCPClient/lib/clientScripts/create_mets_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_v2.py
@@ -52,6 +52,7 @@ from archivematicaCreateMETSTrim import getTrimFileAmdSec
 from archivematicaFunctions import escape
 from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import normalizeNonDcElementName
+from create_mets_dataverse_v2 import create_dataverse_sip_dmdsec, create_dataverse_tabfile_dmdsec
 from custom_handlers import get_script_logger
 import namespaces as ns
 from sharedVariablesAcrossModules import sharedVariablesAcrossModules
@@ -69,7 +70,9 @@ globalState = None
 class MetsGlobalState():
     def __init__(self, globalAmdSecCounter=0, globalTechMDCounter=0, globalDigiprovMDCounter=0):
         self.globalFileGrps = {}
-        self.globalFileGrpsUses = ["original", "submissionDocumentation", "preservation", "service", "access", "license", "text/ocr", "metadata"]
+        self.globalFileGrpsUses = [
+            "original", "submissionDocumentation", "preservation", "service",
+            "access", "license", "text/ocr", "metadata", "derivative"]
         for use in self.globalFileGrpsUses:
             grp = etree.Element(ns.metsBNS + "fileGrp")
             grp.set("USE", use)
@@ -869,7 +872,7 @@ def createFileSec(job,
                 if original_file is not None:
                     GROUPID = 'Group-' + original_file.uuid
 
-            elif use in ("preservation", "text/ocr"):
+            elif use in ("preservation", "text/ocr", "derivative"):
                 # Derived files should be in the original file's group
                 try:
                     d = Derivation.objects.get(derived_file_id=f.uuid)
@@ -915,6 +918,18 @@ def createFileSec(job,
                         admidApplyTo.set("DMDID", ids)
                     else:
                         dspaceMetsDMDID = ids
+
+            # Special Dataverse processing. If there's .tab file, check if
+            # there's a Dataverse METS with additional metadata.
+            if f.originallocation.endswith('.tab'):
+                dv_metadata = create_dataverse_tabfile_dmdsec(
+                    job, baseDirectoryPath,
+                    os.path.basename(f.originallocation))
+                globalState.dmdSecs.extend(dv_metadata)
+                ids = ' '.join([ds.get('ID') for ds in dv_metadata])
+                if ids != "":
+                    fileDiv.attrib['DMDID'] = \
+                        fileDiv.attrib.get('DMDID', '') + ids
 
             if GROUPID == "":
                 sharedVariablesAcrossModules.globalErrorCount += 1
@@ -1370,6 +1385,14 @@ def call(jobs):
                         # Attach the DC metadata to the top level SIP div
                         # See #9822 for details
                         structMapDiv.set('DMDID', ID)
+                    root.append(dmdSec)
+
+                # Look for Dataverse specific descriptive metatdata.
+                dv = create_dataverse_sip_dmdsec(job, baseDirectoryPath)
+                for dmdSec in dv:
+                    dmdid = dmdSec.attrib['ID']
+                    dmdids = structMapDivObjects.get("DMDID", '') + ' ' + dmdid
+                    structMapDivObjects.set("DMDID", dmdids)
                     root.append(dmdSec)
 
                 for dmdSec in globalState.dmdSecs:
--- a/src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
+++ b/src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""Parse the transfer METS file created as part of a Dataverse transfer and
+validate against the objects expected to be part of the SIP generated during
+transfer.
+"""
+
+import json
+import os
+import uuid
+
+# databaseFunctions requires Django to be set up
+import django
+django.setup()
+from django.utils import timezone
+
+# archivematicaCommon
+from archivematicaFunctions import get_file_checksum
+from custom_handlers import get_script_logger
+import databaseFunctions
+from main.models import Agent, File
+import metsrw
+
+logger = get_script_logger("archivematica.mcp.client.parse_dataverse_mets")
+transfer_objects_directory = "%transferDirectory%objects"
+
+
+class ParseDataverseError(Exception):
+    """Exception class for failures that might occur during the execution of
+    this script.
+    """
+
+
+def get_db_objects(job, mets, transfer_uuid):
+    """
+    Get DB objects for files in METS.
+
+    This also validates that files exist for each file asserted in the
+    structMap
+
+    :param mets: Parse METS file
+    :return: Dict where key is the FSEntry and value is the DB object
+    """
+    mapping = {}
+    for entry in mets.all_files():
+        if entry.type == 'Directory' or entry.label == 'dataset.json':
+            continue
+        # Retrieve the item name from the database. The proof-of-concept for
+        # Dataverse extracts objects from a bundle (zip) which can then be
+        # found on the root path of the transfer (the objects folder), but
+        # that means the initial lookup of the file might not resolve as
+        # anticipated as the directory structure of the bundle is reflected in
+        # the original path. We try the base name for the item below.
+        file_entry = None
+        try:
+            item_path = os.path.join(
+                transfer_objects_directory, entry.path)
+            logger.info(
+                "Looking for file type: '%s' using relative path: %s",
+                entry.type, item_path)
+            file_entry = File.objects.get(
+                originallocation=item_path, transfer_id=transfer_uuid)
+        except File.DoesNotExist:
+            logger.info(
+                "Could not find file type: '%s' in the database: %s",
+                entry.type, entry.path)
+        except File.MultipleObjectsReturned as err:
+            logger.info("Multiple entries for `%s` found. Exception: %s",
+                        entry.path, err)
+        try:
+            # Attempt to find the original location through just its filename
+            # as it may be sitting in the root item/objects directory of the
+            # transfer.
+            if file_entry is None:
+                base_name = os.path.basename(entry.path)
+                item_path = os.path.join(transfer_objects_directory, base_name)
+                logger.info("Looking for file type: '%s' using "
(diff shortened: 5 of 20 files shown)


src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,474 @@
...
+            logger.error(
+                "Restricted dataset files may not have transferred "
+                "correctly: %s", contact_information)
+        if file_json.get("dataFile").get("filename").endswith(".tab"):
+            # A Tabular Data File from Dataverse will consist of an original
+            # tabular format submitted by the researcher plus multiple
+            # different representations. We need to map that here.
+            bundle = create_bundle(job, file_json)
+            if bundle:
+                sip.add_child(bundle)
```
sevein, line 376: If you return here then the else in L18 would not be needed.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,474 @@
...
+            else:
+                logger.error(
+                    "Create Dataverse transfer METS failed. "
+                    "Bundle returned: %s", bundle)
+                return None
+        else:
+            path_ = None
```
sevein, line 383: Why `path_` instead of `path`?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,474 @@
...
+        with open(json_path, "r") as f:
+            return json.load(f)
+    except IOError as e:
+        logger.error("Error opening dataset metadata: %s", e)
+        return None
+
+
+def map_(job, unit_path, unit_uuid, dataset_md_name="dataset.json",
+         output_md_path=None, output_md_name=None):
+    """Docstring..."""
```
sevein: Docstring! :D

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class TestDataverseExample(object):
+
+    write_dir = "fixtures/dataverse/dataverse_sources/dataverse_mets/"
+    fixture_path = "fixtures/dataverse/dataverse_sources"
+
+    THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+    FIXTURES_DIR = os.path.join(THIS_DIR, fixture_path)
+    WRITE_DIR = os.path.join(THIS_DIR, write_dir)
```
sevein: Can you use `Pytest.tmpdir` instead, please? That way we don't have to write inside our working tree which is a problem sometimes, and you don't need the extra entry in `.gitignore`.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -58,7 +58,8 @@
     'zipped bag': 'baggitZippedDirectory',
     'dspace': 'Dspace',
     'maildir': 'maildir',
-    'TRIM': 'TRIM'
+    'TRIM': 'TRIM',
+    'dataverse': 'dataverseTransfer'
```
sevein: Nitpick: missing trailing comma in the last item of the dict.
Related: https://docs.python.org/dev/faq/design.html#why-does-python-allow-commas-at-the-end-of-lists-and-tuples.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
sevein: I see you're doing this in some other places. FYI this is something we only do to avoid clashes with builtin functions, e.g. `zip_` instead of `zip`.

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+        eventIdentifierUUID=event_id,
+        eventType='fixity check',
+        eventDateTime=date,
+        eventDetail=event_detail,
+        eventOutcome=event_outcome,
+        eventOutcomeDetailNote=detail_note,
+    )
+
+
+def parse_(job, unit_path, unit_uuid):
```
sevein: Why `parse_`? Shouldn't be `parse` or `_parse` (private)?
Even better, `_parse_thethingyouareparsing`?

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+        transfer_dir = job.args[1]
+        transfer_uuid = job.args[2]
+        logger.info("Parse Dataverse METS with dir args: '%s' and transfer "
+                    "uuid: %s", transfer_dir, transfer_uuid)
+        return parse_(job, transfer_dir, transfer_uuid)
+    except IndexError:
+        logger.error(
+            "Problem with the supplied arguments to the function "
+            "len: %s", len(job.args))
+        return 1
```
sevein: We have a lot of functions in client scripts doing this same thing, returning exit codes. I'm thinking it'd be much nicer if we separate concerns, e.g. raise an exception instead of returning 1 (`JobContext` will automatically capture and set the status to 1 for us) or just don't return or return `None` which will result in 0. Would it be better in your opinion? I see the same problem in `parse_`, it returns `1` or `0` but I think it'd be more idiomatic if it just raises an exception or just not return at all instead of `return 0`.

sevein (commented):
Sorry @ross-spencer, I realized too late I was adding comments instead of creating a proper GitHub review.

I've also added @jrwdunham to the list of reviewers, it'd be nice to get his feedback in this round. @jrwdunham this PR replaced #1218 - not sure why it had to be replaced.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: This makes sense when `from os import path` has been called.

src/MCPClient/lib/archivematicaClientModules
```diff
@@ -135,3 +135,7 @@ upload-qubit_v0.0 = upload_qubit
 upload-archivesspace_v0.0 = upload_archivesspace
 copyThumbnailsToDIPDirectory_v0.0 = copy_thumbnails_to_dip_directory
 removeDirectories_v0.0 = remove_directories
+
+# Dataverse Scripts
+convertDataverseStructure_v0.0 = convert_dataverse_structure
+parseDataverse = parse_dataverse_mets
```
jrwdunham: Why is there no `_v0.0` on this one?

src/dashboard/src/main/migrations/0060_create_dataverse_transfer_type.py
```diff
@@ -0,0 +1,582 @@
...
+    METS' microservice given a Dataverse transfer type.
+    """
+
+    # 355c22ae (Determine Parse Dataverse METS XML) task to associate with
+    # a MicroServiceChainLink.
+    create_task(
+        apps=apps, task_type_uuid="c42184a3-1a7f-4c4d-b380-15d8d97fdd11",
+        task_uuid="355c22ae-ba5b-408b-a9b6-a01372d158b5",
+        task_desc="Determine Parse Dataverse METS XML",
+        task_config="6f7a2ebd-bd88-44b7-b146-c552ac4e40cb"
```
jrwdunham: I think these UUIDs should be stored in descriptively named constants and not repeated all over the place. The problem with doing it this way is not only that it's redundant but also that it's clearer to the reader when the same UUID is being used in multiple places, which is important to understanding the semantics of these strange database entities.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    latest_version_md = get_latest_version_metadata(json_metadata)
+    if latest_version_md is None:
+        return 1
+
+    # If a dataset is restricted we may not have access to all the files. We
+    # may also want to flag this dataset to the users of this service. We
+    # can do this here and below. We do not yet know whether this microservice
+    # should fail because we don't know how all datasets behave when some
+    # restrictions are placed on them.
+    restricted_access, contact_information = test_access(latest_version_md)
```
jrwdunham: Does your comment above justify the fact that the `restricted_access` var is never used? Unless you are planning to modify this script to use that variable, I think `test_access` should be modified to just return `contact_information`.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    """
+    datasetVersion = json_metadata.get("datasetVersion")
+    if datasetVersion:
+        logger.info(
+            "Dataset seems to have been downloaded from the Dataverse Web UI."
+            "Some features of this method may be incompatible with "
+            "Archivematica at present.")
+        return datasetVersion
+    else:
+        return json_metadata.get("latestVersion")
```
jrwdunham: Pylint says "unnecessary else after return"

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+def map_dataverse(job):
+    """Extract the arguments provided to the script and call the primary
+    function concerned with converting the Dataverse metadata JSON.
+    """
+    try:
+        transfer_dir = job.args[1]
+        transfer_uuid = job.args[2]
+        logger.info("Convert Dataverse Structure with dir args: '%s' "
```
jrwdunham: Nitpick: shouldn't `dir args:` just be `dir:` here?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Also, this function name is confusing to me. Can we give it a more descriptive name? To me, `map_` sounds like something that converts a sequence by applying a supplied function to all its elements, as in the builtin `map` or in functional languages.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    logger.info(
+        "Convert Dataverse structure called with '%s' unit directory and "
+        "'%s' unit uuid", unit_path, unit_uuid)
+
+    json_metadata = load_md_and_return_json(unit_path, dataset_md_name)
+    if json_metadata is None:
+        return 1
+    latest_version_md = get_latest_version_metadata(json_metadata)
+    if latest_version_md is None:
+        return 1
```
jrwdunham: If the JSON has neither a `datasetVersion` or a `latestVersion` attr, then we're gonna fail here with no helpful log message. Maybe one should be added.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+            path="{}/{}citation-ris.ris".format(base_name, base_name),
+            use="metadata",
+            derived_from=original_file,
+            file_uuid=str(uuid.uuid4()),
+        )
+        bundle.add_child(f)
+        return bundle
+
+
+def test_access(latest_version_md):
```
jrwdunham: Why the "test" in `test_access`? It doesn't seem to be "testing" anything...

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Also, it seems odd to have these three final kwargs only being used in tests. Is there a reason that you are not using module-level constants for these? The oddness is compounded by the fact that there are literal `'dataset.json'` values sprinkled throughout the code (e.g., in log msgs) —so it is only in the strange testing scenarios where this JSON file can have any other value.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    # If a dataset is restricted we may not have access to all the files. We
+    # may also want to flag this dataset to the users of this service. We
+    # can do this here and below. We do not yet know whether this microservice
+    # should fail because we don't know how all datasets behave when some
+    # restrictions are placed on them.
+    restricted_access, contact_information = test_access(latest_version_md)
+
+    # Create METS
+    sip = metsrw.FSEntry(
+        path="None", label=get_ddi_title_author(latest_version_md)[0],
```
jrwdunham: Isn't it possible to raise a `TypeError` here just like in the other place where this function is called? Shouldn't this be guarded against here too?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+def get_ddi_title_author(dataset_md_latest):
+    """Retrieve the title and the author of the dataset for the DDI XML
+    snippet to be included in the METS file.
+    """
+    title_text = author_text = None
+    citation = dataset_md_latest.get("metadataBlocks", {}).get("citation")
+    if citation:
+        for field in citation.get("fields"):
```
jrwdunham: Maybe you should just use `citation["fields"]` here since iterating over `None` will fail and `KeyError: 'fields'` is a less opaque error than `TypeError: 'NoneType' object is not iterable`.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    ext_ = os.path.splitext(fname)[1]
+    if ext_.lower() == '.zip':
+        return True
+    return False
+
+
+def add_ddi_xml(job, sip, json_metadata):
+    """Create a DDI XML data block and add this to the METS."""
+    ddi_root = create_ddi(job, json_metadata)
+    if create_ddi is None:
```
jrwdunham: This is a typo I think and should be `if ddi_root is None:`. `create_ddi` is a function and will never be `None`.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+                title_text = field.get("value")
+            if field.get("typeName") == "author":
+                author_text = field.get("value")[0].get("authorName")\
+                    .get("value")
+        return title_text.strip(), author_text.strip()
+
+
+def create_ddi(job, json_metadata):
+    """Create the DDI dmdSec from the JSON metadata."""
+    dataset_md_latest = get_latest_version_metadata(json_metadata)
```
jrwdunham: Small thing, but since you've already called `get_latest_version_metadata` in `map_` and since `map_` is the only (indirect) caller of `create_ddi`, it might be better to just pass `latest_version_md` instead of needlessly calling `get_latest_version_metadata` twice.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+        file_uuid=str(uuid.uuid4())
+    )
+    # Add dataset.json to the metadata fileSec group.
+    md_dir.add_child(f)
+    return sip
+
+
+def add_dataset_files_to_md(job, sip, latest_version_md, contact_information):
+    # Add original files to the METS document.
+    files = {}
```
jrwdunham: This line is not necessary.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+            "Metadata only transfer? There are no file entries in this "
+            "transfer's metadata.")
+
+    for file_json in files:
+        is_restricted = file_json.get("restricted")
+        if is_restricted is True and contact_information:
+            logger.error(
+                "Restricted dataset files may not have transferred "
+                "correctly: %s", contact_information)
+        if file_json.get("dataFile").get("filename").endswith(".tab"):
```
jrwdunham: If you're not protecting against `AttributeError` here, then you might as well just use key access (`__getitem__`) I would think, i.e., `file_json["dataFile"]["filename"]` or better yet maybe `file_json.get("dataFile", {}).get("filename", "")`.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+        .format(checksum_value, checksum_type))
+
+
+def create_bundle(job, tabfile_json):
+    """
+    Creates and returns the metsrw entries for a tabfile's bundle
+    """
+    # Base name is .tab with suffix stripped
+    tabfile_name = tabfile_json.get("label")
+    if tabfile_name is not None:
```
jrwdunham: To reduce the indentation, maybe instead use:

    if tabfile_name is None:
        return None

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+
+    dataaccs = etree.SubElement(stdydscr, ddins + "dataAccs")
+    usestmt = etree.SubElement(dataaccs, ddins + "useStmt")
+    etree.SubElement(usestmt, ddins + "restrctn").text \
+        = ddi_elems["Restriction Text"]
+
+    return ddi_root
+
+
+def display_checksum_for_user(job, checksum_value, checksum_type="MD5"):
```
jrwdunham: The `checksum_type` param seems never to be used. Maybe it should be removed.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Also, it seems like `file_json["dataFile"]` and `file_json["dataFile"]["filename"]` are accessed quite a few times in this area of the code. Why not just get those values from the data structure once and then pass them around?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+        else:
+            path_ = None
+            datafile = file_json.get("dataFile")
+            if datafile:
+                path_ = datafile.get("filename")
+            if path_:
+                if test_if_zip_in_name(path_):
+                    # provide some additional logging around the contents of the
+                    # dataset we're processing.
+                    if not zipped_file:
```
jrwdunham, line 390: so the purpose of this `zipped_file` var is so that we only log the following message once even though there may be multiple "non-bundle .zip" files? Is that right? Just wondering.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+def display_checksum_for_user(job, checksum_value, checksum_type="MD5"):
+    """Provide some feedback to the user that enables them to understand what
+    this script is doing in the Dataverse workflow.
+    """
+    job.pyprint(
+        "Checksum retrieved from dataset.json: {} ({})"
+        .format(checksum_value, checksum_type))
+
+
+def create_bundle(job, tabfile_json):
```
jrwdunham, line 174: So if the `dataFile.filename` is a `.tab` file, that means that we know the Dataverse package contains a bunch of derivative files and metadata files with predictable names and we are documenting them in the "bundle" (i.e., `metsrw.FSEntry`) being constructed here? Is that right? Where does this knowledge of what is in a Dataverse JSON file come from? Can we provide a link to some relevant authoritative document in the docstring to this function and maybe also a bit more description in this docstring of what this func is doing?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+        metadata_name = "METS.xml"
+    mets_path = os.path.join(metadata_path, metadata_name)
+
+    # Write the data structure out to a file and ensure that the encoding is
+    # purposely set to UTF-8. This pattern is used in ```create_mets_v2.py```.
+    # Given the opportunity we should add an encoding feature to the metsrw
+    # package.
+    mets_f = metsrw.METSDocument()
+    mets_f.append_file(sip)
+    tree = etree.fromstring(mets_f.tostring())
```
jrwdunham: I think if you called `mets_f.serialize()` instead of `mets_f.tostring()` here then you could avoid unnecessarily round-tripping this XML to a string and back to an lxml object.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+    ddi_elems = {
+        "Title": "",
+        "Author": "",
+        "PID Type": "",
+        "IDNO": "",
+        "Version Date": "",
+        "Version Type": "",
+        "Version Number": "",
+        "Restriction Text": "",
+        "Distributor Text": ""}
```
jrwdunham: I think this is redundant since you are setting all of these values to at least the empty string in the lines below. That is, you could replace lines 61-70 with `ddi_elems = {}` and there would be no change in behaviour.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
...
+django.setup()
+
+from custom_handlers import get_script_logger
+import metsrw
+
+
+logger = get_script_logger("archivematica.mcp.client.convert_dataverse_struct")
+
+
+THIS_DIR = os.path.abspath(os.path.dirname(__file__))
```
jrwdunham: This `THIS_DIR` seems never to be used. Is it needed?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,499 @@
+#!/usr/bin/env python2
```
jrwdunham, line 1: Could you add a module-level docstring explaining what this client script does? My reading is that it takes a Dataverse-created JSON metadata file and uses it to produce a METS XML file that it writes to disk at a certain location in the SIP. I think this should be documented here. If doing so would be redundant, then maybe the module docstring here could point to another location (in the code or AM docs) where the purpose and high-level behaviour of this client script is documented.

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# This file is part of Archivematica.
+#
+# Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
```
jrwdunham: Maybe update `2013` to `2018` here.

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Create a module level logger.
+logger = get_script_logger("archivematica.mcp.client.createMETSDataverse")
+
+
+def create_dataverse_sip_dmdsec(job, sip_path):
+    """
+    Return SIP-level Dataverse dmdSecs for inclusion in the AIP METS.
+
+    :param str: Path to the SIP
```
jrwdunham: I think the convention here would be `:param str sip_path: ...`

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Related to the above, I think that in `create_mets_dataverse_v2.py` you are assuming that the metadata file names can only ever be `'METS.xml'`.

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
```diff
@@ -0,0 +1,96 @@
...
+            continue
+
+        # Get SIP-level DDI
+        for f in mets.all_files():
+            # TODO get better query API
+            if f.type == "Directory" and f.dmdsecs:
+                # Serialize
+                ret += [d.serialize() for d in f.dmdsecs]
+    # Return list
+    return ret
```
jrwdunham, line 63: Not sure all of the comments in the above function are helpful to the reader.

Also, are you sure that you are only getting SIP-level dmdsecs here? Is it not possible that other sub-directories in the parsed METS could have dmdsecs and that you would not want to collect and return those here but this logic would do that anyways? Hope the question makes sense.

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -916,6 +919,17 @@ def createFileSec(job,
                     else:
                         dspaceMetsDMDID = ids
 
+            # Special Dataverse processing. If there's .tab file, check if
+            # there's a Dataverse METS with additional metadata.
+            if f.originallocation.endswith('.tab'):
+                dv_metadata = create_dataverse_tabfile_dmdsec(
+                    job, baseDirectoryPath,
+                    os.path.basename(f.originallocation))
+                globalState.dmdSecs.extend(dv_metadata)
```
jrwdunham, line 928: It is possible for `create_dataverse_tabfile_dmdsec` to return `None`, in which case this line will trigger `TypeError: 'NoneType' object is not iterable`. Is that desirable? If not, would avoiding it by always having `create_dataverse_tabfile_dmdsec` return some kind of list be preferable?

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
```diff
@@ -0,0 +1,96 @@
...
+
+    :param str: Path to the SIP
+    :return: List of dmdSec Elements
+    """
+    logger.info("Create dataverse sip dmdsec %s", sip_path)
+
+    # Find incoming external METS
+    metadata_mets_paths = archivematicaFunctions.find_metadata_files(
+        sip_path, 'METS.xml', only_transfers=True)
+    # If doesn't exist, return None
```
jrwdunham: Doesn't actually return `None`; returns empty list.

src/dashboard/src/components/filesystem_ajax/views.py
jrwdunham: Same also applies to https://github.com/artefactual/archivematica/pull/1242/files#diff-cd921d566b4b0c45fc437feb8dea55f9R70 (if that dict is not replaced with an empty dict, as per my suggestion)

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Same question applies to the analogous part of parse_dataverse_mets.py

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+        # Dataverse extracts objects from a bundle (zip) which can then be
+        # found on the root path of the transfer (the objects folder), but
+        # that means the initial lookup of the file might not resolve as
+        # anticipated as the directory structure of the bundle is reflected in
+        # the original path. We try the base name for the item below.
+        f = None
+        try:
+            item_path = os.path.join(transfer_objects_directory, entry.path)
+            logger.info(
+                "Looking for for file type: '{}' using relative path: {}"
```
jrwdunham: Typo `for for`

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+                        entry.path, e)
+
+        try:
+            # Attempt to find the original location through just its filename
+            # as it may be sitting in the root item/objects directory of the
+            # transfer.
+            if f is None:
+                base_name = os.path.basename(entry.path)
+                item_path = os.path.join(transfer_objects_directory, base_name)
+                logger.info("Looking for for file type: '{}' using "
```
jrwdunham: Typo `for for`

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+        mapping[entry] = f
+
+    return mapping
+
+
+def update_file_use(job, mapping):
+    """
+    Update the file's USE for files in mets.
+
+    :param mets: Parse METS file
```
jrwdunham: There is no such param

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+        if created:
+            job.pyprint('Added agent', agent)
+        else:
+            job.pyprint('Agent already exists', agent)
+        agent_id = agent_id or a.id
+
+    return agent_id
+
+
+def create_derivatives(job, mapping, dataverse_agent_id):
```
jrwdunham: Naming nitpick: doesn't really create derivatives. Actually creates `Event` and `Derivation` rows in the databases. The actual derivatives (bitstreams) were created by Dataverse, right?

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+    mapping = get_db_objects(job, mets, unit_uuid)
+    if mapping is None:
+        logger.error(
+            "Exiting. Returning the database objects for our Dataverse "
+            "files has failed.")
+        return 1
+    update_file_use(job, mapping)
+    agent = add_external_agents(job, unit_path)
+    create_derivatives(job, mapping, agent)
+    validate_checksums(job, mapping, unit_path)
```
jrwdunham, line 252: `update_file_use`, `create_derivatives` and `validate_checksums` all iterate over all elements in `mapping`. I wonder if you could not do all of their work in one iteration?

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+            os.makedirs(cls.WRITE_DIR)
+        except OSError as e:
+            print(
+                "Creating fixture directory failed: {}".format(e),
+                file=sys.stderr)
+
+    def teardown_class(cls):
+        shutil.rmtree(cls.WRITE_DIR)
+
+    @pytest.mark.parametrize(
```
jrwdunham, line 101: There seems to be some redundancy in these `parametrize` decorators. Why not just access `FIXTURES_DIR` from `self` within the test methods? Same question for `WRITE_DIR`. Also, why repeatedly call `"METS.{}.dataverse.xml".format` when it's constructed the same way every time? Similarly, why have `DataverseMDIndex.url` if you are just repeating those values in the decorator of the third test method below? I think the second argument to `parametrize` here could simply be the following list: `[dv_1, dv_2, dv_3, dv_4, dv_5, dv_6, dv_7, dv_8]`. No?

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
jrwdunham: Another question: why even pass `unit_uuid` to this client script (and to `map_`) at all if it is only ever used to log a message? Does that use justify making it an argument?

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+        [(WRITE_DIR, "METS.{}.dataverse.xml".format(dv_1.fname), 13, 3, 10),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_2.fname), 43, 6, 37),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_3.fname), 5, 2, 3),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_4.fname), 10, 3, 7),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_5.fname), 11, 2, 9),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_6.fname), 18, 3, 15),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_7.fname), 10, 3, 7),
+         (WRITE_DIR, "METS.{}.dataverse.xml".format(dv_8.fname), 10, 3, 7),
+         ])
+    def test_generated_mets(self,
```
jrwdunham: It was confusing to me that `test_generated_mets` will only pass if `test_parse_dataverse` is run first. This interdependency between test methods seems not ideal. Why not just remove `test_generated_mets` and put its assertions into `test_parse_dataverse`?

Also, the `all_file_count`, `dir_count`, and `item_count` data points seem like things that would make sense as `DataverseMDIndex` attributes. I'd say the same about `ddi_title`, `pid_type`, and `ddi_count` below.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+        mets_path = os.path.join(mets_output_path, mets_name)
+        if not os.path.isfile(mets_path):
+            pytest.fail("Fixtures were not previously setup correctly.")
+
+        try:
+            mets = metsrw.METSDocument.fromfile(mets_path)
+        except mets.MetsError:
+            pytest.fail("Could not parse mets %s", mets_path)
+
+        mets_root = et.fromstring(mets.tostring())
```
jrwdunham: I think `mets.serialize()` would get you the root without the tostring/fromstring calls, no?

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # Setup namespaces to search for in the document.
+        ns = {'ddi': 'http://www.icpsr.umich.edu/DDI'}
+
+        # Test for a single instance of the ddi codebook.
+        codebook = mets_root.findall('.//ddi:codebook', ns)
+        assert len(codebook) == 1, ("Incorrect number of codebook entries "
+                                    "discovered: '%s' expected 1."
```
jrwdunham: I think you're mixing string formatting language conventions here (`%s` should be `{}`).

jrwdunham: The above comment applies to the other assertion failure strings below also.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,263 @@
...
+        # they are found as well.
+        for map_ in mets_root.findall("{http://www.loc.gov/METS/}structMap"):
+            for entry in map_:
+                if entry.get("Type") == "Directory" and \
+                        entry.get("DMDID") is not None:
+                    assert entry.get("LABEL") == ddi_title, \
+                        ("Title not found in METS struct maps where expected")
+
+        # Finally, the metadata should contain two mdrefs at this point. Look
+        # for them here.
```
jrwdunham: Nitpick: not really "finally" since there are more assertions below.

Not really "two" mdrefs in this assertion because `ddi_count` changes per permutation.

Also, line 252 could be shortened to `assert ddi_count >= len(refs) > 0`

src/dashboard/src/main/migrations/0060_create_dataverse_transfer_type.py
```diff
@@ -0,0 +1,582 @@
...
+    create_parse_dataverse_mets_microservice(apps)
+    create_dataverse_unit_variables_and_initial_tasks(apps)
+    create_convert_dataverse_link_pull(apps)
+    create_parse_dataverse_mets_link_pull(apps)
+
+
+def data_migration_down(apps, schema_editor):
+    """Reverse the changes made to the database in order to create a Dataverse
+    transfer type.
+    """
```
jrwdunham: Nice. Probably was overkill, but to test this I did the following.

First, I removed the new migration and rebuilt the database:

    $ mv src/dashboard/src/main/migrations/0060_create_dataverse_transfer_type.py ~/
    $ make bootstrap-dashboard-db

Then I generated the JSON links workflow as json-links-init.json:

    $ docker-compose ps -q archivematica-dashboard
    5e9dc12e620c8b85ab0b6f6f9371973a8322831487aa3c1c5351321fa475418e
    $ make manage-dashboard ARG=json_links
    /var/archivematica/sharedDirectory/tmp/json-links-kMsuqr.json
    $ docker cp 5e9dc12e620c8b85ab0b6f6f9371973a8322831487aa3c1c5351321fa475418e:/var/archivematica/sharedDirectory/tmp/json-links-kMsuqr.json ./
    $ mv json-links-kMsuqr.json json-links-init.json

Then I returned the Dataverse migration, rebuilt the database again, and
generated the new JSON links workflow as json-links-dataverse.json:

    $ mv ~/0060_create_dataverse_transfer_type.py src/dashboard/src/main/migrations/
    $ make bootstrap-dashboard-db
    $ make manage-dashboard ARG=json_links
    /var/archivematica/sharedDirectory/tmp/json-links-qdxpMI.json
    $ docker cp 5e9dc12e620c8b85ab0b6f6f9371973a8322831487aa3c1c5351321fa475418e:/var/archivematica/sharedDirectory/tmp/json-links-qdxpMI.json ./
    $ mv json-links-qdxpMI.json json-links-dataverse.json

Then I reverted the Dataverse migration and generated the new JSON links
workflow as json-links-reverted.json:

    $ make manage-dashboard ARG='migrate main 0059_siparrange_longblob'
    $ make manage-dashboard ARG=json_links
    /var/archivematica/sharedDirectory/tmp/json-links-rMIh8e.json
    $ docker cp 5e9dc12e620c8b85ab0b6f6f9371973a8322831487aa3c1c5351321fa475418e:/var/archivematica/sharedDirectory/tmp/json-links-rMIh8e.json ./
    $ mv json-links-rMIh8e.json json-links-reverted.json

Showed that the reverted and initial workflow JSON documents are identical:

    $ diff json-links-init.json json-links-reverted.json
    $

jrwdunham (changes requested):
Big addition! Overall I think it looks good. I have found a handful of real issues, and have some questions, suggestions and nitpicks.

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
```diff
@@ -0,0 +1,96 @@
...
+    Returns dmdSec associated with the given tabfile, if one exists.
+    """
+    logger.info("Create Dataverse tabfile dmdsec %s", sip_path)
+
+    # Find incoming external METS
+    metadata_mets_paths = archivematicaFunctions.find_metadata_files(
+        sip_path, 'METS.xml', only_transfers=True)
+    # If doesn't exist, return None
+    if not metadata_mets_paths:
+        return None
```
jrwdunham: I don't think you can return `None` here. If no METS.xml files are found, then the statement `globalState.dmdSecs.extend(dv_metadata)` in `create_mets_v2.py` will raise `TypeError: 'NoneType' object is not iterable`.

jrwdunham (changes requested):

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
jrwdunham: I guess maybe when no METS file can be found then you are already on an error path so you are ok with raising an exception maybe? I'm realizing that the way I was able to trigger this was by doing a Standard-type transfer on a Dataverse transfer source.

src/MCPClient/tests/test_create_dataverse_mets.py
ross-spencer: @sevein are you sure about that approach? From the docs, and some initial attempts, it doesn't seem like I can setup a tmpdir neatly in a single setup call for the module? https://docs.pytest.org/en/latest/tmpdir.html do you have any examples from our codebase you can point me at?

src/dashboard/src/components/filesystem_ajax/views.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: I'm not sure it is so will remove. Tests past without, will double check when I re-deploy.

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
ross-spencer: I think your suggestion is neater so will try that out in these scripts. It'll be good to find the consistency in the other scripts over time.

ross-spencer: When you take a look at the modified PR, I'm not sure if I've addressed all your concerns though, but happy to make changes accordingly.

src/MCPClient/tests/test_create_dataverse_mets.py
sevein: Here's an example: https://github.com/artefactual/archivematica/pull/1208/commits/c7e90d2269bb69f6927917ffa92e6571c7a9a316.

sevein:
```python
def test_json_csv_conversion_with_int_val(tmpdir):
    json_path = os.path.join(str(tmpdir), 'metadata.json')
    csv_path = os.path.join(str(tmpdir), 'metadata.csv')
    with open(json_path, 'w') as jsonfile:
        jsonfile.write(JSON_INT_VAL)
    job = Job("stub", "stub", ["", json_path])
    json_metadata_to_csv.call([job])
    with open(csv_path) as csvfile:
        csvdata = csvfile.read()
    assert csvdata == CSV_INT_VAL
```

src/MCPClient/lib/archivematicaClientModules
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: > Also, it seems odd to have these three final kwargs only being used in tests. Is there a reason that you are not using module-level constants for these? The oddness is compounded by the fact that there are literal 'dataset.json' values sprinkled throughout the code (e.g., in log msgs) —so it is only in the strange testing scenarios where this JSON file can have any other value.

I'm not sure if you'd agree, but prior to the Hudmol changes when these modules had a `main(...)` I was simply choosing to honor the idea that the script could be run standalone to convert any file purporting to be in the same format as `dataset.json`. I'm not sure how sensible this was! But it was useful in testing. Happy to reconsider and set testing variables at the module level if you think that would be best.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: Nice spot! @sevein gave me a good link about this the other day. Will keep it in mind more often.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: There's a little bit of gold-plating going on here, where Dataverse is using more checksum types now, since the original PoC, and we might want to expand this code to take advantage of that. Right now, it is used to display the hard coded "MD5" but it could easily display the different algorithm name if we change the calling code earlier on.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: That's right. We experienced a few issues with zips and while I'm not sure those issues are still the same now that we're handling bundles better, I wanted  to make sure we log enough about the characteristics of a download in the UI so that there is plenty for the user to debug for themselves if necessary. A breadcrumb if you will.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/create_mets_dataverse_v2.py
ross-spencer: :+1:

src/MCPClient/lib/clientScripts/create_mets_v2.py
ross-spencer: I didn't spend a lot of time on `create_dataverse_mets.py` which was created for the PoC and can see the inconsistency there now too. I think returning the empty list from there like in one of its other two functions is a good idea.

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,265 @@
...
+                .format(entry.type, item_path))
+            f = File.objects.get(originallocation=item_path,
+                                 transfer_id=transfer_uuid)
+        except File.DoesNotExist:
+            logger.info(
+                "Could not find file type: '{}' in the database: {}"
+                .format(entry.type, entry.path))
+        except File.MultipleObjectsReturned as e:
+            logger.info("Multiple entries for `%s` found. Exception: %s",
+                        entry.path, e)
```
sevein: Here you're using the optional keyword arguments to expand `%s`, this is good.
However, in L56, L70, L76 and maybe in other places you're using `str.format()` instead.

sevein (changes requested):
Small observation about logging messages.

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
ross-spencer: I've logged this here just in case it cannot be addressed in time.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,525 @@
...
+# -*- coding: utf-8 -*-
+
+"""Given a transfer type Dataverse access the metadata submission object
+```dataset.json``` to generate a transfer METS.xml file.
```
jrwdunham: Nitpick: I'm pretty sure Python docstrings assume reStructuredText not Markdown, so code samples should be enclosed in double (not triple) backticks. See https://devguide.python.org/documenting/#inline-markup

jrwdunham: I think there should be a comma after "access". I had trouble parsing this sentence.

jrwdunham: Thanks for the docstring. I think this is helpful. One question. You say "to generate a transfer METS.xml file." Am I correct in my understanding that this is an additional transfer METS file, which is distinct from the other transfer METS file that AM is already creating? If that is true, I think it should be mentioned. I was confused by that.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,525 @@
...
+        job, unit_path, dataset_md_name="dataset.json", output_md_path=None,
+        output_md_name=None):
+    """Create a transfer METS file from a Dataverse's dataset.json file"""
+    logger.info(
+        "Convert Dataverse structure called with '%s' unit directory",
+        unit_path)
+
+    json_metadata = load_md_and_return_json(unit_path, dataset_md_name)
+    if json_metadata is None:
+        return 1
```
jrwdunham: It seems odd to be raising exceptions everywhere else but returning 1 here.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,223 @@
...
+
+        mets_path = os.path.join(
+            self.WRITE_DIR, self.mets_file_name.format(fixture.fname))
+        if not os.path.isfile(mets_path):
+            pytest.fail("Fixtures were not previously setup correctly.")
+
+        try:
+            mets = metsrw.METSDocument.fromfile(mets_path)
+        except mets.MetsError:
+            pytest.fail("Could not parse mets %s", mets_path)
```
jrwdunham: I don't think `pytest.fail` has the signature that you are assuming here. Here's what I get when I call `help(pytest.fail)` in the REPL:

    fail(msg='', pytrace=True)
        explicitly fail an currently-executing test with the given Message.
        
        :arg pytrace: if false the msg represents the full failure information
                      and no python traceback will be reported.

jrwdunham: If I'm correct about the above, then there are a couple more instances of this error in this test file.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+from collections import namedtuple
+import os
+import sys
+
+import pytest
+import shutil
```
jrwdunham: Nitpick: shutil is in the stdlib so it should be with the above group of imports.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,223 @@
...
+
+    mets_file_name = "METS.{}.dataverse.xml"
+
+    def setup_class(cls):
+        try:
+            os.makedirs(cls.WRITE_DIR)
+        except OSError as e:
+            print(
+                "Creating fixture directory failed: {}".format(e),
+                file=sys.stderr)
```
jrwdunham: Nitpick: these are instance methods not class methods so `cls` should by convention be `self`. Just relaying a pylint complaint.

src/MCPClient/lib/clientScripts/convert_dataverse_structure.py
```diff
@@ -0,0 +1,525 @@
...
+            path="None", label=get_ddi_title_author(dataset_md_latest)[0],
+            use=None, type="Directory"
+        )
+    except TypeError as err:
+        citation_msg = (
+            "Unable to gather citation data from dataset.json: %s", err)
+        logger.error(citation_msg)
+        raise ConvertDataverseError(citation_msg)
+    except ConvertDataverseError as err:
+        raise
```
jrwdunham: What's the point of catching this exception and then just raising it again?

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,223 @@
...
+        METS as anticipated. A high-level test to quickly understand if
+        something has gone wrong while refactoring.
+        """
+        job = Job("stub", "stub", ["", ""])
+        ret = convert_dataverse_structure.convert_dataverse_to_mets(
+            job=job, unit_path=self.FIXTURES_DIR,
+            dataset_md_name=fixture.fname, output_md_path=self.WRITE_DIR,
+            output_md_name=self.mets_file_name.format(fixture.fname))
+        assert ret is None, ("Creation of Dataverse METS failed with return "
+                             "code {}".format(ret))
```
jrwdunham: If you remove the incongruous `return 1` in `convert_dataverse_to_mets` as I have suggested in a previous comment, then this assertion would no longer make sense since all failure paths would raise exceptions. I think this assert could either be removed or you could add specific fixtures to your tests that are expected to result in certain exceptions being raised and assert such using `pytest.raises`.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,223 @@
...
+                .format(title[0].text, fixture.ddi_title))
+
+        # Test that we have a single PID and that it matches what is expected.
+        pid = mets_root.findall('.//ddi:IDNo', ns)
+        assert len(pid) == 1
+        # Agency is synonymous with Type in our use of the PID Agency value.
+        # N.B. The agency providing the persistent identifier.
+        pid_agency = pid[0].get("agency")
+        assert pid_agency == \
+            fixture.pid_type, ("PID type: '{}'' is not what was expected: '{}'"
```
jrwdunham: Typo: extra single quote.

jrwdunham (commented):
@ross-spencer this is almost ready now. I have a few outstanding questions and suggestions, and a couple of nitpicky comments that you can ignore at your discretion. I think once you make these changes or respond to the comments we can wrap this up.

src/MCPClient/tests/test_create_dataverse_mets.py
ross-spencer: Great spot!

src/MCPClient/tests/test_create_dataverse_mets.py
ross-spencer: You're right. For expediency, I will remove this, and note this as a future enhancement to these tests. If that sounds okay?

jrwdunham: sounds good

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,286 @@
...
+django.setup()
+from django.utils import timezone
+
+# archivematicaCommon
+from archivematicaFunctions import get_file_checksum
+from custom_handlers import get_script_logger
+import databaseFunctions
+from main.models import Agent, File
+import metsrw
+from six.moves.urllib.parse import unquote_plus, urlparse, urlunparse
```
sevein: AM does not install `six` - it only works because it's a dependency of a dependency.
We use `django.utils.six` instead, see:

```
$ git grep -e six --and -e import
src/MCPClient/lib/archivematicaClient.py:from django.utils import six
src/MCPClient/lib/clientScripts/create_processed_structmap.py:from django.utils import six
src/MCPClient/lib/clientScripts/extract_maildir_attachments.py:from django.utils import six
src/MCPClient/lib/clientScripts/json_metadata_to_csv.py:from django.utils import six
src/MCPClient/lib/clientScripts/transcoder.py:from django.utils import six
src/MCPClient/lib/clientScripts/upload_archivists_toolkit.py:from django.utils.six.moves import input as raw_input
src/MCPServer/lib/archivematicaMCP.py:from django.utils import six
src/MCPServer/lib/jobChainLink.py:from django.utils import six
src/archivematicaCommon/lib/archivematicaFunctions.py:from django.utils import six
src/archivematicaCommon/lib/databaseFunctions.py:from django.utils import six, timezone
src/archivematicaCommon/lib/dicts.py:from django.utils import six
src/archivematicaCommon/lib/elasticSearchFunctions.py:from django.utils.six.moves import xrange
src/archivematicaCommon/lib/executeOrRunSubProcess.py:from django.utils import six
src/archivematicaCommon/lib/externals/extractMaildirAttachments.py:from django.utils import six
src/archivematicaCommon/lib/xml2obj.py:from django.utils import six
src/archivematicaCommon/tests/test_dicts.py:from django.utils import six
src/dashboard/src/components/administration/management/commands/graph_links.py:from django.utils import six
src/dashboard/src/components/administration/management/commands/json_links.py:from django.utils import six
src/dashboard/src/components/administration/views.py:from django.utils.six.moves import map
src/dashboard/src/components/filesystem_ajax/views.py:from django.utils import six
src/dashboard/src/main/management/commands/__init__.py:from django.utils.six.moves import input
src/dashboard/src/main/models.py:from django.utils import six
src/dashboard/tests/test_dashboardsetting_manager.py:from django.utils import six
```

Also, the import should be moved to the second group of imports.

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,286 @@
...
+
+    :param mets: Parse METS file
+    :return: Dict where key is the FSEntry and value is the DB object
+    """
+    mapping = {}
+    for entry in mets.all_files():
+        if entry.type == 'Directory' or entry.label == 'dataset.json':
+            continue
+        # TODO remove this once RData files are returned
+        if urldecode(entry.path).endswith('.RData'):
```
sevein: What would happen if urldecode raises `ValueError`? E.g.:

```
urlparse("http://foo[bar.com/hello")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/urllib/parse.py", line 368, in urlparse
    splitresult = urlsplit(url, scheme, allow_fragments)
  File "/usr/lib/python3.6/urllib/parse.py", line 418, in urlsplit
    raise ValueError("Invalid IPv6 URL")
ValueError: Invalid IPv6 URL
```

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,229 @@
...
+
+class TestDataverseExample(object):
+    """Dataverse test runner class."""
+
+    write_dir = "fixtures/dataverse/dataverse_sources/dataverse_mets/"
+    fixture_path = "fixtures/dataverse/dataverse_sources"
+
+    THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+    FIXTURES_DIR = os.path.join(THIS_DIR, fixture_path)
+    WRITE_DIR = os.path.join(THIS_DIR, write_dir)
```
sevein: The only thing that it's not obvious from the behaviour of `tmpdir` is that it does not return a string but a special kind of object, but you can convert it into a string to get its physical path.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,229 @@
...
+                "Directory count incorrect: '{}', expected: '{}'"
+                .format(dir_counter, fixture.dir_count))
+        assert item_counter == \
+            fixture.item_count, (
+                "Item count incorrect: '{}', expected: '{}'"
+                .format(item_counter, fixture.item_count))
+
+    @pytest.mark.parametrize(
+        "fixture", [dv_1, dv_2, dv_3, dv_4, dv_5, dv_6, dv_7, dv_8])
+    def test_mets_content(self, fixture):
```
sevein: Something I've noticed about this test is that it depends on the outcome of `test_generated_mets_`, e.g. the following command does not work but it should:

     docker-compose run --workdir /src/MCPClient --rm --user=root --entrypoint=py.test archivematica-mcp-client -x --reuse-db -vv /src/MCPClient/tests/test_create_dataverse_mets.py -k test_mets_content

sevein: Here's a patch to your test file where `tmpdir` replaces `WRITE_DIR` and there's no shared state between tests: https://gist.github.com/sevein/2e224b41ba8f275a7096644c5c664044. With this you don't need `src/MCPClient/tests/fixtures/dataverse/dataverse_sources/dataverse_mets` to be ignored in `.gitignore`.

.gitignore
```diff
@@ -24,3 +24,6 @@
 # frontend builds
 src/dashboard/src/static/
 src/dashboard/src/media/js/build
+
+# Dataverse: Tests output directory
+src/MCPClient/tests/fixtures/dataverse/dataverse_sources/dataverse_mets
```
sevein: I've suggested a [solution](https://gist.github.com/sevein/2e224b41ba8f275a7096644c5c664044) below in `test_create_dataverse_mets.py` that makes this new entry in `.gitignore` unnecessary.

sevein (changes requested):

src/MCPClient/lib/clientScripts/parse_dataverse_mets.py
```diff
@@ -0,0 +1,291 @@
...
+    """Exception class for failures that might occur during the execution of
+    this script.
+    """
+
+
+def urldecode(url):
+    """Decode percent encoding introduced per RFC3986 when the Dataverse METS
+    was first written in this workflow:
+    https://tools.ietf.org/html/rfc3986#section-2.1.
+    """
```
sevein: Did you mean to use that URL? I'm confused about the word you used: "workflow".

What would happen if urldecode raises `ValueError`? E.g.:

```python
>>> urlparse("http://foo[bar.com/hello")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/urllib/parse.py", line 368, in urlparse
    splitresult = urlsplit(url, scheme, allow_fragments)
  File "/usr/lib/python3.6/urllib/parse.py", line 418, in urlsplit
    raise ValueError("Invalid IPv6 URL")
ValueError: Invalid IPv6 URL
```

sevein (changes requested):

jrwdunham (approved):

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -0,0 +1,213 @@
...
+        job = Job("stub", "stub", ["", ""])
+        convert_dataverse_structure.convert_dataverse_to_mets(
+            job=job, unit_path=self.FIXTURES_DIR,
+            dataset_md_name=fname, output_md_path=str(tmpdir),
+            output_md_name=mets_file_name)
+        return str(tmpdir.join(mets_file_name))
+
+    @pytest.mark.parametrize(
+        "fixture", [dv_1, dv_2, dv_3, dv_4, dv_5, dv_6, dv_7, dv_8])
+    def test_generated_mets_(self, fixture, tmpdir):
```
sevein, line 103: Thanks for doing this!

sevein (approved):
> Resolves #1147
> Resolves #1163
> Resolves #1149
> Resolves #1121
> Resolves #1119
> Resolves #1148
> Resolves archivematica/Issues#59
> Resolves archivematica/Issues#152
> Resolves archivematica/Issues#187

Are you sure you want to add all the `Resolves #1234` references to the pull request / commit description? Isn't that going to close them automatically? Also in the pull request description the issues https://github.com/archivematica/Issues/issues/57, https://github.com/archivematica/Issues/issues/104 and https://github.com/archivematica/Issues/issues/131 are not included - the others do.

:+1:

PR #1248, opened 2018-08-30
Adjust async poll timeout

marktriggs:
If the storage service is busy running background tasks, a poll event might take a little more than 5 seconds to return, and that's not really a problem.  Wait a bit longer instead of aborting straight away.

Connects https://github.com/archivematica/Issues/issues/114.

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -200,11 +200,17 @@ def browse_location(uuid, path):
     return browse
 
 
-def wait_for_async(response, poll_seconds=2):
+def wait_for_async(response, poll_seconds=2, poll_timeout_seconds=600):
     """
-    Poll for results on an async endpoint.  `response` should have a HTTP 202
-    (Accepted) status code, and is expected to contain a Location header telling
-    us where to get our results from.
+    Poll for results on an async endpoint.
+
+    `response` should have a HTTP 202 (Accepted) status code, and is expected to
+    contain a Location header telling us where to get our results from.
+
+    `poll_seconds` controls how long we wait between poll requests.
+
+    `poll_timeout_seconds` controls how long we wait for a poll request to
+    complete before giving up and throwing an exception.
 
     Returns a tuple like (<result>, <error>), where <error> will be None if
     everything went well.
@@ -214,7 +220,7 @@ def wait_for_async(response, poll_seconds=2):
     poll_url = response.headers['Location']
 
     while True:
-        poll_response = _storage_api_session().get(poll_url)
+        poll_response = _storage_api_session(timeout=poll_timeout_seconds).get(poll_url)
         poll_response.raise_for_status()
 
         response_json = poll_response.json()


src/archivematicaCommon/lib/storageService.py
```diff
@@ -200,7 +200,7 @@ def browse_location(uuid, path):
     return browse
 
 
-def wait_for_async(response, poll_seconds=2):
+def wait_for_async(response, poll_seconds=2, poll_timeout=600):
     """
     Poll for results on an async endpoint.  `response` should have a HTTP 202
     (Accepted) status code, and is expected to contain a Location header telling
```
ross-spencer: Unless a reader looks back over the issue then understanding the poll_timeout is in seconds will take you back to the request lib docs to ensure that 600 is 10 mins, and not say, 0.6ms (I believe anyway, I couldn't see it anywhere so just followed that process). While it is unlikely to be the latter! That being said, can it be made explicit in the variable name, or the docstring what the defaults are in this case?

ross-spencer (changes requested):
Looks good @marktriggs. Just a small change requested while there is an opportunity. I'll just double check with @sromkey about the process of getting this into `qa/1.x` right now.

sromkey:
Thanks so much @ross-spencer and @marktriggs! I've added the 1.8 milestone.

src/archivematicaCommon/lib/storageService.py
marktriggs: Thanks @ross-spencer, good thought.  I've done both: renamed the variable to make the units obvious, and added a description of the two optional parameters to the docstring.

ross-spencer (approved):

sevein:
@marktriggs I'm not sure if I understand the conditions in which a "poll event might take a little more than 5 seconds to return". SS uses gevent so I would have thought that a Gunicorn worker busy doing I/O (including threads created in async_manager since the threading module is monkey patched) would still be able to handler other web requests promptly.

Where is the slowness happening? Is is the database query in the Async model?

marktriggs:
Hi @sevein,

It was definitely fiddly to track down :)  There's a bit more background in the comment on the issue (https://github.com/archivematica/Issues/issues/114#issuecomment-417168663), but ultimately I had to strace the storage service to see what was going on.

What I observed was that the async poll request always ran quickly (it's a pretty simple query), but when the timeouts occurred it was because the request itself had been queued at the gunicorn level for about 5 seconds prior to actually being run.  I.e. the client opened the connection at T=0, but the connection didn't get accepted until T=5.  Once the request was actually being run it was fast as always, but by then the client had hit the timeout and closed the socket from its side.

During that 5 seconds where the request wasn't being handled, the background `copy_file` thread was busy doing XML generation, so it was CPU-bound rather than waiting on I/O.  I guessed that it was holding onto the GIL during that time and locking out other threads.

So my thinking was that Python would periodically yield control back to the request thread, and we just need a timeout long enough to accommodate that.  I haven't dug into the native code that drives the XML generation... I guess it has to yield the GIL periodically, and maybe it just holds onto it a bit longer than regular Python code would.

sevein:
@marktriggs that makes sense. I'll merge soon, thank you!

Here's maybe a reason why we should be using processes instead of threads in [`async_manager`](https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/storage_service/locations/models/async_manager.py), would you agree? The overhead of starting new processes is probably fine since the duration of the jobs that we process there are usually in the order of seconds. We could also look into a separate process but that's going to be trickier.

marktriggs:
Yep, I think processes would fix this issue.  I opted for threads in the first instance because I wasn't sure whether database connections would survive across a fork.  Looking at https://code.djangoproject.com/ticket/19117, it sounded like some special care might need to be taken to avoid messing up those connections.

I guess ideally you would want to start the multiprocessing pool after all the application imports were run but before django had connected to the DB.  Not sure how tricky that ends up being :)

PR #1250, opened 2018-08-30
Add tests to MCPServer

sevein:
This pull request makes changes to our code so it's possible to add tests to MCPServer. It also updates the configuration of Travis CI and tox so the tests are executed. I've added a simple test to prove that it's working. For our Compose users, `make test-mcp-server` can now be used. Also, for faster runs:

    docker-compose run \
        --no-deps --workdir /src/MCPServer --rm \
        --user=root --entrypoint=py.test \
            archivematica-mcp-server --reuse-db -vv

If `pytest-cov` is installed (it is in the MCPServer container as a side effect of installing the requirements of the Dashboard) you can also add flags like `--cov=/src/MCPServer/lib` or even `--cov-report html:cov_html`.

This is connected to https://github.com/archivematica/Issues/issues/88.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ before_script:
 
 script: tox
 
+install: true
+
 notifications:
   email: false
 
@@ -17,11 +19,23 @@ matrix:
   fast_finish: true
   include:
 
+    # Dashboard
+
     - python: "2.7"
       env: TOXENV=py27
 
     - python: "3.6"
       env: TOXENV=py36
+
+    # MCPServer
+
+    - python: "2.7"
+      env: TOXENV=mcpserver
+
+    - python: "3.6"
+      env: TOXENV=mcpserver
+
+    # Linters
 
     - python: "2.7"
       env: TOXENV=py27-flake8
@@ -33,3 +47,6 @@ matrix:
 
     - python: "3.6"
       env: TOXENV=py36
+
+    - python: "3.6"
+      env: TOXENV=mcpserver
--- a/src/MCPServer/lib/jobChainLink.py
+++ b/src/MCPServer/lib/jobChainLink.py
@@ -35,19 +35,23 @@ from linkTaskManagerUnitVariableLinkPull import linkTaskManagerUnitVariableLinkP
 
 from databaseFunctions import auto_close_db, logJobCreatedSQL, getUTCDate
 
-from main.models import Job, MicroServiceChainLink, MicroServiceChainLinkExitCode, TaskType
+from main.models import Job, MicroServiceChainLink, MicroServiceChainLinkExitCode
+
 
 LOGGER = logging.getLogger('archivematica.mcp.server')
 
-# Constants
-constOneTask = TaskType.objects.get(description="one instance").pk
-constTaskForEachFile = TaskType.objects.get(description="for each file").pk
-constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").pk
-constGetReplacementDic = TaskType.objects.get(description="get replacement dic from user choice").pk
-constlinkTaskManagerGetMicroserviceGeneratedListInStdOut = TaskType.objects.get(description="Get microservice generated list in stdOut").pk
-constlinkTaskManagerGetUserChoiceFromMicroserviceGeneratedList = TaskType.objects.get(description="Get user choice from microservice generated list").pk
-constlinkTaskManagerSetUnitVariable = TaskType.objects.get(description="linkTaskManagerSetUnitVariable").pk
-constlinkTaskManagerUnitVariableLinkPull = TaskType.objects.get(description="linkTaskManagerUnitVariableLinkPull").pk
+# This is basically a dump of TaskTypes. It's not worth loading this from the
+# database since it's going to be gone very soon anyways.
+TASK_TYPES = {
+    "36b2e239-4a57-4aa5-8ebc-7a29139baca6": linkTaskManagerDirectories,
+    "a6b1c323-7d36-428e-846a-e7e819423577": linkTaskManagerFiles,
+    "61fb3874-8ef6-49d3-8a2d-3cb66e86a30c": linkTaskManagerChoice,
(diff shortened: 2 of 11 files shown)


src/MCPServer/lib/settings/common.py
```diff
@@ -125,6 +125,24 @@
...
+DATABASES = {
+    "default": {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": ":memory:",
+        "USER": "",
+        "PASSWORD": "",
+        "HOST": "",
+        "PORT": "",
+    },
+}
```
sevein: TODO: test against MySQL instead, available in Travis too.

src/MCPServer/lib/settings/common.py
```diff
@@ -125,6 +125,24 @@
...
+INSTALLED_APPS = (
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
+    'main',
+    'components.administration',
+    'fpr',
+)
```
sevein: TODO: comment the rationale behind this.

src/MCPServer/lib/jobChainLink.py
```diff
@@ -35,19 +35,23 @@
...
 
-# Constants
-constOneTask = TaskType.objects.get(description="one instance").pk
-constTaskForEachFile = TaskType.objects.get(description="for each file").pk
-constSelectPathTask = TaskType.objects.get(description="get user choice to proceed with").pk
-constGetReplacementDic = TaskType.objects.get(description="get replacement dic from user choice").pk
-constlinkTaskManagerGetMicroserviceGeneratedListInStdOut = TaskType.objects.get(description="Get microservice generated list in stdOut").pk
-constlinkTaskManagerGetUserChoiceFromMicroserviceGeneratedList = TaskType.objects.get(description="Get user choice from microservice generated list").pk
-constlinkTaskManagerSetUnitVariable = TaskType.objects.get(description="linkTaskManagerSetUnitVariable").pk
-constlinkTaskManagerUnitVariableLinkPull = TaskType.objects.get(description="linkTaskManagerUnitVariableLinkPull").pk
```
sevein, line 50: I had the following scenario where this wasn't working for me:
1. Wrote a `test_package.py` with a simple unit test that fails because:
2. `test_package.py` imports `package.py` which would import this module as a side effect.
3. `jobChainLink.py` (this snippet above) hits the database which pytest has not loaded yet - and fails!

sevein:
CC @jrwdunham @ross-spencer: I guess this won't be merged anytime soon but I need it anyways for other work I'm doing in the MCPServer where I want to write some tests. So if you want to take a look go ahead and comment your thoughts.

src/MCPServer/lib/jobChainLink.py
```diff
@@ -35,19 +35,23 @@
...
+TASK_TYPES = {
+    "36b2e239-4a57-4aa5-8ebc-7a29139baca6": linkTaskManagerDirectories,
+    "a6b1c323-7d36-428e-846a-e7e819423577": linkTaskManagerFiles,
+    "61fb3874-8ef6-49d3-8a2d-3cb66e86a30c": linkTaskManagerChoice,
+    "9c84b047-9a6d-463f-9836-eafa49743b84": linkTaskManagerReplacementDicFromChoice,
+    "a19bfd9f-9989-4648-9351-013a10b382ed": linkTaskManagerGetMicroserviceGeneratedListInStdOut,
+    "01b748fe-2e9d-44e4-ae5d-113f74c9a0ba": linkTaskManagerGetUserChoiceFromMicroserviceGeneratedList,
+    "6f0b612c-867f-4dfd-8e43-5b35b7f882d7": linkTaskManagerSetUnitVariable,
+    "c42184a3-1a7f-4c4d-b380-15d8d97fdd11": linkTaskManagerUnitVariableLinkPull,
+}
```
sevein, line 54: This solution may not look ideal but I'm suggesting a compromise only because the work in #1251 changes this so the managers are looked up in the workflow data and imported dynamically, e.g.:

```python
class jobChainLink(object):
    # ...
    def _run_task_manager(self):
        """Execute the task manager corresponding to this link."""
        try:
            manager = self.link.manager
        except Exception as err:
            LOGGER.exception('Unsupported task type: %s', err)
            return
        manager(self, self.unit)
```

.travis.yml
```diff
@@ -10,19 +10,33 @@ before_script:
 
 script: tox
 
+install: true
```
sevein, line 13: FYI: this overrides the default `pip install -r requirements.txt` that we don't want Travis to execute because we're using tox anyways.

src/MCPServer/lib/linkTaskManagerFiles.py
```diff
@@ -49,6 +49,9 @@ class linkTaskManagerFiles(LinkTaskManager):
     def __init__(self, jobChainLink, pk, unit):
         super(linkTaskManagerFiles, self).__init__(jobChainLink, pk, unit)
 
+        if jobChainLink.reloadFileList:
+            unit.reloadFileList()
```
sevein, line 53: This used to happen in `createTasks`, just moving it here because it seems a better place.

src/MCPServer/lib/settings/common.py
```diff
@@ -125,6 +125,16 @@
...
 
+# These are all the apps that we need so we can use the models in the
+# Dashboard.
+INSTALLED_APPS = (
+    'django.contrib.auth',
+    'django.contrib.contenttypes',
+    'main',
+    'components.administration',
+    'fpr',
+)
```
sevein, line 136: We're just aiming to import `main.models`, but then you need to make sure the app dependencies are available too or it won't work. Ideally (long term) we wouldn't be sharing models/database at all but that's not what we're trying to achieve here.

jraddaoui (approved):
Hi @sevein, this looks good to me. However I wonder if the `executor` and other changes not directly related to the tests deserve to be in a different PR, or different commits when it gets merged.

sevein:
Good idea, I will not squash. Thank you!

PR #1252, opened 2018-09-05
osdeps: use 1.7.x repository for bionic

scollazo:
The repository for 1.8.x shouldn't be use in 1.7.2.

This connects to archivematica/Issues#119.

### DIFF
--- a/src/MCPClient/osdeps/Ubuntu-18.json
+++ b/src/MCPClient/osdeps/Ubuntu-18.json
@@ -11,7 +11,7 @@
   ],
   "osdeps_repos": [
     "deb [arch=amd64] http://packages.archivematica.org/1.7.x/ubuntu-externals xenial main",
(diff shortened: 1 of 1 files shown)


src/MCPClient/osdeps/Ubuntu-18.json
```diff
@@ -11,7 +11,7 @@
   ],
   "osdeps_repos": [
     "deb [arch=amd64] http://packages.archivematica.org/1.7.x/ubuntu-externals xenial main",
```
sevein, line 13: Why is this needed?

scollazo: Some dependencies have not been rebuilt for 18.04, as the packages from 16.04 work without problems. I would prefer to leave 1.7 as is, and improve the repository management for 1.8

sevein (approved):

Issue #1057, opened 2018-04-17
Problem: #1049 and #1056 not ported to qa/1.x


sevein:
#815 only made it to `stable/1.7.x` - PRs #1049 and #1056 need to be ported, not done yet.

PR #1257, opened 2018-09-12
storageService.create_file: raise HTTP errors, refactor

sevein:
This is a port of some work done in AM17 in order to have better error handling
in the Storage Service client after 4xx or 5xx status codes are returned. The
original pull requests are: https://github.com/artefactual/archivematica/pull/1049 and https://github.com/artefactual/archivematica/pull/1056. Original issue: https://github.com/artefactual/archivematica/issues/815.

Connects to https://github.com/artefactual/archivematica/issues/1057.

### DIFF
--- a/src/MCPClient/lib/clientScripts/move_to_backlog.py
+++ b/src/MCPClient/lib/clientScripts/move_to_backlog.py
@@ -3,14 +3,41 @@
 import logging
 import os
 import shutil
-import sys
 
 # storageService requires Django to be set up
 import django
 django.setup()
 from django.db import transaction
+
 # archivematicaCommon
 import storageService as storage_service
+
+
+class StorageServiceCreateFileError(Exception):
+    pass
+
+
+def _create_file(transfer_uuid, current_location, relative_transfer_path,
+                 backlog, backlog_path, size):
+    try:
+        new_file = storage_service.create_file(
+            uuid=transfer_uuid,
+            origin_location=current_location['resource_uri'],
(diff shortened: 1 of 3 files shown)


sevein:
It hasn't been tested yet.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -13,6 +12,30 @@
...
+                 backlog, backlog_path, size):
+    new_file = storage_service.create_file(
+        uuid=transfer_uuid,
+        origin_location=current_location['resource_uri'],
+        origin_path=relative_transfer_path,
+        current_location=backlog['resource_uri'],
+        current_path=backlog_path,
+        package_type='transfer',  # TODO use constant from storage service
+        size=size,
+    )
```
jrwdunham: `create_file` looks like it can raise `requests.exceptions.RequestException` and `storage_service.ResourceNotFound`. Maybe those should be caught here and re-raised as `StorageServiceCreateFileError`. Or maybe your `except Exception as err:` below should be `except (StorageServiceCreateFileError, storage_service.ResourceNotFound, requests.exceptions.RequestException) as err:`. Either way, the idea would be to be more explicit about which exceptions you expect to be catching.

jrwdunham (approved):
I think this is good. One comment about handling exceptions.

sevein:
Thanks Joel. I've also done some extra testing and both backlog and store AIP are working for me, will merge!

PR #1259, opened 2018-09-13
MCPServer: automate upload DIP config retrieval

sevein:
This pull request updates the `linkTaskManagerReplacementDicFromChoice` task manager to ensure that the user is not prompted when the DIP is ready to be uploaded and the configuration needs to be retrieved from ``DashboardSetting``. We assume that there is only one configuration, MCPServer retrieves it, updates `passVar` and moves on without interrupting the user unnecessarily.

It also includes a small migration to fix the status assigned to the job responsible for retrieving the configuration for AtoM so it's not marked as "Unknown" but "Completed successfully".

This is connected to https://github.com/archivematica/Issues/issues/55.

### DIFF
--- a/src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
+++ b/src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
@@ -46,27 +46,33 @@ class linkTaskManagerReplacementDicFromChoice(LinkTaskManager):
         for i, dic in enumerate(dicts):
             self.choices.append((i, dic.description, dic.replacementdic))
 
-        # This is an attempt to find a choice in DashboardSettings for the
-        # chain link that the current MicroServiceChoiceReplacementDic is
-        # taking us. We're looking up the setting dict by the module name
-        # that the StandardTaskConfig wants to execute, e.g. upload-qubit_v0.0.
+        # There are MicroServiceChoiceReplacementDic links with no
+        # replacements (``self.choices`` has zero elements at this point). This
+        # is true for the following chain links:
         #
-        # Not implemented yet, but this could enable us to store more than a
-        # configuration set a specific chain link.
+        #   - ``Choose Config for Archivists Toolkit DIP Upload``
+        #   - ``Choose config for AtoM DIP upload``
+        #   - ``Choose Config for ArchivesSpace DIP Upload``
         #
-        # DashboardSettings does not belong to MCPServer. We currently have
-        # direct access to the database but that may not be always possible.
-        try:
-            mscl = MicroServiceChainLink.objects.get(id=jobChainLink.pk)
-            task_id = mscl.defaultnextchainlink.currenttask.tasktypepkreference
-            stc = StandardTaskConfig.objects.get(id=task_id)
-        except (MicroServiceChainLink.DoesNotExist, StandardTaskConfig.DoesNotExist, AttributeError):
-            pass
-        else:
-            args = DashboardSetting.objects.get_dict(stc.execute)
-            if args:
-                args = {'%{}%'.format(key): value for key, value in args.items()}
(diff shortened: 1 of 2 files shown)


sevein:
@jraddaoui, do you want to take a look at this one?

src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
```diff
@@ -46,27 +46,33 @@ def __init__(self, jobChainLink, pk, unit):
         for i, dic in enumerate(dicts):
             self.choices.append((i, dic.description, dic.replacementdic))
```
sevein, line 47: FYI: self.choices is the list that fills up the drop-down in the user interface when the workflow needs intervention from the user.

src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
```diff
@@ -46,27 +46,33 @@ def __init__(self, jobChainLink, pk, unit):
...
+        # user decide between multiple configurations, e.g. more than one
+        # AtoM instance is available and the user wants to decide which one is
+        # going to be used.
+        rdict = self._get_dashboard_setting_choice()
+        if rdict and not self.choices:
+            LOGGER.debug('Found Dashboard settings for this task, proceed.')
+            self.update_passvar_replacement_dict(rdict)
+            self.jobChainLink.linkProcessingComplete(
+                0, passVar=self.jobChainLink.passVar)
+            return
```
sevein, line 75: This is the new magic. Basically, if we know there's just a single choice we just ask the engine to go with it instead of prompting the user - that's why we're calling `linkProcessingComplete` and return immediately.

src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
```diff
@@ -46,27 +46,33 @@ def __init__(self, jobChainLink, pk, unit):
...
-            mscl = MicroServiceChainLink.objects.get(id=jobChainLink.pk)
-            task_id = mscl.defaultnextchainlink.currenttask.tasktypepkreference
-            stc = StandardTaskConfig.objects.get(id=task_id)
-        except (MicroServiceChainLink.DoesNotExist, StandardTaskConfig.DoesNotExist, AttributeError):
-            pass
-        else:
-            args = DashboardSetting.objects.get_dict(stc.execute)
-            if args:
-                args = {'%{}%'.format(key): value for key, value in args.items()}
-                self.choices.append((len(self.choices), stc.execute, str(args)))
```
sevein, line 69: I've moved this look up down to `_get_dashboard_setting_choice`, much cleaner that way!

src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
jraddaoui: So, the dashboard look up can't return multiple choices at the moment, it's just a dict with a single choice in every case, right?

jraddaoui (approved):
Looks good to me!

Just a question.

src/MCPServer/lib/linkTaskManagerReplacementDicFromChoice.py
sevein: Exactly, just a single dict that becomes part of `passVar` with all the values that the next script needs to upload the DIP to the access system, e.g. `%host%`, `%user%`, etc...

sevein:
Ops looks like I did things in the wrong order. For the record:

- **Migrations: fix exitmessage in MSCL ExitCode** - 429e8ab05995cbce38e842218c10cdffd70aeebf
- **MCPServer: automate upload DIP config retrieval** - 8d573098c3efd8208bacc6217fa4eacb92546028

PR #1260, opened 2018-09-14
Handle multiple occurrences of bag-info metadata elements

helrond:
This adds a loop to check if the value returned from a bag-info key is a list. 

It's not a very DRY approach, but breaking that loop out into a separate function caused the Update METS.xml document microservice (in Transfer) to fail, and I couldn't get any meaningful debug messages out of Archivematica to help me figure out where things went wrong. Any advice would be more than welcome!

Connects to archivematica/Issues#173.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETS2.py
@@ -1102,21 +1102,16 @@ def create_object_metadata(struct_map, baseDirectoryPath):
         xmldata = etree.SubElement(mdwrap, ns.metsBNS + 'xmlData')
         bag_metadata = etree.SubElement(xmldata, "transfer_metadata")
         for key, value in bagdata.items():
-            if isinstance(value, (list,)):
-                for v in value:
-                    try:
-                        bag_tag = etree.SubElement(bag_metadata, key)
-                    except ValueError:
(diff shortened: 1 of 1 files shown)


sevein (commented):
Hi @helrond, the script you're editing is a big mess so thanks for spending your time figuring things out.

I'm wondering if it'd be more elegant if we convert `value` into a list when it's not already one, that way the loop would always be relevant?

```python
for key, value in bagdata.items():
    if not isinstance(value, list):
        value = [value]
    for v in value:
        try:
            bag_tag = etree.SubElement(bag_metadata, key)
        except ValueError:
            print("Skipping bag key {}; not a"
                  " valid XML tag name".format(key), file=sys.stderr)
            continue
        bag_tag.text = v
```

helrond:
Thanks @sevein - I knew there had to be a better way of doing this but I was having trouble figuring it out! I've made the change as you've suggested above.

sevein:
Thanks, @helrond. I'd say this is ready to merge but we need to hold on until a milestone is added to the issue.

sromkey:
Thanks for this fix @helrond ! We're in bug-fixing mode for 1.8 so I added that as the milestone.

helrond:
Great, thanks @sevein and @sromkey. I should have a Contributor Agreement on file with you already but let me know if I need to re-up.

sevein:
f96df2f7e6ca9007fe6cd4b1029914aab219814e in `qa/1.x`

PR #1261, opened 2018-09-15
Include dc.json in SIP to load transfer metadata

sevein:
PR #676 stopped including the `dc.json` in the created SIP with the
unintended consequence of breaking `load_dublin_core`, i.e. DC metadata
introduced in Transfer would never make it to the METS.

This commits changes Archivematica so `dc.json` is instead removed from
`load_dublin_core` once the metadata has been populated into the
database. `copy_transfer_metadata_and_logs` is untouched - `dc.json` is
still included in the transfer metadata folder of the AIP, e.g.:

    /data/objects/metadata/transfers/Transfer1234-2d08bb8d-58ab-40f1-8d63-818ce400320d/dc.json

Connects to archivematica/Issues#140.


sevein:
@jraddaoui hey you may want to CR this one because we both were involved in the original approach that tried to solve the problem described by https://projects.artefactual.com/issues/11013. Thank you in advance.

jraddaoui (approved):
Ouch! Thanks @sevein, it looks good to me.

PR #1272, opened 2018-09-27
API: handle bogus units gracefully

sevein:
Handle exceptions raised by `get_unit_status` gracefully.

I've added some tests.

Connects to https://github.com/archivematica/Issues/issues/216.

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -234,7 +234,12 @@ def status(request, unit_uuid, unit_type):
     response['uuid'] = unit_uuid
 
     # Get status (including new SIP uuid, current microservice)
-    status_info = get_unit_status(unit_uuid, unit_type)
+    try:
+        status_info = get_unit_status(unit_uuid, unit_type)
+    except IndexError as err:
+        msg = "Unable to determine the status of the unit {}".format(unit_uuid)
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Looks great @sevein!

A minor suggestion, in terms of code coverage, we could add an extra test for one of the `completed` endpoints in which one of the units raises an error getting its status, to hit the `status_err` assignation and the warning logging. Not a big deal, as it will probably require a new fixture, but I thought it was worth mention ;)

sevein:
> A minor suggestion, in terms of code coverage, we could add an extra test for one of the completed endpoints in which one of the units raises an error getting its status, to hit the status_err assignation and the warning logging. Not a big deal, as it will probably require a new fixture, but I thought it was worth mention ;)

Thanks for the suggestion. I believe that `test_completed_ingests_with_bogus_sip` and `test_completed_transfers_with_bogus_transfer` are doing that. They're not confirming that the event is logged but they're verifying that the exception is handled (not raised to the caller like it used to be). If I wanted to test that the function is logging I think we could use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs but it's not available in Py2. I've looked into https://github.com/aurzenligl/pytest-logger but I'm not sure I want to add another dependency. pytest 3.x also brings `caplog` but we're still using pytest 2.x...

PR #1282, opened 2018-10-03
osdeps: remove mariadb-devel

hakamine:
Remove mariadb-devel due to dependency conflicts when using a database
server different from MariaDB.

Connected to https://github.com/archivematica/Issues/issues/22


mamedin (approved):
LGTM, I tested this change several times when updating CentOS/RH artefactual percona role

PR #1283, opened 2018-10-04
Processing: fix new config button

sevein:
A recent change broke the ability to add a new processing configuration because it was trying to load from disk when the name of the config was `None` which always returned a 404 as a reaction to `IOError`. In this commit, the form is rendered without initial data when `name` equals `None`.

Connects to https://github.com/archivematica/Issues/issues/242.


ablwr:
I can make a new config, save it, download it, edit it.

Also if it has no name, I'm prompted:

screenshot from 2018-10-03 21-23-00

ablwr (approved):

PR #1287, opened 2018-10-08
uses split instead of replace to get directory

helrond:
Instead of using `replace()` with a dependency on `get_modified_standard_transfer_path()` (which now returns None if no transfer type is provided) this uses `split()` to parse the filepath and get the directory that way.

Connects to archivematica/Issues#252.


sevein:
I've merged https://github.com/artefactual/archivematica/pull/1291 because I thought it would bring less risk given that we're very close to the release. But @helrond thanks so much for taking your time to do this, I hope we can work on more improvements on the next release cycle!

PR #1294, opened 2018-10-18
Convert Dataverse jobs to sentence-case

ross-spencer:
Following some refining with @sallain and suggestion on the OCUL channel from @joel-simpson that we won't be able to test the fix for the Dataverse issue with extract packages for the `1.8` milestone, this is a discrete commit that corrects all Dataverse tasks and jobs to sentence-case. The commit also brings all the strings out of the weeds of the code so it is easier to maintain. 

The migration has been tested locally. Rebased with the latest commit in `qa/1.x` and including a typo fix suggested by @sevein. 

Connected to archivematica/Issues#266


jraddaoui (approved):
:+1:

ross-spencer:
Hi @sromkey there was some discussion around this on Slack. I didn't see a final commitment but it sounded like @sevein and @sallain were largely in favor of setting the `1.8` milestone once the details on its impact was a bit clearer. I'll wait for your final decision on tagging it before doing anything more right now. We have a merge approval from @jraddaoui so as soon as we have a direction I can do that work.

ablwr (approved):

PR #1295, opened 2018-10-19
api.views: fix get_unit_status for SIPs

hakamine:
Fix get_unit_status for SIPs, to use the job that is the last in the
microservice chain, not the one with the latest created time.

Unit tests fixed/added

Connected to https://github.com/archivematica/Issues/issues/262

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -158,7 +158,24 @@ def get_unit_status(unit_uuid, unit_type):
     :return: Dict with status info.
     """
     ret = {}
-    job = models.Job.objects.filter(sipuuid=unit_uuid).filter(unittype=unit_type).order_by('-createdtime', '-createdtimedec')[0]
+    # get jobs for the current unit ordered by created time
+    unit_jobs = models.Job.objects.filter(sipuuid=unit_uuid).filter(unittype=unit_type).order_by('-createdtime', '-createdtimedec')
+
+    # tentatively choose the job with the latest created time to be the current/last for the unit
+    job = unit_jobs[0]
+
+    # if a SIP, check if there could be another job that is the actual current/last
+    # by looking at the microservicechainlink object (ref. issue 262)
+    if unit_type == 'unitSIP':
+        while job.microservicechainlink.defaultnextchainlink is not None:
+            next_matches = [x for x in unit_jobs
+                            if x.microservicechainlink ==
+                            job.microservicechainlink.defaultnextchainlink]
+            if next_matches:
+                job = next_matches[0]
+            else:
+                break
+
     ret['microservice'] = job.jobtype
     if job.currentstep == models.Job.STATUS_AWAITING_DECISION:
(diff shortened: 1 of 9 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -158,7 +158,24 @@ def get_unit_status(unit_uuid, unit_type):
...
+    # tentatively choose the job with the latest created time to be the current/last for the unit
+    job = unit_jobs[0]
+
+    # if a SIP, check if there could be another job that is the actual current/last
+    # by looking at the microservicechainlink object (ref. issue 262)
+    if unit_type == 'unitSIP':
+        while job.microservicechainlink.defaultnextchainlink is not None:
+            next_matches = [x for x in unit_jobs
+                            if x.microservicechainlink ==
+                            job.microservicechainlink.defaultnextchainlink]
```
jraddaoui, line 173: Hi @hakamine, the code looks good and I'm not sure this is necessary, but I wonder if we could follow the path like it's done in [here](https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPServer/lib/jobChainLink.py#L111) instead of using the `defaultnextchainlink` job.

jraddaoui (commented):
Thanks @hakamine, just a comment/question.

src/dashboard/src/components/api/views.py
hakamine: @jraddaoui thank you for the review. I am not sure if I understand: do you mean to follow the path just by querying the `MicroServiceChainLink` objects directly (i.e., the `MicroServiceChainLinks` table), instead of following within the job objects? Please note that I am doing the latter because I think it's possible that there is not yet a job created for the next link in the chain.

hakamine:
Note: due to release scheduling constraints, hold off on merging until after AM 1.8 release

src/dashboard/src/components/api/views.py
jraddaoui: Hi @hakamine, this is still new to me, and I'm not sure if I'm understanding the fix correctly.

Looking at your changes, it seems like you're looking at the default path of the chain, I wonder if you could check the jobs output to follow the real path it took.

hakamine: @jraddaoui thank you, I think I understand it now (have to check `MicroServiceChainLinkExitCode` objects as well) This seems to be a little more involved than initially thought... I'll try to fix

hakamine: Hi @jraddaoui, after thinking a bit more about this, it seems to me that trying to follow the microservice chains would add too much complexity to the function. Instead I just created a workaround that is much simpler, in https://github.com/artefactual/archivematica/pull/1313 (the workaround in the new PR covers the case that affects automation tools causing it to block, mentioned in https://github.com/archivematica/Issues/issues/262 )

hakamine:
Closing. Replacing with https://github.com/artefactual/archivematica/pull/1313

PR #1296, opened 2018-10-20
Verify AIP: fix checksum verification

sevein:
When verifying checksums, omit files listed under the file group
`manualNormalization`.

This is connected to https://github.com/archivematica/Issues/issues/287.


jraddaoui (approved):
@sevein, looks good! But Flake8 doesn't like it, it wants the `or` before the line breaks. Feel free to merge after that ;)

Issue #1044, opened 2018-04-05
Problem: pressing Enter in Archival Storage tab search does not execute the current search


jrwdunham:
it executes the previous one.

ross-spencer:
Marking as `Severity:High` for discussion for inclusion in `1.9` as a PR from @replaceafill exists to resolve this issue.

ross-spencer:
This is a very popular issue! Closing in favor of https://github.com/archivematica/Issues/issues/271 simply because it is in he newer issues repo.

PR #1297, opened 2018-10-20
Dashboard: fix use of Return key in archival storage search

replaceafill:
This PR fixes the search in the Archival Storage tab when the Return key is used to "submit" the form instead of clicking the `Search archival storage` button. The problem is caused by the query input not having its value set before the new redirect URL is calculated. Forcing the blur event on it fixes the problem.

There was a related problem when additional rows were added to the search but the event handler wasn't attached to them. Using `.on()` instead fixes the problem.

**Note**: I wasn't sure about the merge workflow so I based the PR against the `qa/1.x` branch.

Connected to https://github.com/artefactual/archivematica/issues/1044
Connected to archivematica/issues#271

### DIFF
--- a/src/dashboard/src/media/js/archival_storage/archival_storage_search.js
+++ b/src/dashboard/src/media/js/archival_storage/archival_storage_search.js
@@ -122,8 +122,10 @@ $(document).ready(function() {
     return false;
   });
 
-  $('.aip-search-query-input').keypress(function (e) {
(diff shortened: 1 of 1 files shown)


sevein:
Thanks for this, @replaceafill. We're at the point in our release cycle that non-critical fixes are going to wait for the next release. We'll look at this as soon as possible!

sevein (approved):

PR #1299, opened 2018-10-23
Dashboard: pull translations

sevein:
This also includes appraisal-tab and transfer-browser.

If you want to test this, make sure you run `manage.py compilemessages` and the frontends are rebuilt!

Connects to https://github.com/archivematica/Issues/issues/231.


src/dashboard/src/main/templatetags/lang.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@register.filter
+def standardize_lang_code(language_code):
+    """Django uses language codes like es-es or pt-br. This will convert them
+    into es_ES and pt_BR."""
+    head, sep, tail = language_code.partition("-")
+    if sep == '':
+        return head
+    return "{}_{}".format(head, tail.upper())
```
sevein, line 13: I had to do this so `pt-br` shows up as `pt_BR` in `window.DashboadConfig` - otherwise `angular-gettext` wouldn't understand the value as it expects the latter format.

src/dashboard/src/settings/base.py
```diff
@@ -156,10 +156,13 @@
...
 # https://docs.djangoproject.com/en/dev/ref/settings/#languages
 LANGUAGES = [
-    ('fr', _('French')),
     ('en', _('English')),
+    ('fr', _('French')),
     ('es', _('Spanish')),
+    ('ja', _('Japanese')),
     ('pt', _('Portuguese')),
+    ('pt-br', _('Brazilian Portuguese')),
+    ('sv', _('Swedish')),
```
sevein, line 165: These are the new languages supported!

jraddaoui (approved):

jraddaoui:
Hi @sevein, 

We could take the chance to change the language code in the `html` tag attributes.

sevein:
> We could take the chance to change the language code in the html tag attributes.

@jraddaoui good catch. I'm going to capture that bit as a new issue just to make sure it's carefully considered (https://github.com/archivematica/Issues/issues/297).

sallain:
Connected to archivematica/Issues#297

PR #1300, opened 2018-10-24
WIP Enable use of identifiers.json for PIDs in PREMIS

ross-spencer:
:construction: :construction: :construction:

Where an identifiers.json file is found in a transfer metadata folder
and Bind PIDs is configured to 'Yes'; parse the identifiers.json file
and record the discovered PIDs in PREMIS.

If the Bind PIDs is configured to Yes and the handle server
configuration is missing or incorrect then the identifiers.json will
still be used.

Identifiers can be recorded for the SIP, directories, and files.

Some redundancy has been removed between Bind PID modules though some
may remain.

New steps have been taken to make sure services exit more gracefully
if there are issues around configuration, e.g. if the required fields
have not been populated for the handle server, the the suggestion is
that there is zero configuration. This should not result seemingly in
a failure for day-to-day users.

Flake8 and Pylint changes have been included in this commit to help
improve the robustness of the code.

Connected to archivematica/issues#133
Connected to archivematica/Issues#285
Connected to archivematica/Issues#284
Connected to IISH#1
Connected to IISH#3

### DIFF
--- a/src/MCPClient/lib/clientScripts/bind_pid.py
+++ b/src/MCPClient/lib/clientScripts/bind_pid.py
@@ -49,9 +49,11 @@ from django.db import transaction
 import django
 django.setup()
 # dashboard
-from main.models import DashboardSetting, File, Identifier
+from main.models import DashboardSetting, File
 # archivematicaCommon
 import bindpid
+from bind_pid_helpers import validate_handle_server_config, \
+    _add_custom_pid_to_mdl_identifiers
 from custom_handlers import get_script_logger
 from archivematicaFunctions import str2bool
 
@@ -75,6 +77,9 @@ def exit_on_known_exception(func):
     """
     @wraps(func)
     def wrapped(*_args, **kwargs):
+        """Try the function following the decorator and return as appropriate
+        if we fall foul of an exception.
+        """
         try:
             func(*_args, **kwargs)
         except (BindPIDException, BindPIDWarning) as exc:
@@ -94,6 +99,8 @@ def _get_bind_pid_config(file_uuid):
     _args = {'entity_type': 'file',
              'desired_pid': file_uuid}
     _args.update(DashboardSetting.objects.get_dict('handle'))
+    if not validate_handle_server_config(_args):
+        raise BindPIDWarning
     _args['pid_request_verify_certs'] = str2bool(
         _args.get('pid_request_verify_certs', 'True'))
     return _args
@@ -115,9 +122,9 @@ def _update_file_mdl(file_uuid, naming_authority, resolver_url):
         # redundant identifiers for the file.
         matches = [True for id_ in existing_ids
                    if id_.type == id_type and id_.value == id_val]
-        if len(matches) == 0:
-            idfr = Identifier.objects.create(type=id_type, value=id_val)
-            file_mdl.identifiers.add(idfr)
+        if len(matches) is 0:
+            _add_custom_pid_to_mdl_identifiers(
+                mdl=file_mdl, scheme=id_type, value=id_val)
 
 
 @exit_on_known_exception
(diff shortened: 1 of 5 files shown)


ThomasEdvardsen:
Hi. Thomas from the National Library of Norway here.

We have started testing Archivematica for legally deposited music, and would like to get the functionality with external provided identifiers described in this issue.

When could we expect this to be released @ross-spencer ?

Best regards,
Thomas Edvardsen

ross-spencer:
Hi @ThomasEdvardsen I have just submitted a PR to replace this one, but it is complete pending CR and so will be released in line with the Archivematica `1.10` release schedule. 

The new work effort can be found here: https://github.com/artefactual/archivematica/pull/1410 and we'd love your feedback as you start making use of it. 

One part of that work that I can point you in that might be useful is the sample data that describes the file naming convention and the format we're expecting folk to use. If you have any comments on that, please let me know. https://github.com/artefactual/archivematica-sampledata/pull/53

ThomasEdvardsen:
Hi @ross-spencer. I found some code in index_aip.py, which indicates that it is already possible to import identifiers from a mods file in submissionDocumentation. 

Is this functionallity documented, and how is this different from using identifiers.json ?

[Link to Importing identifiers from Mods](https://github.com/artefactual/archivematica/blob/stable/1.9.x/src/MCPClient/lib/clientScripts/index_aip.py#L33)

ross-spencer:
Hi @ThomasEdvardsen that's an interesting question. 

From what I can tell, this is undocumented, but I'll log an issue and see if we can resolve that. 

There are fixtures you can look at which will show you the MODS examples we were testing against. OBJID is returned from [this one](https://github.com/artefactual/archivematica/blob/a67d9d30832a313e49a8242d869739690b35e86c/src/archivematicaCommon/tests/fixtures/test-identifiers-islandora-METS.xml#L2) and in the second one, [these identifiers](https://github.com/artefactual/archivematica/blob/a67d9d30832a313e49a8242d869739690b35e86c/src/archivematicaCommon/tests/fixtures/test-identifiers-MODS-METS.xml#L26-L31) look like the ones that will be indexed.

I'll try and ascertain which workflows these currently belong to, and who might use them, but Fedora/Islandora seem to be two examples. 

The functionality is limited to indexing and so will appear in Elasticsearch. I can't see corresponding code where MODS identifiers are taken from METS and then added to the AIP METS created by Archivematica. 

When [identifiers.json](https://github.com/artefactual/archivematica-sampledata/blob/2f700f6c4a1bf95253159916260fbe5a4c6fb346/SampleTransfers/DemoTransferCSV/metadata/identifiers.json) is parsed it is done so in the Bind PIDs group of Microservice Jobs. It will appear in the AIP METS as a `premis:objectIdentifier`:
```xml
<premis:objectIdentifier>
   <premis:objectIdentifierType>ULID</premis:objectIdentifierType>
   <premis:objectIdentifierValue>01D9T8MJM9NEQ390H239VT50CQ</premis:objectIdentifierValue>
</premis:objectIdentifier>
<premis:objectIdentifier>
    <premis:objectIdentifierType>EXID</premis:objectIdentifierType>
    <premis:objectIdentifierValue>https://example.com/AIP/id/1348554167</premis:objectIdentifierValue>
</premis:objectIdentifier>
```

We missed the existing capability in our analysis, so thank you for bringing it to our attention. I'll log an issue about gathering more documentation around the MODS identifiers support. Hopefully you see some relative merit in being able to adopt one over the other. One other approach; folks could also adopt [dc:identifier](http://dublincore.org/specifications/dublin-core/dcmi-terms/2012-06-14/?v=elements#identifier) as well which would both be indexed, and appear in the AIP METS with the descriptive metadata. We recently did some testing with a client using this approach. It's safe to say there are options!

ross-spencer:
MODS identifiers issue logged [here](https://github.com/archivematica/Issues/issues/687).

PR #1301, opened 2018-10-24
Enable use of CONTENTIDS for nested paths in Struct Maps

ross-spencer:
:construction: :construction: :construction: 

Previously the FILEID field used by the user-supplied structmap was
limited to the NCName type which is a non-colonized. Slashes which
denote nested directory structures are not allowed in an NCName.

We can use CONTENTIDS as a workaround. This commit allows a user to
create an user-supplied structmap for a transfer whereby the
CONTENTIDS field can described a fully-qualified path inside the SIP.

The CONTENTIDS field is expected to match up with the contents of the
DB. When a file is discovered, the FILEID field in the structmap is
replaced with the UUID of the matching file as previously
anticipated.

Connected to archivematica/issues#283


ross-spencer:
Closing via https://github.com/artefactual/archivematica/pull/1404

PR #1302, opened 2018-10-25
storageService: add timeout retry to async poll

sevein:
We've experienced random read timeouts when polling the Async SS
endpoint. This commit introduces a retry with up to four attempts and a
one second backoff.

We're planning to confirm this fix in QA before merging.

This is connected to https://github.com/archivematica/Issues/issues/280.

### DIFF
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -4,11 +4,13 @@ import logging
 import os
 import platform
 import requests
-from requests.auth import AuthBase
 import urllib
 import time
 
 from django.conf import settings as django_settings
+from requests.auth import AuthBase
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 # archivematicaCommon
 from archivematicaFunctions import get_setting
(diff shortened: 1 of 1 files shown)


src/archivematicaCommon/lib/storageService.py
```diff
@@ -28,6 +30,18 @@ class WaitForAsyncError(Error):
     pass
 
 
+# The default retry for idempotent endpoints.
+DEFAULT_RETRY = Retry(total=4, backoff_factor=1)
+
+# This retry can be used in POST requests where we want to make multiple
+# attempts in case of communication errors. It is advisable not to use this
+# retry unless you can afford having then endpoint run more than once.
```
jraddaoui: `then` -> `the`

jraddaoui (approved):
Nice!

sevein:
This doesn't solve the problem, unfortunately. I'll leave it around because it can still be useful. See https://github.com/archivematica/Issues/issues/280 for more.

sevein:
I'm going to close this for now. It's perfectly valid but it does not solve the problem reported by #280 so I don't want to add a new feature at this point.

PR #1303, opened 2018-10-29
MCPClient: use instance attribute in rights_from_csv

sevein:
`rights_from_csv` uses an internal data structure that was being stored
as an attribute of the class instead of as an attribute of the instance.
With the new MCPClient inline execution of client scripts, the data was
being shared causing the script to misbehave.

This is connected to https://github.com/archivematica/Issues/issues/305.

### DIFF
--- a/src/MCPClient/lib/clientScripts/rights_from_csv.py
+++ b/src/MCPClient/lib/clientScripts/rights_from_csv.py
@@ -42,7 +42,6 @@ class RightCsvReader(object):
 
     metadata_applies_to_type = None
     current_row = None
-    object_basis_act_usage = {}
     rows_processed = 0
 
     required_column_names = [
@@ -77,6 +76,7 @@ class RightCsvReader(object):
         self.transfer_uuid = transfer_uuid
         self.rights_csv_filepath = rights_csv_filepath
         self.job = job
+        self.object_basis_act_usage = {}
 
     def parse(self):
         """Read and parse rights CSV file."""


src/MCPClient/lib/clientScripts/rights_from_csv.py
```diff
@@ -42,7 +42,6 @@ class RightCsvReader(object):
 
     metadata_applies_to_type = None
     current_row = None
-    object_basis_act_usage = {}
     rows_processed = 0
```
ross-spencer, line 45: It seems to me that these variables should be in the constructor as well, is there a reason you elected to leave them out with this fix?

src/MCPClient/lib/clientScripts/rights_from_csv.py
```diff
@@ -77,6 +76,7 @@ def __init__(self, job, transfer_uuid, rights_csv_filepath):
         self.transfer_uuid = transfer_uuid
         self.rights_csv_filepath = rights_csv_filepath
         self.job = job
+        self.object_basis_act_usage = {}
```
ross-spencer, line 79: :+1:

ross-spencer (changes requested):
I think some clarification around why the other member variables were left out of the constructor might be helpful. `rows_processed` looks like a simple one that looks like it will just keep being incremented every time the script is run. I could be reading that wrong though.

src/MCPClient/lib/clientScripts/rights_from_csv.py
sevein: I guess I was just trying to reduce the scope of the fix in order to minimize risk. If you look at the way that other class attributes are used, `object_basis_act_usage` is really the only one with a mutable value type that is being manipulated.

I found this clarifying: https://timothyawiseman.wordpress.com/2012/10/06/class-and-instance-variables-in-python-2-7/. This one gave me a good overview too: https://www.toptal.com/python/python-class-attributes-an-overly-thorough-guide.

I'm open about it though, if you feel that it should be changed at this point feel free to push it and I'll do it.

ross-spencer (approved):

src/MCPClient/lib/clientScripts/rights_from_csv.py
ross-spencer: Thanks @sevein. I had a read, and gave the code a whirl. Looks good to me. No reason to expand the scope :+1:

sevein: Thanks for taking the time to do that!

PR #1307, opened 2018-11-08
handle transfer source copy errors

jlj5aj:
This addresses part a) of the Expected Behavior section in https://github.com/archivematica/Issues/issues/325

Connects to https://github.com/archivematica/Issues/issues/325.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -267,8 +267,15 @@ def start_transfer(transfer_name, transfer_type, accession, paths, row_ids):
             filepath = os.path.join(temp_dir, target)
 
         transfer_relative = transfer_dir.replace(SHARED_DIRECTORY_ROOT, '', 1)
-        copy_from_transfer_sources([path], transfer_relative)
         filepath = archivematicaFunctions.unicodeToStr(filepath)
+
+        error, message = copy_from_transfer_sources([path], transfer_relative)
+
+        if error:
+            shutil.rmtree(temp_dir)
+            logger.exception("Error copying %s from transfer source: %s", filepath, message)
+            raise storage_service.StorageServiceError('Error copying {} from transfer source.'.format(filepath))
+
         try:
             destination = copy_to_start_transfer(
                 filepath=filepath,


sallain:
Connected to archivematica/Issues#325

ross-spencer (changes requested):
Hi @jlj5aj - I have been looking at this today. Archivematica has shifted a little since this PR. I'm not sure if you feel this is really needed any more, but if so, feel free to rebase, and we can include it. 

What I am seeing:

If I create a new space in the storage service with an artificially low amount of space:

```
dd if=/dev/zero of=20MB_HDD.img bs=1M count=20
fdisk 20MB_HDD.img
    --> with options: g, n, default, default, default, w
mkfs.ext4 20MB_HDD.img 
sudo losetup -Pf --show 20MB_HDD.img 
mkdir /mnt/small-storage-space
mount /dev/loop0 -o loop /mnt/small-storage-space
```

If I configure the storage service to use this space as a currently processing location and then start a 4gb transfer (the space is only 20MB) then I don't get any unusual behavior in the UI. 

If I make use of the 1.7 API with your code, then it do get a user friendly message:

```
http --pretty=format \
    -f \
    POST "http://51.75.94.10/api/transfer/start_transfer/" \
    Authorization:"ApiKey analyst:..." \
    name="api-demo-1" \
    type="standard" \
    paths[]="\
    [$(echo -n 'a329a708-6ab3-4c9e-a053-e83155b58650:/home/centos/archivematica-sampledata/one-file' \
    | base64 -w 0)]"  
```

```
HTTP/1.1 500 INTERNAL SERVER ERROR
Connection: keep-alive
Content-Language: en
Content-Type: application/json
Date: Fri, 08 Nov 2019 16:59:57 GMT
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Language, Cookie

{
    "error": true,
    "message": "Error copying /mnt/small-storage-space/tmp/tmpveoWe8/api from transfer source."
}
```

Without your code:

```
HTTP/1.1 500 INTERNAL SERVER ERROR
Connection: keep-alive
Content-Language: en
Content-Type: application/json
Date: Fri, 08 Nov 2019 16:58:19 GMT
Server: nginx
Transfer-Encoding: chunked
Vary: Accept-Language, Cookie

{
    "error": true,
    "message": "Error starting transfer /mnt/small-storage-space/tmp/tmpO_wmfD/api: Filepath /mnt/small-storage-space/tmp/tmpO_wmfD/api does not exist."
}
```

What I think is important is that in between the time of `1.7` and `1.10` the transfer browser now uses the [package endpoint](https://github.com/artefactual/archivematica/blob/ef3af9fb58ac700dfca2a350095ef13a0c396e9c/src/dashboard/src/components/api/views.py#L799-L805) which seems to be graceful enough, and doesn't result in a folder being created in a watched directory for approval as a transfer. 

The `start_transfer` endpoint is only maintained as we want to remain backward compatible to older callers of this function, but normally we'd suggest the v2 alternative. 

Let us know what you think, we can include this small change from an API perspective, I believe, but otherwise, we can probably close this PR and the original issue if you seem satisfied with AM 1.10's handling of this situation.

PR #1310, opened 2018-12-01
Return Exceptions in JSON response

helrond:
Connects to https://github.com/archivematica/Issues/issues/354.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -255,9 +255,9 @@ def start_transfer(transfer_name, transfer_type, accession, access_id, paths, ro
                 type=transfer_type, accession=accession,
                 access_id=access_id,
                 transfer_metadata_set_row_uuid=row_id)
-        except Exception:
+        except Exception as e:
             logger.exception("Error copying %s to start of transfer", filepath)
-            raise storage_service.StorageServiceError('Error copying {} to start of transfer.'.format(filepath))
+            raise Exception('Error starting transfer: {}'.format(filepath, e))
 
     shutil.rmtree(temp_dir)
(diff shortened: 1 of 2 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -255,9 +255,9 @@ def start_transfer(transfer_name, transfer_type, accession, access_id, paths, ro
                 type=transfer_type, accession=accession,
                 access_id=access_id,
                 transfer_metadata_set_row_uuid=row_id)
-        except Exception:
+        except Exception as e:
             logger.exception("Error copying %s to start of transfer", filepath)
-            raise storage_service.StorageServiceError('Error copying {} to start of transfer.'.format(filepath))
+            raise Exception('Error starting transfer: {}'.format(filepath, e))
```
sevein: The string is not using the second argument `e`, can you remove it?

sevein (changes requested):
@helrond thanks for doing this! I've only added a tiny observation.

Thanks for improving this area of the application, it's definitely very improvable! Slightly unrelated, but I find somewhat cumbersome that `start_transfer` is dealing with concerns like returning API objects or logging. Even its location in `filesystem_ajax.views` is questionable. It could use a facelift for sure.

helrond:
Thanks @sevein! I changed the exception and the logging error so they're both the same, and also both return the actual exception thrown. 

I agree with your observation about `filesystem_ajax.views` - a lot of what's in that file doesn't, IMO, have anything to to with views. Making sure the Archivematica API works is high on my list of priorities, so I'd be interested in looking at this some more if that aligns with the roadmap and other ideas about the API.

sevein (approved):

PR #1311, opened 2018-12-11
Return full stderr when MediaConch has a policy XML error

ablwr:
Resolves https://github.com/archivematica/Issues/issues/345. 

Only one line matters in this PR, and that's LN:236. The rest is me cleaning the file up according to the very selective flake8 spec (although admittedly ignoring some of its sillier complaints against the structure of code descriptions).

### DIFF
--- a/src/MCPClient/lib/clientScripts/policy_check.py
+++ b/src/MCPClient/lib/clientScripts/policy_check.py
@@ -35,6 +35,7 @@ FAIL_CODE = 1
 
 
 def main(job, file_path, file_uuid, sip_uuid, shared_path, file_type):
+    """Entry point for policy checker."""
     setup_dicts(mcpclient_settings)
 
     policy_checker = PolicyChecker(
@@ -46,7 +47,8 @@ logger = get_script_logger("archivematica.mcp.client.policyCheck")
 
 
 class PolicyChecker(object):
-    """Checks whether a given file conforms to all of the MediaConch policies
+    """Checks a file against one or more policies.
+    Checks whether a given file conforms to all of the MediaConch policies
     that the system is configured to run against that type of file, given the
     file's format and its purpose, i.e., whether it is intended for access or
     preservation. Usage involves initializing on a file and then calling the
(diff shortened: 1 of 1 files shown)


ablwr:
Sorry @sromkey for working on this even though it's outside of our milestone spec :grimacing: !!!

src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -231,7 +233,7 @@ def _execute_rule_command(self, rule):
         except ValueError:
             logger.exception(
                 'Unable to load an object from the malformed JSON in\n%s',
-                stdout)
+                stderr)
```
sevein, line 236: I haven't tested but would this be the message logged after your change?

```
Unable to load an object from the malformed JSON in
Validation generated an internal error:    Invalid expression
    compilation error: element when
    xsl:when : could not compile test expression 'mi:MediaInfo/mi:track[@type='General'][*]/mi:FormatIs equal (=)'MPEG-4''
```
If that's the case, shouldn't the end of our sentence end with a colon or period? E.g.:

    Unable to load an object from the malformed JSON:
    [...]

ablwr: Nice! I just (force-)pushed a simplified update with the added colon.

src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -230,8 +232,8 @@ def _execute_rule_command(self, rule):
             output = json.loads(stdout)
         except ValueError:
             logger.exception(
-                'Unable to load an object from the malformed JSON in\n%s',
-                stdout)
+                'Unable to load an object from the malformed JSON: ',
+                stderr)
```
sevein, line 236: This will probably raise an exception because the `%s` is now missing. I would have thought `PolicyChecker` had coverage, but it doesn't seem so.

sevein (changes requested):

src/MCPClient/lib/clientScripts/policy_check.py
ablwr: :woman_facepalming:

ablwr: Sorry! Dunno what I was thinking

sevein (approved):

PR #1251, opened 2018-09-04
MCPServer: introduce JSON-encoded workflow

sevein:
This pull request introduces a new [`workflow`](https://github.com/artefactual/archivematica/blob/dev/issue-1101-workflow-json/src/MCPServer/lib/workflow.py) module in order to expose JSON-encoded workflows to MCPServer. The Django models are being deleted.

Workflow: [`src/MCPServer/lib/assets/workflow.json`](https://github.com/artefactual/archivematica/blob/dev/issue-1101-workflow-json/src/MCPServer/lib/assets/workflow.json).
Schema: [`src/MCPServer/lib/assets/workflow-schema-v1.json`](https://github.com/artefactual/archivematica/blob/dev/issue-1101-workflow-json/src/MCPServer/lib/assets/workflow-schema-v1.json).

This is connected to #1101.

PR #1313, opened 2018-12-12
api.views: Add get_unit_status workaround for SIPs

hakamine:
For SIPs, in case the job with the latest created time is not the one
that closes(completes) the chain, check if there is a job that does so.
Unit test added.

Connected to archivematica/Issues#262


sevein (approved):
This is great, @hakamine! Thank you. Perhaps we can revisit the solution once #1251 is merged. I think we could try to find find new ways to annotate the workflow to signal completion.

PR #1319, opened 2018-12-20
Bring fpr and disconnect from workflow

sevein:
This pull request imports the `fpr` app and remove the bits from `IDCommand` that manipulated the workflow data in preparation for the merge of https://github.com/artefactual/archivematica/pull/1251 where the workflow data is removed from the database.

And you may be wondering... why 120K lines of code? Well that's just mostly the initial data that we populate to users via fixtures, mostly related to PRONOM updates!

We'd still need to figure out what's the impact of removing this feature:
- Were users relying on it? It's definitely not documented!
- How could we work around it? Once #1251 is merged, we could explain how to alter the workflow in according to new identification commands. This could easily be done in the scope of AM19.
- What would be a good solution in the long term? E.g. implement a workflow link that can list the identification commands made available via fpr, similar to the link that lists SS locations during AIP/DIP storage.

My suggestion: merge this, file a new issue to tackle the above.

- Connects to https://github.com/archivematica/Issues/issues/213
   `Problem: fpr app import and writes to main.models`
- Connects to https://github.com/archivematica/Issues/issues/181
    `Problem: it's hard to maintain fpr-admin as a separate repo`


jraddaoui:
Hi @sevein, this looks good to me, but maybe wait for @ross-spencer to check it out after you end up with #1251.

ablwr (approved):

PR #1323, opened 2018-12-25
Update to PREMIS 3

ablwr:
This updates all PREMIS we make to PREMIS 3.

This is connected to https://github.com/archivematica/Issues/issues/370.

Tests need to be updated still, which seems like it will have to be a total overhaul. Which is probably good to do!

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
@@ -44,6 +44,30 @@ def formatDate(date):
     return date
 
 
+def _add_start_end_date_complex_type(
+        name, elem, start_date, end_date, end_date_open):
+    """Add new ``startAndEndDateComplexType`` subelement to ``elem``.
+
+    The following elements in PREMIS use it:
+
+      - ``copyrightApplicableDates``
+      - ``licenseApplicableDates``
+      - ``otherRightsApplicableDates``
+      - ``statuteApplicableDates``
+      - ``termOfGrant``
+      - ``termOfRestriction``
+
+    """
+    if not any([start_date, end_date, end_date_open]):
+        return
+    dates = etree.SubElement(elem, ns.premisBNS + name)
+    etree.SubElement(dates, ns.premisBNS + "startDate").text = \
+        formatDate(start_date) if start_date else ""
+    end_date = "OPEN" if end_date_open else formatDate(end_date)
+    if end_date:
+        etree.SubElement(dates, ns.premisBNS + "endDate").text = end_date
+
+
 def archivematicaGetRights(job, metadataAppliesToList, fileUUID, state):
     """[(fileUUID, fileUUIDTYPE), (sipUUID, sipUUIDTYPE), (transferUUID, transferUUIDType)]"""
     ret = []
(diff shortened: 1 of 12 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETSRights.py
```diff
@@ -97,6 +97,8 @@ def createRightsStatement(job, statement, fileUUID, state):
             copyrightApplicableDates = etree.SubElement(copyrightInformation, ns.premisBNS + "copyrightApplicableDates")
             if copyright.copyrightapplicablestartdate:
                 etree.SubElement(copyrightApplicableDates, ns.premisBNS + "startDate").text = formatDate(copyright.copyrightapplicablestartdate)
+            else:
+                etree.SubElement(copyrightApplicableDates, ns.premisBNS + "startDate").text = ""
```
sevein: If this is needed, wouldn't that be true for every `startAndEndDateComplexType`?

```xsd
<xs:complexType name="startAndEndDateComplexType">
  <xs:sequence>
    <xs:element ref="startDate"/>
    <xs:element ref="endDate" minOccurs="0"/>
  </xs:sequence>
</xs:complexType>
```

I've confirmed by looking at the [XSD](https://www.loc.gov/standards/premis/v3/premis-v3-0.xsd) that `copyrightApplicableDates` is a `startAndEndDateComplexType` but also `licenseApplicableDates`, `statuteApplicableDates`, `termOfGrant` and `termOfRestriction`.

src/MCPClient/requirements/base.txt
```diff
@@ -6,7 +6,7 @@ django-extensions==1.1.1
 mysqlclient==1.3.9
 gearman==2.0.2
 lxml==3.5.0
-metsrw==0.2.4
+git+https://github.com/artefactual-labs/mets-reader-writer@8186e8e3da4a777cf212d0e6f7796343613ca294#egg=metsrw
```
sevein: I wanted to add a comment on `store_aip.py` but GitHub won't let me do that because the file is unchanged in this pull request I believe. See the following two blocks:

https://github.com/artefactual/archivematica/blob/e240825332b76f3cc50b66646921fd9b1ab84d4b/src/MCPClient/lib/clientScripts/store_aip.py#L251

https://github.com/artefactual/archivematica/blob/e240825332b76f3cc50b66646921fd9b1ab84d4b/src/MCPClient/lib/clientScripts/store_aip.py#L287

I think that they also need some tweaking for PREMIS 3, i.e. use `PREMIS_3_META` instead and update `event_detail` according to the new name.

src/dashboard/src/components/archival_storage/atom.py
```diff
@@ -87,11 +87,11 @@ def add_prop_from_xml(dict_, name, el, xpath):
...
+            add_prop_from_xml(attrs, 'size', amdsec, './/{http://www.loc.gov/premis/v3}objectCharacteristics/{http://www.loc.gov/premis/v3}size')
+            add_prop_from_xml(attrs, 'format_name', amdsec, './/{http://www.loc.gov/premis/v3}objectCharacteristics/{http://www.loc.gov/premis/v3}format/{http://www.loc.gov/premis/v3}formatDesignation/{http://www.loc.gov/premis/v3}formatName')
+            add_prop_from_xml(attrs, 'format_version', amdsec, './/{http://www.loc.gov/premis/v3}objectCharacteristics/{http://www.loc.gov/premis/v3}format/{http://www.loc.gov/premis/v3}formatDesignation/{http://www.loc.gov/premis/v3}formatVersion')
+            add_prop_from_xml(attrs, 'format_registry_name', amdsec, './/{http://www.loc.gov/premis/v3}objectCharacteristics/{http://www.loc.gov/premis/v3}format/{http://www.loc.gov/premis/v3}formatRegistry/{http://www.loc.gov/premis/v3}formatRegistryName')
+            add_prop_from_xml(attrs, 'format_registry_key', amdsec, './/{http://www.loc.gov/premis/v3}objectCharacteristics/{http://www.loc.gov/premis/v3}format/{http://www.loc.gov/premis/v3}formatRegistry/{http://www.loc.gov/premis/v3}formatRegistryKey')
```
sevein: This is a good place to use `metsrw.fsentry.get_premis_objects` because you don't have to tell it the namespace you want to use. That would let us read from both v2 and v3. It returns a `PREMISObject` that makes very easy to access the properties, e.g. `premis_object.size`, `premis_object.format_name`, etc...

sevein (commented):
@ablwr this is looking really good. I've made a few comments.

I think that the tests under `test_reingest_mets` are breaking because they now expect PREMIS 3 since we've made that the new default by changing the value of `premis` in  `namespaces.NSMAP`. So I think what's missing is to update `archivematicaCreateMETSReingest` (the module used to generate the METS during reingest) so it converts PREMIS 2 into PREMIS 3 when it updates existing objects from previous METS.
I looked into using `premisrw` to generate new PREMIS 3 objects but it's hard to work with its tuples when you want to make changes too. Here's an alternative that mutates the original element properties: https://github.com/artefactual/archivematica/commit/20e2977c80fd17ad4e0fe22f80942988cff59abf#diff-291d3b6400c5b8c37393c13f538d0643R17.

ablwr:
I think this is looking excellent and appreciate the scoping-down to not include the pointer at this time.

sevein:
I'm going to continue the development in [dev/issue-24-handle-old-aips](https://github.com/artefactual/archivematica/commits/dev/issue-24-handle-old-aips) because some other issues are arising when working with old namespaceless METS.

Issue #1007, opened 2018-03-21
AIC number doesn't persist in metadata entry form


sallain:
Part of the AIC workflow requires users to add the AIC number to the transfer, usually through the metadata entry form in the GUI:

![screenshot of archivematica dashboard metadata entry form -](https://user-images.githubusercontent.com/4612276/37736908-6c969a14-2d0f-11e8-8189-caf771fdc0a7.jpg)

Entering the data and clicking save creates a new metadata entry, as expected. However, if you then edit the form, the AIC number has disappeared.

The AIC number IS making it to the AIP METS, however. It seems to be the form that is malfunctioning, not the creation of the AIC number itself.

Tested on AM 1.7 (centOS and ubuntu) and on Chrome and Firefox.

sallain:
Closed in favour of archivematica/Issues#311

ross-spencer:
(Was never previously closed, closing now!)

PR #1325, opened 2019-01-04
Dashboard: fix the edit metadata form for transfers

replaceafill:
This PR saves the DublinCore metadata for transfers through the [DublinCoreMetadataForm](https://github.com/artefactual/archivematica/blob/e240825332b76f3cc50b66646921fd9b1ab84d4b/src/dashboard/src/components/ingest/forms.py#L23) instead of saving each field individually.

Connects to https://github.com/archivematica/Issues/issues/311
Connected to https://github.com/artefactual/archivematica/issues/1007

### DIFF
--- a/src/dashboard/tests/test_transfer.py
+++ b/src/dashboard/tests/test_transfer.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+import os
+
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+from django.test.client import Client
+
+from components import helpers
+from main.models import DublinCore
+
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class TestTransferViews(TestCase):
+    fixture_files = ['test_user.json', 'transfer.json']
+    fixtures = [os.path.join(THIS_DIR, 'fixtures', p) for p in fixture_files]
+
+    def setUp(self):
+        self.client = Client()
+        self.client.login(username='test', password='test')
+        helpers.set_setting('dashboard_uuid', 'test-uuid')
+
+    def test_metadata_edit(self):
+        """Test the metadata form of a transfer"""
+        transfer_uuid = '3e1e56ed-923b-4b53-84fe-c5c1c0b0cf8e'
+        url = reverse(
+            'components.transfer.views.transfer_metadata_edit',
+            args=[transfer_uuid],
+        )
+        # Post metadata in Spanish
+        response = self.client.post(url, {
+            'title': u'Mi pequeña transferencia',
+            'is_part_of': u'1234aéiou',
+            'creator': 'El Creador',
+            'subject': 'Un Tema',
+            'description': u'La Descripción',
+            'publisher': 'El Publicista',
+            'contributor': 'Un colaborador',
+            'date': '2019-01-01',
+        })
+        # Verify changes
+        transfer_metadata = DublinCore.objects.get(
+            metadataappliestoidentifier=transfer_uuid)
+        assert transfer_metadata.title == u'Mi pequeña transferencia'
+        assert transfer_metadata.is_part_of == u'AIC#1234aéiou'
+        assert transfer_metadata.creator == 'El Creador'
+        assert transfer_metadata.subject == 'Un Tema'
+        assert transfer_metadata.description == u'La Descripción'
+        assert transfer_metadata.publisher == 'El Publicista'
+        assert transfer_metadata.contributor == 'Un colaborador'
+        assert transfer_metadata.date == '2019-01-01'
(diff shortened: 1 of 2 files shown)


src/dashboard/tests/test_transfer.py
```diff
@@ -0,0 +1,52 @@
...
+        assert transfer_metadata.title == 'My First Transfer'
+        assert transfer_metadata.is_part_of == 'AIC#1234'
+        assert transfer_metadata.creator == 'The Creator'
+        assert transfer_metadata.subject == 'A Subject'
+        assert transfer_metadata.description == 'Transfer description'
+        assert transfer_metadata.publisher == 'The Publisher'
+        assert transfer_metadata.contributor == 'A contributor'
+        assert transfer_metadata.date == '2019-01-01'
+        # Verify form redirects to the metadata list after saving
+        assert response.url == 'http://testserver/transfer/{}/metadata/'.format(transfer_uuid)
```
sevein: Should that value use `reverse` too?

replaceafill: `reverse` doesn't ouput the `http://testserver` part iirc?

replaceafill: Maybe:

```
        redirect_url = 'http://testserver{}'.format(reverse(
            'components.transfer.views.transfer_metadata_list',
            args=[transfer_uuid],
        ))
```

src/dashboard/tests/test_transfer.py
```diff
@@ -0,0 +1,56 @@
+import os
```
ross-spencer, line 2: Something I am considering more in my testing, although I haven't had much opportunity to practice it lately, is making sure there are some Unicode strings incorporated into tests so that there is good coverage for folks wanting to use AM in different languages. You'd need an encoding declaration `# -*- coding: utf-8 -*-` first. Up to you if you'd like to add some strings in the below form to test against.

src/dashboard/tests/test_transfer.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        response = self.client.post(url, {
+            'title': 'My First Transfer',
+            'is_part_of': '1234',
+            'creator': 'The Creator',
+            'subject': 'A Subject',
+            'description': 'Transfer description',
+            'publisher': 'The Publisher',
+            'contributor': 'A contributor',
+            'date': '2019-01-01',
```
ross-spencer, line 40: An aside, but it is interesting the `date` field doesn't validate as a date. Nothing to fix here now, but thanks to your tests I was quickly able to try out a few different strings to see the behavior, nice :+1:

ross-spencer (approved):
I think this looks good @replaceafill. It's always good to see complexity reduced and new tests too. I left a comment about potentially testing some Unicode strings in the form, feel free to address that as you like. Everything passes as expected on my system. 

BTW, you should should be a member of the Artefactual organisations now? You should be good to branch from `artefactual:qa/1.x` in future. Have a chat with @sevein or maybe @scollazo if you still need access.

In terms of process, merges into `qa/1.x` will become part of the next release of Archivematica. Have a chat with @sromkey about a potential `1.9` milestone on this ticket and we will be able to merge it into `qa/1.x` prior to release so it can be tested, otherwise, we can keep this PR around for `1.10`. Once merged, we'll add the `in review` label so that it gets picked up for testing.

ablwr:
Great work right off the bat @replaceafill!! You'll get a feel for our process during the meeting today for sure. :smile:

src/dashboard/tests/test_transfer.py
```diff
@@ -0,0 +1,56 @@
...
+        assert transfer_metadata.subject == 'A Subject'
+        assert transfer_metadata.description == 'Transfer description'
+        assert transfer_metadata.publisher == 'The Publisher'
+        assert transfer_metadata.contributor == 'A contributor'
+        assert transfer_metadata.date == '2019-01-01'
+        # Verify form redirects to the metadata list after saving
+        redirect_url = 'http://testserver{}'.format(reverse(
+            'components.transfer.views.transfer_metadata_list',
+            args=[transfer_uuid],
+        ))
```
sevein, line 57: Would this work?

```python
        redirect_url = response.wsgi_request.build_absolute_uri(reverse(
            'components.transfer.views.transfer_metadata_list',
            args=[transfer_uuid],
        )))
```

replaceafill: Thanks! :+1:

src/dashboard/tests/test_transfer.py
replaceafill: I'll make sure to take this into account when writing tests from now on. I updated the PR to use some non-ascii characters.

PR #1326, opened 2019-01-07
api.views: add get_unit_status workaround for SIPs

hakamine:
For SIPs, in case the job with the latest created time is not the one
that closes the chain, check if there is a job that does so.

Connects to https://github.com/archivematica/Issues/issues/262.

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -82,7 +82,11 @@ def get_unit_status(unit_uuid, unit_type):
     :return: Dict with status info.
     """
     ret = {}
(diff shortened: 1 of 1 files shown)


hakamine:
This is a cherry pick of #1313 on stable/1.6.x (required for a client deployment) (#1313 has already been merged to qa/1.x). I tried to merge to stable/1.6.x, but it failed with this error:
```
remote: error: GH006: Protected branch update failed for refs/heads/stable/1.6.x.
remote: error: At least 1 approving review is required by reviewers with write access.
To github.com:artefactual/archivematica
 ! [remote rejected]   stable/1.6.x -> stable/1.6.x (protected branch hook declined)
```

sevein (approved):

PR #1328, opened 2019-01-10
ArchivesSpace: take base_url instead of host/port

sevein:
Update the ArchivesSpace configuration page so it takes a URL instead of host/port.

The following changes have been made:

- agentarchives updated [v0.5.0](https://github.com/artefactual-labs/agentarchives/releases/tag/v0.5.0) (brings full support for URL values in `host` parameter)
- ArchivesSpace configuration form - showing new `ArchivesSpace base URL` field
- ArchivesSpace task CLI parameters (see migration 0066)
- ArchivesSpace config dictionary (see migration 0066)

See screenshot:

image

This is connected to https://github.com/archivematica/Issues/issues/409.

### DIFF
--- a/src/MCPClient/lib/clientScripts/dip_generation_helper.py
+++ b/src/MCPClient/lib/clientScripts/dip_generation_helper.py
@@ -30,8 +30,7 @@ def create_archivesspace_client():
 
     try:
         client = archivesspace.ArchivesSpaceClient(
-            host=config['host'],
-            port=config['port'],
+            host=config['base_url'],
             user=config['user'],
             passwd=config['passwd'],
             repository=config['repository']
--- a/src/MCPClient/lib/clientScripts/post_store_aip_hook.py
+++ b/src/MCPClient/lib/clientScripts/post_store_aip_hook.py
@@ -47,7 +47,7 @@ def dspace_handle_to_archivesspace(job, sip_uuid):
     # POST Dspace handle to ArchivesSpace
     # Get ArchivesSpace config
     config = models.DashboardSetting.objects.get_dict('upload-archivesspace_v0.0')
-    archivesspace_url = 'http://' + config['host'] + ':' + str(config['port'])
+    archivesspace_url = config["base_url"]
 
     # Log in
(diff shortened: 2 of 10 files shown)


ablwr (approved):

hakamine:
@sevein I have a question related the migration. As you pointed out in redmine 12694, there is a need to port this fix to AM 1.7.x (this port without migration, requiring instead a manual fix of the AS config information?) . However I was wondering if, in the future, when migrating this client's 1.7.x to qa/1.x (1.9.x?), there could be issues (such as the migration failing because data such as host and port is not present). Do we need to worry about it?

sevein:
> @sevein I have a question related the migration. As you pointed out in redmine 12694, there is a need to port this fix to AM 1.7.x (this port without migration, requiring instead a manual fix of the AS config information?) . However I was wondering if, in the future, when migrating this client's 1.7.x to qa/1.x (1.9.x?), there could be issues (such as the migration failing because data such as host and port is not present). Do we need to worry about it?

Yeah right I had the same thought and I think the plan should be to port it so AM17 still relies on the `host` configuration property (from `DashboardSetting.objects.get_dict('upload-archivesspace_v0.0'`) instead of replacing it with `base_url` as I've done here, but making it so it behaves the same way. `port` can be entirely removed since my migration can handle that, i.e. missing value. I don't think it's perfect, e.g. if we were releasing v1.7.3 I'd add in the release notes a mention to this change and recommending people to review their AS settings, but overall is okay maybe?

UPDATE: we decided to go with a slightly different and perhaps more conservative approach in older releases. I'll submit a new PR soon!

J4bbi:
There is a problem in that the agentarchives update will break DSpace REST location functionality in storage service if it is configured to use ArchivesSpace.

https://github.com/artefactual/archivematica-storage-service/blob/8863492193a8f07e89fa6f1a27ad9889be4de822/storage_service/locations/models/dspace_rest.py#L361

sevein:
Thanks @J4bbi, you're totally right. We'll have to make similar tweaks in SS when we update it to use agentarchives v0.5.0.

PR #1285, opened 2018-10-05
Upgrade ES support to version 6.x

jraddaoui:
- Upgrade `elasticsearch` dependency in dashboard and MCPClient.
- Split multi document type indexes to single document indexes:
  - `aips` and `aipfiles` indexes for AIP documents.
  - `transfers` and `transferfiles` indexes for Transfers documents.
- Assume indexes are created correctly and only check existence.
- Avoid the use of `doc_type` and use a default one in all indexes.
- Pass `job.pyprint` to index AIP and files function and normalize
  its use in the index Transfer and files call.
- Improve dashboard commands to rebuild ES indexes.
- Change string fields declaration, use `text` and `keyword`.
- Add `raw` subfields to sort over text fields.
- Check `SEARCH_ENABLED` setting (enhanced in #1172) to create indexes.
- Move indexes check and creation to setup call.
- Create single call to index AIP and AIP files and to index Transfer
  and Transfer files, instead of making it separately.
- Do not check shards after resource deletion.
- Cosmetic changes and reorganization of `elasticSearchFunctions`.
- Reduce max query size to 10000 (default ES limit).
- Add custom analyzer to filenames and paths to create tokens splitting
  on selected chars.
- Properly raise exception in `_try_to index` after max attempts.
- Use `_source` instead of `fields` to get and access the selected
  fields (it doesn't include single values in lists).
- Stop using the `_all` field and checking the mapping to get sub-fields.
- Move filters inside boolean query.
- Improve boolean query generation.
- Use `sum` aggregation to get total AIPs size.
- Install `unar` in dashboard Docker image to allow AIP extraction
  while running the rebuild AIPs indexes command.

**TODO:**

- [x] Modify tests.
- [x] Add task to migrate from 1.x cluster.

**TODO?:**

1. Improve ES hosts declaration:

To allow setting a list of hosts separated by comma.

2. Modify `elasticSearchFunctions` to a class:

To avoid using global variables and passing the client around. The class could be initiated in `wsgi.py` like it's being setup now.

3. Normalize field names to `snake_case` in all indexes:

Field names are quite a mess. I didn't want to change this as external tools may be using the indexes but, since the indexes have changed anyway to be single document indexes, it may be a good moment to do so.

4. Remove ES dependency in dashboard:

Only used to wrap some ES errors, but I think we could find a way to do it via `archivematicaCommon.elasticSearchFunctions` or fully remove it.

5. Avoid index check and creation on the fly:

[...]

PR #1329, opened 2019-01-11
Fix AIP metadata indexing in ES

replaceafill:
This PR fixes the ElasticSearch indexing of files listed in `metadata.csv`. The implementation looks in the `structMap` of the `METS.xml` to get to the correct `dmdSec` for each file.

Connects to https://github.com/archivematica/Issues/issues/274

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -541,13 +541,6 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName, 
     # TODO add a conditional to toggle this
     remove_tool_output_from_mets(tree)
 
-    # get SIP-wide dmdSec
-    dmdSec = root.findall("mets:dmdSec/mets:mdWrap/mets:xmlData", namespaces=ns.NSMAP)
-    dmdSecData = {}
-    for item in dmdSec:
-        xml = ElementTree.tostring(item)
-        dmdSecData = xmltodict.parse(xml)
-
     # Extract isPartOf (for AIPs) or identifier (for AICs) from DublinCore
     dublincore = root.find('mets:dmdSec/mets:mdWrap/mets:xmlData/dcterms:dublincore', namespaces=ns.NSMAP)
     aic_identifier = None
@@ -571,7 +564,7 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName, 
         'isPartOf': is_part_of,
         'AICID': aic_identifier,
         'METS': {
-            'dmdSec': rename_dict_keys_with_child_dicts(normalize_dict_values(dmdSecData)),
+            'dmdSec': {},
             'amdSec': {},
         },
         'origin': get_dashboard_uuid(),
@@ -610,6 +603,32 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName, 
             xml = ElementTree.tostring(amdSecInfo)
             indexData['METS']['amdSec'] = rename_dict_keys_with_child_dicts(normalize_dict_values(xmltodict.parse(xml)))
 
+        # Get the parent division for the file pointer
+        # by searching the physical structural map section (structMap)
+        file_id = file_.attrib.get('ID', None)
+        file_pointer_division = root.find(
+            "mets:structMap[@TYPE='physical']//mets:fptr[@FILEID='{}']/..".format(
+                file_id
+            ),
+            namespaces=ns.NSMAP
+        )
+        if file_pointer_division is not None:
+            # If the parent division has a DMDID attribute then index
+            # its data from the descriptive metadata section (dmdSec)
+            dmd_section_id = file_pointer_division.attrib.get('DMDID', None)
+            if dmd_section_id is not None:
+                dmd_section_info = root.find(
+                    "mets:dmdSec[@ID='{}']/mets:mdWrap/mets:xmlData".format(
+                        dmd_section_id
+                    ),
+                    namespaces=ns.NSMAP
+                )
+                xml = ElementTree.tostring(dmd_section_info)
+                data = rename_dict_keys_with_child_dicts(
+                    normalize_dict_values(xmltodict.parse(xml))
+                )
+                indexData['METS']['dmdSec'] = data
+
         indexData['FILEUUID'] = fileUUID
 
         # Get file path from FLocat and extension
@@ -623,9 +642,11 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName, 
         wait_for_cluster_yellow_status(client)
         try_to_index(client, indexData, index, type_)
 
-        # Reset fileData['METS']['amdSec'], since it is updated in the loop
-        # above. See http://stackoverflow.com/a/3975388 for explanation
+        # Reset fileData['METS']['amdSec'] and fileData['METS']['dmdSec'],
+        # since they are updated in the loop above.
+        # See http://stackoverflow.com/a/3975388 for explanation
         fileData['METS']['amdSec'] = {}
+        fileData['METS']['dmdSec'] = {}
 
     print('Indexed AIP files and corresponding METS XML.')
 
--- a/src/archivematicaCommon/tests/fixtures/test_index_metadata-METS.xml
+++ b/src/archivematicaCommon/tests/fixtures/test_index_metadata-METS.xml
@@ -1,0 +1,402 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!--
+  This METS file is a cut-down version of the AIP METS produced
+  using the SampleTransfers/DemoTransfer
+-->
+<mets:mets xmlns:mets="http://www.loc.gov/METS/" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.loc.gov/METS/ http://www.loc.gov/standards/mets/version111/mets.xsd">
+  <mets:metsHdr CREATEDATE="2019-01-11T09:23:21"/>
+  <mets:dmdSec ID="dmdSec_2">
+    <mets:mdWrap MDTYPE="DC">
+      <mets:xmlData>
+        <dcterms:dublincore xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dcterms="http://purl.org/dc/terms/" xsi:schemaLocation="http://purl.org/dc/terms/ http://dublincore.org/schemas/xmls/qdc/2008/02/11/dcterms.xsd">
+          <dc:title>Morning view from lookout over Queenstown towards the Remarkables in spring</dc:title>
+          <dc:creator>Pseudopanax at English Wikipedia</dc:creator>
+        </dcterms:dublincore>
+      </mets:xmlData>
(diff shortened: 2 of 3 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -610,6 +610,34 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName,
             xml = ElementTree.tostring(amdSecInfo)
             indexData['METS']['amdSec'] = rename_dict_keys_with_child_dicts(normalize_dict_values(xmltodict.parse(xml)))
 
+        # Get the parent division for the file pointer
+        # by searching the physical structural map section (structMap)
+        file_id = file_.attrib.get('ID', None)
```
ross-spencer, line 608: The default for `.get(...)` in this instance is `None` so you don't necessarily have to be explicit here but I think this is good nontheless.

src/archivematicaCommon/tests/test_elasticsearch_functions.py
```diff
@@ -88,3 +89,116 @@ def test_set_tags(self):
...
             elasticSearchFunctions.set_file_tags(self.client, 'no_such_file', [])
+
+    @mock.patch('elasticSearchFunctions.get_dashboard_uuid')
+    @mock.patch('elasticSearchFunctions.wait_for_cluster_yellow_status')
+    @mock.patch('elasticSearchFunctions.try_to_index')
+    def test_index_mets_file_metadata(
+            self,
+            dummy_try_to_index,
+            dummy_wait_for_cluster_yellow_status,
+            dummy_get_dashboard_uuid,
```
ross-spencer, line 100: I've wrestled quite a lot with spacing in Python the last few months. I think I've settled on a style more along the lines of:

``` python
    def test_index_mets_file_metadata(
            self, dummy_try_to_index, dummy_wait_for_cluster_yellow_status,
            dummy_get_dashboard_uuid,):
```

I think it's a bit easier to scan more per line, and I find it a bit easier to read as a function definition. 

What do you think?

The formatting issue should become fairly academic soon as I think we're going to be putting `black` formatting into the GitHub hooks, so that will be :+1:

src/archivematicaCommon/tests/test_elasticsearch_functions.py
```diff
@@ -88,3 +89,116 @@ def test_set_tags(self):
...
+        identifiers = []
+        indexed_files_count = elasticSearchFunctions.index_mets_file_metadata(
+            self.client,
+            uuid,
+            mets_file_path,
+            index,
+            type_,
+            sipName,
+            identifiers,
+        )
```
ross-spencer, line 135: In relation to my layout comment above, when it domes make more sense to adopt a vertical layout,  I find being explicit with the parameter names can also help the reader, so something like:

``` python
        indexed_files_count = elasticSearchFunctions.index_mets_file_metadata(
            client=self.client,
            uuid=uuid,
            metsFilePath=mets_file_path,
            index=index,
            type_=type_,
            sipName=sipName,
            identifiers=identifiers,
        )
```

Renders quite nicely in syntax highlighters. It probably helps more when the variable names are much more different than the parameters they're being used for,.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -538,16 +538,16 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName,
     tree = ElementTree.parse(metsFilePath)
     root = tree.getroot()
 
+    # Map each element's parent because ElementTree doesn't
+    # provide a direct way to do it
+    parent_elements = {}
+    for parent in root.iter():
+        for child in parent:
+            parent_elements[child] = parent
+
```
jraddaoui: I wonder if the parents could be taken using the `..` XPath expression. It looks like it should work as long as the path doesn't attempt to reach the ancestors of the start element:

2.x: https://docs.python.org/2/library/xml.etree.elementtree.html#supported-xpath-syntax
3.x: https://docs.python.org/3/library/xml.etree.elementtree.html#supported-xpath-syntax

If that doesn't work, this mapping seems a bit excessive for this matter. We could iterate over the `mets:structMap[@TYPE='physical']` only and create a mapping with just the `FILEID -> DMDID` relations.

ross-spencer (approved):
:tada: :tada: :tada: 

This is really good Douglas. The code works well. I've learned more about testing reading this. Very minor style comments below around parameters and function definitions, none of which you need to address if you feel they don't add enough value. 

This will need to be rebased with `origin/qa/1.x` before merging.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -623,9 +651,11 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName,
         wait_for_cluster_yellow_status(client)
         try_to_index(client, indexData, index, type_)
 
-        # Reset fileData['METS']['amdSec'], since it is updated in the loop
-        # above. See http://stackoverflow.com/a/3975388 for explanation
+        # Reset fileData['METS']['amdSec'] and fileData['METS']['dmdSec'],
+        # since they are updated in the loop above.
+        # See http://stackoverflow.com/a/3975388 for explanation
         fileData['METS']['amdSec'] = {}
+        fileData['METS']['dmdSec'] = {}
```
jraddaoui, line 649: Nice!

src/archivematicaCommon/tests/test_elasticsearch_functions.py
```diff
@@ -88,3 +89,116 @@ def test_set_tags(self):
...
+        # Neither will the generated files during the transfer process
+        generated_files = (
+            'dc.json',
+            'directory_tree.txt',
+        )
+        for filename in generated_files:
+            path = 'objects/metadata/transfers/DemoTransfer-{}/{}'.format(
+                mets_object_id, filename
+            )
+            assert indexed_data[path] is None
```
jraddaoui, line 204: Great test! It will require some changes after the ES upgrade work is merged but it shouldn't be hard.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -610,6 +610,34 @@ def index_mets_file_metadata(client, uuid, metsFilePath, index, type_, sipName,
...
 
+        # Get the parent division for the file pointer
+        # by searching the physical structural map section (structMap)
+        file_id = file_.attrib.get('ID', None)
+        file_pointer = root.find(
+            "mets:structMap[@TYPE='physical']//mets:fptr[@FILEID='{}']".format(
+                file_id
+            ),
+            namespaces=ns.NSMAP
+        )
```
jraddaoui, line 614: If the `..` syntax works, we could try to get the parent element in here ...

```
file_pointer_division = root.find(
    "mets:structMap[@TYPE='physical']//mets:fptr[@FILEID='{}']/..".format(
        file_id
    ),
    namespaces=ns.NSMAP
)
```

Otherwise, with the `FILEID` to `DMDID` mapping, we could avoid this query and pass directly to the one bellow to get the related `mets:dmdSec[@ID='{}']/mets:mdWrap/mets:xmlData` element.

src/archivematicaCommon/tests/fixtures/test_index_metadata-METS.xml
```diff
@@ -0,0 +1,6790 @@
...
+            <premis:relationship>
+              <premis:relationshipType>derivation</premis:relationshipType>
+              <premis:relationshipSubType>has source</premis:relationshipSubType>
+              <premis:relatedObjectIdentification>
+                <premis:relatedObjectIdentifierType>UUID</premis:relatedObjectIdentifierType>
+                <premis:relatedObjectIdentifierValue>040e2611-8704-4912-a5f9-d59ecf8699c9</premis:relatedObjectIdentifierValue>
+              </premis:relatedObjectIdentification>
+              <premis:relatedEventIdentification>
+                <premis:relatedEventIdentifierType>UUID</premis:relatedEventIdentifierType>
+                <premis:rel
```
jraddaoui, line 402: I don't like the idea of adding an entire METS file to the repository for the test. However, I see the point of having it as I think that test is a great addition. Could we reduce it for just this test case while it's still a valid METS file? Other ideas? @sevein?

jraddaoui (changes requested):
Hi @replaceafill, great work!

I've added a few comments bellow. Also, we're currently working in an upgrade to support ES 6.x in #1285 and this changes will conflict a little with that work. I'm not sure if we started the tests for that upgrade as we wanted to do it over the dev. branch before merging, but I can take care of the conflicts if this gets merged before the other PR or help you in here if it happens the other way around.

ross-spencer:
Hi @jraddaoui @replaceafill funnily enough, I just realized about the conflicts and started a re-based branch here as I thought you were still away: https://github.com/artefactual/archivematica/tree/dev/issue-1171-es-upgrade-rebased-spencer

@currmie and @scollazo (and @mamedin) may already have done the same but I need to check, otherwise the original branch is still behind `qa/1.x` about 21 commits. 

The new branch could be deployed and used to test, or if you wanted  to rebase yours it'll work automatically. Then we can cherry-pick @replaceafill's changes to test on as well. What do you think?

RE: METS - I made some cut down METS files for a PR yesterday. That's a good idea if possible. What I liked about this particular METS is that is aligns with the analyst's testing, and so it mirrors the bug and the fix 1:1.

src/archivematicaCommon/tests/test_elasticsearch_functions.py
replaceafill: Yes, this is something I always have trouble being consistent with. I prefer the single line approach because it makes a bit easier to spot changes in diffs. When you have multiple identifiers in a single line and you change or delete only one identifier you need to scan the whole line. I also add commas after the last parameters in case you add extra parameters later, again making the diff easier to read. Hopefully using `black` will put us all in sync :)

src/archivematicaCommon/tests/test_elasticsearch_functions.py
replaceafill: :+1: I'll change the code accordingly.

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: I tried extracting only the physical struct map but had namespace problems when I tried to search inside of it. It looked like the current namespaces mapping only work off the root element. I'll make the change again and show you.

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: Thanks! It worked like a charm.

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: No need for iterating the physical struct map since the `..` syntax worked!

src/archivematicaCommon/tests/fixtures/test_index_metadata-METS.xml
sevein: I think @ross-spencer took that exact same approach in https://github.com/artefactual/archivematica/pull/1332 and it worked great.

jraddaoui (approved):

jraddaoui:
Thanks @replaceafill!

Please, make sure to merge in a single commit.

PR #1331, opened 2019-01-13
Extract multiple Dataverse authors

ross-spencer:
This commit will see the extraction of multiple author entities from
the Dataverse metadata dataset.json file. Affiliation is also added
to the DDI segment. There is a small refactor to separate some of the
responsibilities of the code to better enable future refactoring.
Tests have been added to test the addition of multiple authors.

Connected to archivematica/issues#278

### DIFF
--- a/src/MCPClient/tests/test_create_dataverse_mets.py
+++ b/src/MCPClient/tests/test_create_dataverse_mets.py
@@ -19,70 +19,110 @@ sys.path.append(
 
 import convert_dataverse_structure
 
-# List of dataverse metadata fixtures. We use a namedtuple to provide some
+# List of Dataverse metadata fixtures. We use a named-tuple to provide some
 # structure to this index so that we can keep track of information regarding
-# dataverse over time, e.g. dataverse version number, the dataset uri and so
+# Dataverse over time, e.g. Dataverse version number, the dataset URI and so
 # forth.
-DataverseMDIndex = namedtuple('DataverseMDIndex',
-                              'dv_version, ddi_title, pid_value, pid_type, '
-                              'fname, all_file_count, dir_count, item_count, '
-                              'ddi_count')
+DataverseMDIndex = namedtuple(
+    'DataverseMDIndex', 'dv_version, ddi_title, pid_value, pid_type, '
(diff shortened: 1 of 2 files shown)


src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -211,3 +251,31 @@ def test_mets_content(self, fixture, tmpdir):
...
+        the list compares 1:1 in order.
+        """
+        mets_path = self._create_mets(fixture.fname, tmpdir)
+        if not os.path.isfile(mets_path):
+            pytest.fail("Fixtures were not previously setup correctly.")
+
+        try:
+            mets = metsrw.METSDocument.fromfile(mets_path)
+        except mets.MetsError:
+            pytest.fail("Could not parse mets {}".format(mets_path))
```
sevein, line 268: `metsrw.MetsError` would be the exception error. This is happening in other tests in this same file.

There is this tension between simpler tests and friendlier error messages. You could just let it raise, the developer is going to be focusing on the traceback anyways.

src/MCPClient/tests/test_create_dataverse_mets.py
```diff
@@ -211,3 +251,31 @@ def test_mets_content(self, fixture, tmpdir):
...
+
+        mets_root = mets.serialize()
+
+        # Setup namespaces to search for in the document.
```
sevein: This is a grammar question from a ESL out of curiosity, not really suggesting a change.
Shouldn't this read `Set up` (the verb) instead of `Setup`?

sevein (approved):
I've found what's maybe a small bug in `test_create_dataverse_mets` but I think it'll be an easy fix.

:+1:

src/MCPClient/tests/test_create_dataverse_mets.py
ross-spencer: For a native speaker, my English is pretty bad @sevein. It's more fluid than perhaps it should be for something with pretty clear rules and structures! (probably like some the code you see from me! :innocent: ) So this is a great question, and, today I discovered a Stack Overflow for the english language! https://english.stackexchange.com/a/3665 - and you are correct! :nerd_face:

PR #1332, opened 2019-01-13
Enable the deletion of packages in a Dataverse transfer

ross-spencer:
This commit makes sure that when Archivematica is used with `Extract Packages: On` and `Remove extracted packages: On` a Dataverse transfer will not fail, e.g. if it contains an `RData` file. 

A modest number of tests have been added where there was less coverage before. 

Logging has been cleaned up somewhat too. If something goes wrong, users should be able to get a clearer view of what is going on. 

Connected to archivematica/issues#269

### DIFF
--- a/src/MCPClient/tests/test_parse_dataverse.py
+++ b/src/MCPClient/tests/test_parse_dataverse.py
@@ -8,8 +8,8 @@ import sys
 
 from django.test import TestCase
 import metsrw
-from job import Job
 
+from job import Job
 from main import models
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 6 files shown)


src/MCPClient/tests/test_parse_dataverse.py
```diff
@@ -252,3 +265,33 @@ def test_validate_checksums(self):
         assert events.event_outcome == 'Pass'
         assert events.event_outcome_detail == \
             'Dataverse checksum 4ca2a78963445bce067e027e10394b61 verified'
+
+    def test_get_db_objecrs_returns(self):
```
sevein: Typo.

sevein (approved):

PR #1324, opened 2018-12-28
WIP: Replace hashdeep with coreutils hash utilities

ross-spencer:
:construction: :construction: :construction: :construction: :construction: :construction: 

Work to replace hashdeep in Archivematica. 

TODO:

* Push the docker-base image (currently a test one is on the server)
* Change the script name and module name (merge post i18n)
* Documentation about other hashes if we want them, e.g. including sha512
* Documentation that describes more specific requirements for the hash files supplied with Archivematica, i.e. current demo-transfer won't work with its current configuration.

Open questions:

* Do we still want the log output from the shell script ?
* It is easy  now to add other hashing algorithms, do we want to add sha512sum?
* What would automated testing look like?

Connected to archivematica/Issues#346
Connected to artefactual/archivematica#1061

Output currently looks as followss in the UI"

image

image

And in the PREMIS:
```
<premis:event xmlns:premis="info:lc/xmlns/premis-v2" xsi:schemaLocation="info:lc/xmlns/premis-v2 http://www.loc.gov/standards/premis/v2/premis-v2-2.xsd" version="2.2">
  <premis:eventIdentifier>
    <premis:eventIdentifierType>UUID</premis:eventIdentifierType>
    <premis:eventIdentifierValue>7be84e31-6f2c-41e6-966a-a7995a58470a</premis:eventIdentifierValue>
  </premis:eventIdentifier>
  <premis:eventType>fixity check</premis:eventType>
  <premis:eventDateTime>2018-12-30T17:04:06.276741+00:00</premis:eventDateTime>
  <premis:eventDetail>sha1sum (GNU coreutils) 8.28</premis:eventDetail>
  <premis:eventOutcomeInformation>
    <premis:eventOutcome>Pass</premis:eventOutcome>
    <premis:eventOutcomeDetail>
      <premis:eventOutcomeDetailNote></premis:eventOutcomeDetailNote>
    </premis:eventOutcomeDetail>
  </premis:eventOutcomeInformation>
  <premis:linkingAgentIdentifier>
    <premis:linkingAgentIdentifierType>preservation system</premis:linkingAgentIdentifierType>
    <premis:linkingAgentIdentifierValue>Archivematica-1.9</premis:linkingAgentIdentifierValue>
  </premis:linkingAgentIdentifier>
  <premis:linkingAgentIdentifier>
    <premis:linkingAgentIdentifierType>repository code</premis:linkingAgentIdentifierType>
[...]

Issue #1061, opened 2018-04-25
Problem: verify transfer checksums succeeding without verifying the transfer checksums


bb233:
System: Archivematica 1.6.1, Storage Service 0.10.1, on CentOS 7

What I am doing: 

Testing the ability of Archivematica to detect corrupted files in standard transfers, using the metadata/checksum.sha256 file in the transfer. Specifically, we are creating a checksum.sha256 file and editing the checksums in the file. We also inadvertently replaced the spaces separating checksum from filename with a tab.

What I expected: 

The verify transfer checksums microservice to fail with some sort of indication that there was an invalid checksum.

What I saw: 

The verify transfer checksums microservice to succeed without visible error, and the transfer to proceed.
Digging into the verify transfer checksum microservice and looking at the generated report for the "verify metatdata directory checksums" task, I found the following STDERR output:

> sha256deep: /var/archivematica/sharedDirectory/currentlyProcessing/bb_bad_checksum-055fa458-2fe6-4bed-a179-8868e63b907d/metadata/checksum.sha256: Unable to find any hashes in file, skipped.
> sha256deep: Unable to load any matching files.
> Try `sha256deep -h` for more information.
> sha256deep: /var/archivematica/sharedDirectory/currentlyProcessing/bb_bad_checksum-055fa458-2fe6-4bed-a179-8868e63b907d/metadata/checksum.sha256: Unable to find any hashes in file, skipped.
> sha256deep: Unable to load any matching files.
> Try `sha256deep -h` for more information.

What else I tried:

I get similar results when using md5 checksums with tab separators, when using UTF-16 encoding, when using UTF-8 with BOMs, and other cases where the format _looks_ right in a text editor

What I think is going on:

The issue seems to be that if sha256deep fails to parse the checksum file, because it is in an unsupported format, it reports "unable to find hashes, skipped", and Archivematica isn't seeing that as an error.

Why I think this is a problem:

The checksum files are supposed to ensure transfer fixity, and yet when they exist but are themselves too corrupted to read, Archivematica silently ignores them and claims to have verified fixity when it hasn't.

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Artefactual (internal)](https://trello.com/c/eCHD7L5W/14-artefactual-internal)

ross-spencer:
Marking as `Severity:High` for discussion about inclusion in `1.9`. (The ticket was already previously marked as `Severity:Critical`.

[...]

PR #1333, opened 2019-01-15
Replace hashdeep with hashsum commands

ross-spencer:
This commit provides the scaffolding to replace the current checksum
verification job in Archivematica with the hashsum suite of
utilities. Functionality should be nearly equivalent to that of the previous
job with additional tests where previously there were none.

* The biggest difference is that now there won't be a log file output as part of checksum comparison, but this can of course be added if the Job logging isn't enough.
* There would need to be a minor adjustment to the documentation and future transfer structures (n.b. demo transfers too) as the manifest needs to something like as follows (note we add `objects` in front of all paths):
```
c93db26f0a2a2c118278d3998c077dfa  objects/afile.txt
0bee89b07a248e27c83fc3d5951213c1  objects/nested/abc.md
b026324c6904b2a9cb4b88d6d61c81d1  objects/dupe/2.md
1b2297c171a91450d184871ccf6c9ad4  objects/nested/def.md
3e7d4585a1d42237fbbe16c6be1e1952  objects/bfile.txt
b026324c6904b2a9cb4b88d6d61c81d1  objects/dupe/1.md
```

If something does go wrong (or right) for a user, they may see a mixture of logging such as follows:
```
Standard output (stdout):
-------------------------
md5: Comparison was OK
sha256: Comparison was OK

Standard error (stderr):
------------------------
Comparing transfer checksums with the supplied sha512 file
sha512: Comparison failed with 1 checksum lines and 6 transfer files
Comparing transfer checksums with the supplied sha1 file
sha1: comparison exited with status: 1 check the file's formatting
sha1: objects/dupe/2.md: FAILED
sha1: objects/bfile.txt: FAILED
Comparing transfer checksums with the supplied md5 file
Comparing transfer checksums with the supplied sha256 file
```

Connected to archivematica/issues#346 
Connected to #1061 
Replaces https://github.com/artefactual/archivematica/pull/1324
Sample data for testing: https://github.com/artefactual/archivematica-sampledata/pull/54


src/MCPClient/lib/clientScripts/verify_md5.py
```diff
@@ -1,20 +1,186 @@
...
+    try:
+        transfer_dir = job.args[1]
+        event_id = job.args[0]
+        transfer_uuid = job.args[0]
+    except IndexError:
+        logger.error("Cannot access expected module arguments: %s", job.args)
+        return 1
+    ret = 0
+    for hashfile in Hashsum.COMMANDS:
+        os.chdir(transfer_dir)
```
sevein: I wouldn't use `chdir` because it can have unintended consequences since it's a process-level state change. You may want to look into pathlib (pathlib2 is a backport compatible with py27).

sevein:
@ross-spencer, does this replace https://github.com/artefactual/archivematica/pull/1324?

FYI: I've added `triage-release-1.10` to https://github.com/artefactual/archivematica/issues/1061 and https://github.com/archivematica/Issues/issues/346.

ross-spencer:
@sevein it will do, yep :+1: Thanks for adding the triage label.

src/MCPClient/lib/clientScripts/verify_md5.py
ross-spencer: @sevein I had a look but this library didn't have anything equivalent. If you do take another look at this PR and this is a problem, we'll need to take a bit more time to find an alternative, possibly even be more verbose in the args we're passing the hash tools.

sevein: Ok, that sounds like a plan. Path objects are nice in that they have a nice syntax and it's the idiomatic way in Python 3.

src/MCPClient/lib/archivematicaClientModules
```diff
@@ -114,7 +114,7 @@ createsipsfromtrimtransfercontainers_v0.0 = create_sips_from_trim_transfer_conta
 determineaipversionkeyexitcode_v0.0 = determine_aip_version_key_exit_code
 dipGenerationHelper = dip_generation_helper
 emailfailreport_v0.0 = email_fail_report
-verifymd5_v0.0 = verify_md5
+verifyChecksum_v1.0 = verify_checksum
```
replaceafill: Nitpick: I guess we're not very consistent separating terms here: copyThumbnailsToDIPDirectory (camelCase), upload-archivesspace (dashes). But the majority are multiple lowercase terms, so maybe we could go with `verifychecksum`?

Question: could you explain your rationale for setting this as version `1.0`? I don't understand the `#the version numbers are versions of the protocol, not versions of the programs themselves.` line at the top of this file quite well.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
+# -*- coding: utf-8 -*-
+
+"""Verify Checksum Job
+
+Archivematica wraps the hashsum utility to verify checksums provided to the
+system as part of a transfer, e.g. checksum.md5 in the transfer metadata
+folder.
+"""
```
replaceafill: Not sure if this is documented but license before module docstring seems more common.

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
+# -*- coding: utf-8 -*-
+
+"""Test Verify Checksum Job in Archivematica.
+
+Tests for the verify checksum Job in Archivematica which makes calls out to the
+hashsum checksum utilities. We need to ensure that the output of the tool is
+mapped consistently to something that can be understood by users when
+debugging their preservation workflow.
+"""
```
replaceafill: Not sure if this is documented but license before module docstring seems more common.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+from custom_handlers import get_script_logger
+from databaseFunctions import insertIntoEvents
+
+
+def concurrent_instances():
+    """Function docstring. N.B. Delete me."""
+    return 1
+
+
+logger = get_script_logger("archivematica.mcp.client.compare_hashes")
```
replaceafill: It seems like most of the loggers use the name of the script in some form for the logger: `archivematica.mcp.client.verify_checksum` or `archivematica.mcp.client.verifyChecksum`

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class NoHashCommandAvailable(Exception):
+    """Provide feedback to the user if the checksum command cannot be found
+    for the provided checksum file.
+    """
+
+
+class PREMISFailure(Exception):
```
replaceafill, line 60: I noticed this is raised in two different places so maybe it needs some context message.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+    """Provide feedback to the user if there is a problem writing PREMIS event
+    information to the database.
+    """
+
+
+class Hashsum(object):
+    """Class to capture various functions around calling Hashsum as a mechanism
```
replaceafill, line 67: Is the `Hashsum` you refer in this docstring a different component or package? Or does it mean the class itself?

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+    transfer_dir = None
+    transfer_uuid = None
+    try:
+        transfer_dir = job.args[1]
+        transfer_uuid = job.args[2]
+    except IndexError:
+        logger.error("Cannot access expected module arguments: %s", job.args)
+        return 1
+    ret = 0
+    for hashfile in Hashsum.COMMANDS:
```
replaceafill: Can't think of a better name alternative for `COMMANDS` but this iteration seems "off?" (sorry can't think of a better word either!), like "shouldn't this be iterating through commands?".

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+
+    OKAY_STRING = ": OK"
+    FAIL_STRING = ": FAILED"
+    ZERO_STRING = "no properly formatted"
+    IMPROPER_STRING = "improperly formatted"
+    FAILED_OPEN = ": FAILED open or read"
+    EXIT_NON_ZERO = "returned non-zero exit status 1"
+
+    def __init__(self, path, job=print):
+        """Class constructor."""
```
replaceafill: I don't see much value in this docstring.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+            raise NoHashCommandAvailable()
+
+    def _call(self, *args):
+        """Make the call to Python subprocess and record the command being
+        called.
+        """
+        self.command_called = (self.COMMAND,) + args
+        return self._decode(subprocess.check_output(self.command_called))
+
+    def count_and_compare_lines(self, hashfile, objectsdir):
```
replaceafill: If I understand the code paths correctly, `hashfile` here is the same `path` the constructor received? If that's the case why is it necessary to pass it again?

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        if lines == objects:
+            return True
+        self.job.pyprint(
+            "{}: Comparison failed with {} checksum lines and {} "
+            "transfer files".format(self.get_ext(hashfile), lines, objects),
+            file=sys.stderr,
+        )
+        return False
+
+    def compare_hashes(self, hashfile, objectsdir):
```
replaceafill: If I understand the code paths correctly, `hashfile` here is the same `path` the constructor received? If that's the case why is it necessary to pass it again?

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+                    continue
+                if line.endswith(self.OKAY_STRING):
+                    continue
+                if (
+                    line.endswith(self.FAIL_STRING)
+                    or self.ZERO_STRING in line
+                    or self.IMPROPER_STRING in line
+                ):
+                    self.job.pyprint(
+                        "{}: {}".format(self.get_ext(hashfile), line), file=sys.stderr
```
replaceafill: I haven't run the script so I don't know how these conditions work together, but if the `EXIT_NON_ZERO` warning doesn't always come before a message like this it could be difficult to track the issue back to the `verify_checksum` script just from the logs. Maybe it's worth adding some more context to this message.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+                    )
+                if line.endswith(self.FAILED_OPEN):
+                    self.job.pyprint(
+                        "{}: {}".format(self.get_ext(hashfile), line), file=sys.stderr
```
replaceafill: Same as previous comment.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        try:
+            return self._call("--version")[0]
+        except subprocess.CalledProcessError:
+            return self.COMMAND
+
+    def get_command_detail(self):
+        """Provide some way for the user to get information out of the class to
+        write METS/PREMIS provenance information.
+        """
+        if not self.command_called:
```
replaceafill, line 173: What's the code path for this condition to succeed? I noticed `get_command_detail` is only called after checking the result status is 0 so the command should have been saved.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        """Decode string output in Py2 or Py3 and return a list of lines to be
+        parsed elsewhere.
+        """
+        try:
+            return str(out, "utf8").split("\n")
+        except TypeError:
+            return out.decode("utf8").split("\n")
+
+    @staticmethod
+    def get_ext(path):
```
replaceafill, line 191: If I understand the code paths correctly, `path` here is the same `path` the constructor received? If that's the case why is it necessary to pass it again and have this to be a `staticmethod`?

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        """Return the extension of the checksum file provided"""
+        ext = os.path.splitext(path)[1]
+        if not ext:
+            return path
+        return ext.replace(".", "")
+
+    @staticmethod
+    def _count_lines(path):
```
replaceafill, line 199: Same as previous comment.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        """Count the number of lines in a checksum file."""
+        count = 0
+        with open(path) as hashfile:
+            for count, _ in enumerate(hashfile):
+                pass
+        # Negate zero-based count.
+        return count + 1
+
+    @staticmethod
+    def _count_files(path):
```
replaceafill, line 209: Same as previous comment.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+            insertIntoEvents(
+                fileUUID=fileUUID,
+                eventType=event_type,
+                eventDateTime=datetime.datetime.now()
+                .replace(microsecond=0)
+                .isoformat(),
+                eventDetail=event_detail,
+                eventOutcome=event_outcome,
+                eventOutcomeDetailNote="",
+            )
```
replaceafill, line 245: I noticed this logic is very similar to the `createEventsForGroup.py` script. Is there a reason why that can't be adapted or reused? Or maybe it should be removed if it's not needed anymore?

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+    transfer_uuid = None
+    try:
+        transfer_dir = job.args[1]
+        transfer_uuid = job.args[2]
+    except IndexError:
+        logger.error("Cannot access expected module arguments: %s", job.args)
+        return 1
+    ret = 0
+    for hashfile in Hashsum.COMMANDS:
+        os.chdir(transfer_dir)
```
replaceafill: Why can't the logic of the script work off the `transfer_dir` joining absolute paths? I also noticed @sevein [raised this as an issue](https://github.com/artefactual/archivematica/pull/1333#discussion_r253208026).

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+        objectsdir = "objects/"
+        hashsum = None
+        if os.path.exists(hashfile):
+            try:
+                hashsum = Hashsum(hashfile, job)
+            except NoHashCommandAvailable:
+                job.pyprint(
```
replaceafill, line 273: Even if this doesn't stops the execution of the script, should this be redirected to `file=sys.stderr`? I noticed other "warnings" doing it.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,276 @@
...
+            job.pyprint(
+                "Comparing transfer checksums with the supplied {} file".format(
+                    Hashsum.get_ext(hashfile)
+                ),
+                file=sys.stderr,
+            )
+            result = hashsum.compare_hashes(hashfile=hashfile, objectsdir=objectsdir)
+            # Add to PREMIS on success only.
+            if result == 0:
+                job.pyprint("{}: Comparison was OK".format(Hashsum.get_ext(hashfile)))
```
replaceafill, line 289: Should this be redirected to `file=sys.stderr` too?

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def test_invalid_initialisation(self):
+        """Test that we don't return a Hashsum object if there isn't a tool
+        configured to work with the file path provided.
+        """
+        try:
+            Hashsum("checksum.invalid_hash")
+            assert False
+        except NoHashCommandAvailable:
+            assert True
```
replaceafill: Not sure if our pytest version supports it but this could be more clear using [pytest.raises](https://docs.pytest.org/en/3.0.3/assert.html#assertions-about-expected-exceptions)

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
...
+            ("metadata/checksum.sha512", None),
+            ("metadata/checksum.b2", None),
+            ("metadata/checksum_md5", False),
+            ("metadata/checksum_sha1", False),
+            ("metadata/checksum_sha256", False),
+            ("metadata/checksum_sha512", False),
+            ("metadata/checksum_b2", False),
+        ],
+    )
+    def test_valid_init(self, fixture):
```
replaceafill: `test_valid_initialisation` to match the previous (invalid) method?

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
...
+            b"objects/file1.bin: OK\n"
+            b"objects/file2.bin: FAILED\n"
+            b"objects/nested/\xe3\x83\x95\xe3\x82\xa1\xe3\x82\xa4\xe3\x83\xab"
+            b"3.bin: FAILED\n"
+            b"objects/readonly.file: FAILED open or read"
+        )
+        exception_string = (
+            "sha256: comparison exited with status: 1. Please check the formatting of the checksums or integrity of the files.\n"
+            "sha256: objects/file2.bin: FAILED\n"
+            "sha256: objects/nested/ファイル3.bin: FAILED\n"
```
replaceafill, line 162: Nice!

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
...
+            returncode=1, cmd=toolname, output=output_string
+        )
+        ret = hashsum.compare_hashes(hash_file, "")
+        mock.assert_called_once_with("-c", "--strict", hash_file)
+        assert ret == 1
+        print(job.get_stderr().decode("utf8").strip())
+        print(exception_string)
+        assert job.get_stderr().decode("utf8").strip() == exception_string
+
+    def test_compare_hashes_w_bad_files(self, mocker):
```
replaceafill: Why not use `_with_`?

src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -0,0 +1,285 @@
...
+        """Test that the microservice job connects to the database as
+        anticipated, writes its data, and that data can then be retrieved.
+        """
+        # Values the job writes.
+        group_type = "transfer_id"
+        event_type = "fixity_check"
+        event_outcome = "pass"
+        # Values we will write.
+        package_uuid = "e95ab50f-9c84-45d5-a3ca-1b0b3f58d9b6"
+        event_detail = "suma de verificación validada: OK"
```
replaceafill: What's the black magic that makes this localized string possible?!?! :)

replaceafill (changes requested):
Nice work @ross-spencer! I left a few questions and learned a lot from reviewing it :)

src/MCPClient/lib/archivematicaClientModules
ross-spencer: I'll go with lower-case :+1:. The version numbering I took a leaf out of Joel D's book, e.g. https://github.com/artefactual/archivematica/commit/4045ff387ba1aadfc72c0b3df198d687df29f561#diff-044d43ce529873bb3a061d1c782a04cf so it's just a second major version of a client script is my understanding, but it probably doesn't tell us very much, and I've broken that connection a little too by changing the name of the script as well...

src/MCPClient/tests/test_verify_checksum.py
ross-spencer: Ahh! None! You're right. But I want to keep on top of Unicode strings wherever possible. Who knows, we might write this out one day to the event details! :nerd_face: 

(But yep, really just pushing the database functions a little here so hope that's okay)

ross-spencer:
Thanks for your time on this @replaceafill I think it is ready for re-review. 

I might have committed a CR sin - i have added additional statements to the unit tests (you'll see) so that my asserts output something useful if they fail. (One of my favorite things!). I hope that they a) look okay, b) it's not too much to add to the review. I also removed some errant logging from the unit tests too. 

Let me know, and I'll get any follow up changes in asap.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I've updated the `docstring` do you think it is enough to make it clearer to the reader?
```
"""Verify Checksum Job

Archivematica wraps the hashsum utility to verify checksums provided to the
system as part of a transfer, e.g. checksum.md5 in the transfer metadata
folder. We wrap this coreutils utility by using a Hashsum class which enables
us to easily call each of the tools packaged against its different algorithms:

    * MD5
    * SHA1
    * SHA256
    * SHA512
    * Blake2
"""
```

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I see what you mean. I wrestled with "COMMANDS" "ALGORITHMS" but I have settled on "HASHFILES_COMMANDS" with an additional comment associated with the data structure it is referring to. It borrows a bit from database lookup table naming to say it is a set of key-values, hash-files and commands? Will that work do you think?

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: Hmm, that's a good point. I think there was a reasonable origin for this, where you could create an object of type, `md5sum` maybe, and allowing other files to be validated against it, but it doesn't make sense in this context. And I didn't commit to that design. I will remove the arg and setup the attribute in the constructor. :+1:

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I might not quite understand but we can talk more. Part of what I was hoping to achieve here was a balance in user-friendly output vs. logging output. Without giving the user too much information to have to dive into as well. 

I have created some [sample data](https://github.com/artefactual/archivematica-sampledata/tree/ed7fc9ae8361308f3215cbd32f970bea8dd58509/training-resources/fixityCheckShouldFail) to specifically test failures, the output on `stderr` for each of the checksums for different failures will look as follows:
```
Comparing transfer checksums with the supplied sha512 file
sha512: comparison exited with status: 1. Please check the formatting of the checksums or integrity of the files.
Comparing transfer checksums with the supplied sha1 file
sha1: Comparison failed with 2 checksum lines and 3 transfer files
Comparing transfer checksums with the supplied b2 file
b2: comparison exited with status: 1. Please check the formatting of the checksums or integrity of the files.
b2: objects/Nemastylis_geminiflora_Flower.PNG: FAILED
b2: objects/oakland03.jp2: FAILED
Comparing transfer checksums with the supplied md5 file
md5: comparison exited with status: 1. Please check the formatting of the checksums or integrity of the files.
md5: objects/Nemastylis_geminiflora_Flower.PNG: FAILED
md5: objects/oakland03.jp2: FAILED
md5: WFPC01.GIF: FAILED open or read
Comparing transfer checksums with the supplied sha256 file
sha256: Comparison failed with 1 checksum lines and 3 transfer files
```
So on the one hand I think we've covered a lot. But it would be good to understand a bit more about the context that might be possible to add?

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: It's just gold-plating. We've given the caller an object to work with and one of the things the caller can do with that object is call this function independently of doing any actual hash comparison and so `self.command_called` will still be `None`. I've added context to the exception, and don't believe it's doing too much harm. But I can remove it?

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: In this instance I'm just using them as helper functions so I call them within the class and outside of the class. So take for example this output I generate:
```python
        if os.path.exists(hashfile):
            try:
                hashsum = Hashsum(hashfile, job)
            except NoHashCommandAvailable:
                job.pyprint(
                    "Nothing to do for {}. No command available.".format(
                        Hashsum.get_ext(hashfile)
                    )
                )
                continue
```
If I can't initialize the object, I'd still like to return some information about what I was doing, so do it via this helper. I guess it's a bit of an unholy mix of OO and something else really? (Archivematica Microservice Style? :wink: ) Although they are independent in their own way and are hidden by the implementation. I can rework it if there's a strong opinion either way.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I'm not sure, my intent was to keep the output as user friendly/intuitive as possible. We'll end up with something along the lines of:
image

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: A fair question. I pretty much created this as a nice to know, but nothing to worry about so while the code is calling an exception, it's not really anything I would ask the user to worry about. It's unlikely that it will ever be triggered. The script is just being made as robust as possible. The output will look as follows:

image

I feel like I may have confused components of this PR by designing something to the way the user interface acts, vs. what we might expect as technicians. I'd definitely like to discuss this more. It would be great to see some questions like this fall out of user-acceptance-testing as well.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: Good catch. This was on my list of [todos](https://github.com/artefactual/archivematica/pull/1324#issuecomment-451204944). I will remove it.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I might be missing something, and I probably wasn't clear enough in my response to Sevein, but as far as my reading and tests have gone, as far as I can work out, the check function of the tool we are looking to replace the `md5deep`-like commands with functions relative to the digest listing. 

This doesn't give us many options:

* Users create an absolute path up front.
* We run the tool in the script as we are.
* We manipulate paths in the digest and find a way of comparing an absolute path with a checksum? (this looks like our own custom checksum comparison script at this point, or finding another tool)
* Find a way of running the tool from a staging location? 

The paths we have in the microservice are:

* The path the script is executed from (`cwd`): `/`
* The path we need to be on to run the `--check` comparison: `/var/archivematica/sharedDirectory/currentlyProcessing/cs1-456358ec-5c28-45d7-842c-7e9c2cca6cfa/`

And trying this with our sample data when the objects are expected to be at: `/home/ross-spencer/git/artefactual-labs/am/src/archivematica-sampledata/TestTransfers/fixityCheckShouldFail/objects`:

```
ross-spencer@artefactual:~$ md5sum -c /home/ross-spencer/git/artefactual-labs/am/src/archivematica-sampledata/TestTransfers/fixityCheckShouldFail/metadata/checksum.md5
md5sum: ./WFPC01.GIF: No such file or directory
./WFPC01.GIF: FAILED open or read
md5sum: ./oakland03.jp2: No such file or directory
./oakland03.jp2: FAILED open or read
md5sum: ./Nemastylis_geminiflora_Flower.PNG: No such file or directory
./Nemastylis_geminiflora_Flower.PNG: FAILED open or read
md5sum: WARNING: 3 listed files could not be read
```

As I noted to Sevein, if this is a blocker we need to find an alternative. And/or:

* Figure out how to use the tool the way that works better for us. 
* Understand and prove the risk that might be caused by changing the directory and setting it back (some additional information on this would help me). 
* Understand what other alternatives we have for this function in Archivematica.

sevein: > Understand and prove the risk that might be caused by changing the directory and setting it back (some additional information on this would help me).

I think that there's very little risk today because client scripts are executed in single-threaded context since AM 1.8. However I think it's better to write thread-safe code where possible because that doesn't lock us in a particular execution model. E.g. if we started using threads again, context switches will happen and the use of `chdir` will likely cause race conditions.

I'm not able to prove it formally but here's a very quick example that illustrates the issue:

```python
>>> import os
>>> import threading
>>> 
>>> def change_chdir():
...     os.chdir("/etc")
... 
>>> os.chdir("/")
>>> threading.Thread(target=change_chdir).start()
>>> print(os.getcwd())
/etc
```

This example is non-deterministic as the threads are going to be scheduled by the operative system. I think that makes clearer that the context in which your code is executed is unpredictable.

> Understand what other alternatives we have for this function in Archivematica.

If all you care about is the state of the child process, you could try passing `cwd` to `subprocess.check_output`.

https://docs.python.org/2/library/subprocess.html#subprocess.Popen
> If `cwd` is not `None`, the child’s current directory will be changed to `cwd` before it is executed. Note that this directory is not considered when searching the executable, so you can’t specify the program’s path relative to `cwd`.

replaceafill (approved):
Thank you for your follow up comments @ross-spencer. After testing the script in my dev instance most of my questions about error output are now clear. The new `fixityCheckShouldFail` test transfer was very helpful to understand the overall functionality. Also, now I understand why it's not possible to use absolute paths when checking the checksums.

ross-spencer:
Thanks @replaceafill! Nearly there! 

I've rebased the work so it is ready to go but will try and address @sevein's comments before we make a final decision with everyone about merging this work.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,291 @@
...
+
+def write_premis_event_per_file(transfer_uuid, event_detail):
+    """Write a PREMIS event for every file in the transfer on success."""
+    group_type = "transfer_id"
+    event_type = "fixity_check"
+    event_outcome = "pass"
+    # Retrieve file UUIDs from the database.
+    kwargs = {"removedtime__isnull": True, group_type: transfer_uuid}
+    with transaction.atomic():
+        file_uuids = File.objects.filter(**kwargs).values_list("uuid")
```
cole: Since this function is called once per checksum type (so 5x), this is going to do `SELECT * FROM File WHERE...` that many times. I think some issues that have come up recently show that we need to be thinking about the case where there are 50k files, which is where extra queries like this can get a bit slow.

I think there are at least two ways to get this to one single select:

1. Pass a queryset of file objects into write_premis_event_per_file, and construct it before calling (outside the loop). This seems like a relatively quick fix that might make it into this PR.
2. Change the job structure so that it iterates over files, and writes multiple hash events per file, rather than iterating over hashes and the files per hash. I'm guessing that is probably too broad a change for this stage in the PR.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,291 @@
...
+        if not file_uuids:
+            err_str = "Unable to find the transfer objects for the SIP: '{}' in the database".format(
+                transfer_uuid
+            )
+            raise PREMISFailure(err_str)
+        for (fileUUID,) in file_uuids:
```
cole: I was going to suggest `file_uuids.iterator()` here ([iterator docs](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#iterator)) to avoid loading all results into memory _but_ that doesn't make sense in combination with the suggestion to pass in a queryset to the function above. If we're not going to pass in a queryset that can then be cached (so with the function as is), I think iterator should improve memory usage and therefor performance here (though maybe not by much).

cole (commented):
I have some efficiency related points, based on the work we've done in sanitize names, etc, to optimize for case where we have a _lot_ of files. I realize it's pretty late in the PR process to bring up things like this, so if you want to defer for possible later fix (or to wait and see if performance is actually a problem), that's fine with me.

src/MCPClient/lib/clientScripts/verify_checksum.py
```diff
@@ -0,0 +1,291 @@
...
+def write_premis_event_per_file(transfer_uuid, event_detail):
+    """Write a PREMIS event for every file in the transfer on success."""
+    group_type = "transfer_id"
+    event_type = "fixity_check"
+    event_outcome = "pass"
+    # Retrieve file UUIDs from the database.
+    kwargs = {"removedtime__isnull": True, group_type: transfer_uuid}
+    with transaction.atomic():
+        file_uuids = File.objects.filter(**kwargs).values_list("uuid")
+        if not file_uuids:
```
cole: Oops, I missed a comment, sorry. If you do go with `iterator` below, this should be changed to use ~`file_uuids.count()`~ `file_uuids.exists()`.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: Thanks @sevein that was really helpful. I've made the `check_output` change here; https://github.com/artefactual/archivematica/pull/1333/commits/886d2c356e634cc44c07843e6514eaadd64be20b hopefully that's better for this microservice.

src/MCPClient/lib/clientScripts/verify_checksum.py
ross-spencer: I went with 1! https://github.com/artefactual/archivematica/pull/1333/commits/502a3c489cd1965600191936e81e35675720993b Hopefully that looks okay. 2. would be a bigger conversation (a pretty different microservice, but could definitely be a good one to discuss in future).

ross-spencer:
Hi @replaceafill I am hoping you can review the last two commits for me? I have addressed @sevein's and @cole's replacements independently in each. :crossed_fingers: it all looks okay.

replaceafill:
@ross-spencer I tested this again using the DemoTransferCSV from https://github.com/artefactual/archivematica-sampledata/pull/54 and the last commit keeps failing like this (tested after flushing my dev environment and rebuilding the containers):

Captura de pantalla de 2019-06-24 23-04-07

If I set your branch to the previous commit 886d2c356e634cc44c07843e6514eaadd64be20b and flush my dev environment the checksums microservice works again:

Captura de pantalla de 2019-06-24 23-09-24

ross-spencer:
@replaceafill love that error error message! :wink: And also sorry about that. I have added a commit against this to fix it. I was passing a dict as `kwargs` but not indicating they were indeed `kwargs` by using `**` I have:

* Added two new tests to show it working and not working, i.e. raising the PREMIS exception.
* I have moved the try/except block around to make it more logical. 
* I have tested against against demo transfer csv now too (the problem I ran into was testing against a set that fails where we don't write to the database).

replaceafill (approved):
LGTM! And it works now :+1:

ross-spencer:
> LGTM! And it works now +1

I always try for one of the two :laughing: Thanks @replaceafill! :tada:

PR #1335, opened 2019-01-28
Fix pip dependency installation issue (brotli)

sevein:
Recent versions of pip couldn't get brotli to build.
Updating brotli to the latest release fixed the problem.

This is connected to https://github.com/archivematica/Issues/issues/455.


sevein:
setuptools fixed the problem upstream, see https://github.com/archivematica/Issues/issues/455 for more details.

PR #1336, opened 2019-01-28
Remove query/filters from elasticSearchFunctions

jraddaoui:
The match all query and backlog filters attributes are prone to be
modified externally causing issues when they're obtained after for other
queries.

Connects to https://github.com/archivematica/Issues/issues/452.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -49,41 +49,6 @@ from elasticsearch import Elasticsearch, ImproperlyConfigured
 
 logger = logging.getLogger('archivematica.common')
 
-MATCH_ALL_QUERY = {
-    "query": {
-        "match_all": {}
-    }
-}
-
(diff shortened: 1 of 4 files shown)


ross-spencer (approved):
LGTM! :+1: 

I thought this was an interesting issue. Do you think the other values in `elasticsearchfunctions` are likely to be impacted in a similar way? e.g. `BACKLOG_FILTER` 

I notice this isn't building in Python 2.7 but that looks to be because of the recent updates also being discussed in the Archivematica channel.

jraddaoui:
The tests are failing due to a dependency installation issue, which I think @sevein is addressing in here alongside https://github.com/artefactual/archivematica-storage-service/pull/437.

jraddaoui:
Good point @ross-spencer! The backlog filters are not being modified anywhere, but they're also prone to the same issue in the future and, considering their use, it's a good idea to remove them in here too.

jraddaoui:
Back to you @ross-spencer. Sorry, I had to amend the commit.

ross-spencer (approved):
¡Perfecto! :+1:

PR #1340, opened 2019-02-05
Dashboard: update get_as_system_client

sevein:
`host` was removed, use `base_url` instead.

This connects to https://github.com/archivematica/Issues/issues/483.


hakamine (approved):
Thank you @sevein. I tested on the am19xenial.qa server and it's working (no errors, and able to connect to the AS server instance)

PR #1342, opened 2019-02-07
Workflow: introduce default IDCommand

sevein:
In this commit, `fpr.IDCommand` is changed so only one entry can be enabled at
a time. The following workflow links have been updated so they don't list a
hard-coded set of tools but a simple Yes/No question instead:

- Perform file format identification (Transfer)
  `f09847c2-ee51-429a-9478-a860477f6b8d`
  Question: Do you want to perform file format identification?
  Responses: Yes, No
- Perform file format identification (Ingest)
  `7a024896-c4f7-4808-a240-44c87c762bc5`
  Question: Do you want to perform file format identification?
  Responses: Yes, No
- Perform file format identification (Submission documentation & metadata)
 `087d27be-c719-47d8-9bbb-9a7d8b609c44`
  Question: Do you want to perform file format identification?
  Responses: Yes, No

The client script `identify_file_format` performs file format identification
using the command currently enabled in FPR.

This is connected to https://github.com/archivematica/Issues/issues/485.
This is connected to https://github.com/archivematica/Issues/issues/212.

---

There are two extra commits solving other issues found:
- **Fix variable shadowing issue** - the processing config page wasn't matching all the choices in the workflow (e.g. when choosing Normalize for access, the XML should list two entries)
- **MCPServer: support Django lang code format** - show workflow translations when choosing languages like `pt-br` (internally we use the `pt_BR` format hence the discrepancy)


src/MCPClient/lib/clientScripts/identify_file_format.py
```diff
@@ -159,7 +170,13 @@ def main(job, command_uuid, file_path, file_uuid, disable_reidentify):
 
 def call(jobs):
     parser = argparse.ArgumentParser(description='Identify file formats.')
+
+    # Since AM19 the accepted values are "True" or "False" since the ability to
+    # choose the command from the workflow has been removed. Instead, this
+    # script will look up in FPR what's the preferred command.
+    # This argument may be renamed later.
     parser.add_argument('idcommand', type=str, help='%IDCommand%')
```
sevein, line 181: I've decided to repurpose `%IDCommand%` because I want to minimize the number of changes since we're pretty close to the release date. The values expected now are "True" and "False" - determined in the workflow data (`workflow.json`).

src/dashboard/src/fpr/views.py
```diff
@@ -375,17 +374,13 @@ def idcommand_delete(request, uuid):
         if 'disable' in request.POST:
             command.enabled = False
             messages.info(request, _('Disabled.'))
-            for obj in dependent_objects:
-                obj['value'].enabled = False
-                obj['value'].save()
```
sevein, line 380: This was disabling `IDRule` objects when the associated `IDCommand` was disabled.

ablwr:
Technically this new workflow works fine, and I am able to update the default (Siegfried) to FIDO or another option. I think this is definitely cleaner from a workflow perspective, but I have a lot of concerns from a UX perspective, mostly around just cleaning up some of our previous trash we left behind. I think this requires an additional quick conversation with @sromkey as Product Owner. From a technical point, though, the code is fine and the workflow is fine too, and an upgrade works (when done properly -- as I found out).

ablwr:
Gracias jajaja!

ablwr (approved):

sromkey:
Hey @sevein , just a question, in Ingest if you choose No for file ID, does it use the existing data from Transfer or do you just not end up with file identification? It would be desirable to maintain the option to ID in Transfer and not have to re-perform in Ingest.

sevein:
> Hey @sevein , just a question, in Ingest if you choose No for file ID, does it use the existing data from Transfer or do you just not end up with file identification?

"No" in ingest is equivalent to the old "Skip file identification" where the file ID data was reused. I've just renamed it. So I suspect it's the former: existing data is used.

> It would be desirable to maintain the option to ID in Transfer and not have to re-perform in Ingest.

Ok, I suggest to confirm that's the actual behaviour but I haven't really changed that part of the code.

I'm wondering if that affects the defaults in the processing config. In both `default` and `automated` I've set it as:

image

Would you like to skip file ID in Ingest?

sromkey:
There wasn't previously a "skip" option in ingest so I'm not sure what the behaviour will be... when she's back online @ablwr probably won't mind testing for us since she has it running locally!

If our suspicion is correct, and the existing data is used if "No" in ingest, then I would suggest that "No" be the ingest default.

sevein:
Right, the option in Ingest was "Use existing data".  I can confirm that's still the behaviour when you choose "No". Should maybe "No" be renamed as "No, use existing data"?

sevein:
To be clear,

image

sromkey:
perfecto @sevein !

PR #1346, opened 2019-02-12
Zip / deflate support for AIP compression

helenst:
Provide zip/deflate capability in the AIP compression option

This connects to https://github.com/archivematica/Issues/issues/493.

### DIFF
--- a/src/MCPClient-base.Dockerfile
+++ b/src/MCPClient-base.Dockerfile
@@ -60,6 +60,7 @@ RUN set -ex \
 		ufraw \
 		unrar-free \
 		uuid \
+		unzip \
 	&& rm -rf /var/lib/apt/lists/*
 
 # Download ClamAV virus signatures
--- a/src/MCPClient/lib/clientScripts/compress_aip.py
+++ b/src/MCPClient/lib/clientScripts/compress_aip.py
@@ -92,7 +92,19 @@ def compress_aip(
             'algorithm="{}"\; '
             'version="$((pbzip2 -V) 2>&1)"'.format(compression_algorithm)
         )
-
+    elif program == "zip":
+        compressed_location = uncompressed_location + ".zip"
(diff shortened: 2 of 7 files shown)


src/MCPClient/lib/clientScripts/compress_aip.py
```diff
@@ -69,6 +69,17 @@ def compress_aip(job, compression, compression_level, sip_directory, sip_name, s
             'echo program="7z"\; '
             'algorithm="{}"\; '
             'version="`7z | grep Version`"'.format(compression_algorithm))
+    elif program == 'zip':
+        compressed_location = uncompressed_location + ".zip"
+        command = '/usr/bin/7z a -bd -tzip -y -mm={algorithm} -mx={level} -mtc=on -mmt=on "{compressed_location}" "{uncompressed_location}"'.format(
```
alexwlchan, line 97: At a quick glance, it looks like `uncompressed_location` is built from the user-provided name of the SIP, which mean something suitably annoying could escape this quoting.

Maybe `shlex.quote()` could be a safer way to escape those filenames?

(But I acknowledge that breaks the convention established in the rest of the file, so feel free to ignore me!)

src/MCPClient-base.Dockerfile
```diff
@@ -62,6 +62,7 @@ RUN set -ex \
 		ufraw \
 		unrar-free \
 		uuid \
+		unzip \
```
sevein, line 63: Is this needed by 7z?

helenst: It's needed by `atool` when it tries to decompress during AIP verification.

sevein: Oh I had no idea, thanks!

helenst:
Closing this since the related issue was closed.

PR #1347, opened 2019-02-12
Write external identifier to bag-info.txt

helenst:
Since the bag tool doesn't support specifying this field explicitly, it gets written to a `bag-info.txt` in a temporary directory which is passed in on the command line.

This connects to https://github.com/archivematica/Issues/issues/492.

### DIFF
--- a/src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
+++ b/src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
@@ -22,7 +22,9 @@
 
 import argparse
 import os
+import shutil
 import sys
+import tempfile
 
 import django
 django.setup()
@@ -83,7 +85,8 @@ def create_directories(base_dir, dir_list):
             pass
 
 
-def bag_with_empty_directories(job, operation, destination, sip_directory, payload_entries, writer, algorithm):
+def bag_with_empty_directories(job, operation, destination, sip_directory,
+                               payload_entries, writer, algorithm, sip_uuid):
     """ Run bagit create bag command, and create any empty directories from the SIP. """
     # Get list of directories in SIP
     dir_list = get_sip_directories(job, sip_directory)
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
```diff
@@ -83,7 +85,7 @@ def create_directories(base_dir, dir_list):
             pass
 
 
-def bag_with_empty_directories(job, operation, destination, sip_directory, payload_entries, writer, algorithm):
+def bag_with_empty_directories(job, operation, destination, sip_directory, payload_entries, writer, algorithm, sip_uuid):
```
ross-spencer: It might be a good opportunity to bring this line here, and `139` below into line with PEP8 style guidelines (lines > 79 characters), given that we're adding to it here. 

* https://github.com/artefactual/archivematica/blob/stable/1.9.x/CONTRIBUTING.md#working-with-legacy-code

We also hope to bring a code formatting tool into this process soon though.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -4554,7 +4554,7 @@
             "config": {
                 "@manager": "linkTaskManagerDirectories",
                 "@model": "StandardTaskConfig",
-                "arguments": "create \"%SIPDirectory%%SIPName%-%SIPUUID%\" \"%SIPDirectory%\" \"logs/\" \"objects/\" \"METS.%SIPUUID%.xml\" \"README.html\" \"thumbnails/\" \"metadata/\" --writer filesystem",
+                "arguments": "create \"%SIPDirectory%%SIPName%-%SIPUUID%\" \"%SIPDirectory%\" \"logs/\" \"objects/\" \"METS.%SIPUUID%.xml\" \"README.html\" \"thumbnails/\" \"metadata/\" --writer filesystem --sipUUID \"%SIPUUID%\"",
```
ross-spencer: Thanks for escaping this! We recently went through and escaped all arguments for the `1.9` release too.

ross-spencer (approved):
Tested, and the resulting AIP has an external identifier included which is the SIP/AIP UUID. 

I wonder, as you start adding to the output of an Archivematica Bag whether you would consider adding to the bag documentation here: https://www.archivematica.org/en/docs/archivematica-1.9/user-manual/archival-storage/aip-structure/#bagit-doc

It would be great for users to know from reading the docs what information they can access that might be useful to them without having to try-it and see. 

Otherwise, LGTM :+1:

helenst:
Thanks for reviewing @ross-spencer. I've tweaked that line length and will put in a PR to the docs repo.

PR #1348, opened 2019-02-13
Update default config

ablwr:
For @sallain. This is connected to https://github.com/archivematica/Issues/issues/255.


sallain (approved):
LGTM!

PR #1349, opened 2019-02-22
FileID: persist boolean for use in later chain

sevein:
We used to persist the `IDCommand` in `UnitVariable` to pass it down to
further processing chains. Instead, this commit only persists whether fileID
is enabled/disabled since the command is now looked up in FPR on demand.

This is a follow up for 3e958317f60a88c87073b5254499e2ee661c0efc.

Connects to https://github.com/archivematica/Issues/issues/512.


ross-spencer:
Tested and this works as expected. 

:+1: LGTM

ross-spencer (approved):

sevein:
Thanks for the quick response, @ross-spencer.

PR #1354, opened 2019-02-26
MCPServer: link activeAgent to transfer

sevein:
Ensure that a new transfer created by the new `package` module has a linked
"activeAgent".

This should be enough to have the logged-in user recorded as an
Agent for Events during transfer.

This is connected to https://github.com/archivematica/Issues/issues/529.

PR #1357, opened 2019-02-27
MCPClient: set activeAgent fallback in ingest

sevein:
Set `activeAgent` to the value used in Transfer to ensure the existence of the
user agent in PREMIS events. This value may change as soon as a different user
interacts with the user interface in processing, but this commit is useful when
interactions do not occur.

Follow-up for https://github.com/artefactual/archivematica/pull/1354.

Connects to https://github.com/archivematica/Issues/issues/529.

### DIFF
--- a/src/MCPClient/lib/clientScripts/create_sip_from_transfer_objects.py
+++ b/src/MCPClient/lib/clientScripts/create_sip_from_transfer_objects.py
@@ -28,7 +28,7 @@ import django
 django.setup()
 from django.db import transaction
 # dashboard
-from main.models import File, Directory, SIP, Transfer
+from main.models import File, Directory, SIP, Transfer, UnitVariable, Agent
 
 # archivematicaCommon
 import archivematicaFunctions
@@ -84,6 +84,24 @@ def call(jobs):
                     sip_uuid = databaseFunctions.createSIP(
                         lookup_path, UUID=sip_uuid, sip_type=sip_type, diruuids=diruuids, printfn=job.pyprint)
                     sip = SIP.objects.get(uuid=sip_uuid)
+
+                # Set activeAgent using the value in Transfer. This ensures
+                # that events generated in Ingest can fall to this value in
+                # scenarios where the processing config does not require user
+                # interfactions, e.g. in the "automated" processing config.
+                try:
+                    unit_variable = UnitVariable.objects.get(
+                        unittype="Transfer", unituuid=transferUUID,
+                        variable="activeAgent")
+                except UnitVariable.DoesNotExist:
+                    unit_variable = None
+                if unit_variable:
+                    try:
+                        agent = Agent.objects.get(id=unit_variable.variablevalue)
+                    except Agent.DoesNotExist:
+                        pass
+                    else:
+                        sip.update_active_agent(agent.userprofile.user_id)
 
                 # Move the objects to the SIPDir
                 for item in os.listdir(objectsDirectory):
--- a/src/MCPServer/lib/unit.py
+++ b/src/MCPServer/lib/unit.py
@@ -77,6 +77,8 @@ class unit:
         upsert `UnitVariable` rows for this unit. This is triggered by
         execution of a workflow chain link of type
         'linkTaskManagerSetUnitVariable'.
+
+        TODO: use Transfer/SIP model methods.
         """
         if not variableValue:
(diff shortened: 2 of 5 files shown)


src/MCPServer/lib/unit.py
```diff
@@ -77,6 +77,8 @@ def setVariable(self, variable, variableValue, microServiceChainLink):
         upsert `UnitVariable` rows for this unit. This is triggered by
         execution of a workflow chain link of type
         'linkTaskManagerSetUnitVariable'.
+
+        TODO: use Transfer/SIP model methods.
         """
```
sevein, line 82: I didn't want to touch this method before the release. But we could have the method removed as long as the `unit` in MCPServer holds a property with the instance of `Transfer` or `SIP`, so we can call the new `set_active_agent` method.

src/dashboard/src/main/models.py
```diff
@@ -871,6 +877,26 @@ class Meta:
...
+        return self.get_queryset().update_or_create(
+            unittype=unit_type, unituuid=unit_uuid,
+            variable=variable, defaults=defaults)
+
+    def set_active_agent(self, unit_type, unit_id, user_id):
+        """Set active agent given the user ID."""
+        return self.set_variable(
+            unit_type, unit_id,
+            "activeAgent", User.objects.get(id=user_id).userprofile.agent.id)
+
```
sevein, line 900: There's probably more and better ways to do this.
I'm definitely not used to work with Django models.

src/MCPClient/lib/clientScripts/create_sip_from_transfer_objects.py
```diff
@@ -85,6 +85,19 @@ def call(jobs):
...
 
+                # Set activeAgent using the value in Transfer. This ensures
+                # that events generated in Ingest can fall to this value in
+                # scenarios where the processing config does not require user
+                # interfactions, e.g. in the "automated" processing config.
+                try:
+                    agent = Agent.objects.get(id=UnitVariable.objects.get(
+                        unittype="Transfer", unituuid=transferUUID,
+                        variable="activeAgent").variablevalue)
+                except (Agent.DoesNotExist, UnitVariable.DoesNotExist):
```
cole: Nitpicky but I would break this out
```suggestion
                try:
                    unit_variable = UnitVariable.objects.get(
                        unittype="Transfer", unituuid=transferUUID, variable="activeAgent")
                except UnitVariable.DoesNotExist:
                    unit_variable = None
                
                if unit_variable:
                    try:
                        agent = Agent.objects.get(id=unit_variable.variablevalue)
                    except Agent.DoesNotExist:
```
(or alternately, get rid of the if and lookup the agent in the else clause of the try/except).
It's a bit clunkier (I guess, depends on your perspective) but I don't like having multiple lookups per statement.

src/dashboard/src/main/models.py
cole: I don't actually think there is really...but maybe I'm misunderstanding things. I like that you're using a manager here, that's the correct place for logic that operates on this level (across multiple model instances). I have a few more nitpicky suggestions: 
* I would break out `User.objects.get(id=user_id).userprofile.agent.id` as above
* I guess you can make the user query more efficient with `select_related` and just accessing the agent as `userprofile.agent_id` (something like `User.objects.select_related('userprofile').get(id=user_id).userprofile.agent_id` which I haven't tested 😄 )
* I think as a general rule it's better to use `update_foo` or `save_foo` or something vs `set_foo` when dealing with the ORM — makes it clearer that calling the method hits the DB and it's not just a setter/getter. So minor though.

cole (commented):
I have some minor Django feedback, but overall this seems like a good change given the scope we can tackle for the release. In general I'm not really clear on why transfers don't just have a direct relationship to agents, that seems like it would be much clearer.

src/dashboard/src/main/models.py
sevein: Thank you! Nice to learn about `select_related`!

sevein:
> In general I'm not really clear on why transfers don't just have a direct relationship to agents, that seems like it would be much clearer.

Good point. We should look into that at some point.

Thanks for the feedback, really useful.
I've also added some tests.

cole (approved):
Looks good! With the caveat that I haven't actually tested it.

PR #1359, opened 2019-03-01
FPR: update Inkscape normalization command

sevein:
This commit updates the command so it does not use sudo.

Connects to https://github.com/archivematica/Issues/issues/537.

### DIFF
--- a/src/dashboard/src/fpr/migrations/0029_update_inkscape_svg_command.py
+++ b/src/dashboard/src/fpr/migrations/0029_update_inkscape_svg_command.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+"""Stop using sudo in the `Transcoding to plain svg with inkscape` command."""
+
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+INKSCAPE_TOOL_UUID = "34afccc3-2cfd-4780-ba56-06221f1bfe18"
+
+# Normalization rules using `Transcoding to plain svg with inkscape`.
+INKSCAPE_SVG_RULES = (
+    "363f452e-a41b-46f3-865b-80a249493dfd",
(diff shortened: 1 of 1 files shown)


src/dashboard/src/fpr/migrations/0029_update_inkscape_svg_command.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-ç
```
ross-spencer: `ç` -  typo (I think). Not sure it affects anything...

ross-spencer (approved):
One typo I think, but looks good. I appreciate you didn't split the command over multiple lines to bring it inline with 80 characters. Will we run `black` over the migrations do you think?

src/dashboard/src/fpr/migrations/0029_update_inkscape_svg_command.py
sevein: Ha! Fixed, thanks.

sevein:
> I appreciate you didn't split the command over multiple lines to bring it inline with 80 characters. Will we run `black` over the migrations do you think?

I guess so. I just tried to see what black would do and this is what happened:

```diff
     # The only difference with the new command is that it's not using sudo.
     new_cmd = (
-        '/usr/bin/inkscape \"%fileFullName%\" --export-plain-svg=\"%outputDirectory%%prefix%%fileName%%postfix%.svg\"'
-        '\nchmod 777 \"%outputDirectory%%prefix%%fileName%%postfix%.svg\"'
+        '/usr/bin/inkscape "%fileFullName%" --export-plain-svg="%outputDirectory%%prefix%%fileName%%postfix%.svg"'
+        '\nchmod 777 "%outputDirectory%%prefix%%fileName%%postfix%.svg"'
     )
```

Which makes sense, I was using single quotes so no need to escape double quotes.

PR #1360, opened 2019-03-01
FPR: fix downgrade path in migration fpr.0029

sevein:
I had the data `data_migration_down` added but I didn't test it. It turned out to be broken.

---

This commit fixes the `data_migration_down` function in the migration
introduced by the previous commit to ensure that the order of the operations
is incorrect and the migration can be properly reverted.

Connects to https://github.com/archivematica/Issues/issues/537.


ross-spencer (approved):
Shoot, that makes sense. I should have taken an extra minute to test too. 

Good spotting, looks good. :+1:

PR #1362, opened 2019-03-06
Support BagIt in transfers

sevein:
Update the backlog workflow so it creates BagIt packages instead of plain transfers.
Restore from backlog is updated so both formats are supported.

Connects to https://github.com/archivematica/Issues/issues/246.
Connects to https://github.com/archivematica/Issues/issues/444.
Connects to https://github.com/archivematica/Issues/issues/446.


src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
```diff
@@ -83,47 +61,44 @@ def create_directories(base_dir, dir_list):
...
-    # Goal: bagit <operation> <destination> <flattened payload list> <optional args>
-    bagit_args = [operation, destination]
-    bagit_args.extend(payload_entries)
-    bagit_args.extend(['--writer', writer, '--payloadmanifestalgorithm', algorithm])
-
-    # Run bagit bag creator
-    run_bag(job, bagit_args)
+    os.mkdir(destination)
+    for item in payload_entries:
+        shutil.move(item, destination)
```
sevein, line 88: bagit-java had the ability to create the bag in a different destination while in the Python library the package is created in situ so I'm moving the contents to the actual destination expected by the workflow before the bag is made.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -60,29 +193,28 @@ def main(job, transfer_uuid, transfer_path):
...
 
-    message = "Transfer moved to backlog: {}".format(new_file)
-    logging.info(message)
-    job.pyprint(message)
-
-    # TODO update transfer location?  Files location?
-
-    # Delete transfer from processing space
+    logger.info("Deleting transfer from processing space (%s)...",
+                pprint.pformat(transfer_path))
```
cole: Isn't this just a string? Why pformat?

src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
```diff
@@ -85,62 +61,44 @@ def create_directories(base_dir, dir_list):
...
-    tmp_dir = tempfile.mkdtemp()
-    baginfo_filename = os.path.join(tmp_dir, 'bag-info.txt')
-    with open(baginfo_filename, 'a') as baginfo_file:
-        baginfo_file.write('External-Identifier: %s\n' % sip_uuid)
-
-    bagit_args.extend(['--baginfotxt', baginfo_filename])
-
-    # Run bagit bag creator
-    run_bag(job, bagit_args)
+    os.mkdir(destination)
```
cole, line 86: If destination already exists, something has gone wrong, right? Just making sure that's not a case we need to handle.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -1,27 +1,53 @@
...
+from databaseFunctions import insertIntoEvents
+import elasticSearchFunctions
+from main.models import Agent, File, UnitVariable
 import storageService as storage_service
 
+from bagit import make_bag
+import metsrw
+
+
+logger = get_script_logger('archivematica.mcp.client.move_tob_backlog')
```
cole: typo I think, should be **move_to_backlog**?

`get_script_logger` does some global logging config and then just returns a logger with the name you pass. We should move away from that in future I think, although it probably needs a separate issue. We can continue to use it in this PR I think.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -40,11 +66,118 @@ def _create_file(transfer_uuid, current_location, relative_transfer_path,
...
+        )
+    return metsrw.plugins.premisrw.data_to_premis(
+        premis_data, metsrw.plugins.premisrw.PREMIS_3_0_VERSION)
+
+
+def _transfer_agents(transfer_id):
+    """Return the agents associated to a transfer.
+
+    This is similar to ``databaseFunctions.getAMAgentsForFile`` but it returns
+    a ``QuerySet`` instead. TODO: move to model?
```
cole, line 133: Agree this should be moved to the model but probably out of scope for this PR.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -40,11 +66,118 @@ def _create_file(transfer_uuid, current_location, relative_transfer_path,
...
+        transfer_path, 'metadata', 'submissionDocumentation', 'METS.xml')
+    mets = metsrw.METSDocument().fromfile(mets_path)
+
+    # Run all_files once, convert into a dict for faster lookups.
+    fsentries = {entry.file_uuid: entry for entry in mets.all_files()}
+
+    # Assuming the same agents apply to all files.
+    agents = _transfer_agents(transfer_id)
+
+    for file in File.objects.filter(transfer_id=transfer_id).iterator():
```
cole: `file` is a builtin so it is preferable to avoid using it as a var name — I usually use `file_obj` in this case, every alternative seems a bit awkward though so it doesn't really matter.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -40,11 +66,118 @@ def _create_file(transfer_uuid, current_location, relative_transfer_path,
...
+            unituuid=transfer_id,
+            variable='activeAgent')
+    except UnitVariable.DoesNotExist:
+        pass
+    else:
+        query |= Q(id=var.variablevalue)
+    return Agent.objects.filter(query)
+
+
+def _record_backlog_event(transfer_id, transfer_path, created_at):
```
cole, line 147: Do we want to update the METS document here? I don't think we've discussed really the ideal for this. There are two designs I can see that make sense:

1. Each client script adds its own events to the METS xml file, if it exists.
2. METS updating is only done in the various update METS client scripts, and they are added to the workflow where required.

1 is kind of cool in that the mets is always up to date — but since we're not doing stream parsing, if the METS file is large, it may be problematic

src/MCPClient/requirements/test.txt
```diff
@@ -1,7 +1,7 @@
 -r base.txt
 
-pytest>=2,<3
-pytest-django>=2,<3
+pytest>=4,<5
```
cole, line 3: 👏 👏 👏

src/MCPClient/tests/test_move_to_backlog.py
```diff
@@ -0,0 +1,84 @@
...
+    assert ret.get(identifiertype='repository code')
+    assert ret.get(identifiertype='Archivematica user pk')
+
+
+@pytest.mark.django_db
+def test_record_backlog_event(tmp_path):
+    transfer = Transfer.objects.create(uuid="756db89c-1380-459d-83bc-d3772f1e7dd8")
+    user = User.objects.create(id=1)
+    transfer.update_active_agent(user_id=user.id)
+    file = File.objects.create(
```
cole: same, file is a builtin.

cole (commented):
I think this is all good — it's a big one! A few comments/questions.

src/MCPClient/lib/clientScripts/move_to_backlog.py
sevein: I've clarified this with @cole in PM. `move_to_backlog` is executed after the transfer package has already been created so we don't have a better alternative atm.

src/MCPClient/lib/clientScripts/move_to_backlog.py
sevein: Ops, I wanted to use with the line above, `new_file`, that one is a complex object likely with long lines that would be nice format for readability.

src/MCPClient/lib/clientScripts/move_to_backlog.py
sevein: Agree! Typo fixed.

src/MCPClient/tests/test_move_to_backlog.py
sevein: Right! Thanks, fixed.

src/MCPClient/requirements/test.txt
sevein: Zero issues! So nice.

src/MCPClient/lib/clientScripts/move_to_backlog.py
sevein: 👍

src/MCPClient/lib/clientScripts/bag_with_empty_directories.py
sevein: Yeah, I wouldn't expect that and if it happens to raise I think that's what we want, the JobContext context manager would capture the error and set the job as failed. I'm always tempted to do some logging around these unlikely events but I don't think it's worth the effort?

cole: I agree, I think it's fine to raise the error. Just checking because I don't really know what's expected to be on the file system.

cole (approved):

Issue #998, opened 2018-03-18
Problem: Validation events aren't making it into the Transfer METS.xml on Transfer


ross-spencer:
I have created a SIP on Augustus and sent it to [backlog](http://augustus.qa.archivematica.org/tasks/bc9f9ae4-18bb-415f-8c2e-48d734e21012/) the log confirmats that files that have been passed onto JHOVE for validation but if I review the METS in Backlog ([download link](http://augustus.qa.archivematica.org/backlog/download/74367005-4470-4fa8-900b-1ce8760c533e)), the validation events for these objects aren't there. 

Can anyone help me confirm this?

(Further, but this might be a misunderstanding reading the METS at this stage, but I see PREMIS events for identification and virus scan, but they don't seem to be linked back to the digital objects properly? - Am I reading this incorrectly too?)

**NB.** I am seeing this in my own QA environment as well.

ross-spencer:
This does not seem to be a regression, and can be observed in the 1.6 release via http://sandbox.archivematica.org. It seems we may need to review the content of the METS/PREMIS at this phase of the workflow where certain events related to the digital objects have already happened.

ross-spencer:
Further, reviewing the AIP, the information does seem to be present there (1.6). 

It seems that another difference between the Transfer METS and AIP METS is that `            <premis:originalName>%transferDirectory%objects/Club-Report.doc</premis:originalName>` doesn't seem to exist in the Transfer METS which (I think) means the PREMIS events can't be linked back to their original objects - not until the Backlog SIP is turned into an AIP.

peterVG:
I can confirm on AM 1.8 that, while Validate events are successfully occurring during transfer, they are not being written to the backlog METS file as premis:eventTypes. They do, however, appear in the AIP METS xml as premis:eventTypes. Other transfer premis:eventTypes are being written to the transfer backlog METS file including ingestion, message digest calculation, fixity check, virus check, and format identification.

peterVG:
Also, while the `premis:originalName` value is not included in the transfer METS, the PREMIS events in the Transfer METS can still be linked to their original objects via the `mets:fileSec` which links the original object to its amdSec via an `ADMID`.

ross-spencer:
[...]

PR #24, opened 2014-03-17
views_processing: Update comments for preconfigured choice settings

mistydemeo:

PR #1363, opened 2019-03-08
Use metsrw for transfer METS creation

cole:
Not ready for review yet.

Connects to https://github.com/artefactual/archivematica/issues/998.
Connects to https://github.com/archivematica/Issues/issues/442.


src/MCPClient/lib/clientScripts/create_transfer_mets.py
```diff
@@ -0,0 +1,351 @@
...
+                    file_obj.modificationtime.strftime("%Y-%m-%d"),
+                ),
+            ),
+        ),
+        ("original_name", escape(file_obj.originallocation)),
+    ) + get_premis_relationship_data(
+        file_obj.related_is_source_of, file_obj.related_has_source
+    )
+
+    return metsrw.plugins.premisrw.data_to_premis(premis_data)
```
sevein:
```suggestion
    return metsrw.plugins.premisrw.data_to_premis(
        premis_data, metsrw.plugins.premisrw.PREMIS_3_0_VERSION)
```

src/MCPClient/lib/clientScripts/create_transfer_mets.py
```diff
@@ -0,0 +1,351 @@
...
+    Converts an File model object to a PREMIS event object via metsrw.
+
+    Returns:
+        metsrw.plugins.premisrw.premis.PREMISObject
+    """
+
+    premis_digest_algorithm = convert_to_premis_hash_function(file_obj.checksumtype)
+    premis_data = (
+        "object",
+        metsrw.plugins.premisrw.PREMIS_2_2_META,
```
sevein:
```suggestion
        metsrw.plugins.premisrw.PREMIS_3_0_META,
```

sevein (commented):
I've added an example on how to use PREMIS 3 which we need (see [#24](https://github.com/archivematica/Issues/issues/24) for context).

src/MCPClient/lib/clientScripts/create_transfer_mets.py
cole: Thanks! I was looking at the current behaviour of createprocessedstructmap. Will change.

src/MCPClient/lib/clientScripts/create_transfer_mets.py
cole: Thanks! I was looking at the current behaviour of createprocessedstructmap. Will change.

sevein (commented):
I've started using your branch for my rebuild work. I have a couple of comments:

1. The pipeline UUID is not being included in `metsHdr`. I only need this because when the transfer is sent to backlog, Storage Service [raises an error](https://github.com/artefactual/archivematica-storage-service/blob/4987a8c682d16ea38089acb7a9200a0c5f2200e5/storage_service/locations/models/package.py#L1784-L1790) when the METS doesn't include it. E.g.:
   ```xml
   <mets:metsHdr CREATEDATE="2015-12-16T22:38:48">
     <mets:agent OTHERTYPE="SOFTWARE" ROLE="CREATOR" TYPE="OTHER">
       <mets:name>39461beb-22eb-4942-88af-848cfc3462b2</mets:name>
       <mets:note>Archivematica dashboard UUID</mets:note>
     </mets:agent>
   </mets:metsHdr>
    ```

2. I've found that `Flocat` does not include the full relative path, e.g. I'd expect something like:
   ```xml
   <FLocat xlink:href="objects/V00154.MPG" LOCTYPE="OTHER" OTHERLOCTYPE="SYSTEM"/>
   ```

   But the XML that you're generating only reads `xlink:href="V00154.MPG"`. Not sure if the problem is here or in metsrw. When reading, I'd expect `fsentry.path` to be `objects/V00154.MPG` - at least that's how it's always been AFAIK. E.g. see https://github.com/artefactual/archivematica-storage-service/blob/4987a8c682d16ea38089acb7a9200a0c5f2200e5/storage_service/locations/models/package.py#L1799.

sevein (approved):
This is great!

I've been testing this and it's all working excepting the missing `metsHdr` agent details.

PR #1364, opened 2019-03-08
Tag and click arrangement

replaceafill:
This PR:

Connects to https://github.com/archivematica/Issues/issues/369

- Synchronizes tag display/add/removal in all the panes of the Appraisal tab
- Adds **tag-n-click** functionality to the Appraisal tab allowing selection of all the files with the same tag to be included in a SIP

Connects to https://github.com/archivematica/Issues/issues/507

- Merges the two Angular apps (appraisal-tab and transfer-browser) into a single dashboard app, synchronizing their dependencies and adds browser configurations to the dashboard.Dockerfile to allow running JS tests with Karma

Connects to https://github.com/archivematica/Issues/issues/472

- Fixes tag removal in the File list pane of the Appraisal tab


src/dashboard.Dockerfile
```diff
@@ -19,13 +19,64 @@ RUN set -ex \
 		locales \
 		locales-all \
 		unar \
+		wget \
```
sevein: @replaceafill do you think it's worth to move the Chrome/FF installer snippets over to use curl so we don't need to install wget too?

src/dashboard.Dockerfile
```diff
@@ -19,13 +19,64 @@ RUN set -ex \
...
+	&& apt-get -qqy install ${CHROME_VERSION:-google-chrome-stable} \
+	&& rm /etc/apt/sources.list.d/google-chrome.list \
+	&& rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+ARG CHROME_DRIVER_VERSION="latest"
+RUN CD_VERSION=$(if [ ${CHROME_DRIVER_VERSION:-latest} = "latest" ]; then echo $(wget -qO- https://chromedriver.storage.googleapis.com/LATEST_RELEASE); else echo $CHROME_DRIVER_VERSION; fi) \
+	&& echo "Using chromedriver version: "$CD_VERSION \
+	&& wget --no-verbose -O /tmp/chromedriver_linux64.zip https://chromedriver.storage.googleapis.com/$CD_VERSION/chromedriver_linux64.zip \
+	&& rm -rf /opt/selenium/chromedriver \
+	&& unzip /tmp/chromedriver_linux64.zip -d /opt/selenium \
```
sevein: I'm seeing this error:

    /bin/sh: 1: unzip: not found

It disappear when I had `unzip` installed. It'd also be possible to rewrite the snippet so it uses `unar` instead, which is already installed by us (see L21).

sevein (commented):
Neat! It works, including tests!

👏🎉👏🎉👏🎉👏🎉👏🎉 

I've added a couple of minor comments.

replaceafill:
@sevein I just addressed your observations.

I also fixed the `npm install` command. I probably didn't noticed before because I still had the old `frontend/appraisal-tab` and `frontend/transfer-browser` directories lying around.

sevein:
> I also fixed the npm install command. I probably didn't noticed before because I still had the old frontend/appraisal-tab and frontend/transfer-browser directories lying around.

I guess it happen the same to me, but I wasn't testing the new functionality you've added so I guess that's why I missed it! I'll give it another try tonight! Thanks you!

.travis.yml
```diff
@@ -43,6 +43,26 @@ matrix:
...
 
+    # Dashboard JS
+    - language: node_js
+      node_js: "8"
+      cache: npm
+      before_install:
+        - export DISPLAY=:99.0
+        - sh -e /etc/init.d/xvfb start
+        - sleep 3 # give xvfb some time to start
+        - cd ./src/dashboard/frontend/
```
sevein, line 52: Wow @replaceafill that was fast. Why do we still need xvfb?

.travis.yml
```diff
@@ -43,6 +43,26 @@ matrix:
...
+      install: npm install
+      before_script:
+        - '' # override global before_script
+      script: npm run "test-single-run"
+      addons:
+        apt:
+          sources:
+            - google-chrome
```
sevein: Out of curiosity, why is this needed?

.travis.yml
replaceafill: We don't. And this made me realize we don't need the browser drivers either in the `dashboard.Dockerfile`. I just removed them in https://github.com/artefactual/archivematica/pull/1364/commits/36bba916ca5930caa9f2aad13a070048f67dd988

.travis.yml
replaceafill: I was just trying to mimic what we do in the `dashboard.Dockerfile` (download key, add source, etc). But I just simplified it in https://github.com/artefactual/archivematica/pull/1364/commits/36bba916ca5930caa9f2aad13a070048f67dd988 to:

```
addons:
  chrome: stable
```

sevein:
Thanks @replaceafill! 1 min 44 secs for build time is pretty great.

sevein (approved):
:heart_eyes:

ablwr:
So awesome, can't wait to check this out!

PR #1369, opened 2019-03-18
Reduce the need to access the disk in ES scripts

ross-spencer:
This commit removes the need to test the existence of an AIP on disk
when being indexed by our elasticsearch scripts by sending the
statistics that we are already in possession of when that script is
called. Additionally variables are cleaned up to perhaps be a bit
more intuitive to future readers of this code.

Connected to archivematica/issues#559


jraddaoui (approved):

ross-spencer:
Thanks @jraddaoui  :+1:

jraddaoui:
@ross-spencer don't forget to cherry-pick the commit to qa/1.x, thanks!!

PR #1370, opened 2019-03-18
Reduce the need to access the disk in ES scripts

ross-spencer:
This commit removes the need to test the existence of an AIP on disk
when being indexed by our elasticsearch scripts by sending the
statistics that we are already in possession of when that script is
called. Additionally variables are cleaned up to perhaps be a bit
more intuitive to future readers of this code.

Connected to archivematica/issues#559
Connected to archivematica/issues#585
Cherry-picked from 1ec81430466688ae989f2413228d16d5849c9fee

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -351,10 +351,10 @@ def _get_index_settings():
 def index_aip_and_files(
     client,
     uuid,
-    path,
-    mets_path,
+    aip_stored_path,
+    mets_staging_path,
     name,
-    size=None,
+    aip_size,
     aips_in_aic=None,
     identifiers=[],
     encrypted=False,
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Thanks @ross-spencer!

I personally think there is no need to create a PR to cherry-pick a commit from the closest stable branch, unless there is a lot of conflicts in the process and you want an extra pair of eyes.

ross-spencer:
Thanks @jraddaoui - I agree. I'll try the other way with the DSpace fix I've been working on. I have a feeling though our repository settings will block without approval though, so if you see a PR come through for that we'll know why :)

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -364,35 +364,28 @@ def index_aip_and_files(
...
-    :param size: optional AIP size.
+    :param aip_size: AIP size.
     :param aips_in_aic: optional number of AIPs stored in AIC.
     :param identifiers: optional additional identifiers (MODS, Islandora, etc.).
     :param identifiers: optional AIP encrypted boolean (defaults to `False`).
     :param printfn: optional print funtion.
     :return: 0 is succeded, 1 otherwise.
     """
-    # Stop if AIP or METS file don't not exist
+    # Stop if AIP or METS file don't not exist.
```
alexwlchan: Is this comment still correct? I don’t see a check for the existence of the AIP file.

ross-spencer: Your're right. I missed that comment. I'll remove it before merging. Thanks @alexwlchan.

ross-spencer: Fix it rather. The METS check is still there.

PR #1371, opened 2019-03-19
Dashboard: remove SIP arrange from Ingest tab

replaceafill:
This PR removes the SIP arrange functionality from the Ingest tab and makes the following changes to the Appraisal tab:

* Search form: Adds **Show metadata & logs directories?** checkbox that affects the results of the Backlog pane. The metadata/log files can't be dragged to the Arrangement pane but can be downloaded from the File List pane
* Arrangement pane: **Edit Metadata** button allows to set levels of description from AtoM
* Arrangement pane: Adds ability to drag folders and files

Connects to https://github.com/archivematica/Issues/issues/536


sevein (commented):
Looking really good, @replaceafill!

src/dashboard/frontend/app/arrangement/arrangement.controller.js
```diff
@@ -294,6 +296,92 @@ controller('ArrangementController', ['$scope', 'gettextCatalog', 'Alert', 'Trans
     }
   };
 
+  var getCookie = function(name) {
+    var value = "; " + document.cookie;
+    var parts = value.split("; " + name + "=");
+    if (parts.length == 2) return parts.pop().split(";").shift();
+  };
```
sevein: Could this be replaced by the [$cookies](https://docs.angularjs.org/api/ngCookies/service/$cookies) service?
Same in `transfer_browser_transfer.service.js`.

src/dashboard/frontend/app/front_page/appraisal_tab.html
```diff
@@ -1,8 +1,10 @@
 <div ng-controller="SearchController as vm">
-  <form id='search_form'>
+  <form id='search_form' class='form-inline'>
   <div id='search_form_container'></div>
   <div id='search_form_submit_container'>
     <input class='form-control btn-primary' type='button' id='search_submit' value='{{ "Search transfer backlog" | translate }}'>
+    <span style='margin-left: 10px'>Show metadata &amp; logs directories?</span>
```
sevein: `| translate` not used here!

src/dashboard/frontend/app/front_page/appraisal_tab.html
```diff
@@ -343,7 +346,7 @@
         {{ node.title }} <span ng-if="node.properties">({{ node.properties.display_string }})</span>
       </span>
       <span tree-draggable file-type="arrange" file-path="{{ node.path }}" ng-if="!node.directory">
-        <span>{{ node.title }} <span class="tag" ng-repeat="tag in node.properties.tags">{{ tag }} <span ng-click="vm.remove_tag($event, node, tag)"><i class="fa fa-minus-square remove-button"></i></span></span></span>
+        <span>{{ node.title }} <span ng-if="node.properties.display_string">({{ node.properties.display_string }})</span><span class="tag" ng-repeat="tag in node.properties.tags">{{ tag }} <span ng-click="vm.remove_tag($event, node, tag)"><i class="fa fa-minus-square remove-button"></i></span></span></span>
```
sevein: Could this kind of oneliner be formatted like the following? E.g.:

```html
<span>
  {{ node.title }}&nbsp;
  <span ng-if="node.properties.display_string">({{ node.properties.display_string }})</span>
  <span class="tag" ng-repeat="tag in node.properties.tags">
    {{ tag }}&nbsp;
    <span ng-click="vm.remove_tag($event, node, tag)">
      <i class="fa fa-minus-square remove-button"></i>
    </span>
  </span>
</span>
```

src/dashboard/src/components/ingest/views.py
```diff
@@ -514,7 +514,14 @@ def _es_results_to_appraisal_tab_format(
...
+            # check if the transfer is a bag (contains a data directory)
+            if node_parts[1] == "data":
+                if len(node_parts) > 2:
+                    node_not_draggable = not_draggable or _is_hidden(node_parts[2])
+                else:
+                    node_not_draggable = not_draggable
+            else:
+                node_not_draggable = not_draggable or _is_hidden(node_parts[1])
         else:
             node_not_draggable = not_draggable
```
sevein: Fiddly! I wonder if we can clear this a bit by adding a function or nested function like ` def _is_draggable(parts)` so we can replace the block above with something like:

```python
draggable = _is_draggable(node_parts) if is_transfer else not_draggable
```

And then flip the boolean below like:

```python
{"not_draggable": not draggable}
```

And use `_is_draggable` again below in L547?

src/dashboard/src/templates/ingest/grid.html
```diff
@@ -24,21 +24,12 @@
 
   <!-- browser related -->
   <script type="text/javascript" src="{% static 'js/vendor/base64.js' %}"></script>
-  <script type="text/javascript" src="{% static 'js/file-explorer.js' %}"></script>
-  <script type="text/javascript" src="{% static 'js/file_browser.js' %}"></script>
-  <script type="text/javascript" src="{% static 'js/ingest/directory-metadata-form-dialog.js' %}"></script>
-  <script type="text/javascript" src="{% static 'js/ingest/ingest_file_browser.js' %}"></script>
 
   <script type="text/javascript" src="{% static 'js/advanced-search-query-creator.js' %}"></script>
```
sevein: Is this still needed?

sevein: And what about directory_picker.css and friends? Is that stuff still needed because of the review dialogs?

src/dashboard/src/templates/ingest/grid.html
```diff
@@ -71,32 +62,6 @@
...
-        <span id='arrange_create_directory_button' class='btn btn-default'>{% trans "Add directory" %}</span>
-        <span id='arrange_edit_metadata_button' class='btn btn-default'>{% trans "Edit metadata" %}</span>
-        <span id='arrange_delete_button' class='btn btn-default'>{% trans "Delete" %}</span>
-        <span id='arrange_create_sip_button' class='btn btn-primary'>{% trans "Create SIP" %}</span>
-      </div>
-    </div>
-    <br clear='all' />
-    <br />
-  {% endif %}
-
```
sevein, line 99: From a usability perspective, considering that the old SIP arrangement widget is being removed, I wonder if we should have some kind of link at the top of the Ingest grid linking users to Appraisal? I'm not really a user but I'd definitely miss it since I didn't really use Appraisal myself.

Probably too simple but something like the following could be helpful:

---

image

---

CC @sallain @sromkey @ablwr @ross-spencer @peterVG @evelynPM @currmie @corinnemrogers @kellyannewithane - thoughts?

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -1103,7 +1104,14 @@ def download_by_uuid(request, uuid, preview_file=False):
             % {"uuid": uuid},
         }
         return helpers.json_response(response, status_code=404)
-    relative_path = f.currentlocation.replace("%transferDirectory%", "")
+    # check if the transfer is in the backlog and if it's a bag
+    transfer = models.Transfer.objects.get(uuid=f.transfer.uuid)
```
sevein: Is this a good use case for Django's [`select_related(*fields)`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#select-related)? I've learned about it recently!

```python
    f = models.File.objects.select_related('transfer').get(uuid=uuid)

File: src/dashboard/src/components/filesystem_ajax/views.py [hunk 8]
```diff
@@ -1103,7 +1104,14 @@ def download_by_uuid(request, uuid, preview_file=False):
...
+    transfer_name = os.path.basename(transfer.currentlocation.rstrip("/"))
+    transfer_dir = os.path.join(ORIGINAL_DIR, transfer_name)
+    if os.path.isdir(transfer_dir) and is_bag(transfer_dir):
```
sevein, line 1110: This feels a bit hacky and I take blame for that...

I wonder if `Transfer` should learn if it's in backlog and its format. Something to consider in the future.

src/dashboard/frontend/app/locale/translations.sv.po
```diff
@@ -463,6 +467,7 @@ msgstr ""
 "Kunde inte starta SIP; inga filer ordnade i arkivbeståndet \"{{title}}\"."
 
 #: app/archivesspace/archivesspace.controller.js:162
+#: app/arrangement/arrangement.controller.js:362
```
sevein: Should we add `src/dashboard/frontend/app/locale/*.po` to `.gitattributes`?
See https://help.github.com/en/articles/customizing-how-changed-files-appear-on-github.

sevein (changes requested):
Nicely done @replaceafill. It feels so good to have some of that old code removed. I've added some comments.

For future work I'd suggest, as a reviewer, to break big chunks of work into separate commits, e.g. in this case I think that translations, sip arrangement removal, metadata/log checkbox, edit metadata button, drag folder/files could have all been broken into individual commits? See https://github.com/artefactual/archivematica/blob/stable/1.9.x/CONTRIBUTING.md#commits-should-be-specific-and-atomic.

src/dashboard/frontend/app/arrangement/arrangement.controller.js
replaceafill: Nice! :+1:

src/dashboard/frontend/app/front_page/appraisal_tab.html
replaceafill: :+1:

src/dashboard/frontend/app/front_page/appraisal_tab.html
replaceafill: :+1:

src/dashboard/src/components/ingest/views.py
replaceafill: I agree. I'll try to do that.

src/dashboard/src/templates/ingest/grid.html
replaceafill: Yes. I tried not to delete files which content was still referenced in the templates. I can take another look and check.

src/dashboard/src/components/filesystem_ajax/views.py
replaceafill: Ah nice!

src/dashboard/src/components/filesystem_ajax/views.py
replaceafill: It does feel hacky but couldn't find an alternative.

src/dashboard/frontend/app/locale/translations.sv.po
replaceafill: Thanks for the reference. I added the `*.po` and `translation.json` to `.gitattributes`.

replaceafill:
@sevein I still have a couple of comments to address, but I'll take the individual/atomic commits comment into account from now on. Sorry that so far I've been doing it backwards (get rid of the little commits before requesting CR).

.gitattributes
```diff
@@ -1,3 +1,4 @@
 *.py ident
 *.sh ident
-
+src/dashboard/frontend/app/locale/*.po linguist-generated=true
+src/dashboard/frontend/app/locale/translations.json linguist-generated=true
```
sevein, line 5: Should we also add the following?

    src/dashboard/frontend/app/locale/*.pot linguist-generated=true

replaceafill: Done.

replaceafill:
@sevein Let me know if I missed anything, but I think this is ready for another check.

sevein (approved):

PR #1373, opened 2019-03-19
Ensure file paths aren't turned into dir paths

ross-spencer:
Previously when transfers of individual files - possible for DSpace
and Zipped Bag transfers were sent to Archivematica they were
incorrectly turned into paths that would be interpreted as
directories preventing approval of a transfer for not being able to
locate that directory. Here we re-enable individual DSpace file
transfers and galvanize the check to also test whether or not we're
looking at a file at the same time.

For expediency of `1.9.1` this is just rectifying the previously introduced issues. I haven't written new unit tests which would take me a few hours to work out and write. I am hoping more coverage for the API can be done in the near future.

Connected to archivematica/issues#468

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -42,6 +42,7 @@ from main import models
 from processing import install_builtin_config
 
 LOGGER = logging.getLogger('archivematica.dashboard')
+SHARED_PATH_TEMPLATE_VAL = "%sharedPath%"
 SHARED_DIRECTORY_ROOT = django_settings.SHARED_DIRECTORY
 UUID_REGEX = re.compile(r'^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$', re.IGNORECASE)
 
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -449,20 +449,22 @@ def approve_transfer(request):
...
         return _error_response("Invalid transfer type.", status_code=500)
-
-    if transfer_type == 'zipped bag':
-        db_transfer_path = os.path.join(modified_transfer_path, directory)
+    watched_path = os.path.join(modified_transfer_path, directory)
+    if (transfer_type == 'zipped bag' or transfer_type == "dspace") \
+            and os.path.isfile(watched_path
+                               .replace(SHARED_PATH_TEMPLATE_VAL,
+                                        SHARED_DIRECTORY_ROOT)):
+        db_transfer_path = watched_path
```
sevein, line 464: Do you think the following would be clearer?

```python
    transfer_file = watched_path.replace(
        SHARED_PATH_TEMPLATE_VAL, SHARED_DIRECTORY_ROOT
    )
    if transfer_type in ("zipped bag", "dspace") and os.path.isfile(transfer_file):
        db_transfer_path = watched_path
```

sevein (approved):
Thanks for doing this! :clap:

I see you don't like blank lines? PEP 8 has something to say about that:
> Use blank lines in functions, sparingly, to indicate logical sections.

I personally don't care, black is flexible on that too apparently. My only suggestion would be for future work to do that kind of stuff on separate commits so the readers are not distracted with cosmetics - does that sound like a good idea to you?

ross-spencer:
@sevein I appreciate the review. RE: blank lines, my GitHub diff must look a lot different as I can only see `327` and `452` are perhaps deleted? But GitHub otherwise highlights the added and removed intellectual payload? Ultimately I agree within reason which is where I made the distinction in the two commits here.

PR #1375, opened 2019-03-20
Bump version to 1.10.0

sevein:
The "preservation system" agent is still a database object, this is the usual
migration that bumps its version (see #383).

Connects to https://github.com/archivematica/Issues/issues/582.


ross-spencer (approved):
:+1:

PR #1376, opened 2019-03-20
Fix AIP bag creation

sevein:
The recent update to bagit-python was not including the METS payload and
moving instead of copying was causing other microservices to fail since the
originals were still expected to be present in currentlyProcessing.

Connects to https://github.com/archivematica/Issues/issues/246.


sevein:
Sorry, I think there's still something wrong with it. Investigating...

replaceafill (approved):
Thanks! AMAUAT black box tests pass again!

hakamine:
@sevein Thank you for the fix. Tested by doing an ingest and now the METS file that had been missing is included

PR #1379, opened 2019-03-20
Add import support for self-describing backlog transfers

sevein:
This pull requests updates the `rebuild_transfer_backlog` management command to support restoring the new self-describing transfer packages. The biggest change is that it can be done over any pipeline even when it's freshly installed.

Directories are not populated yet - this work and other improvements may be completed in following pull requests.

Connects to https://github.com/archivematica/Issues/issues/446.
Connects to https://github.com/archivematica/Issues/issues/586.

### DIFF
--- a/src/dashboard/tests/test_rights.py
+++ b/src/dashboard/tests/test_rights.py
@@ -12,17 +12,13 @@ from components.rights import load
 from main.models import File, Transfer, SIP, MetadataAppliesToType
 
 
-_RIGHTS_STATEMENT_BASE = (
-    "rights_statement",
-    premisrw.PREMIS_META,
-    (
-        "rights_statement_identifier",
-        ("rights_statement_identifier_type", "UUID"),
-        ("rights_statement_identifier_value", "3a9838ac-ebe9-4ecb-ba46-c31ee1d6e7c2"),
-    ),
+RIGHTS_STATEMENT_IDENTIFIER = (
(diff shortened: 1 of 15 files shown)


src/dashboard/tests/test_rights.py
```diff
@@ -0,0 +1,336 @@
...
+
+from metsrw.plugins import premisrw
+import pytest
+import uuid
+
+from components.rights import load
+from main.models import File, Transfer, SIP, MetadataAppliesToType
+
+
+_RIGHTS_STATEMENT_IDENTIFIER = (
```
cole: This is all personal preference, but I don't know if we need to worry about "private" constants in test — I think it should be obvious to everybody not to import from a test file.

cole (approved):
This looks really good! Thanks for fixing all of the things 🙂 

What is the plan for the [TMP] commits, I'm assuming we don't want to merge them?

src/dashboard/tests/test_rights.py
sevein: You're right, fixed.

sevein:
> What is the plan for the [TMP] commits, I'm assuming we don't want to merge them?
- The docker fix has been fixed by Radda, but it needs to be merged.
- I'm going to be uploading the instructions somewhere else and tell Peter and the client.
- pytest test collection improvement: I'll file a new issue for that!

PR #1380, opened 2019-03-20
Search: store AIP size in MB

sevein:
This commit reverts a recent change where bytes were used - although I think
it's probably better to use bytes we have to keep using MB unless we update
consumers.

Found by @replaceafill.
Connects to https://github.com/archivematica/Issues/issues/559.
Once merged, it should also go to stable/1.9.x before the v1.9.1 release!


replaceafill (approved):
:+1:

sevein:
e16eded0 in stable/1.9.x

PR #1381, opened 2019-03-21
Fix index_aip_and_files() call from mgmt cmd

hakamine:
Fix index_aip_and_files() invocation from
rebuild_elasticsearch_aip_index_from_files management command.

Connected to https://github.com/archivematica/Issues/issues/595


jraddaoui (approved):
Thanks @hakamine!

Sorry, I totally missed this call in the review of the commit that added the regression. Please, cherry-pick it to stable/1.9.x too.

PR #1383, opened 2019-03-22
Fix index_aip_and_files() call from mgmt cmd (stable/1.9.x)

hakamine:
Fix index_aip_and_files() invocation from
rebuild_elasticsearch_aip_index_from_files management command.

Connected to https://github.com/archivematica/Issues/issues/595

Cherry-pick of https://github.com/artefactual/archivematica/pull/1381 (protected branch requires approving review)

### DIFF
--- a/src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py
+++ b/src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py
@@ -164,10 +164,10 @@ def processAIPThenDeleteMETSFile(path, temp_dir, es_client,
     return elasticSearchFunctions.index_aip_and_files(
         client=es_client,
         uuid=aip_uuid,
-        path=path,
-        mets_path=path_to_mets,
+        aip_stored_path=path,
+        mets_staging_path=path_to_mets,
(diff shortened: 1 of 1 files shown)


jraddaoui (approved):
Thanks @hakamine! I personally think there is no need to create a PR to cherry-pick an already reviewed commit, unless you want an extra pair of eyes in case there was a lot of conflicts involved.

hakamine:
@jraddaoui Thank you. The reason I created a separate PR is because github did not let me just cherry-pick onto stable/1.9.x and push, got the following error:
<pre>
remote: error: GH006: Protected branch update failed for refs/heads/stable/1.9.x.
remote: error: At least 1 approving review is required by reviewers with write access.
To github.com:artefactual/archivematica
 ! [remote rejected]   stable/1.9.x -> stable/1.9.x (protected branch hook declined)
error: failed to push some refs to 'git@github.com:artefactual/archivematica'
</pre>
Do you know a way to avoid this problem?

jraddaoui:
Thanks for letting me know @hakamine, I didn't know about that hook.

jraddaoui:
And I guess we probably don't want to avoid it, it's a good security measure.

PR #1385, opened 2019-03-25
Add rights data to transfer METS

cole:
Connects to https://github.com/archivematica/Issues/issues/587

### DIFF
--- a/src/MCPClient/lib/clientScripts/create_transfer_mets.py
+++ b/src/MCPClient/lib/clientScripts/create_transfer_mets.py
@@ -24,34 +24,31 @@ from __future__ import unicode_literals
 import argparse
 import logging
 import os
+import uuid
 
 import django
 import scandir
 from lxml import etree
-from django.db.models import Prefetch
+from django.db.models import Prefetch, Q
 
 django.setup()
 import metsrw
 
 # archivematicaCommon
 from archivematicaFunctions import get_dashboard_uuid, escape
+from countryCodes import getCodeForCountry
 
 # dashboard
-from main.models import Derivation, File, FPCommandOutput
+from main.models import Derivation, File, FPCommandOutput, RightsStatement
 
 
 PREMIS_META = metsrw.plugins.premisrw.PREMIS_3_0_META
 
+
 logger = logging.getLogger(__name__)
 
 
-def write_mets(
-    mets_path,
-    transfer_dir_path,
-    base_path_placeholder,
-    identifier_group,
-    identifier_uuid,
-):
+def write_mets(mets_path, transfer_dir_path, base_path_placeholder, transfer_uuid):
     """
     Writes a METS XML file to disk, containing all the data we can find.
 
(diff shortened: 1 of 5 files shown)


src/MCPClient/lib/clientScripts/create_transfer_mets.py
```diff
@@ -376,6 +601,74 @@ def event_to_premis(event):
...
 
 
+def rights_to_premis(rights, file_uuid):
+    """
+    Converts an RightsStatement model to a PREMIS event object via metsrw.
+
+    Returns:
+        lxml.etree._Element
+    """
+    VALID_RIGHTS_BASES = ["copyright", "institutional policy", "license", "statute"]
```
cole, line 611: I'd prefer to move this up to module level, but` ensure_no_mutable_globals` complains (even if I make this an _immutable_ `frozenset`, lol). I don't think thats a valid complaint, but I'll submit an issue for that.

src/MCPClient/lib/clientScripts/create_transfer_mets.py
```diff
@@ -197,6 +208,33 @@ def lookup_file_data(relative_path, db_base_path, lookup_kwargs):
     return file_obj
 
 
+def lookup_file_rights(file_uuid, transfer_uuid):
+    TRANSFER_RIGHTS_LOOKUP_UUID = "45696327-44c5-4e78-849b-e027a189bf4d"
+    FILE_RIGHTS_LOOKUP_UUID = "7f04d9d4-92c2-44a5-93dc-b7bfdf0c1f17"
```
cole, line 213: I'd prefer not to use hardcoded UUIDs here, but that's how the model is. In the meantime, I'd prefer to move this up to module level, but` ensure_no_mutable_globals` complains.

sevein (commented):
I've found some validation issues - via http://pim.fcla.edu.
Full report here: https://pastebin.com/ZWfmFAE4.

* Invalid content was found starting with element 'premis:eventDetail'.

  I think that in PREMIS 3 the element has been renamed.

* The type definition cannot be abstract for element premis:object.

  ```xml
  <mets:techMD ID="techMD_1" CREATED="2019-03-26T18:02:02" STATUS="current">
    <mets:mdWrap MDTYPE="PREMIS:OBJECT">
      <mets:xmlData>
        <premis:object
          xmlns:premis="http://www.loc.gov/premis/v3"
          xsi:schemaLocation="http://www.loc.gov/premis/v3
          http://www.loc.gov/standards/premis/v3/premis.xsd"
          version="3.0">
          [...]
  ```

  Missing `type`?

* Element `premis:object` must have no character or element information item [children], because the type's content type is empty.
  
  I think this means that `xsi:type="file"` is missing!

  ```xml
  <premis:object xmlns:premis="http://www.loc.gov/premis/v3" xsi:schemaLocation="http://www.loc.gov/premis/v3 http://www.loc.gov/standards/premis/v3/premis.xsd" version="3.0">
    [...]
  </premis:object>
  ```

* Element 'premis:objectCharacteristicsExtension' cannot have character [children], because the type's content type is element-only.
  
  There's somethign wrong with the XML we're creating, see the `object_characteristics_extension` block - shouldn't be there.
  
  ```xml
  <premis:objectCharacteristicsExtension>
    <ffprobe>
    </ffprobe>object_characteristics_extension...
  <premis:objectCharacteristicsExtension>
  ```

* Attribute 'version' is not allowed to appear in element 'premis:object'.<br />
  Attribute 'version' is not allowed to appear in element 'premis:rightsStatement'.

  The error is clear, but why?

sevein:
I'll be addressing the validation issues in a different pull request.
The code is working, the new transfer METS is generated and it includes rights!

sevein (approved):

PR #1388, opened 2019-03-28
ES: increase total fields limit in all indexes

jraddaoui:
Connects to https://github.com/archivematica/Issues/issues/608.


ross-spencer (approved):
Nice change. LGTM @jraddaoui.

PR #1390, opened 2019-04-01
Dashboard: Improve appraisal arrangement performance

replaceafill:
This fixes the **tag and click** functionality of the Appraisal tab when many tagged files are included in the arrangement. It uses [Django's bulk_create](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create) to create arrange directories and files to avoid hitting the database too fast too many times which causes `OperationalError: (1040, 'Too many connections')` errors and updates the JS code accordingly.

Related to https://github.com/archivematica/Issues/issues/549

Connects to https://github.com/archivematica/Issues/issues/610


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -622,20 +624,37 @@ def create_arrange_directory(path):
...
     error = None
 
-    path = base64.b64decode(request.POST.get("path", ""))
-
-    if path:
-        try:
-            create_arrange_directory(path)
-        except ValueError as e:
-            error = str(e)
+    if "path[]" in request.POST:
```
cole: I'm not sure why the behaviour needs to be different with one vs many directories. It seems like we could just always do `request.POST.getlist("path", []))`, and we don't need both `create_arrange_directories` and `create_arrange_directory`

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -622,20 +624,37 @@ def create_arrange_directory(path):
...
 
+def create_arrange_directories(paths):
+    arranges = [
+        models.SIPArrange(
+            original_path=None,
+            arrange_path=os.path.join(path, ""),  # ensure ends with /
+            file_uuid=None,
+        )
+        for path in paths
+        if path.startswith(DEFAULT_ARRANGE_PATH)
```
cole: I think we should probably replicate the error behaviour here (it's more of a pain with multiple but I think you can just do a guard at the start of the function, something like:
```
        for path in paths:
            if not path.startswith(DEFAULT_ARRANGE_PATH):
                raise ValueError(_("All directories must be within the arrange directory.")
```

cole (commented):
Nice! This should be way faster. I think we can simplify the python stuff a bit, comments inline.

src/dashboard/src/components/filesystem_ajax/views.py
replaceafill: It doesn't. I removed `create_arrange_directory` and updated the callers.

src/dashboard/src/components/filesystem_ajax/views.py
replaceafill: :+1:

cole (approved):
Nice work!

replaceafill:
Thanks a lot for your help @cole!

PR #1392, opened 2019-04-01
ES: set indexes depth limit to 1000

jraddaoui:
Connects to https://github.com/archivematica/Issues/issues/608.

PR #1393, opened 2019-04-01
ES: set indexes depth limit to 1000

jraddaoui:
Connects to https://github.com/archivematica/Issues/issues/608.


jraddaoui:
Hi @replaceafill, this is the same as #1392 but for stable/1.9.x. Thanks!

replaceafill (approved):
:+1:

PR #1394, opened 2019-04-01
Add premis:intellectualEntity entries for directories

cole:
Also add any extra identifiers for all PREMIS objects.

Connects to https://github.com/archivematica/Issues/issues/442.


sevein (approved):
It's working! :clap:

PR #1396, opened 2019-04-02
Create METS continue on error

helenst:
Addresses https://github.com/archivematica/Issues/issues/620

### DIFF
--- a/src/MCPClient/lib/clientScripts/create_mets_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_v2.py
@@ -1033,6 +1033,7 @@ def createFileSec(
                     'No uuid for file: "', directoryPathSTR, '"', file=sys.stderr
                 )
                 state.error_accumulator.error_count += 1
+                continue
 
             use = f.filegrpuse
             label = f.label
--- a/src/MCPClient/tests/fixtures/custom_structmaps/model/files.json
+++ b/src/MCPClient/tests/fixtures/custom_structmaps/model/files.json
@@ -270,6 +270,24 @@
     }
 },
 {
+    "pk": "707696d4-5526-4908-b746-345356209bd3",
+    "model": "main.file",
+    "fields": {
+        "filegrpuuid": "",
+        "sip": "3a915449-d1bb-4920-b274-c917c7bb5929",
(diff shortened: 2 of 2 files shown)


ross-spencer (commented):
LGTM! :+1:

ross-spencer (approved):

ross-spencer:
Hi @helenst I think to merge this into `qa/1.x` I need to ask you to rebase against the current `qa/1.x` and then I'll merge from the rebased branch. I can't think of another way to do it without creating another branch on here, but let me know if you have another approach?

helenst:
After a rebase, it started breaking a (presumably newly added) test, `test_create_file_sec`. Which was failing because the number of file entries in the xml didn't match the number of files on disk. It was passing in `qa/1.x` because the `FileDoesNotExist` exception was thrown due to a missing entry in the fixture and because this was on an iteration other than the first one, `f` was already defined but holding the entry from the previous iteration, resulting in incorrect xml output (rather than outright failure). I've fixed it by adding the entry to the fixture. I don't know if this is the correct fix, what do you think?

Also, now it is failing due to linting issues which come from qa/1.x.

ross-spencer:
@helenst yep, looks like the correct fix to me. That was a new test, and also, that structmap happened to be the last structmap xml I added to the test! I hope we get the opportunity to better isolate these tests/fixtures but it's gonna be fiddly! 

Will try and rebase and merge this work asap!

ross-spencer (approved):

ross-spencer:
Hi @helenst I've been wrestling with Git to try and rebase and merge this without either:

1. Asking you to rebase it again so I can merge it.
2. Creating a new PR against it to then merge it.

Both seem sub-optimal with the work already in place. I can't figure out another way though so can I ask you to do 1. again for me? 

I'll try and get this merged asap again today. 

Do you have a preferred approach to merges in future? Is this approach okay? Can you think of another way to do this with the protections we currently have on the qa branch?

helenst:
@ross-spencer This is now rebased against the latest qa/1.x - I'm really not sure how to do things any better though in terms of merging :)

ross-spencer:
Merged!! :tada: :tada: :tada: 

As long as it's not too patience challenging then it's okay I guess and our branches aren't always moving so fast.

PR #1397, opened 2019-04-02
Optimize db access in create_transfer_mets

cole:
On my laptop, 200 seconds for a transfer containing 5k simple files, vs 284 before.

Connects to https://github.com/archivematica/Issues/issues/315.


sevein (approved):
I've just tested this.
It works, it's faster and more readable!

👏👏👏

PR #1398, opened 2019-04-02
ES Transfer: try to guess ingest_date

sevein:
Before, it was using datetime.today() at all times. This pull request changes the code so it tries to guess the date when there's data associated to the transfer in the database.

Connects to https://github.com/archivematica/Issues/issues/367.

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -33,7 +33,7 @@ import sys
 import time
 from lxml import etree as ElementTree
 
-from django.db.models import Q
+from django.db.models import Min, Q
 from django.utils.six.moves import xrange
 from main.models import File, Transfer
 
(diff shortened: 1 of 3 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -628,25 +628,39 @@ def index_transfer_and_files(client, uuid, path, printfn=print):
...
+        transfer_name = transfer.currentlocation.split("/")[-2]
+        # It doesn't seem that Archivematica records the ingestion date
+        # associated with the Transfer but we can look at the earliest file
+        # entry instead - as long as there is a match which may not always be
+        # the case.
+        ingest_date = (
+            str(
+                File.objects.filter(transfer=transfer)
+                .aggregate(Min("enteredsystem"))["enteredsystem__min"]
+                .date()
```
cole: I think it's pretty unlikely, but if you have no files, I think `enteredsystem__min` will be None, which will raise AttributeError.

cole (approved):
Nice, looks good. I think there's maybe an issue with a Transfer without files (can that happen?), noted inline.

src/archivematicaCommon/lib/elasticSearchFunctions.py
sevein: riiiight

cole (approved):

PR #1400, opened 2019-04-04
Validation API

sevein:
This pull request adds the [proposed](https://adr.archivematica.org/0005-new-api-endpoint-for-csv-validation.html) validation API.

* URL: `/api/v2beta/validate/avalon`
* Content-Type: `text/csv; charset=utf-8`
* Body: `<document>`

Usage example: (assuming that `avalon.csv` exists)

```bash
curl http://127.0.0.1:62080/api/v2beta/validate/avalon \
  --data-binary "@avalon.csv" \
  --header "Authorization: ApiKey test:test" \
  --header "Content-Type: text/csv; charset=utf-8"
```

Note: do not use `--data`, use `--data-binary` instead.

Response examples:

- **200 OK**
  ```json
  {
    "valid": true
  }
  ```

- **400 Bad Request**
  (expect `reason` to include different validation errors)
  ```json
  {
    "valid": false,
    "reason": "Administrative data must include reference name and author."
  }
  ```

- **404 Not Found**
  ```json
  {
    "error": true,
    "message": "Unknown validator. Accepted values: avalon"
  }

See also: [ADR-0005](https://adr.archivematica.org/0005-new-api-endpoint-for-csv-validation.html).
Connects to https://github.com/archivematica/Issues/issues/618.

### DIFF
--- a/src/dashboard/src/components/api/validators.py
+++ b/src/dashboard/src/components/api/validators.py
@@ -1,24 +1,32 @@
-"""Validation of transfer documents."""
+"""Validation of transfer documents.
 
-from base64 import b64decode
+Usage example::
+
+    >> validator = get_validator("avalon")
+    >> validator.validate(b"...")
+    True
+
+Expect the ``validate`` method to raise ``ValidationError`` when validation
+does not pass. The exception may include a message with more details.
+
+New validators must be added to the `` _VALIDATORS`` registry.
+"""
+
 import collections
 import csv
 from io import BytesIO
 
 
 class ValidationError(Exception):
-    pass
+    """Validators should raise this exception when the input is invalid."""
 
 
 class BaseValidator(object):
-    def _decode(self, s):
-        try:
-            return b64decode(s)
-        except Exception as err:  # Py3: binascii.Error?
-            raise ValidationError("Base64 decoding failed: {}".format(err))
+    def validate(self, string):
+        """Validator must implement this method.
 
-    def validate(self, s):
-        """Implementors are expected to raise ``ValidationError`` if needed."""
+        Raise ``ValidationError`` when a validation error occurs. Otherwise,
+        return ``True``."""
         raise NotImplementedError
 
 
@@ -206,9 +214,11 @@ class AvalonValidator(BaseValidator):
                         )
                     )
 
-    def validate(self, s):
-        csvr = csv.reader(BytesIO(super(AvalonValidator, self)._decode(s)))
+    def validate(self, string):
+        csvr = csv.reader(BytesIO(string))
+        empty = True
         for i, row in enumerate(csvr):
+            empty = False
             if i == 0:
                 self._check_admin_data(row)
             if i == 1:
@@ -219,12 +229,15 @@ class AvalonValidator(BaseValidator):
             if i >= 2:
                 self._check_file_exts(row, file_cols)
                 self._check_op_fields(row, op_cols)
+        if empty:
+            raise ValidationError("The document is empty.")
+        return True
 
 
 _VALIDATORS = {"avalon": AvalonValidator}
 
 
-class ValidatorNotAvailableError(Exception):
+class ValidatorNotAvailableError(ValueError):
     default = "Unknown validator. Accepted values: {}".format(
         ",".join(_VALIDATORS.keys())
     )
@@ -235,7 +248,7 @@ class ValidatorNotAvailableError(Exception):
         super(ValidatorNotAvailableError, self).__init__(*args, **kwargs)
 
 
-def validator(name):
+def get_validator(name):
     try:
         klass = _VALIDATORS[name]
     except KeyError:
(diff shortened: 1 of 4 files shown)


src/dashboard/src/components/api/validators.py
```diff
@@ -0,0 +1,243 @@
...
+                self._check_field_pairs(row)
+            if i >= 2:
+                self._check_file_exts(row, file_cols)
+                self._check_op_fields(row, op_cols)
+
+
+_VALIDATORS = {"avalon": AvalonValidator}
+
+
+class ValidatorNotAvailableError(Exception):
```
cole: It may make more sense to subclass [ValueError]( https://docs.python.org/2/library/exceptions.html#exceptions.ValueError) here.

src/dashboard/src/components/api/views.py
```diff
@@ -824,3 +829,29 @@ def _package_create(request):
         LOGGER.error("{}: {}".format(msg, err))
         return helpers.json_response({"error": True, "message": msg}, 500)
     return helpers.json_response({"id": id_}, 202)
+
+
+@_api_endpoint(expected_methods=["POST"])
+def validate(request):
+    try:
+        body = json.loads(request.body)
+    except Exception:
```
cole: `    except ValueError:`

src/dashboard/src/components/api/views.py
```diff
@@ -824,3 +829,29 @@ def _package_create(request):
...
+    if not validator_name or not payload:
+        return _error_response("Missing parameters")
+    try:
+        validator = validators.validator(validator_name)
+    except validators.ValidatorNotAvailableError as err:
+        return _error_response(six.text_type(err))
+    try:
+        validator.validate(payload)
+    except validators.ValidationError as err:
+        return _ok_response({"valid": False, "reason": six.text_type(err)})
```
cole: This should be HTTP 400 response.

cole (changes requested):
Nice, this looks pretty good!

The main thing I see here I'm still confused by what we are achieving with base64 encoding — it seems like an unnecessary extra step. Do we need to support other encodings than UTF-8? I would highly recommend only supporting UTF-8, in which case I think it makes sense to fail early if the encoding is invalid.

There's also a few minor things inline.

ablwr:
I think the base64 encoding provides value. People come up with the weirdest, most unexpected things...

sevein:
Thanks @cole and @ablwr for the feedback. I've updated my changes so the document can be embedded in the body of the request. See the updated PR description for usage examples.

Now the encoding and mime type can be determined using the `Content-Type` header - for now I'm just expecting `"text/csv; charset=utf-8"` but that should give us some flexibility in the future.

src/dashboard/src/components/api/views.py
cole: I should clarify this comment a bit for posterity — there is a case to be made either way, like all REST API status codes, it's a bit of a judgement call. You can make the argument that validation succeeded and the result was failure, but I think an error status code is more useful for clients if validation failed.

cole (approved):
Nice, I like this solution. Looks good!

ablwr (approved):

ablwr:
Tagging @helrond in case Hillel is curious as to the implementation of this. Could you see this being extended for a rights.csv or metadata.csv?

helrond:
@ablwr Yes, this looks _exactly_ like what I had in mind for validating those other CSV files! Am I right in thinking that there isn't validation logic elsewhere in the application that could be leveraged for rights.csv and/or metadata.csv, so this would have to be extended with additional validators?

ablwr:
@helrond Yeah, that's right. The only validation logic right now is for Avalon Media System. I hope we can add other validators in the future, though! I spend a lot of time helping people puzzle over metadata.csv.

helrond:
@ablwr This looks like it will be super easy to extend, and this might actually be a place where we could jump in with a PR if that's desirable. Thanks for bringing it to my attention!

ablwr (changes requested):
Getting incorrect results here when testing -- investigating and will follow up with more feedback.

sevein:
> Getting incorrect results here when testing -- investigating and will follow up with more feedback.

Ok @ablwr thanks for testing. I'm curious to hear more.

Here's the command that I've been using to just run the tests under the new TestValidate class:

    docker-compose exec archivematica-dashboard bash -c "cd /src/dashboard && py.test -p no:cacheprovider --reuse-db -x -v -k TestValidate"

ablwr:
I am getting `{"valid": true}` whenever I pass in a CSV that is not valid.

If I pass in something that doesn't exist, I get the expected error. 

It's funny, I am getting inconsistent results and trying to replicate them, because previously I was receiving an error, but of the wrong type, and for everything I passed into it. It is almost as if the API is getting stuck in some sort of way -- could it always be testing on the first file passed in instead of new data? I don't know exactly what is wrong, but CSVs that should fail are just returning as valid:true.

ablwr:
OK, the issue was cURL not being happy with linebreaks, and @sevein changing the example from `--data` to `--data-binary` sorted it all out! I also tested out this most recent update, in which @sevein checks that a file has at least three rows. :tada:

ablwr (approved):

Issue #1118, opened 2018-06-04
Problem: "cleanup names" produces too much output


jrwdunham:
# Problem

The "cleanup names" micro-service runs once per package and will print at least one line to stdout and stderr for each file in the package. For very large transfers/SIPs, this can mean many MBs of output that must be stored in the Tasks table. This leads to a bloated Tasks table and, worse, can lead to dropped connections to the MySQL server, which Archivematica's MCP Server does not handle gracefully: the entire daemon process exits.  This causes all processing to halt and the user must restart the Archivematica services and restart the transfer.

# Workaround

The `max_allowed_packet` setting in MySQL is 1MB by default. Setting a higher value for the `max_allowed_packet` variable in the mysql config file will provide an immediate work around to this issue.  Based on tests with a transfer containing 3174 files, the maximum packet created was just under 6MB.  Assuming a desired transfer size of 21,000, it is conceivable that the current version of Archivematica could attempt to construct a packet that is up to 7 times larger (35MB).  Tests were run successfully with the value set to 32MB.  It is probably safe to set it as high as 128MB, but it may be better to also look at some of the longer term solutions listed below, and stick with a value like 32MB or 64MB.

The steps we followed to implement this workaround were:

1. Stop services
2. open (as root) the mysql config file at /etc/my.cnf.d/server.conf for editing (we used vi).
3. add the following entries to the section labelled [mariadb]:

       [mariadb]
       innodb_use_sys_malloc = 0
       innodb_log_file_size = 32M
       innodb_log_files_in_group = 4
       innodb_buffer_pool_size = 512M
       innodb_log_buffer_size = 2M
       innodb_file_per_table = 1
       join_buffer_size = 256K
       thread_cache_size = 8
       query_cache_size = 0
       query_cache_type = 0
       query_cache_limit = 2M
       max_allowed_packet = 32M

   Note that only the last variable (`max_allowed_packet`) has to be added.  All of the others were mysql tuning parameters recommended by mysqltuner.pl (https://github.com/major/MySQLTuner-perl/).  These additional changes did not significantly affect our test results.

4. restart services

# Longer term solutions

1. Upgrade to 1.7.1 and disable output capture.
2. New development work to reduce the output generated by the most verbose preservation tasks, such as "clean up names" and "characterize and extract metadata".

ross-spencer:
[...]

Issue #1111, opened 2018-05-30
Problem: filenameCleanup.log isn't always written to following extract packages


ross-spencer:
Related to #1104. 

Attached is an AIP where the sanitization of cafè happens but isn't found in the log.

Stderr shows:
```
sanitizeObjectNames.py: INFO      2018-05-30 15:58:35,706  archivematica.mcp.client.sanitizeObjectNames:sanitize_object_names:59:  sanitizations: /var/archivematica/sharedDirectory/watchedDirectories/workFlowDecisions/extractPackagesChoice/enc_5-6c305197-b4e7-4d39-9160-dbcfcc8d4a88/objects/ascii_cp437.zip-2018-05-30T15_58_32.317841_00_00/ascii_cp437/caf? -> /var/archivematica/sharedDirectory/watchedDirectories/workFlowDecisions/extractPackagesChoice/enc_5-6c305197-b4e7-4d39-9160-dbcfcc8d4a88/objects/ascii_cp437.zip-2018-05-30T15_58_32.317841_00_00/ascii_cp437/caf
sanitizeObjectNames.py: INFO      2018-05-30 15:58:35,706  archivematica.mcp.client.sanitizeObjectNames:sanitize_object_names:59:  sanitizations: /var/archivematica/sharedDirectory/watchedDirectories/workFlowDecisions/extractPackagesChoice/enc_5-6c305197-b4e7-4d39-9160-dbcfcc8d4a88/objects/ascii_cp437.zip-2018-05-30T15_58_32.317841_00_00/ascii_cp437/caf/caf? -> /var/archivematica/sharedDirectory/watchedDirectories/workFlowDecisions/extractPackagesChoice/enc_5-6c305197-b4e7-4d39-9160-dbcfcc8d4a88/objects/ascii_cp437.zip-2018-05-30T15_58_32.317841_00_00/ascii_cp437/caf/caf
[...]

PR #1401, opened 2019-04-05
Sanitize names faster

cole:
~~WIP, but so far seems to be about 3x faster (around 3 vs 10 secs for 5k corrections).~~ ~~This is a big change, it needs some better tests.~~ ~~We can also improve perf further probably by removing more logging (as I think @sevein noted somewhere that is a big delay) and removing the use of `updateFileLocation` which does another (completely unnecessary) SELECT query per file right now.~~
EDIT: 🤦‍♂️ Due to a path mismatch I've been profiling _not_ sanitizing 5k files. It's more like 200 secs to actually sanitize them, and `updateFileLocation` is the source of most of the slowness. This PR will need some revision.

Connects to https://github.com/archivematica/Issues/issues/315.

### DIFF
--- a/src/MCPClient/lib/clientScripts/sanitize_names.py
+++ b/src/MCPClient/lib/clientScripts/sanitize_names.py
@@ -38,10 +38,18 @@ REPLACEMENT_CHAR = "_"
 
 
 def sanitize_name(basename):
-    unicode_name = unidecode(strToUnicode(basename))
-    # Handle the case where '' is returned by unidecode
+    if basename == "":
+        raise ValueError("sanitize_name recieved an empty filename.")
+    unicode_basename = strToUnicode(basename)
+    unicode_name = unidecode(unicode_basename)
+    # We can't return  an empty string here because it will become the new filename.
+    # However, in some cases unidecode just strips out all chars (e.g.
+    # unidecode(u"🚀") == ""), so if that happens, we to replace the invalid chars with
+    # REPLACEMENT_CHAR. This will result in a filename of one or more underscores,
+    # which isn't great, but allows processing to continue.
     if unicode_name == "":
-        unicode_name = strToUnicode(basename)
+        unicode_name = unicode_basename
+
     return ALLOWED_CHARS.sub(REPLACEMENT_CHAR, unicode_name)
 
 
--- a/src/MCPClient/lib/clientScripts/sanitize_names.py
+++ b/src/MCPClient/lib/clientScripts/sanitize_names.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python2
-
+# -*- coding: utf8
 # This file is part of Archivematica.
 #
 # Copyright 2010-2013 Artefactual Systems Inc. <http://artefactual.com>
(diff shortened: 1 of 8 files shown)


src/MCPClient/tests/test_sanitize.py
```diff
@@ -128,3 +130,71 @@ def test_sanitize_object_names(self):
         finally:
             # Delete files
             shutil.rmtree(transfer_path)
+
+
+@pytest.mark.parametrize(
+    "basename, expected_name", [("helloworld", "helloworld"), (u"a\x80b", "ab")]
```
alexwlchan: Maybe throw in `Smörgåsbord.txt` as well, as a regression test for #352?

(I haven’t tested it with the old code; `a\x80b` is just what I happened to discover.)

Also I would double-check if my code does anything sensible with the case where all the characters get sanitised away, e.g. `\x80`. I’d want to test it doesn’t return an empty string in that case.

cole: So...it does return an empty string, but it did before your changes (`unidecode(u"\x80") == ""`). I'll put a check in for that case.

cole:
OK, round 2 is done and ready for review. Seems to take around 20% of the time (40 vs 200 seconds for 5k files), although test times are pretty variable on my system.

src/MCPClient/lib/clientScripts/sanitize_names.py
```diff
@@ -21,100 +21,74 @@
...
-            ret += c
-        else:
-            ret += replacementChar
-    return ret.encode("utf-8")
+def sanitize_name(basename):
+    unicode_name = unidecode(strToUnicode(basename))
+    # Handle the case where '' is returned by unidecode
+    if unicode_name == "":
+        unicode_name = strToUnicode(basename)
+    return ALLOWED_CHARS.sub(u"_", unicode_name)
```
alexwlchan: Should this be

```suggestion
    return ALLOWED_CHARS.sub(REPLACEMENT_CHAR, unicode_name)
```

?

src/MCPClient/lib/clientScripts/sanitize_object_names.py
```diff
@@ -20,140 +20,251 @@
...
+
+        self.dirs_index[old_path] = new_path
+
+        if len(self.dirs_index) >= self.BATCH_SIZE:
+            self.apply_dir_updates()
+
+    def sanitize_objects(self):
+        """
+        Iterate over the filesystem, sanitizing as we go. Updates made on disk
+        are batched and then applied to the datab
```
alexwlchan: “applied to the datab” – is this sentence incomplete?

alexwlchan (commented):
Two suggestions. I'm not familiar enough with Django database stuff to comment on that code, but the renaming logic looks pretty good. 👍

src/MCPClient/lib/clientScripts/sanitize_object_names.py
cole: Whoops! Thanks, I'll finish that thought.

sevein (commented):
I've rebased and squashed. I've also tested, it seems to run much faster!

One subtle change in `sanitizeObjectNames_v0.0` is that it stopped logging the filename changes, e.g.:

    logger.info("Sanitized name: %s -> %s", old_location, new_location)
    job.pyprint("Sanitized name:", old_location, " -> ", new_location)

That was captured in the database and (specified in workflow.json) also in `logs/filenameCleanup.log`.
Generally, output capturing as implemented today is not scalable, so having chatty client scripts is a problem regardless if they're also written to disk like `logs/filenameCleanup.log`.

So I'd say that this is a good move but we should probably consult this change around not having `logs/filenameCleanup.log`. Do we really need these records even when PREMIS events are recorded too? Note: dirname changes are not recorded as PREMIS events, but that was already true before this PR I believe.

Other issues relevant to my comment are:
- https://github.com/archivematica/Issues/issues/324
- https://github.com/artefactual/archivematica/issues/1111
- https://github.com/artefactual/archivematica/issues/1118

I'll bring this up to the analysts.

src/MCPClient/lib/clientScripts/sanitize_names.py
```diff
@@ -21,100 +21,74 @@
...
-    basename = transliterate(basename)
-    for c in basename:
-        if c in valid:
-            ret += c
-        else:
-            ret += replacementChar
-    return ret.encode("utf-8")
+def sanitize_name(basename):
+    unicode_name = unidecode(strToUnicode(basename))
+    # Handle the case where '' is returned by unidecode
```
alexwlchan: Does this comment add much? It would be more useful to explain why we don’t want to return an empty string here.

Are there circumstances where `strToUnicode` can return an empty string? What’s the best thing to do in that case?

alexwlchan (commented):
One comment about a comment.

sevein:
@cole, after a quick discussion in Slack, we've thought that the removal of `filenameCleanup.log` (and other log files too) is something that needs to be investigated further. Would it be possible to restore that bit?

src/MCPClient/lib/clientScripts/sanitize_names.py
cole: Yeah, on review, that's a pretty useless comment. Good catch.

The return value of this function becomes the new filename, so we don't want to return "" (there are also some assumptions made when checking for an existing file that will break down).

[strToUnicode](https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/archivematicaFunctions.py#L93) shouldn't return an empty string unless it's passed one. If that is happening, things are pretty broken, but I've added a ValueError for that case.

The if statement this comment is referring to handles the case where there is nothing in the filename that unidecode recognizes, e.g. if the filename is just an emoji: `🚀`. That's obviously a pretty rare case — the old function didn't bother handling it.

I've updated with a more detailed comment, let me know if it's clear.

cole:
@sevein OK, I've restored the original logging.

sevein (approved):
Thank you!

src/MCPClient/lib/clientScripts/sanitize_names.py
```diff
@@ -21,100 +21,82 @@
...
-from archivematicaFunctions import unicodeToStr
+
+from archivematicaFunctions import strToUnicode
 
 VERSION = "1.10." + "$Id$".split(" ")[1]
-valid = "-_.()" + string.ascii_letters + string.digits
-replacementChar = "_"
 
+# Letters, digits and a few punctuation characters
+ALLOWED_CHARS = re.compile(r"[^a-zA-Z0-9\-_.\(\)]")
```
ross-spencer, line 36: As part of archivematica/issues#315 it looks like we need to update the third paragraph [here](https://github.com/artefactual/archivematica-docs/blob/1.9/user-manual/transfer/clean-up-names.rst) in the docs. The regular expression translates to the same set of characters as I read it so it's likely just the line we link to.

cole: PR: https://github.com/artefactual/archivematica-docs/pull/292

PR #1404, opened 2019-04-11
Fixes to enable custom structmaps to be more expressive

ross-spencer:
**Fixes to enable custom structmaps to be more expressive**

Previously a user would need to submit a custom mets:structMap in
Archivematica using a FILEID attribute for a file path which would
be replaced during METS generation. The FILEID is only capable of
storing a limited set of values and so here we replace that
capability and ask users to use the CONTENTIDS attribute of a custom
structMap which can encode a larger range of values to ensure that
the final METS document remains valid.

Unit tests have been created to support maintenance of this feature
on an ongoing basis.

Validation of structMap input is completed during transfer and ingest
and this validation was previously been done by an external command
called xmllint. Given the opportunity to make the validation of METS
documents in Archivematica more consistent these calls in
Archivematica jobs to xmllint have been replaced with more idiomatic
calls to Python lxml validation methods.

Validation of mets:structMap files is now completed against
METS v1.12.

Other small fixes in this commit include the ability for the custom
structMap identifier to be updated dynamically where previously it
would always be structMap_2 which would be incompatible if other
automatically generated structMaps were included in the METS. Users
should note that they can also label a custom structMap with
something more meaningful to them using their own label and this will
not be replaced.

**From the original commit history**

This PR is separated into multiple commits and it is recommended to tackle them one by one for the code reviewer:

1. We implement new unit tests to demonstrate the existing capability.
2. We enhance the `create_mets_v2.py` code to support mode dynamic testing and add those tests.
3. Make sure FILEIDs are all handled.
4. We fix custom structmap numbering.
5. We add the fixes to `embed_custom_structmap` to enable more complex structures.
6. Add mets validation as structmap should be valid in its own right.
7. We do some minor refactoring to `create_mets_v2.py` to separate responsibilities somewhat.
8. We replace the verification shell script calling xmllint with Python.
9. We didn't correctly anticipate all tests required to accurately reflect a transfer layout, this commit deals with that.
10. Finally, we missed the sanitized names use-case where user submitted names might not anticipate AM's handling of that, so we deal with that here.

Feature: [Draft documentation](https://github.com/artefactual/archivematica-docs/blob/2a2df2ff140e53b1947c5b125a8f2f786f35731e/user-manual/transfer/import-metadata.rst#importing-structural-metadata-with-mets_structmapxml)
Sample-data for testing can be found here: https://github.com/artefactual/archivematica-sampledata/pull/52

Connected to archivematica/Issues#283
Connected to archivematica/Issues#633
Connected to archivematica/issues#245
Connected to https://github.com/IISH/archivematica/issues/5
Connected to https://github.com/IISH/archivematica/issues/4


src/MCPClient/lib/clientScripts/verify_mets.py
```diff
@@ -1,20 +1,34 @@
...
-            job.write_output(std_out)
-            job.set_status(exit_code)
+            mets_structmap = os.path.join(job.args[1], "metadata", "mets_structmap.xml")
+            mets_xsd = job.args[2]
+            if not os.path.isfile(mets_structmap):
+                job.pyprint("Custom structmap not supplied with package")
+                return
+            if not os.path.isfile(mets_xsd):
+                raise (VerifyMETSException("METS asset is unavailable"))
+            xmlschema = etree.XMLSchema(etree.parse(mets_xsd))
```
cole, line 30: Just to confirm, does `metsrw` not yet work for this case?

src/MCPClient/tests/test_create_aip_mets.py
```diff
@@ -1,7 +1,9 @@
 # -*- coding: utf8
+from __future__ import unicode_literals
```
cole, line 2: 👍

src/MCPClient/tests/test_create_aip_mets.py
```diff
@@ -744,3 +753,242 @@ def test_create_rights_granted(self):
...
+        os.path.join(THIS_DIR, "../lib/assets/mets/mets.xsd")
+    )
+
+    @staticmethod
+    def count_dir_objects(path):
+        """Count all objects on a given path tree."""
+        return sum([len(files) for _, dir_, files in os.walk(path)])
+
+    @staticmethod
+    def validate_mets(mets_xsd, mets_structmap):
```
cole, line 776: Again, just to confirm, `metsrw` can't be used here?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -872,39 +873,61 @@ def getAMDSec(
     return ret
 
 
-def getIncludedStructMap(job, baseDirectoryPath, state):
+def _fixup_path_input_by_user(job, path):
+    """Fix-up paths submitted by a user, e.g. in custom structmap examples so
+    that they don't have to anticipate the Archivematica normalization process.
+    """
+    return os.path.join(
+        "", *[sanitizeName(name.encode("utf8")) for name in path.split(os.path.sep)]
```
cole, line 881: The idea here is that these files will already have been sanitized, so we're just matching up to the current state on the filesystem/database, correct?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -872,39 +873,61 @@ def getAMDSec(
...
+        structMapXmlPath = os.path.join(dirPath, custom_structmap)
         if not os.path.isdir(dirPath):
             continue
         if os.path.isfile(structMapXmlPath):
             tree = etree.parse(structMapXmlPath)
-            root = (
-                tree.getroot()
-            )  # TDOD - not root to return, but sub element structMap
-            # print etree.tostring(root)
+            root = tree.getroot()
```
cole, line 906: Just to confirm, does metsrw not yet work for this case?

cole (commented):
This code looks good! A few questions inline.

I have one overall concern. If I understand [ADR-0001](https://adr.archivematica.org/0001-unification-of-mets-creation.html) correctly, we'd like to (over time) remove most of the XML generation code from `create_mets_v2`, and replace it with `metsrw` usage. It seems like this PR does not advance that goal. It doesn't make that task too much harder, but it seems like if we're touching this script in a significant way, it's a good opportunity to move in that direction. I do understand though that `metsrw` has some serious issues, that may prevent it from being used at all here, so maybe we need to just understand that while we have that goal, we can't advance it here (I don't know the urgency of this change, etc).

ross-spencer:
Hi @cole thanks for the quick input. For `metsrw` I investigated the possibility, and came up with this as the primary issue for not using it: https://github.com/archivematica/Issues/issues/636 and getting to this stage in the PR used the majority of the budgeted time as much as I would love to work my way through it.

In prepping the way for `metsrw`:
* We have removed a shell script calling `xmllint` so brought scripts into line using `lxml`.
* We use the same validation call in the microservices as we do in the unit tests. 
* Created unit tests for the new feature to help with refactoring to use `metsrw` in future.

I approached it as taking one step forward.

cole:
@ross-spencer fair enough! Moving create_mets_v2 to metsrw is a big job, maybe it's wishful thinking that we can tackle it incrementally. Thanks for the detailed issues you've created (https://github.com/archivematica/Issues/issues/636 and related).

ross-spencer:
@cole yeah, but I'm looking forward to making it happen though :nerd_face: . If there's an appetite to have `lxml` and `metsrw` mixing with each other for a few releases, then that's useful to know too.

cole (approved):

src/MCPClient/lib/clientScripts/create_mets_v2.py
ross-spencer: Sorry i missed this @cole, yep! That's right. We'll have to keep this in mind when you complete your work on that microservice job.

PR #1405, opened 2019-04-16
Fix aips and aipfiles ES mappings

hakamine:
elasticSearchFunctions now use lxml.etree instead of ElementTree, which
does not replace namespace prefixes with its own ones. Fix mappings
accordingly.

Connected to https://github.com/archivematica/Issues/issues/619

### DIFF
--- a/src/archivematicaCommon/lib/elasticsearch/aipfiles_mets_mapping.json
+++ b/src/archivematicaCommon/lib/elasticsearch/aipfiles_mets_mapping.json
@@ -2,18 +2,18 @@
   "properties": {
     "dmdSec": {
       "properties": {
-        "ns0:xmlData_dict_list": {
+        "mets:xmlData_dict_list": {
           "properties": {
-            "@xmlns:ns1": {
+            "@xmlns:premis": {
               "type": "text"
             },
-            "@xmlns:ns0": {
+            "@xmlns:mets": {
               "type": "text"
             },
             "@xmlns:xsi": {
               "type": "text"
             },
-            "ns1:dublincore_dict_list": {
+            "premis:dublincore_dict_list": {
               "properties": {
                 "dc:available": {
                   "type": "text"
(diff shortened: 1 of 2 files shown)


jraddaoui (changes requested):
Thanks for looking into this @hakamine!

While I really like the mapping changes, I'm not sure if we should make them at the moment. The main reason is that we can't fully trust in the re-index commands and we'll need another upgrade path to reflect this changes in existing indexes.

The current indexes mapping, as you sure noticed, requires a few more enhances to be more more sustainable and I think we should try to tackle them together in https://github.com/archivematica/Issues/issues/404 alongside other ES setup changes. However, I don't know if we'll have the time to do it before 1.10, but if we have it, this is a great start.

If we don't have the time to modify the indexes right now, I wonder if moving from `xml.etree.ElementTree` to `lxml.etree` in `elasticSearchFunctions` was completely required after the PREMIS 3 upgrade, maybe not doing it stills introduces namespace changes (for example: ns1 gets moved to ns2). If reverting to `xml.etree.ElementTree` is not possible or does not work, we have some [normalization functions](https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/elasticSearchFunctions.py#L821-L878) that may be extended to transform the new namespace names to the old ones.

Please, let me know your thoughts.

hakamine:
Hi @jraddaoui, thank you very much, I agree with your comments.

The move from `xml.etree.ElementTree` to `lxml.etree` in `elasticSearchFunctions` was done because the PREMIS3/old AIP handling code fixes ([AM issue 24](https://github.com/archivematica/Issues/issues/24)) somehow broke the indexing of 
AIP files. Changing the library (among other changes) fixed this, but this also changes the indexes producing the issues you mentioned.

I will close this PR. I opened a new [PR 1408](https://github.com/artefactual/archivematica/pull/1408/files) (still WIP) to try to fix the indexing problem while reverting the move from `xml.etree.ElementTree` to `lxml.etree` in `elasticSearchFunctions`

PR #1410, opened 2019-04-18
PID declaration feature

ross-spencer:
This feature is broken into different commits as logically as was possible:

* [x] [Introduce unit tests to the PID binding feature to avoid feature shift.](https://github.com/artefactual/archivematica/pull/1410/commits/c2df783d27f324f86b813100efff7e3db367769b)
* [x] [Introduce a new declare PIDs microservice and changes to unit tests in reaction to understanding Bind PIDs better as a result.](https://github.com/artefactual/archivematica/pull/1410/commits/de4cf440569a08efd98485017abedeaf61dd2f0b) 
* [x] [Introduce additional configuration checks for PID binding so that the microservice doesn't fail if accidentally enabled with zero-configuration.](https://github.com/artefactual/archivematica/pull/1410/commits/3e9113d7bd0326794b0692df73340cb21bb5ad4b)
* [x] [Pylint changes prior to CR](https://github.com/artefactual/archivematica/pull/1410/commits/99a112fca386513b24374c29cb4dc63f18744fee)
* [x] [First round of CR changes](https://github.com/artefactual/archivematica/pull/1410/commits/a39783e0dde1615c6fa792cf9dac3aa1a2c82be4).

Connected to archivematica/issues#133
Replaces https://github.com/artefactual/archivematica/pull/1300 
Sample data here: https://github.com/artefactual/archivematica-sampledata/pull/53

### DIFF
--- a/src/MCPClient/lib/clientScripts/pid_declaration.py
+++ b/src/MCPClient/lib/clientScripts/pid_declaration.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python2
+# -*- coding: utf8
+
+"""pid_declaration.py
+
+Given an identifiers.json file, supplying third-party persistent identifiers
+(PIDS), such as handle (hdl) or doi, identifiers; associate those identifiers
+with the objects in the transfer, so be translated to PREMIS objects in the
+AIP METS.
+"""
+from __future__ import unicode_literals
+from functools import wraps
+import json
+import os
+import sys
+
+import django
+
+django.setup()
+
+from main.models import Directory, File, SIP
+
+from sanitize_names import sanitizeName
+
+
+class DeclarePIDsException(Exception):
+    """Exception to raise if there's a more fundamental issue with the running
+    of this script and we want to exit and stop the workflow from continuing.
+    """
+
+
+class DeclarePIDsExceptionNonCritical(Exception):
+    """Exception to raise when we have problems trying to open identifiers.json
+    or the file doesn't exist.
+    """
+
+    exit_code = 0
+
+
+class DeclarePIDs(object):
+    """Class to wrap PID declaration features and provide some mechanism of
+    recording state.
+    """
+
+    SIP_IDENTIFIER = "SIP"
+    SIP_DIRECTORY = "%SIPDirectory%"
+
+    def __init__(self, job):
+        self.job = job
+        self.identifier_count = 0
+        self.actual_count = 0
+
+    def exit_on_known_exception(func):
+        """Decorator to allows us to raise an exception but still exit-zero when
+        the exception is cleaner to return than ad-hoc integer values.
+        """
+
+        @wraps(func)
+        def wrapped(*args, **kwargs):
+            try:
+                func(*args, **kwargs)
+            except (DeclarePIDsExceptionNonCritical) as err:
+                return err.exit_code
+
+        return wrapped
+
+    def _fixup_path_input_by_user(self, path):
+        """Fix-up paths submitted by a user, e.g. in custom structmap examples
+        so that they don't have to anticipate the Archivematica normalization
+        process.
+        """
+        return "{}{}".format(
+            self.SIP_DIRECTORY,
+            os.path.join(
+                "",
+                *[sanitizeName(name.encode("utf8")) for name in path.split(os.path.sep)]
+            ),
+        )
+
+    def _log_identifer_to_stdout(self, mdl, identifier_type, identifier):
+        """Create a consistent string to log information about our identifiers
+        to stdout for the user.
+        """
+        try:
+            if isinstance(mdl, SIP):
+                msg = "Identifier {}: {} added for: {}: {}".format(
+                    identifier_type,
+                    identifier,
+                    "SIP",
+                    os.path.basename(os.path.split(mdl.currentpath)[0]),
+                )
+            elif isinstance(mdl, Directory):
+                msg = "Identifier {}: {} added for: {}: {}".format(
+                    identifier_type,
+                    identifier,
+                    "Directory",
+                    os.path.basename(os.path.split(mdl.currentlocation)[0]),
+                )
+            else:
+                msg = "Identifier {}: {} added for: {}: {}".format(
+                    identifier_type,
+                    identifier,
+                    "File",
+                    os.path.basename(mdl.currentlocation),
+                )
+            self.job.pyprint(msg)
+        except KeyError:
+            self.job.pyprint(
+                "Identifier {}: {} added to {}".format(identifier_type, identifier, mdl)
+            )
+
+    def _validate_identifier(self, mdl, id_):
+        """Provide some feedback for the user if the identifier information is
+        somehow incomplete.
+        """
+        identifier_type = id_.get("identifierType", None)
+        identifier = id_.get("identifier", None)
+        if identifier_type is None:
+            self.job.pyprint(
+                "None value returned for identifier type: {} on object: {}".format(
+                    id_, mdl
+                ),
+                file=sys.stderr,
+            )
+            return (False,)
+        if identifier is None:
+            self.job.pyprint(
+                "None value returned for identifier: {} on object: {}".format(id_, mdl),
+                file=sys.stderr,
+            )
+            return (False,)
+        return identifier_type, identifier
+
+    def _add_identifier_to_model(self, mdl, id_):
+        """Add custom identifier to the model we found a match for. We treat
+        incomplete information as a warning which then allows the script to
+        continue. This behavior will be easy to change if required with further
+        usage and greater understanding of this feature.
+        """
+        identifier_type, identifier = self._validate_identifier(mdl, id_)
+        if identifier_type:
+            mdl.add_custom_identifier(identifier_type, identifier)
+            self._log_identifer_to_stdout(mdl, identifier_type, identifier)
+            self.actual_count += 1
+
+    def parse_and_attach_identifiers(self, unit_uuid, json_data):
+        """Take our identifiers and add them to the model objects that we want
+        them to belong to by the time we create the AIP METS.
+        """
+        self.identifier_count = len(json_data)
+        for file_object in json_data:
+            file_path = file_object.get("file")
+            identifiers = file_object.get("identifiers", [])
+            objects_dir, base_name = os.path.split(file_path)
+            if (objects_dir == "objects" or objects_dir == "") and not base_name:
+                file_path = self.SIP_IDENTIFIER
+            else:
+                # Run name normalization to allow us to pair our data with
+                # information already in the database.
+                file_path = self._fixup_path_input_by_user(file_path)
+            for id_ in identifiers:
+                if file_path == self.SIP_IDENTIFIER:
+                    mdl = SIP.objects.get(uuid=unit_uuid)
+                    self._add_identifier_to_model(mdl, id_)
+                    continue
+                try:
+                    mdl = File.objects.get(sip_id=unit_uuid, currentlocation=file_path)
+                    self._add_identifier_to_model(mdl, id_)
+                    continue
+                except File.DoesNotExist:
+                    pass
+                try:
+                    mdl = Directory.objects.get(
+                        sip_id=unit_uuid, currentlocation=os.path.join(file_path, "")
+                    )
+                    self._add_identifier_to_model(mdl, id_)
+                except Directory.DoesNotExist:
+                    pass
+        self.job.pyprint(
+            "{} identifiers added for {} objects in the package".format(
+                self.actual_count, self.identifier_count
+            )
+        )
+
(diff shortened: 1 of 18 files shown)


ross-spencer:
**Potential location for a PID declaration microservice**

In-between `Remove empty normalization directories` and `Bind PIDs?`:

image

cole (approved):

src/MCPClient/lib/clientScripts/pid_component_helpers.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
cole: This is definitely an existing pattern (for example I think sanitize names uses it), but it seems like having `clientScripts` not include any helper code (just for actual scripts) might ease the complexity of the import relationships in AM. I'm not really sure where these helpers should go then... maybe common?

Another option in this case is since these are relatively simple helpers, is just duplicating them.

Flagging mostly as a point of discussion, I'm not sure this needs to change in this PR.

src/MCPClient/lib/clientScripts/pid_component_helpers.py
```diff
@@ -0,0 +1,26 @@
...
+def _add_custom_pid_to_mdl_identifiers(mdl, scheme, value):
+    """Create an identifier with scheme:value and add the row to the
+    Identifiers table. Reference the new identifier in the given model (mdl)
+    table.
+    """
+    if scheme is None or value is None:
+        msg = "None value returned for identifier scheme: '{}'' or value: '{}'".format(
+            scheme, value
+        )
+        raise AssignPIDsException(msg)
```
cole: I think this can be more useful as two separate exceptions/messages, so callers know which is missing. Also, the custom exception pattern is pretty widespread in AM, but `ValueError` would probably also be fine here.

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,212 @@
...
+class DeclarePIDs(object):
+    """Class to wrap PID declaration features and provide some mechanism of
+    recording state.
+    """
+
+    SIP_IDENTIFIER = "SIP"
+    SIP_DIRECTORY = "%SIPDirectory%"
+
+    def __init__(self, job):
+        """Class constructor"""
```
cole: I think this docstring can just be removed, it doesn't really tell us anything.

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,212 @@
...
+        self.identifier_count = 0
+        self.actual_count = 0
+
+    def exit_on_known_exception(func):
+        """Decorator to allows us to raise an exception but still exit-zero when
+        the exception is cleaner to return than ad-hoc integer values.
+        """
+
+        @wraps(func)
+        def wrapped(*_args, **kwargs):
```
cole: `*_args` -> `*args`? It doesn't make any difference, I'm mostly just confused by the underscore, what does that signify?

src/MCPClient/lib/clientScripts/pid_component_helpers.py
```diff
@@ -0,0 +1,26 @@
...
+    identifier = Identifier.objects.create(type=scheme, value=value)
+    mdl.identifiers.add(identifier)
```
cole: You can use `mdl.identifiers.create(type=scheme, value=value)` here.

src/MCPClient/lib/clientScripts/pid_component_helpers.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+import bindpid
+from main.models import Identifier
+
+
+class AssignPIDsException(Exception):
+    """Exception to raise if a validation or assignment error occurs."""
+
+
+def _add_custom_pid_to_mdl_identifiers(mdl, scheme, value):
```
cole: It seems like this would be most useful as a model method on the `SIP` `Directory`, and `File` models (so you can do `sip.add_custom_pid` or similar).

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,208 @@
...
+from sanitize_names import sanitizeName
+
+
+class DeclarePIDsException(Exception):
+    """Exception to raise if there's a more fundamental issue with the running
+    of this script and we want to exit and stop the workflow from continuing.
+    """
+
+
+class DeclarePIDsWarning(Exception):
```
cole: I'm not totally clear on the behaviour desired — it seems like we want to short circuit the script but still exit with success? If that's the case it seems like we don't need the warning and context manager, we can just add an except for the error in `call`.

I find the naming on this a bit confusing, if something is named `Warning`, I'd expect something inheriting from [the builtin Warning](https://docs.python.org/2/library/warnings.html).

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,208 @@
...
+                if file_path == self.SIP_IDENTIFIER:
+                    mdl = SIP.objects.get(uuid=unit_uuid)
+                    self._add_identifier_to_model(mdl, id_)
+                    continue
+                try:
+                    mdl = File.objects.get(sip_id=unit_uuid, currentlocation=file_path)
+                    self._add_identifier_to_model(mdl, id_)
+                    continue
+                except File.DoesNotExist:
+                    pass
```
cole, line 171: Maybe worth logging here?

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,208 @@
...
+                try:
+                    mdl = Directory.objects.get(
+                        sip_id=unit_uuid, currentlocation=os.path.join(file_path, "")
+                    )
+                    self._add_identifier_to_model(mdl, id_)
+                except Directory.DoesNotExist:
+                    pass
```
cole, line 178: Maybe worth logging here?

src/MCPClient/tests/test_parse_dataverse.py
```diff
@@ -65,8 +65,18 @@ def setUp(self):
     @staticmethod
     def test_fixture():
         """Test that the fixture has been loaded as expected."""
-        assert models.Transfer.objects.count() == 1
-        assert models.File.objects.count() == 12
+        assert (
```
cole: Maybe even just remove this method? It doesn't seem like a useful test, as it's something that Django should be taking care of testing.

src/MCPClient/tests/test_pid_components.py
```diff
@@ -0,0 +1,421 @@
...
+            ), "Identifier value count is incorrect"
+            for key, value in dict(zip(id_type, id_value)).items():
+                if key == self.pid_exid:
+                    assert example_uri in value, "Example URI not preserved"
+                if key == self.pid_ulid:
+                    assert len(example_ulid) == len(value)
+
+    @pytest.mark.django_db
+    def test_pid_declaration_exceptions(self, mocker):
+        """Ensure that the PID declaration feature exits when the JSOn cannot
```
cole, line 394: *JSON

src/MCPClient/lib/clientScripts/pid_declaration.py
```diff
@@ -0,0 +1,208 @@
...
+        """Create a consistent string to log information about our identifiers
+        to stdout for the user.
+        """
+        try:
+            if isinstance(mdl, SIP):
+                msg = "Identifier {}: {} added for: {}: {}".format(
+                    identifier_type,
+                    identifier,
+                    "SIP",
+                    os.path.basename(os.path.split(mdl.currentpath)[0]),
```
cole, line 90: It seems like we can use the `__unicode__` method defined on the models which already includes path information as you've done in the KeyError case? Or is there a need for this specific format?

cole (changes requested):
Looks pretty good overall! I've noted a few suggestions and questions. I don't have a great handle on this feature yet, so hopefully what I've said makes sense.

src/MCPClient/lib/clientScripts/pid_declaration.py
ross-spencer: True, if someone has tried to assign a custom identifier to something that doesn't exist then this will be useful. Will put this in, and for the directory too.

src/MCPClient/lib/clientScripts/pid_declaration.py
ross-spencer: I adopted this approach as a matter of style perhaps, vs. substance. The output will look like this:
```
Identifier EXID: https://example.com/file/id/2026496916 added for: File: MARBLES.TGA
```
vs.

```
Identifier EXID: https://example.com/file/id/2026496916 added to File efd059dd-7050-4bf4-bae5-bf1bd47082f2:%transferDirectory%objects/artwork/MARBLES.TGA now at %SIPDirectory%objects/artwork/MARBLES.TGA
```

I thought it a bit cleaner for the user, but what do you think? 

We end up with something like this:

image

src/MCPClient/lib/clientScripts/pid_component_helpers.py
ross-spencer: In this case, your other suggestions I think have helped me collapse this into the other modules, so I think we've managed to improve things :+1: We hadn't really discussed an approach before, next PR I'll consider Archivematica Common instead.

src/MCPClient/lib/clientScripts/pid_component_helpers.py
ross-spencer: I don't think we've been using model methods like this very often, but this is a nice suggestion. I like it.

src/MCPClient/lib/clientScripts/pid_component_helpers.py
ross-spencer: I'll try the `ValueError`! :) I've moved this back into the pid_declaration code too. I'm a bit happier with that as well.

src/MCPClient/lib/clientScripts/pid_declaration.py
ross-spencer: I wonder if I am missing something, but if I add the `except` we can't prevent the code from exiting when the exception is first raised. I haven't managed to return an exception back to `call` yet. 

The model we've adopted seems to have evolved over time, with some of our scripts trying to do a lot of scripts doing strange wrangling with states and asking permission. 

This is my first time trying this, adopting Joel D's approach in the similar bind_pids work: https://github.com/artefactual/archivematica/blob/84661775836cf8037ad3b48feb8e02bb80335f0f/src/MCPClient/lib/clientScripts/bind_pids.py#L75-L85 

But yeah, this just allows us to short-circuit as you say, from any point in the code and return back to the MCP server with a non-positive status if we're just trying to exit when something's non-critical. 

If you can help me get it to work using an `except` in call that would be great.

src/MCPClient/lib/clientScripts/pid_declaration.py
ross-spencer: I didn't want to introduce too much logic/state so as an alternative, I log the things that we can confirm as added, and then provide a count afterwards. I do think we can put a greater emphasis on validation of these files up-front so that users have a better chance of their input matching, but that's a bigger problem with a number of special files like this. Same with directories. 

IDK, is that okay? It will be pretty easy to add if it comes up through user-testing.

src/dashboard/src/main/models.py
```diff
@@ -409,6 +409,10 @@ def __unicode__(self):
     def update_active_agent(self, user_id):
         UnitVariable.objects.update_active_agent("SIP", self.uuid, user_id)
 
+    def add_custom_identifier(self, scheme, value):
+        """Allow callers to add custom identifiers to the model's instance."""
+        self.identifiers.add(Identifier.objects.create(type=scheme, value=value))
```
cole: I think I suggested before in a different spot, but you should be able to use `self.identifiers.create` here. I think this does the same thing, so it's not a big deal, it's just shorter the other way. If my suggestion isn't working please let me know and I'll stop suggesting it 😄

src/MCPClient/lib/clientScripts/pid_declaration.py
cole: Yeah, that's totally fine. It's always a trade off between noise and information that's needed in logs.

src/MCPClient/lib/clientScripts/pid_declaration.py
cole: kk, makes sense.

cole (approved):
Awesome, looks good! One niggle, but I'm happy to merge as is.

src/dashboard/src/main/models.py
ross-spencer: Ahhh! Okay, I was just being slow! Unit tests and a quick play-through show that this works a treat! I'm surprised it worked the way I put it together at all! :smile:

PR #1411, opened 2019-05-02
Arrangement: avoid local access to determine BagIt format

sevein:
When a SIP is arranged, we need to ask SS which files are copied to the
local staging path.

`logs` and `metadata` are also copied because they're used in Ingest to
build the final SIP. However these entries are not listed in `SIPArrange`
so we need to guess their paths which is going to be different depending
on the format of the transfer package, e.g. `data/logs` instead of `logs`
when it's using the BagIt format.

This commit changes the logic that guesses the format of the package.
Originally, we assumed that physical access to the transfer package was
possible to determine its format. As that's not always true, we're now
guessing the format by looking at existing entries listed by `SIPArrange`.

Connects to https://github.com/archivematica/Issues/issues/671.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -36,9 +36,8 @@ from main import models
 
 import archivematicaFunctions
 import databaseFunctions
+import elasticSearchFunctions
 import storageService as storage_service
-
-from bag import is_bag
 
 
 logger = logging.getLogger("archivematica.dashboard")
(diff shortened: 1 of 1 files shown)


hakamine (approved):
In an environment where the pipeline is separated from the storage service (am191bionic.qa.archivematica.org and ss191bionic.qa.archivematica.org)  did the following tests and confirmed this is working (had errors before the fixes were applied):
- Appraisal tab: from a backlogged transfer (on the Backlog pane) dragged some files to a directory created on the Arrangement pane and clicked "Create SIP". A SIP was created and showed on the Ingest tab
- Appraisal tab: from a backlogged transfer (on the Backlog pane), selected some files so they appear in the File List pane. Then, in the File List pane clicked on a file so it appears in the Analysis pane. Finally, in the Analysis pane clicked on Download to download the file.

sevein:
Thank you, @hakamine!

peterVG:
I've QA'ed and confirmed as well on am191bionic.qa, including the "tag and select" SIP creation.

PR #1415, opened 2019-05-12
Increase timeout value for dashboard MCP RPC calls

hakamine:
Quick workaround for archivematica/Issues#624

Connected to archivematica/Issues#624


ross-spencer (approved):
LGTM! :+1:

PR #1417, opened 2019-05-21
Remove verify structmap job from ingest workflow

ross-spencer:
Removes the Verify mets_structmap.xml compliance job from the
ingest workflow where its use is no longer part of a supported
workflow.

Before:

image

After:

image

Connected to archivematica/issues#286


replaceafill (approved):
:+1: Tested it in my dev instance and the microservice and job are removed from the Ingest tab.

ross-spencer:
Thanks @replaceafill :+1:

PR #1420, opened 2019-05-22
Improvements to move_or_merge

helenst:
Addreses https://github.com/archivematica/Issues/issues/704

This was originally a PR against the wellcometrust fork: https://github.com/wellcometrust/archivematica/pull/4

From that PR:

```
It seems like move_or_merge.py gets upset if it encounters nested directories. If there's a nested directory in dst that matches one in src, it throws an error even if they're identical! This means it should error less often.

It also means it will error out if it finds existing files in dst that don't match. Plus a pile of tests.
```

### DIFF
--- a/src/MCPClient/lib/clientScripts/move_or_merge.py
+++ b/src/MCPClient/lib/clientScripts/move_or_merge.py
@@ -1,7 +1,69 @@
 #!/usr/bin/env python2
 
+import errno
+import filecmp
 import os
 import shutil
+
+
+def mkdir_p(path):
+    """Create a directory if it doesn't already exist."""
+    try:
+        os.makedirs(path)
+    except OSError as err:
+        if err.errno == errno.EEXIST:
+            pass
+        else:
+            raise
+
+
+def _move_file(src, dst):
+    """
+    Move an individual file from ``src`` to ``dst``.
+    """
+    # Ensure the destination directory exists
+    mkdir_p(os.path.dirname(dst))
+
+    # If the file already exists at the destination, check if it's the same.
+    # If so, we can clean up the original file and we're done.  If not, we're
+    # at risk of losing data, so error out.
+    if os.path.isfile(dst):
+        if filecmp.cmp(src, dst, shallow=False):
+            os.unlink(src)
+        else:
+            raise RuntimeError(
+                "Tried to move src=%s to dst=%s, but dst exists and is different"
+                % (src, dst)
+            )
+    else:
+        shutil.move(src, dst)
+
+
+def move_or_merge(src, dst):
+    """
+    Move a file/directory to a new location, or merge two directories.
+
+    If ``dst`` doesn't exist, it's a simple move: ``src`` is moved to the same
+    path as ``dst``.
+
+    If ``dst`` does exist and is a directory, the two directories are merged by
+    moving the contents of ``src`` into ``dst``.
+    """
+    if os.path.isfile(src):
+        _move_file(src, dst)
+    elif os.path.isdir(src):
+        # This loop walks the tree looking for files.  For every file in ``src``,
+        # it finds the path relative to the top of ``src``, then moves it
+        # to the corresponding path in ``dst``.
+        for root, _, filenames in os.walk(src):
+            rel_root = os.path.relpath(root, start=src)
+            for f in filenames:
+                _move_file(
+                    src=os.path.join(src, rel_root, f),
+                    dst=os.path.join(dst, rel_root, f),
+                )
+
+        shutil.rmtree(src)
 
 
 def main(src, dst):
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -1,9 +1,72 @@
...
+    If ``dst`` doesn't exist, it's a simple move: ``src`` is moved to the same
+    path as ``dst``.
+
+    If ``dst`` does exist and is a directory, the two directories are merged by
+    moving the contents of ``src`` into ``dst``.
+    """
+    if os.path.isfile(src):
+        _move_file(src, dst)
+    elif os.path.isdir(src):
+
```
ross-spencer: It would be good to remove this newline during the rebase prior to merging.

src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -1,9 +1,72 @@
...
+    Move an individual file from ``src`` to ``dst``.
+    """
+    # Ensure the destination directory exists
+    mkdir_p(os.path.dirname(dst))
+
+    # If the file already exists at the destination, check if it's the same.
+    # If so, we can clean up the original file and we're done.  If not, we're
+    # at risk of losing data, so error out.
+    if os.path.isfile(dst):
+        if filecmp.cmp(src, dst):
```
ross-spencer: I didn't know about this function! Looking at the source it seems like at the very worst it's going to do a byte-for-byte (in 8k chunks) comparison of the two files. I think that may leave open the slightest chance of a bit-flip:

```
import filecmp

a = "a.txt"
b = "b.txt"

with open(a, "w") as a_file:
	a_file.write("hello")

with open(b, "w") as b_file:
	b_file.write("hel1o")

assert filecmp.cmp(a, b), "a != b, shallow=True"
assert filecmp.cmp(a, b, shallow=False), "a != b, shallow=False"
```

...but likely very low. IDK, what do think? The change is being made for an edge-case anyway, would it be too much of a trade-off to always perform the full check? (The function seems to be called twice in workflow.json.)

src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -1,9 +1,72 @@
...
+    moving the contents of ``src`` into ``dst``.
+    """
+    if os.path.isfile(src):
+        _move_file(src, dst)
+    elif os.path.isdir(src):
+
+        # This loop walks the tree looking for files.  For every file in ``src``,
+        # it finds the path relative to the top of ``src``, then moves it
+        # to the corresponding path in ``dst``.
+        for root, _, filenames in os.walk(src):
```
ross-spencer: I wonder if there's a small optimization here per @cole's [investigation into speeding up AM](https://github.com/artefactual/archivematica/pull/1497)?

src/MCPClient/tests/test_move_or_merge.py
```diff
@@ -0,0 +1,151 @@
+# -*- encoding: utf-8
+
+import pytest
```
ross-spencer, line 3: Always great to have more tests!

ross-spencer (commented):
Hi @helenst this seems like a nice improvement, and worth adding. I've left a comment about the `filecmp.cmp` comparison and then one about how we're walking the directories. 

**NB.** This will need to be rebased as with the other PRs and then we can finalize the review.

src/MCPClient/lib/clientScripts/move_or_merge.py
helenst: I'm not sure. It does seem very much an edge case - so  maybe it is worth doing the deep comparison and taking the performance hit. Worst case if they are the same file and it has to compare the whole file (if different, presumably it'll bail out when it finds a difference).

helenst:
@ross-spencer I've rebased, fixed and responded to comments. Ready for re-review :)

ross-spencer:
Awesome, thanks @helenst!

src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -1,8 +1,72 @@
...
+    else:
+        shutil.move(src, dst)
+
+
+def move_or_merge(src, dst):
+    """
+    Move a file/directory to a new location, or merge two directories.
+
+    If ``dst`` doesn't exist, it's a simple move: ``src`` is moved to the same
+    path as ``dst``.
```
ross-spencer, line 49: The documentation is really nice around this PR. Do you have internal guidelines on this? I feel like it's something we haven't yet found consistency on, so any tips on bringing that consistency into the work would be great!

src/MCPClient/lib/clientScripts/move_or_merge.py
ross-spencer: Thanks @helenst - let's start safe (`shallow=False`) and then we can walk that back later if it becomes a painfully slow situation! Always a tricky place in a CR this! I'll approve because why not?! And then can you make that change and I'll prepare to merge in the next day now this is milestoned.

ross-spencer (approved):
Just a comment about using the deep `cmp` check, but once done we'll get this merged!!

src/MCPClient/lib/clientScripts/move_or_merge.py
helenst: @ross-spencer Agreed - all credit to @alexwlchan for that :-)

helenst:
@ross-spencer I've made that update so we can see how that goes :)

PR #1424, opened 2019-05-29
Remove Archivists' Toolkit DIP upload functionality

sevein:
Connects to https://github.com/archivematica/Issues/issues/174.


ablwr (commented):
Looks like a reasonable ATK-removal surgery procedure to me. The only bit of the code that I don't understand is the change in `src/dashboard/src/components/ingest/views_NormalizationReport.py`.

sevein:
> The only bit of the code that I don't understand is the change in src/dashboard/src/components/ingest/views_NormalizationReport.py.

Ops, sorry, I didn't mean to include that change, it's gone now - good catch :)

ablwr (approved):

PR #1425, opened 2019-05-31
FPR: Treat partial pass validation option as a pass

ablwr:
This is connected to https://github.com/archivematica/Issues/issues/164. It prevents "false failures" in Validation. More description is consensus is in the issue.

A subsequent PR to update the JHOVE command in the FPR is required, but does not block the review and merging of this PR.

### DIFF
--- a/src/MCPClient/lib/clientScripts/validate_file.py
+++ b/src/MCPClient/lib/clientScripts/validate_file.py
@@ -168,6 +168,11 @@ class Validator(object):
             self.job.print_output(
                 'Command "{}" was successful'.format(rule.command.description)
             )
+        elif output.get("eventOutcomeInformation") == "partial pass":
+            self.job.print_output(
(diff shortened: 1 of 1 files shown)


cole (approved):

ablwr:
Am I correct in understanding that it is the linter within Travis that failed?

Also, I have a Black linter on my machine, but it doesn't seem to agree with Travis about some rules -- but that's on me to clear up.

cole:
@ablwr It looks like linting failed, specifically the black code style check. Do you have [pre-commit](https://pre-commit.com/) setup? It should be `pre-commit install` inside the git repo. In theory, that should run the same version of black with the same settings (not that black has many settings) as Travis.

ablwr:
@cole On the CLI it also didn't give me more information about why it was failing but I managed to appease it. You approved this so I think it's still good2merge now that Black approves of me.

PR #1428, opened 2019-05-31
Update JHOVE command with partial pass for byestreams

ablwr:
This is related to https://github.com/archivematica/Issues/issues/164.

It is also associated with https://github.com/artefactual/archivematica/pull/1425.

And it is a bit related to https://github.com/archivematica/Issues/issues/524 as well, because this update will expect 1.20 (but seems to work with older versions too).


replaceafill (approved):
LGTM! :+1:

PR #1429, opened 2019-06-03
Allows users to manage system emails

helrond:
This add functionality for users to choose whether or not they want to receive system-generated emails, such as Transfer Fail reports or Normalization Reports. 

The following changes have been introduced:
- Add a new boolean field to the `UserProfile` model
- Add a new `UserProfile` form and update views and templates accordingly
- Update the querysets used when sending transfer and normalization failure emails

A couple of notes:
- I removed an `exclude()` from the querysets mentioned above because it seemed redundant given the ability to explicitly manage emails.
- I did not add a model manager method as discussed in archivematica/Issues#709 because that queryset touches the `User` model but would have to have been added to the `UserProfile` model. My feeling is that this would be a very counter-intuitive and confusing approach for future developers to understand.
- I didn't update any tests for this, but would welcome any input on modifying existing tests so they validate this new behavior.
- I'm also not sure I have this pointing at the correct branch?

Fixes archivematica/Issues#709


replaceafill:
Hello @helrond :wave: 

> I'm also not sure I have this pointing at the correct branch?

Could you change this to point to the `qa/1.x` branch? I'm not sure if just editing this PR and changing the `base` attribute would be enough or if you will get conflicts. If opening a new PR is easier (and closing this one) that's fine too. Thanks!

helrond:
Hi @replaceafill I have pointed this at `qa/1.x` now and resolved the conflicts. This got a little messy, apologies for the non-optimal commit history. I have also manually updated the migration associated with this PR to avoid the migration conflict.

src/dashboard/src/components/accounts/forms.py
```diff
@@ -115,3 +117,13 @@ class ApiKeyForm(forms.Form):
     regenerate_api_key = forms.CharField(
         widget=forms.CheckboxInput, label="Regenerate API key (shown below)?"
     )
+
+
+class UserProfileForm(forms.ModelForm):
+    class Meta:
+        model = UserProfile
+        fields = ["system_emails"]
+        labels = {"system_emails": "Send system emails?"}
```
replaceafill: Could we make the labels and help text translatable?

src/dashboard/src/main/models.py
```diff
@@ -859,6 +859,7 @@ class UserProfile(models.Model):
 
     user = models.OneToOneField(User)
     agent = models.OneToOneField(Agent)
+    system_emails = models.BooleanField(default=False)
```
replaceafill: I think moving the label and help text here (and making them translatable strings) would be ideal.

src/dashboard/src/main/migrations/0075_userprofile_system_emails.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [("main", "0074_version_number")]
+
+    operations = [
+        migrations.AddField(
+            model_name="userprofile",
+            name="system_emails",
+            field=models.BooleanField(default=False),
```
replaceafill: I'm not sure what's the right approach when the setting is `False` by default here in the migration and in the new model field, but the client scripts are looking for `userprofile__system_emails=True`. Existing users would stop receiving emails, right? @sevein do you have any recommendation on this? Sorry if I'm missing something.

replaceafill (changes requested):
Thanks for updating this @helrond and it looks good overall. The only part I'm not sure how to deal with yet is the new setting being `False` by default and the client scripts looking for its `True` value in order to send emails.

src/dashboard/src/main/migrations/0075_userprofile_system_emails.py
sevein: Good catch, Douglas. To ensure a seamless upgrade, perhaps we could add a new data migration that enables the setting for existing users.

helrond: How about just flipping the default to `True`? That means people will have to opt out of emails rather than opting in, but I'm not sure that's actually any worse...

sevein: CC @sromkey

sromkey: That sounds fine to me, since the current situation is everyone gets the emails by default.

src/dashboard/src/components/accounts/views.py
```diff
@@ -51,15 +52,21 @@ def add(request):
...
             return redirect("components.accounts.views.list")
     else:
         # Clearing out values that are getting inherited from currently logged in user
         data = {"email": ""}
         form = UserCreationForm(initial=data)
+        userprofileform = UserProfileForm()
 
-    return render(request, "accounts/add.html", {"form": form})
+    return render(
+        request, "accounts/add.html", {"form": form, "userprofileform": userprofileform}
```
replaceafill, line 69: If the `UserCreationForm` has a submission error (for example the password being too short), accessing `userprofileform` here results in a `referenced before assignment` exception:

```
UnboundLocalError at /administration/accounts/add/

local variable 'userprofileform' referenced before assignment
```

replaceafill (changes requested):
Thanks for the quick updates @helrond. I tested the new functionality and looks good except for the `UnboundLocalError` when validation fails in the add user form.

helrond:
@replaceafill Thanks for that catch! I've fixed the issue, however...in doing that I noticed that `email` is not a required field in the UserCreationForm but it is the UserChangeForm, which seems...wrong?

Happy to file an issue and PR for that too if that seems like something that's actually an oversight, as opposed to an intentional decision.

src/dashboard/src/components/accounts/views.py
```diff
@@ -53,7 +54,7 @@ def add(request):
             api_key.key = api_key.generate_key()
             api_key.save()
             user_profile = UserProfile.objects.get(user=newuser)
-            userprofileform = UserProfileForm(request.POST, instance=user_profile)
+            userprofileform.instance = user_profile
```
replaceafill: I find this approach to set the instance of a model in a existing form interesting. I've always considered Django forms immutable. Do you know if if this safe doing considering the data initialization of the form (required fields, etc) has been already executed before?

helrond: Hmm you're right, I was getting too clever. [Django docs](https://docs.djangoproject.com/en/3.0/ref/forms/api/#bound-and-unbound-forms) advise to create a new instance of a form if the data changes. That's an easy fix.

replaceafill (approved):

replaceafill:
Thanks a lot for your contribution @helrond!

And please go ahead and file the discrepancy in the requirement of the email field :+1: The team agrees that field shouldn't be required.

PR #1430, opened 2019-06-05
osdeps: don't use external repositories for centos

scollazo:
Connects to https://github.com/archivematica/Issues/issues/254


mamedin (approved):
LGTM

PR #1431, opened 2019-06-06
FPR only shows enabled tools, does not display versions

ablwr:
This is related to https://github.com/archivematica/Issues/issues/394.
This is also related to https://github.com/archivematica/Issues/issues/191. 

Only shows enabled tools, not legacy ones, and hides versions.

Fundamentally this part of the codebase needs more work and I know that. This is more like a patch to clear up confusion to users and prevent misinformation than it is a holistic change, due to time-and-money restraints.

### DIFF
--- a/src/dashboard/src/fpr/models.py
+++ b/src/dashboard/src/fpr/models.py
@@ -368,9 +368,7 @@ class IDTool(models.Model):
     active = Enabled()
 
     def __unicode__(self):
-        return _("%(description)s") % {
-            "description": self.description,
-        }
+        return _("%(description)s") % {"description": self.description}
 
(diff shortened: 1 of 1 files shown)


cole:
@ablwr Does it make sense to just drop the `version` field completely? If it's no longer used and updated it may just cause confusion.

Otherwise this looks good!

ablwr:
I tried that but I think it's used somewhere else in the FPR because Archivematica is unhappy when it's gone. And right now, I think I just want to hide it because versioning is important, and was built into the FPR for a reason, and I think it's likely we will revisit it!

cole (approved):

PR #1438, opened 2019-06-17
storageService: stop using wait_for_async

sevein:
Stop using async endpoints provided by SS until we have a chance to address the
issues with the SS async manager (see https://github.com/archivematica/Issues/issues/730 for more).


mamedin (approved):
LGTM.

PR #1440, opened 2019-06-18
Add microservices and task API endpoints

replaceafill:
This PR adds two endpoints to retrieve information about unit jobs and tasks. Their documentation in the Archivematica API wiki page would be something like this:

## Unit

**List jobs**
* **URL**: `/v2beta/jobs/<unit UUID>/`
* **Verb**: GET
* Returns a list of jobs for the passed unit (transfer or ingest).
* **Parameters**:
  * Optional filters
    * `microservice`: Name of the microservice the jobs belong to
    * `link_uuid`: UUID of the job chain link
    * `name`: Name of the job
* **Response**: JSON body
  * List of dicts with keys:
    * `uuid`: UUID of the job
    * `name`: Name of the job
    * `status`: One of `USER_INPUT`, `PROCESSING`, `COMPLETE`, `FAILED` or `UNKNOWN`
    * `microservice`: Microservice the job belongs to
    * `link_uuid`: UUID of the job chain link
    * `tasks`: List of dicts with information about the microservice's tasks:
      * `uuid`: UUID of the task
      * `exit_code`: Exit code of the task

**Task**
* **URL**: `/v2beta/task/<task UUID>/`
* **Verb**: GET
* Returns information about a task.
* **Response**: JSON body
  * `uuid`: UUID of the task
  * `exit_code`: Exit code of the task
  * `file_uuid`: UUID of the file used for the task
  * `file_name`: File used for the task,
  * `time_created`: String (`YYYY-MM-DD HH:MM:SS`) representing when the task was created
  * `time_started`: String (`YYYY-MM-DD HH:MM:SS`) representing when the task started
  * `time_ended`: String (`YYYY-MM-DD HH:MM:SS`) representing when the task finished
  * `duration`: Task duration in seconds (integer). If the duration is less than a second, this will be a `< 1` string

Connected to https://github.com/archivematica/Issues/issues/722

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -55,6 +55,13 @@ UUID_REGEX = re.compile(
     r"^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$",
     re.IGNORECASE,
 )
+JOB_STATUS_CODE_LABELS = {
+    models.Job.STATUS_UNKNOWN: "UNKNOWN",
+    models.Job.STATUS_AWAITING_DECISION: "USER_INPUT",
+    models.Job.STATUS_COMPLETED_SUCCESSFULLY: "COMPLETE",
+    models.Job.STATUS_EXECUTING_COMMANDS: "PROCESSING",
+    models.Job.STATUS_FAILED: "FAILED",
+}
 
 
 def _api_endpoint(expected_methods):
@@ -858,3 +865,78 @@ def validate(request, validator_name):
             "Unexepected error in the validation process, see the logs for more details."
         )
     return _ok_response({"valid": True})
+
+
+def format_datetime(value):
+    """Convert datetime to string"""
+    try:
+        return value.strftime("%Y-%m-%dT%H:%M:%S")
+    except AttributeError:
+        pass
+
+
+def format_task(task, detailed_output=False):
+    """Format task attributes for endpoint response"""
+    result = {"uuid": task.taskuuid, "exit_code": task.exitcode}
+    if detailed_output:
+        result.update(
+            {
+                "file_uuid": task.fileuuid,
+                "file_name": task.filename,
+                "time_created": format_datetime(task.createdtime),
+                "time_started": format_datetime(task.starttime),
+                "time_ended": format_datetime(task.endtime),
+                "duration": helpers.task_duration_in_seconds(task),
+            }
+        )
+    return result
+
+
+def remove_prefix(s, prefix):
+    """Remove a prefix from a string"""
+    if s.startswith(prefix):
+        return s[len(prefix) :]
+    return s
+
+
+@_api_endpoint(expected_methods=["GET"])
+def unit_jobs(request, unit_uuid):
+    """Return jobs associated with a unit's UUID"""
+    jobs = models.Job.objects.filter(sipuuid=unit_uuid).order_by("createdtime")
+    if not jobs.count():
+        return _error_response("No jobs found for unit: {}".format(unit_uuid))
+    result = []
+    microservice = request.GET.get("microservice")
+    if microservice is not None:
+        microservice = remove_prefix(microservice, "Microservice:")
+        jobs = jobs.filter(microservicegroup=microservice.strip())
+    link_uuid = request.GET.get("link_uuid")
+    if link_uuid is not None:
+        jobs = jobs.filter(microservicechainlink=link_uuid.strip())
+    name = request.GET.get("name")
+    if name is not None:
+        name = remove_prefix(name, "Job:")
+        jobs = jobs.filter(jobtype=name.strip())
+    for job in jobs.prefetch_related("task_set"):
+        tasks = [format_task(task) for task in job.task_set.all()]
+        result.append(
+            {
+                "uuid": job.jobuuid,
+                "name": job.jobtype,
+                "status": JOB_STATUS_CODE_LABELS.get(job.currentstep),
+                "microservice": job.microservicegroup,
+                "link_uuid": job.microservicechainlink,
+                "tasks": tasks,
+            }
+        )
+    return helpers.json_response(result)
+
+
+@_api_endpoint(expected_methods=["GET"])
+def task(request, task_uuid):
+    """Return details of a task"""
+    try:
+        task = models.Task.objects.get(taskuuid=task_uuid)
+    except models.Task.DoesNotExist:
+        return _error_response("Task with UUID {} does not exist".format(task_uuid))
+    return helpers.json_response(format_task(task, detailed_output=True))
--- a/src/dashboard/tests/test_api.py
+++ b/src/dashboard/tests/test_api.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import os
 import tempfile
@@ -7,11 +8,12 @@ from django.core.management import call_command
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test.client import Client
+from django.utils.timezone import make_aware
 from lxml import etree
 
 from components.api import views
 from components import helpers
-from main.models import SIP, Transfer
+from main.models import Job, SIP, Task, Transfer
 from processing import install_builtin_config
 
 THIS_DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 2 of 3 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -55,6 +55,13 @@
     r"^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$",
     re.IGNORECASE,
 )
+JOB_STATUS_CODE_LABELS = {
+    models.Job.STATUS_UNKNOWN: "UNKNOWN",
+    models.Job.STATUS_AWAITING_DECISION: "USER_INPUT",
+    models.Job.STATUS_COMPLETED_SUCCESSFULLY: "COMPLETE",
+    models.Job.STATUS_EXECUTING_COMMANDS: "PROCESSING",
+    models.Job.STATUS_FAILED: "FAILED",
+}
```
ross-spencer, line 64: We could consider replacing the literals in this section [here](https://github.com/artefactual/archivematica/blob/0198d00b616b00f2a70f8e6aad34cb855213f675/src/dashboard/src/components/api/views.py#L191-L222) with a dictionary lookup now we have this. What do you think?

src/dashboard/src/components/api/views.py
```diff
@@ -858,3 +865,65 @@ def validate(request, validator_name):
             "Unexepected error in the validation process, see the logs for more details."
         )
     return _ok_response({"valid": True})
+
+
+def format_datetime(value):
+    try:
```
ross-spencer, line 872: It's only a small detail but I feel like we should endeavor to have doc strings where possible, they don't have to contain too much detail. We could do with a little extra documentation for each of the new functions introduced in this PR.

src/dashboard/src/components/api/views.py
```diff
@@ -858,3 +865,65 @@ def validate(request, validator_name):
...
+                "time_ended": format_datetime(task.endtime),
+                "duration": helpers.task_duration_in_seconds(task),
+            }
+        )
+    return result
+
+
+@_api_endpoint(expected_methods=["GET"])
+def unit_jobs(request, unit_uuid):
+    jobs = models.Job.objects.filter(sipuuid=unit_uuid)
```
ross-spencer:
```suggestion
    jobs = models.Job.objects.filter(sipuuid=unit_uuid).order_by("createdtime")
```
This might be a good way to provide some predictability to the structure returned that can be used by the consumer where we don't provide the `datetime` to work with initially. I could imagine the result of this AIP call being used to create a timeline of jobs other than we replay for the user in the dashboard.

src/dashboard/src/components/api/views.py
```diff
@@ -858,3 +865,65 @@ def validate(request, validator_name):
...
+    if not jobs.count():
+        return _error_response("No jobs found for unit: {}".format(unit_uuid))
```
ross-spencer, line 907: This works nicely in testing.

src/dashboard/src/components/api/views.py
```diff
@@ -858,3 +865,65 @@ def validate(request, validator_name):
...
+    result = []
+    microservice = request.GET.get("microservice")
+    if microservice is not None:
+        jobs = jobs.filter(microservicegroup=microservice)
+    link_uuid = request.GET.get("link_uuid")
+    if link_uuid is not None:
+        jobs = jobs.filter(microservicechainlink=link_uuid)
+    name = request.GET.get("name")
+    if name is not None:
+        jobs = jobs.filter(jobtype=name)
```
ross-spencer: Shall we also allow "Job: " in query strings passed as `name`? We'd just need to spit if if we find it. It might just make it a little [more robust](https://en.wikipedia.org/wiki/Robustness_principle) for the end-user.

src/dashboard/tests/test_api.py
```diff
@@ -298,6 +300,155 @@ def test_completed_ingests_with_bogus_sip(self):
...
+            message=("No jobs found for unit: {}".format(bogus_unit_uuid)),
+        )
+
+    @e2e
+    def test_unit_jobs(self):
+        load_fixture(["jobs-transfer-backlog.json"])
+        sip_uuid = "3e1e56ed-923b-4b53-84fe-c5c1c0b0cf8e"
+        # fixtures don't have any tasks
+        Task.objects.create(
+            taskuuid="12345678-1234-1234-1234-123456789012",
```
ross-spencer, line 319: What are the chances of generating this `uuid`!! :laughing:

ross-spencer (changes requested):
This is really great work @replaceafill. Really excited to see it out in the wild. 

I have tested with the following [`httpie`](https://httpie.org/) reuquests:
```
-----------------
New API endpoints
-----------------

---------------------------
Return jobs for a transfer:
---------------------------

http --pretty=format \
    GET "http://127.0.0.1:62080/api/v2beta/jobs/967db67f-54c3-4e3d-87b8-27d47728a5c0/" \
    Authorization:"ApiKey test:test"

--------------------------    
Return jobs for an ingest:
--------------------------
    
http --pretty=format \
    GET "http://127.0.0.1:62080/api/v2beta/jobs/62598e60-5105-4f99-8a5a-432224bb892b/" \
    Authorization:"ApiKey test:test"    
    
------------------------------------------
Using == syntax to encode query parameters
------------------------------------------
    
http --pretty=format \
    -v \
    GET "http://127.0.0.1:62080/api/v2beta/jobs/62598e60-5105-4f99-8a5a-432224bb892b/" \
    microservice=="Bind PIDs" \
    Authorization:"ApiKey test:test"        

http --pretty=format \
    -v \
    GET "http://127.0.0.1:62080/api/v2beta/jobs/62598e60-5105-4f99-8a5a-432224bb892b/" \
    name=="Index AIP" \
    Authorization:"ApiKey test:test"

-------------------
Different language?
-------------------

http --pretty=format \
    -v  \
    GET "http://127.0.0.1:62080/api/v2beta/jobs/d33afb13-e732-46b0-a4cd-f30d91d49431/" \
    microservice=="Lier les PIDs" \
    Authorization:"ApiKey test:test"

-------------------
Test finding a task
-------------------

http --pretty=format \
    GET "http://127.0.0.1:62080/api/v2beta/task/1c637aaa-c780-4302-8e4c-47f72d5b3fc0/" \
    Authorization:"ApiKey test:test"

---------------------
When task isn't there
---------------------

http --pretty=format \
    GET "http://127.0.0.1:62080/api/v2beta/task/1c637aaa-c780-4302-8e4c-47f72d5b3fc1/" \
    Authorization:"ApiKey test:test"
```
And it all looks great. As do your new tests for the endpoint too.

Some small changes requested, but mostly cosmetic.

You'll see I tested trying to retrieve a microservice's information in French. We don't return anything in that case. 

It might be worth asking @sevein about the `workflow.json` API to see if there is way of retrieving a string via that? Of if there is a way we need to internationalize this endpoint that we haven't considered.

It could also be done in a future iteration. For expedience, I want to make sure the review is in to begin,and we can start that conversation. I also need time to remind myself of how to use the `workflow.json` work so can't comment further just yet. It might be out of scope for now.

src/dashboard/src/components/api/views.py
replaceafill: Hmm. I used the same labels because I kind of thought about it when I created this mapping. However the intended use seemed different. It's like in this mapping there's a 1-1 relationship between the *STATUS* of the job and the label. In the `get_unit_status` function I don't see that 1-1 relationship. The label returned is based on very different conditions and you have additional ones like `REJECTED` not based on status at all. Not sure how to express this point. Sorry about that.

src/dashboard/src/components/api/views.py
replaceafill: :+1: Will do.

src/dashboard/src/components/api/views.py
replaceafill: :+1:

src/dashboard/src/components/api/views.py
replaceafill: Not sure if I understood your idea. See https://github.com/artefactual/archivematica/pull/1440/commits/cb80972acac003dd79d99957880cf06a466609b3. I also implemented "Microservice:"

src/dashboard/tests/test_api.py
replaceafill: *Cough* created it manually *cough* :blush:

replaceafill:
Thank you @ross-spencer! I addressed most of your comments. I'm just not sure about the labels mapping reuse.

Regarding retrieving information in other languages I don't think you can **search** the i18n contents of `workflow.json` efficiently yet. I asked @sevein and you can get the translation strings only if you have the link UUID. This is [how the MCPServer does it](https://github.com/artefactual/archivematica/blob/d9f2d59f1642c1fc644db698352df5e1e7da8572/src/MCPServer/lib/RPCServer.py#L425-L438). So I'd say we leave it out of scope for now and maybe add a note in the wiki when we document the API endpoint that you can only search in English.

src/dashboard/src/components/api/views.py
ross-spencer: Understood perfectly! Thanks @replaceafill!

src/dashboard/src/components/api/views.py
ross-spencer: Hmm, one to ponder. Definitely worth reflecting on the semantics before changing something for the sake of removing small potential redundancy. We still need the other end point to be consistent where this one can change too. And we can still make this change in the future if the two sets of values look closely aligned.

src/dashboard/src/components/api/views.py
```diff
@@ -890,21 +892,31 @@ def format_task(task, detailed_output=False):
...
+    """Return jobs associated with a unit's UUID"""
+    jobs = models.Job.objects.filter(sipuuid=unit_uuid).order_by("createdtime")
     if not jobs.count():
         return _error_response("No jobs found for unit: {}".format(unit_uuid))
     result = []
     microservice = request.GET.get("microservice")
     if microservice is not None:
-        jobs = jobs.filter(microservicegroup=microservice)
+        microservice = remove_prefix(microservice, "Microservice:")
+        jobs = jobs.filter(microservicegroup=microservice.strip())
```
ross-spencer, line 912: This is great. Thanks for making this change and the one for `Job:`. 

If `remove_prefix` performs the `strip()` function too, then this can become:

    jobs = jobs.filter(microservicegroup=remove_prefix(microservice))

But I'm not sure where we all stand on code density! :)

src/dashboard/src/components/api/views.py
```diff
@@ -890,21 +892,31 @@ def format_task(task, detailed_output=False):
...
     link_uuid = request.GET.get("link_uuid")
     if link_uuid is not None:
-        jobs = jobs.filter(microservicechainlink=link_uuid)
+        jobs = jobs.filter(microservicechainlink=link_uuid.strip())
```
ross-spencer, line 915: This is a good idea to do. Nice work.

ross-spencer (approved):
I left one comment for small consideration, but otherwise, looks great! Great final addition to `1.10`. :sunglasses: 

I have a few CR tasks to follow up on for `amclient` and can pick up the `amauat` too as this week progresses, so will look to tackle these later today.

src/dashboard/src/components/api/views.py
replaceafill: I thought about it, but `remove_prefix` seemed like a good and simple name for the intended use and I couldn't think of a better name if I added the `strip` call in it (`extract_name` seemed too generic, etc). And after realizing that `link_uuid` should be stripped too I decided to go with three consistent `strip` calls inside the `filter` calls instead.

currmie:
[original comment removed and re-inserted into related issue [722](https://github.com/archivematica/Issues/issues/722)

PR #1441, opened 2019-06-18
Update migration script to change jhove version too

ablwr:
This is connected to https://github.com/archivematica/Issues/issues/524.

I know jumping in and editing a previously-merged migration is not-so-good, but this one was very newly created -- a couple of weeks ago at most -- and it was missing the bit that involves disabling the previous version of JHOVE and needed a comment removed anyway.

### DIFF
--- a/src/dashboard/src/fpr/migrations/0030_update_jhove_command.py
+++ b/src/dashboard/src/fpr/migrations/0030_update_jhove_command.py
@@ -91,7 +91,7 @@ def data_migration_up(apps, schema_editor):
     FPCommand = apps.get_model("fpr", "FPCommand")
     FPRule = apps.get_model("fpr", "FPRule")
 
-    FPTool.objects.filter(uuid=JHOVE_TOOL_UUID).update(version="1.20", slug="fido-120")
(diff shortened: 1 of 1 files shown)


src/dashboard/src/fpr/migrations/0030_update_jhove_command.py
```diff
@@ -83,15 +84,15 @@
...
-NEW_JHOVE_CMD_UUID = "11036e14-78d9-4449-8360-e2da394279ad"
-
 
 def data_migration_up(apps, schema_editor):
     """Update command to partial pass on bytestream results."""
+    FPTool = apps.get_model("fpr", "FPTool")
     FPCommand = apps.get_model("fpr", "FPCommand")
     FPRule = apps.get_model("fpr", "FPRule")
 
+    FPTool.objects.filter(uuid=JHOVE_TOOL_UUID).update(version="1.20", slug="fido-120")
```
replaceafill: Should this be `jhove-120`?

replaceafill (approved):
Just one observation following our conversation in Slack. Then it's :+1:

PR #1443, opened 2019-06-18
Performance instrumentation

cole:
Adds task duration to prometheus, and a new management command to report on total time spent per SIP/AIP for benchmarking.

A management command is a bit of an awkward way to do this, but it's portable and versioned, vs passing around a SQL query for use in Grafana. I'd like to add other reports (slowest microservices for one) to the command before merging.

Connected to https://github.com/archivematica/Issues/issues/906


ross-spencer:
Hi @cole recognizing this is a work in progress. I am just wondering, I haven't enough knowledge in this area beyond the code itself to know about  the anticipated outcome of this PR. Do you/we have more docs on how to use Prometheus?/Wire things up to test? This is more general curiosity than anything.

cole:
@ross-spencer good questions! This is kind of exploratory, looking at finding some useful ways to benchmark. Prometheus is a service that runs and would scrape metrics from MCP server. [There is a branch of compose setup with it](https://github.com/artefactual-labs/am/blob/dev/instrumentation-v1.8.1/compose/README.instrumentation.md).

cole:
Update: I don't think a management command is going to end up being very helpful here after all. Switched to Prometheus metrics for everything.

cole:
Relates to https://github.com/archivematica/Issues/issues/906.

src/MCPServer/lib/archivematicaMCP.py
```diff
@@ -367,15 +331,15 @@ def _except_hook_log_everything(exc_type, exc_value, exc_traceback):
 if __name__ == "__main__":
 
     # Replace exception handler with one that logs exceptions
-    sys.excepthook = _except_hook_log_everything
+    # sys.excepthook = _except_hook_log_everything
```
replaceafill: Should this line be removed?

src/MCPClient/lib/metrics.py
```diff
@@ -0,0 +1,234 @@
...
+    for script_name, _ in modules_config.items("supportedBatchCommands"):
+        job_counter.labels(script_name=script_name)
+        job_processed_timestamp.labels(script_name=script_name)
+        job_error_counter.labels(script_name=script_name)
+        job_error_timestamp.labels(script_name=script_name)
+        task_execution_time_summary.labels(script_name=script_name)
+
+    failure_types = ("fail", "reject")
+
+    for transfer_type in ("Standard", "Dataverse", "Dspace", "TRIM", "Maildir"):
```
replaceafill, line 131: Is there any technical reason to leave bagged transfers out?

src/archivematicaCommon/lib/databaseFunctions.py
```diff
@@ -31,11 +31,19 @@
 from django.db import close_old_connections
 from django.db.models import Q
 from django.utils import six, timezone
+from prometheus_client import Summary
 from main.models import Agent, Derivation, Event, File, FPCommandOutput, SIP
 
 LOGGER = logging.getLogger("archivematica.common")
 
 
+db_retry_time_summary = Summary(
```
replaceafill, line 40: Is there a reason not to introduce a `metrics.py` module in `archivematicaCommon` like the other packages have?

src/archivematicaCommon/lib/storageService.py
```diff
@@ -57,7 +66,7 @@ def _storage_service_url():
     return storage_service_url
 
 
-def _storage_api_session(timeout=django_settings.STORAGE_SERVICE_CLIENT_QUICK_TIMEOUT):
+def _storage_api_session(timeout=30.0):
```
replaceafill: Is this change intended?

src/archivematicaCommon/lib/storageService.py
```diff
@@ -341,7 +357,8 @@ def create_file(
             ret = response.json()
     else:
         try:
-            session = _storage_api_slow_session()
+            with ss_api_time_summary.labels(function="create_file").time():
```
replaceafill: Should this context manager go a couple of lines below, just before the `session.post` call?

replaceafill (approved):
Really great work @cole! I left a few comments but nothing major. I tested this with the newest `am/compose` environment and everything seems to work right (just a few metrics missing that are addressed in your newest PR).

I hit my instance with lots of AMAUATs to see the nice charts move :smile:

src/MCPServer/lib/archivematicaMCP.py
cole: Uhh, probably not! That looks like an accident. I'll revert that change.

src/MCPClient/lib/metrics.py
cole: No. That list is taken from the values passed to the `archivematicaSetTransferType_v0.0` client script (in workflow.json) — I couldn't find anything more definitive. Is bagged transfer another type in that sense?

src/archivematicaCommon/lib/storageService.py
cole: 🤦‍♂ Nope! Thanks, will fix.

src/archivematicaCommon/lib/databaseFunctions.py
cole: Yeah, that's a good point. I was worried it would be confusing to have two metrics modules in the path — I think the local one (i.e. in MCPClient/Server) would be used, but still there's potential for getting mixed up. Also there are only a couple metrics used in archivematica common.

I could add common_metrics or something though? Thoughts on that?

src/MCPClient/lib/metrics.py
replaceafill: Ah! I need to look but if I remember correctly bagged transfers start on a different approval workflow (or starting point). Maybe they don't pass through `archivematicaSetTransferType_v0.0`.

src/archivematicaCommon/lib/databaseFunctions.py
replaceafill: Ah! I didn't consider the path sharing :grimacing: It just seemed very well abstracted in the other cases, but you're right the MCPServer/Client have a lot more metrics. :+1: on keeping it like this.

src/archivematicaCommon/lib/storageService.py
cole: Yes, it should. Will fix.

src/MCPClient/lib/metrics.py
cole: I don't think that's a type in that sense? I'm going to leave as is for now (it's just initialization to zero) and if "Bagged" starts showing up somewhere, we can add it.

src/archivematicaCommon/lib/databaseFunctions.py
cole: You're right, the other places are a bit cleaner 😅

PR #1445, opened 2019-07-02
Ensure agents are written out for fixity event

ross-spencer:
When events are output for fixity check, we need to ensure that the
linking agents are also output. We make sure that happens here as
well as including tests that should ensure there are fewer
regressions in this area in future.

Connected to archivematica/issues#774

### DIFF
--- a/src/MCPClient/tests/test_verify_checksum.py
+++ b/src/MCPClient/tests/test_verify_checksum.py
@@ -273,7 +273,15 @@ class TestHashsum(object):
     @pytest.fixture(scope="class")
     def django_db_setup(django_db_blocker):
         """Load the various database fixtures required for our tests."""
-        fixture_files = ["transfer.json", "files-transfer-unicode.json"]
+        fixtures_dir = "verify_checksum_microservice"
+        # hashsum_agents and hashsum_unitvars work in concert to return the
+        # Archivematica current user to the result set.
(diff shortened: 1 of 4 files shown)


src/MCPClient/tests/test_verify_checksum.py
```diff
@@ -292,39 +300,84 @@ def test_write_premis_event_to_db(self):
         event_outcome = "pass"
         # Values we will write.
         detail = "suma de verificación validada: OK"
+        # Zero-based index, will be 2 including zero.
```
replaceafill: I find this zero-based need a bit confusing. Maybe you could either use the [`start` parameter](https://docs.python.org/2/library/functions.html#enumerate) of `enumerate` below like `enumerate(event.agents.all(), 1)` or maybe not use the `enumerate` counter and do `number_of_agents == len(event.agents.all())` to verify?

replaceafill (approved):
This looks good @ross-spencer :+1:. Just left a comment about a confusing variable in the test. The `create-aip` AMAUAT passes again!

src/MCPClient/tests/test_verify_checksum.py
ross-spencer: I didn't know one could do that with enumerate! That helps a lot. Thanks @replaceafill :+1:

PR #1446, opened 2019-07-04
Ensure processing config is preserved with zipfile

helenst:
This ensures the processing config is copied across when a zipfile package is moved to an internal Archivematica directory. Some tidyups and tests have been added.

This should fix https://github.com/archivematica/Issues/issues/771

### DIFF
--- a/src/MCPClient/lib/clientScripts/extract_bag_transfer.py
+++ b/src/MCPClient/lib/clientScripts/extract_bag_transfer.py
@@ -33,6 +33,7 @@ from django.db import transaction
 from main.models import Transfer
 
 # archivematicaCommon
+from fileOperations import get_extract_dir_name
 from executeOrRunSubProcess import executeOrRun
 
 
@@ -77,16 +78,9 @@ def call(jobs):
                 sharedPath = job.args[4]
 
                 basename = os.path.basename(target)
-                basename = basename[: basename.rfind(".")]
-
                 destinationDirectory = os.path.join(processingDirectory, basename)
 
-                # trim off '.tar' if present (os.path.basename doesn't deal well with '.tar.gz')
-                try:
-                    tar_extension_position = destinationDirectory.rindex(".tar")
-                    destinationDirectory = destinationDirectory[:tar_extension_position]
-                except ValueError:
-                    pass
+                destinationDirectory = get_extract_dir_name(destinationDirectory)
 
                 zipLocation = os.path.join(
                     processingDirectory, os.path.basename(target)
--- a/src/MCPClient/lib/clientScripts/extract_bag_transfer.py
+++ b/src/MCPClient/lib/clientScripts/extract_bag_transfer.py
@@ -96,14 +96,29 @@ def call(jobs):
                     job.set_status(exit_code)
                     continue
 
-                # checkForTopLevelBag
-                listdir = os.listdir(destinationDirectory)
+                # Ensure that the only thing in the destination dir is the
+                # top level directory from the extracted file, with the
+                # exception of certain files that may have been copied here
+                # previously. (For now this is just the processing config.)
+                # These files will need to be moved down a level.
+                preexisting_files = {"processingMCP.xml"}
+
+                listdir = set(os.listdir(destinationDirectory))
+                to_move = listdir & preexisting_files
+                listdir -= preexisting_files
+
                 if len(listdir) == 1:
-                    internalBagName = listdir[0]
+                    internalBagName = listdir.pop()
+
+                    for filename in to_move:
+                        shutil.move(
+                            os.path.join(destinationDirectory, filename),
+                            os.path.join(destinationDirectory, internalBagName),
+                        )
+
                     # print "ignoring BagIt internal name: ", internalBagName
                     temp = destinationDirectory + "-tmp"
                     shutil.move(destinationDirectory, temp)
-                    # destinationDirectory = os.path.join(processingDirectory, internalBagName)
                     shutil.move(
                         os.path.join(temp, internalBagName), destinationDirectory
                     )
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -201,7 +201,7 @@ def _copy_from_transfer_sources(paths, relative_destination):
     files = {l["uuid"]: {"location": l, "files": []} for l in transfer_sources}
 
     for item in paths:
-        location, path = Path(item).parts()
+        location, path = LocationPath(item).parts()
         if location is None:
             location = _default_transfer_source_location_uuid()
         if location not in files:
@@ -303,8 +303,6 @@ def _move_to_internal_shared_dir(filepath, dest, transfer):
                     extract_dir,
                     e,
                 )
-
-
 
 
 @auto_close_old_connections()
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -25,6 +25,11 @@ from server.jobs import JobChain
 from server.processing_config import copy_processing_config
 from server.utils import uuid_from_path
 
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
 
 logger = logging.getLogger("archivematica.mcp.server.packages")
 
@@ -121,22 +126,27 @@ def _file_is_an_archive(filepath):
 def _pad_destination_filepath_if_it_already_exists(filepath, original=None, attempt=0):
     if original is None:
         original = filepath
+    filepath = Path(filepath)
+    original = Path(original)
+
     attempt = attempt + 1
-    if not os.path.exists(filepath):
+    if not filepath.exists():
         return filepath
-    if os.path.isdir(filepath):
+    if filepath.is_dir():
         return _pad_destination_filepath_if_it_already_exists(
-            original + "_" + str(attempt), original, attempt
+            "{}_{}".format(original, attempt), original, attempt
         )
+
     # need to work out basename
-    basedirectory = os.path.dirname(original)
-    basename = os.path.basename(original)
+    basedirectory = original.parent
+    basename = original.name
+
     # do more complex padding to preserve file extension
     period_position = basename.index(".")
     non_extension = basename[0:period_position]
     extension = basename[period_position:]
     new_basename = "{}_{}{}".format(non_extension, attempt, extension)
-    new_filepath = os.path.join(basedirectory, new_basename)
+    new_filepath = basedirectory / new_basename
     return _pad_destination_filepath_if_it_already_exists(
         new_filepath, original, attempt
     )
@@ -373,7 +383,7 @@ def _capture_transfer_failure(fn):
 def _determine_transfer_paths(name, path, tmpdir):
     if _file_is_an_archive(path):
         transfer_dir = tmpdir
-        p = Path(path).path
+        p = LocationPath(path).path
         filepath = os.path.join(tmpdir, os.path.basename(p))
     else:
         path = os.path.join(path, ".")  # Copy contents of dir but not dir
@@ -478,7 +488,7 @@ def _start_package_transfer(
     _move_to_internal_shared_dir(filepath, starting_point.watched_dir, transfer)
 
 
-class Path(object):
+class LocationPath(object):
     """Path wraps a path that is a pair of two values: UUID and path."""
 
     uuid, path = None, None
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -291,18 +291,14 @@ def _move_to_internal_shared_dir(filepath, dest, transfer):
 
         # Move the processing config into the extraction directory so that it
         # is preserved and used in the workflow
-        processing_config = "processingMCP.xml"
-        config_path = filepath.parent / processing_config
-        if config_path.exists():
-            try:
-                config_path.rename(extract_dir / processing_config)
-            except OSError as e:
-                raise Exception(
-                    "Error moving processing config %s to %s (%s)",
-                    config_path,
-                    extract_dir,
-                    e,
-                )
+        files_to_preserve = {"processingMCP.xml"}
+        for filename in files_to_preserve:
+            path = filepath.parent / filename
+            if path.exists():
+                try:
+                    path.rename(extract_dir / filename)
+                except OSError as e:
+                    raise Exception("Error moving %s to %s (%s)", path, extract_dir, e)
 
 
 @auto_close_old_connections()
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -8,7 +8,6 @@ import ast
 import collections
 import logging
 import os
-import shutil
 from tempfile import mkdtemp
 from uuid import UUID, uuid4
 
@@ -18,6 +17,7 @@ from django.utils import six
 
 import storageService as storage_service
 from archivematicaFunctions import unicodeToStr
+from fileOperations import get_extract_dir_name
 from main import models
 
 from server.db import auto_close_old_connections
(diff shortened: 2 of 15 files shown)


src/archivematicaCommon/lib/fileOperations.py
```diff
@@ -394,3 +394,23 @@ def findFileInNormalizationCSV(
...
+    extract_dir = filename[: filename.rfind(".")]
+
+    # trim off '.tar' if present
+    try:
+        tar_extension_position = extract_dir.rindex(".tar")
+        extract_dir = extract_dir[:tar_extension_position]
+    except ValueError:
+        pass
+
+    return extract_dir
```
alexwlchan: This code isn't especially robust (I know you didn't write it, but while we're here and adding tests we can improve it). How much that matters depends on where we get these filenames from

Two bugs jump out at me:

* What if the filename is in all caps? e.g. `transfer3.TAR.GZ`. Should be case insensitive.
* What if `.tar` occurs somewhere other than the end of the file? e.g. `my.explanation.of.tar.files.txt`. Should only match it at the end of the filename.

Suggestion:

```python
extract_dir, _ = os.path.splitext(filename)

if extract_dir().endswith(".tar"):
    extract_dir = extract_dir[:-len(".tar")]

return extract_dir
```

Plus some extra test cases.

src/MCPServer/tests/test_package.py
```diff
@@ -40,3 +51,80 @@ def test__determine_transfer_paths(name, path, tmpdir, expected):
     assert results[0] == expected[0], "name mismatch"
     assert results[1] == expected[1], "path mismatch"
     assert results[2] == expected[2], "tmpdir mismatch"
+
+
+class TestPadDestinationFilePath:
+    def test_zipfile_does_not_exist(self, tmp_path):
```
alexwlchan: minor: this test name doesn’t quite make sense to me. It sounds like we're testing that a zipfile doesn't exist, but the code below doesn't do any existence checks.

Is it checking the `_pad_destination_filepath_if_it_already_exists` function works if the zip file doesn't exist?

src/MCPServer/tests/test_package.py
```diff
@@ -40,3 +51,80 @@ def test__determine_transfer_paths(name, path, tmpdir, expected):
...
+        transfer_path = tmp_path / "transfer.zip"
+        padded_path = _pad_destination_filepath_if_it_already_exists(str(transfer_path))
```
alexwlchan: Does this method not support path-like objects? :(

src/MCPServer/tests/test_package.py
```diff
@@ -40,3 +51,80 @@ def test__determine_transfer_paths(name, path, tmpdir, expected):
...
+    def test_dir_does_not_exist(self, tmp_path):
+        transfer_path = tmp_path / "transfer/"
+        padded_path = _pad_destination_filepath_if_it_already_exists(str(transfer_path))
+        assert padded_path == str(transfer_path)
+
+    def test_dir_is_padded_if_exists(self, tmp_path):
+        transfer_path = tmp_path / "transfer/"
+        transfer_path.mkdir()
+        padded_path = _pad_destination_filepath_if_it_already_exists(str(transfer_path))
+        assert padded_path == str(tmp_path / "transfer_1")
```
alexwlchan: Minor: I find it odd that you're testing an `_` prefix function; that feels like you're testing an implementation detail. Either make this a public function of a utility package, or treat it as an internal function that doesn't get tested directly.

(LMK if that's not clear, I am struggling at words today.)

src/MCPServer/tests/test_package.py
alexwlchan: To match the others below, how about:

```python
def test_zipfile_is_padded_if_does_not_exist(…)
```

src/MCPServer/tests/test_package.py
```diff
@@ -40,3 +51,80 @@ def test__determine_transfer_paths(name, path, tmpdir, expected):
...
+        padded_path = _pad_destination_filepath_if_it_already_exists(str(transfer_path))
+        assert padded_path == str(transfer_path)
+
+    def test_zipfile_is_padded_if_exists(self, tmp_path):
+        transfer_path = tmp_path / "transfer.zip"
+        transfer_path.touch()
+        padded_path = _pad_destination_filepath_if_it_already_exists(str(transfer_path))
+        assert padded_path == str(tmp_path / "transfer_1.zip")
+
+    def test_dir_does_not_exist(self, tmp_path):
```
alexwlchan: Ditto above.

alexwlchan (commented):
Some comments inline.

src/MCPServer/tests/test_package.py
helenst: No :( I guess it would be nice if it (and others) did. Would need for pathlib2 to be included in main requirements file, not just test.

src/MCPServer/tests/test_package.py
helenst: Hmm, I see what you're saying. It does feel like it's a unit that should be tested, and I'd rather test all the possibilities at this level than work them into more complex tests of the calling functions. I'm inclined to go with the second option, but it'd be good to know what the Artefactual devs think :-)

src/archivematicaCommon/lib/fileOperations.py
helenst: Agreed, now it's extracted as a function it probably is worth improving a little.

src/MCPServer/tests/test_package.py
sevein: I think that's fine. We only started using it recently (in tests as you said) and found it very pleasant.

src/MCPServer/tests/test_package.py
sevein: We don't have a hard rule there. We're still at that point where I believe that any test is better than no test - so I'm personally pretty open about it. Also I trust whatever is your preference and I'm mostly learning here. I haven't had many chances before to see the value of different approaches compared. I've definitely written tests like yours before before, also when trying to understand a function that I have no idea what is doing (legacy code).

helenst:
I've responded to most of the comments here - I've improved `get_extract_dir_name` as well as adding pathlib support in some areas.

ross-spencer:
Hi @helenst, sorry, it just struck me, re: setting expectations. We've reached code-freeze (last-Wednesday) for `1.10` and the release candidates are currently being tested. I suspect this will end up going in the triage bucket for `1.11` and will likely be one of the first off the rank in terms of final review and merging.

src/MCPServer/lib/package.py
```diff
@@ -261,6 +264,29 @@ def _move_to_internal_shared_dir(filepath, dest, transfer):
         )
         transfer.save()
 
+    if not is_dir:
+        # Transfer is not a directory so it is an uploaded zipfile.
+        # Precreate the extraction directory for the uploaded zipfile
+        extract_dir = get_extract_dir_name(dest)
+        try:
+            os.mkdir(extract_dir)
```
ross-spencer: Hi @helenst is this in production at Wellcome, or are there changes upstream not in this PR?

I am working off 8ab2f81ad45ae038c43e803ac0de39c33de26b94 and I am seeing it fail with a message along the lines of:

```
Error validating BagIt package: Expected bagit.txt does not exist: /var/archivematica/sharedDirectory/currentlyProcessing/ZippedBag-718dfe84-fa1f-46b5-a172-c26f906285fa/bagit.txt
Failed bagit compliance. Not restructuring.
```

```
ZippedBag/
├── processingMCP.xml
└── ZippedBag     <--- my bag if it is called 'ZippedBag.zip'
    ├── bag-info.txt
    ├── bagit.txt
    ├── data
    │   ├── forkleaf-sundew.jpg
    │   └── roundleaf-sundew.jpg
    ├── manifest-md5.txt
    └── tagmanifest-md5.txt
```

My understanding is that this `mkdir` changes the behavior of the job entirely so that because this folder exists by the time we reach [here](https://github.com/artefactual/archivematica/blob/stable/1.10.x/src/MCPClient/lib/clientScripts/extract_bag_transfer.py#L47-L49) then 7z will extract the bag to this current directory including the bag's top level directory creating a nested structure. The behavior of the job previously was such that the directory would be created anew by the command. Hence there was never a processing config in there. 

The usual pitfall with bag transfers is that they will nearly always work if you accidentally transfer as a `standard` transfer - there's nothing special about a bag! But this doesn't seem to pass that way which is good. I wonder if there's an interaction with your zip type transfers that makes this work?

I wonder if you can take a look, and perhaps rebase this one now while there is an opportunity so it is up-to-date with the most recent MCP server code, and then I can re-review? 

I'd like to note too, there have been some large-ish changes to the server. If you take a look, and I'll look too as they're new to me, and ask around the team, I'd like to see if there are alternative solutions to this too, e.g. maybe making the server aware of the pre-configuration sooner rather than putting it into the transfer structure.

But let me know about it's use in production above, and maybe I'm not doing something right in this testing. Thanks in advance.

ross-spencer (changes requested):
Hi Helen, a bit of a chunky comment, if you can take a look, we'll try and understand the implications for the triage for 1.11 which we're in the middle of but hope to resolve around the beginning of next week.

src/MCPServer/lib/package.py
helenst: @ross-spencer I've rebased. We have been using it in production with the zip fix too and it works. It doesn't look as if the zip fix should affect this feature, but let me know how you get on with the rebased version and I'll take a closer look.

replaceafill: @helenst @ross-spencer I just tested the `wellcometrust:preserve-processing-config` branch. I had to change [this Path reference](https://github.com/artefactual/archivematica/blob/7c68c90fde617bb085c97834c94f3993560f2dc0/src/MCPServer/lib/server/packages.py#L205) to be [the new LocationPath class](https://github.com/artefactual/archivematica/blob/7c68c90fde617bb085c97834c94f3993560f2dc0/src/MCPServer/lib/server/packages.py#L508) and after following Ross' steps to reproduce (creating a new processing configuration and using `httpie`) I found the same `bagit` error:

```
Error validating BagIt package: Expected bagit.txt does not exist: /var/archivematica/sharedDirectory/currentlyProcessing/ZippedBag-f12c2765-24e6-40dd-a206-7665e27a2da1/bagit.txt
Failed bagit compliance. Not restructuring.
```

helenst: OK - it turned out the bag extraction phase needed a bit more work to ensure the processingMCP file ended up at the right level (and ensure its presence wasn't throwing everything else off)

@ross-spencer @replaceafill Let me know if it seems good to you now?

src/MCPClient/lib/clientScripts/extract_bag_transfer.py
```diff
@@ -96,14 +96,29 @@ def call(jobs):
...
+                # These files will need to be moved down a level.
+                preexisting_files = {"processingMCP.xml"}
+
+                listdir = set(os.listdir(destinationDirectory))
+                to_move = listdir & preexisting_files
+                listdir -= preexisting_files
+
                 if len(listdir) == 1:
-                    internalBagName = listdir[0]
+                    internalBagName = listdir.pop()
```
ross-spencer, line 111: Just wondering if `pop()` is intended? According to the [docs](https://docs.python.org/2/library/sets.html) it will return an arbitrary element, rather than say the first or last. Ultimately we get what we need here with there being only one element to care about.

src/MCPClient/lib/clientScripts/extract_bag_transfer.py
```diff
@@ -96,14 +96,29 @@ def call(jobs):
...
+
+                    for filename in to_move:
+                        shutil.move(
+                            os.path.join(destinationDirectory, filename),
+                            os.path.join(destinationDirectory, internalBagName),
+                        )
```
ross-spencer, line 117: I was struggling to visualize when this would be triggered so I gave it a go but I can't figure out the structure that will make this move necessary with a bag. My structure looks like:

```
SimpleBagWithProcessingMCP/
├── bag-info.txt
├── bagit.txt
├── data
│   ├── LICENSE
│   ├── processingMCP.xml
│   ├── README
│   └── SumiyoshiHonsha.png
├── manifest-sha256.txt
├── manifest-sha512.txt
├── tagmanifest-sha256.txt
└── tagmanifest-sha512.txt
```

And the zip can have the top-level directory as its central directory `SimpleBagWithProcessingMCP`, or the bag contents directly at the top. 

Do you have an example to share? Maybe a zip to post to the PR?

src/MCPServer/lib/server/packages.py
```diff
@@ -454,7 +499,7 @@ def _start_package_transfer(
     _move_to_internal_shared_dir(filepath, starting_point.watched_dir, transfer)
 
 
-class Path(object):
+class LocationPath(object):
```
ross-spencer, line 501: :+1:

src/MCPServer/lib/server/packages.py
```diff
@@ -240,22 +261,46 @@ def _move_to_internal_shared_dir(filepath, dest, transfer):
...
+    except OSError as e:
         raise Exception("Error moving from %s to %s (%s)", filepath, dest, e)
     else:
-        transfer.currentlocation = dest.replace(
+        transfer.currentlocation = str(dest).replace(
             settings.SHARED_DIRECTORY, "%sharedPath%", 1
         )
         transfer.save()
 
+    if not is_dir:
```
ross-spencer, line 283: This works nicely, and the tests are nice to see as well.

ross-spencer (commented):
Thanks @helenst this is a good PR, and I've learned quite a bit following it. i haven't used pathlib so this is a first. 

It's a tricky PR too. In future it would be nice to not have to be doing so much with file paths. Especially for the processing config, one thing in my mind is loading that into memory as quickly as possible so it doesn't have to persist as something on disk we're juggling like this. 

I think where you've put the new functionality to do that juggling makes logical sense. 

I have one (I think just the one) open question in the comments that I'd just like to query before finally approving. Can you see about addressing that? 

RE: sample data, I've some of my own, but if Wellcome wanted to provide a sample for testing, we could look at including that in the sample data repository for the public's use. 

If it helps to answer the question in `extract_bag_transfer.py` my focus is on zipped bags which is where this issue definitely exists for. 

For zipped transfers, the potential to add them is still open, the ticket hasn't been added to `1.11` yet, but I suspect it fell through the gap. Can you take a look at Sarah's comment [here](https://github.com/archivematica/Issues/issues/682#issuecomment-557719450).

I suspect you're already doing much of what Evelyn asked so that issue must be nearly good to go. 

If the answer to my questions in this CR are specific to zip file transfers, then we might want to just make sure that other ticket gets the milestone and then we can merge this in anticipation. 

Let me know.

I'm away for the day, back Thursday so I can follow up with folk then, but obviously if it helps expedite and sure things up just give Sarah a ping in Slack.

src/MCPClient/lib/clientScripts/extract_bag_transfer.py
helenst: It's because I made `listdir` a set rather than a list, so we can perform set operations on it to figure out what needs to be moved.

src/MCPClient/lib/clientScripts/extract_bag_transfer.py
helenst: When MCPServer starts the transfer and needs to include a non-default processing config, it creates a directory named `SimpleBagWithProcessingMCP` and places the config inside it. So we have:
```
SimpleBagWithProcessingMCP/
├── processingMCP.xml
```
When the zipfile is extracted, we end up with
```
SimpleBagWithProcessingMCP/
├── processingMCP.xml
├── SimpleBagWithProcessingMCP
│   ├── bag-info.txt
│   ├── bagit.txt
│   ├── data
│   │   ├── LICENSE
│   │   ├── processingMCP.xml
│   │   ├── README
│   │   └── SumiyoshiHonsha.png
│   ├── manifest-sha256.txt
│   ├── manifest-sha512.txt
│   ├── tagmanifest-sha256.txt
│   └── tagmanifest-sha512.txt
```

But now the processing MCP is in the wrong place and will mess with processing of the bag further on in the workflow. So we move it down a level, to the same place it already ends up with non-zipped transfers.

Does that make sense?

ross-spencer: I think so? The logic certainly. Would I be right in saying this would happen when a user starts a transfer from the API and specifies a processing config there?

helenst: Yes, that's right :)

ross-spencer: Perfect! Thanks @helenst

ross-spencer (approved):
Awesome! :+1:

PR #1447, opened 2019-07-05
Perform validation later in PID pid binding script

ross-spencer:
When validation was added in 97b1ad0c4faa328fd2e9e34724d60743c029c9af it didn't take into account that all 'required' parameters might not be available at the time validation is done. We are interested in the fields of the 'handle' configuration form in Archivematica's dashboard. If some part of that form is incomplete at the time it arrives at validation in this PR, then we can exit early and provide some useful feedback to the user. We can ignore the elements expected to be added to the configuration [here](https://github.com/artefactual/archivematica/blob/85074e77db9dc9d10f44f0d5461d269b739fea6c/src/MCPClient/lib/clientScripts/bind_pids.py#L202) in the same script for example. 

Connected to archivematica/issues#776

### DIFF
--- a/src/MCPClient/lib/clientScripts/bind_pids.py
+++ b/src/MCPClient/lib/clientScripts/bind_pids.py
@@ -58,7 +58,12 @@ from main.models import DashboardSetting, Directory, SIP
 
 # archivematicaCommon
 from archivematicaFunctions import str2bool
-from bindpid import bind_pid, BindPIDException, _validate
+from bindpid import (
+    bind_pid,
+    BindPIDException,
+    _validate_handle_server_config,
+    _validate_entity_type_required_params,
+)
 from custom_handlers import get_script_logger
 import namespaces as ns
 
@@ -200,6 +205,7 @@ def _bind_pid_to_model(job, mdl, shared_path, config):
         job, mdl, is_sip, shared_path, config["handle_archive_pid_source"]
     )
     config.update({"entity_type": entity_type, "desired_pid": desired_pid})
+    _validate_entity_type_required_params(config)
     try:
         msg = bind_pid(**config)
         _add_pid_to_mdl_identifiers(mdl, config)
@@ -224,7 +230,7 @@ def main(job, sip_uuid, shared_path, bind_pids_switch):
         handle_config.get("pid_request_verify_certs", "True")
     )
     try:
-        _validate(handle_config)
+        _validate_handle_server_config(handle_config)
     except BindPIDException as err:
         logger.info(err)
         raise BindPIDsException
--- a/src/archivematicaCommon/lib/bindpid.py
+++ b/src/archivematicaCommon/lib/bindpid.py
@@ -151,14 +151,15 @@ import requests
 
 # Parameters required when requesting the binding of a handle PID.
 REQ_PARAMS = (
-    "entity_type",
-    "desired_pid",
     "naming_authority",
     "pid_web_service_endpoint",
     "pid_web_service_key",
     "handle_resolver_url",
     "pid_request_body_template",
 )
+
+# To bind a PID we require a model and a desired persistent identifier.
+REQ_ENTITY_PARAMS = ("entity_type", "desired_pid")
 
 # Parameters that may be specified in a config file instead of passed on the
 # command line.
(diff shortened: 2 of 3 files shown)


src/MCPClient/lib/clientScripts/bind_pids.py
```diff
@@ -224,7 +224,9 @@ def main(job, sip_uuid, shared_path, bind_pids_switch):
         handle_config.get("pid_request_verify_certs", "True")
     )
     try:
-        _validate(handle_config)
+        # We know entity_type and desired_pid isn't provided until later, we
+        # can't validate without ignoring it here.
```
replaceafill: This comment is a bit confusing if you're not familiar with the code. It's explaining what is about to happen not why (Why are those parameters set in `_bind_pid_to_model`?)

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -206,12 +206,17 @@ class BindPIDException(Exception):
     pass
 
 
-def _validate(argdict):
+def _validate(argdict, ignore_values=()):
```
replaceafill: Could we explain the `ignore_values` parameter in the docstring?

src/archivematicaCommon/lib/bindpid.py
```diff
@@ -206,12 +206,17 @@ class BindPIDException(Exception):
...
     """Validate the argument dictionary ``argdict`` passed to ``bind_pid``."""
     for param in REQ_PARAMS:
+        if param in ignore_values:
+            continue
         val = argdict.get(param)
         if not val:
             raise BindPIDException("A value for parameter {} is required".format(param))
+    # In cases we haven't the entity type we are looking for, return early.
```
replaceafill: This sounds like we're "getting" the value of entity type and we don't like what we get so we return. Like what it's happening a couple of lines below (`et_validation = ENTITY_TYPES.get(entity_type)`). We're returning early because the caller decided we could do so, correct?

replaceafill (approved):
@ross-spencer The change works but is a bit hard to understand. The early return added to `bindpid._validate` didn't help either.

It made me wonder if it'd be better to remove `entity_type` and `desired_pid` from `REQ_PARAMS`. It feels like "they're not really required anymore under **some** circunstances". And I don't see any negative implications to `bind_pid._get_bind_pid_config` which seems to be the only other caller of `bindpid._validate` and which makes sure both parameters are set before the call.

src/MCPClient/lib/clientScripts/bind_pids.py
ross-spencer: Good point. I've decided to rework this and will re-submit for re-review.

src/archivematicaCommon/lib/bindpid.py
ross-spencer: It does. It has made me think twice about this approach.

ross-spencer:
Hi @replaceafill I've reworked the PR into something a little more sensible. Your questions were insightful. I held back a little in the original PR because I didn't want to pull apart the code too much, but I do think the decisions should still be sound. Holding back still and focusing on `bind_pids` I've separated validation responsibilities and so I was able to remove the `ignore=` and then also make sure validation is happening exactly where it should be. I wrapped my two new functions in a new function with the `_validate` function signature so that it doesn't break `bind_pids`. I feel like there is still a little more we might be able to do for these scripts (they're not being introduced in `1.10` so perhaps any more broad changes can be pushed into a later release depending on how significant they look/feel.

Anyhoo!! It's ready to re-review for now!! 

Related, I don't know if you keep your own `Makefile` for am.git, (you might already have made the following modification) but I've added a recipe PR here for testing `archivematica-common`. As this change touches on a module in that component of Archivematica you can test that I haven't broken anything using this. 

* https://github.com/artefactual-labs/am/pull/84

replaceafill:
Thank you @ross-spencer Sorry for putting extra work on you but this is easier to grasp in my opinion. :+1: from me.

My `Makefile` is a bunch of customizations :blush: So far I've been testing with tox directly in a separate virtualenv (similarly to how travis does it).

ross-spencer:
@replaceafill worth it I think too! :) 

RE: `Makefile` we need to have a dev day and share what we're all doing - I feel there's a lot to learn from each other :nerd_face:

PR #1449, opened 2019-07-07
Add METS download button

ross-spencer:
Add a METS download button to the Archival Storage Tab:

image

This commit does a handful of things:

* Adds a new METS download capability, and button to Archival Storage.
* Moves archival storage service tests to their own component.
* Refactors some of the pointer file download work where it overlapped with METS download.
* Adds tests for METS download and pointer file download.
* Adds tests to the views for both those features too. 
* Introduces [AMClient to the dashboard](0009-unification-of-api-handling-in-am.html). 

The tests used for the views seem less sensible to me, but I _think_ they will at least ensure that the interfaces do not change over time without explicitly being changed. Both sets of tests were a bit of a learning experience for me. I tried to figure out what was and wasn't important in two fundamentally very simple functions, but I hope that work paid off. And I am looking for advice on how to improve them as well.

The commits reflect some of that learning, with the later commits much more stable as the tests became easier to write. I can rebase if it helps CR. Commits will be squished into a feature commit once reviewed. 

Connected to archivematica/issues#644


ablwr:
This is cool and extremely useful.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -0,0 +1,199 @@
...
+    )
+    response = client.get(
+        "/archival-storage/download/aip/{}/pointer_file/".format(sip_uuid)
+    )
+    assert isinstance(response, HttpResponse)
+    assert response.status_code == mock_status_code
+    assert json.loads(response.content) == mock_error_message
+
+
+def test_get_mets_known_pointer(client, mocker, django_user_model, username, password):
```
replaceafill: Should this be `test_get_pointer_known_pointer`?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -344,6 +344,20 @@ def aip_file_download(request, uuid):
...
 
 
+def aip_mets_file_download(request, uuid):
+    """Download an individual AIP METS file."""
+    es_client = elasticSearchFunctions.get_client()
+    try:
+        aip = elasticSearchFunctions.get_aip_data(es_client, uuid, fields="name")
+    except IndexError:
+        # TODO: 404 settings for the project do not display this to the user (only DEBUG).
+        raise Http404(_("The AIP package for the requested METS cannot be found"))
```
replaceafill: I'm having trouble understanding the `requested METS` part of this message.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -0,0 +1,199 @@
...
+@pytest.fixture
+def password():
+    return "test"
+
+
+def test_get_mets_unknown_mets(client, mocker, django_user_model, username, password):
+    django_user_model.objects.create_user(username=username, password=password)
+    client.login(username=username, password=password)
+    helpers.set_setting("dashboard_uuid", "test-uuid")
+    mocker.patch("components.archival_storage.views.aip_mets_file_download")
```
replaceafill: You're mocking the actual view function here that I think is the target of these functional tests. This also makes the test module to fail when run in isolation in Docker. If I remove this line (and the pointer equivalent) from each test case the module passes in isolation.

src/dashboard/tests/test_helpers.py
```diff
@@ -0,0 +1,159 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+import __builtin__
+import json
+import os
+import requests
+import tempfile
```
replaceafill: You don't need to import a module from the standard library in order to mock part of it's functionality. You can mock it without import it like this:

```
mocker.patch("os.path.isfile", return_value=False)
```

src/dashboard/tests/test_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+    assert response.status_code == 503
+
+
+def test_stream_mets_from_storage_no_file(mocker, tmp_path):
+    custom_error_message = "METS file cannot be opened"
+    mocker.patch.object(AMClient, "__init__", return_value=None)
+    mocker.patch.object(
+        helpers, "get_setting", return_value="http://storage-service-url/"
+    )
+    mocker.patch.object(AMClient, "extract_file")
```
replaceafill: I think that if you don't care about the return value of `extract_file` here you could mock `AMClient` entirely.

src/dashboard/tests/test_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+def test_stream_mets_from_storage_success(mocker):
+    METS = """<?xml version='1.0' encoding='UTF-8'?>
+    <mets:mets xmlns:mets="http://www.loc.gov/METS/" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.loc.gov/METS/ http://www.loc.gov/standards/mets/version1121/mets.xsd">
+        <mets:metsHdr CREATEDATE="2020-01-20T15:22:15"/>
+    </mets:mets>
+    """
+    test_tmp_dir = "{}{}".format(os.path.sep, os.path.join("test-tmp-dir", ""))
+    sip_uuid = "33333333-3333-3333-3333-333333333333"
+    mets_file = "METS.{}.xml".format(sip_uuid)
+    mets_file_path = "{}{}".format(test_tmp_dir, mets_file)
```
replaceafill: Is this whole path setup necessary? We could use any name (`anything.pdf`) and we just care `open` was called with it, right? I guess it's a matter of taste how "real world" you want to make the test looks like, but we're mocking anyway :man_shrugging:

src/dashboard/tests/test_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+        response_text = "{}{}".format(response_text, response_char)
+    assert response_text == METS_PTR
+    # Make the request again, but ask to view in the browser, i.e. inline.
+    response = helpers.stream_file_from_storage_service(pointer_url, preview_file=True)
+    assert response.get("Content-Disposition") == "inline"
+    assert response_text == METS_PTR
+
+
+def test_stream_pointer_from_storage_no_content_type(mocker):
+    METS_PTR = """<?xml version='1.0' encoding='UTF-8'?>
```
replaceafill: Some parts in these tests are repeated/shared, maybe they could be extracted to separate setup functions?

src/dashboard/tests/test_helpers.py
```diff
@@ -0,0 +1,159 @@
...
+    assert response.get("Content-Type") == "application/xml"
+    assert response.get("Content-Disposition") == content_disposition
+    response_text = ""
+    for response_char in response.streaming_content:
+        response_text = "{}{}".format(response_text, response_char)
+    assert response_text == METS_PTR
+    # Make the request again, but ask to view in the browser, i.e. inline.
+    response = helpers.stream_file_from_storage_service(pointer_url, preview_file=True)
+    assert response.get("Content-Disposition") == "inline"
+    assert response_text == METS_PTR
```
replaceafill: Does this assertion have any additional value than the one a few lines above?

replaceafill (changes requested):
Nice work @ross-spencer! It's very interesting to see the `AMClient` being used in the dashboard :astonished:

I see value in making sure the tests work in isolation. I couldn't figure out why Travis doesn't complain.

src/dashboard/tests/test_helpers.py
ross-spencer: It's simply because I make a request for a second time that I want the tests to guarantee that the response is identical a second time as well. It might change for example if someone had written `preview_file=...` to take the code through an entirely different pathway that affected the way the response was returned.

replaceafill: In that case shouldn't `response_text` be updated as well after the second request is made?

ross-spencer: I see the problem, yes, you are right, I need to generate a new `response_text` here from the streaming content, which makes perfect sense! Good catch!

src/dashboard/tests/test_helpers.py
ross-spencer: I tried, but I'm not sure it would work. what we know is that AMClient itself is a class object whose init doesn't return anything, and whose member functions perform the work. I tried `mocker.patch("amclient.AMClient")` is there another trick I'm missing maybe?

src/dashboard/tests/test_helpers.py
ross-spencer: I'd like to simplify this if possible but ultimately keep this as-is too, my rationale is that our tests can reveal a lot more about the data that we're passing through the system than perhaps is revealed in any of our comments/docs etc. Keeping the tests real for me is a way to show truthfully what we expect and open up opportunities for other devs to consider what to do with that.

ross-spencer: Hmm, in the end I didn't manage to change that much at all, but other parts of the tests are simplified I think/made less redundant. More in the re-review comment!

ross-spencer:
Hi @replaceafill I've placed this in re-review. I'm glad the tests are useful, I wasn't sure of their efficacy. I've tried to remove redundancy as much as possible without making the tests difficult to read, i.e. where values are directly tested inside a test function, I haven't teased those out into functions. That being said, I have introduced some setup functions as well as a fixture or two that will help reduce the cognitive complexity somewhat I hope. Let me know what you think!

src/dashboard/src/components/helpers.py
```diff
@@ -300,32 +303,81 @@ def processing_config_path():
...
+    if stream.status_code != 200:
         response = {
             "success": False,
             "message": error_message.format(stream.status_code),
         }
         return json_response(response, status_code=400)
+    content_type = stream.headers.get("content-type", "text/plain")
+    content_disposition = stream.headers.get("content-disposition")
+    return _prepare_stream_response(
+        payload=stream.content,
```
jraddaoui: Hi @ross-spencer, I wonder if this is breaking the stream flow and loading the entire content on memory to re-stream it later.

src/dashboard/src/components/helpers.py
```diff
@@ -300,32 +303,81 @@ def processing_config_path():
...
+    err_msg = None
+    try:
+        with open(mets_tmp_file, "r") as mets_file:
+            mets = mets_file.read()
+        os.unlink(mets_tmp_file)
+    except IOError as err:
+        err_msg = err
+    if not mets:
+        response = {"success": False, "message": error_message.format(err_msg)}
+        return json_response(response, status_code=404)
```
jraddaoui: How hard would it be to add a `stream_file()` function to AMClient? Temporarily loading the file on the filesystem seems a bit to much to only re-stream it later.

src/dashboard/src/components/helpers.py
ross-spencer: I think I see what you mean @jraddaoui with the initial get requesting a stream any suggestions how to resolve? Or test? I can revert the behavior as well, but what do you think?

ross-spencer: For reference, the original function: is [here](https://github.com/artefactual/archivematica/blob/1a47b3f9227eef8e8d7a9901215fdeab52423977/src/dashboard/src/components/helpers.py#L311-L328).

src/dashboard/src/components/helpers.py
ross-spencer: @replaceafill asked about this also, but I don't believe it to be in scope. I noted to him that this feels like an incremental step towards enhancing both the behavior of Archivematica and AMClient and so it is something I believe we can react to if there proves to be a problem with this mechanism. Although admittedly, AMClient weaknesses have previously been noted too, it's just the first time we're able to find a good reason to integrate it :slightly_smiling_face:

src/dashboard/src/components/helpers.py
jraddaoui: I'd try to implement `_prepare_stream_response` using a stream instead of the full content. And I'd add a `stream_file` function to AMClient (as noted below) from where you can get a stream and pass it to `_prepare_stream_response` in a similar manner.

src/dashboard/src/components/helpers.py
jraddaoui: I understand this may not be in scope and I'm sorry I'm late to the code review. Nevertheless, from a SAD perspective, this is far from ideal. In general, the current process would be:

- The user clicks the view METS file button.
- The file is extracted and downloaded from the SS to the filesystem. Luckily, the file is not fully loaded on memory [on that process](https://github.com/artefactual-labs/amclient/blob/ee66a2e22a76e89ad1f9c47012eb2336c6a78a76/amclient/amclient.py#L599-L630).
- However, the file is fully loaded in memory in here before deleting it from the filesystem.
- A response is finally sent to the user, streaming the METS file content.

Now that I see the `extract_file` function, I think you could easily achieve the stream behavior. It doesn't have to be a new function, it could be an optional parameter on that function, `return_stream` for example. When it's set to `true` the response is returned directly [here](https://github.com/artefactual-labs/amclient/blob/ee66a2e22a76e89ad1f9c47012eb2336c6a78a76/amclient/amclient.py#L608) before the file is loaded on the filesystem.

src/dashboard/src/components/helpers.py
ross-spencer: Hi @jraddaoui as discussed in Slack how is this for an initial compromise? The flag allows us to not stream for the METS while we develop out AMClient, and once that functionality is available we'll be able to switch that flag and remove the code that relies on it. Tests now reflect the data we're anticipating better.

The commit is: 7ac55a035b771c059e365a462aa5954e4cfb8b3d

src/dashboard/tests/test_helpers.py
replaceafill: You could mock the `AMClient` class from the `helpers` module (or anything else in that module) like this:
 
```python
def mock_amclient_details(mocker, side_effect=None):
    mocker.patch(
        "components.helpers.get_setting", return_value="http://storage-service-url/"
    )
    amclient_mock = mocker.Mock(**{"extract_file.side_effect": side_effect})
    mocker.patch("components.helpers.AMClient", return_value=amclient_mock)
````

File: src/dashboard/src/components/helpers.py (hunk 9 above)
Comment (replaceafill) [reply src/dashboard/src/components/helpers.py]: Very glad that @jraddaoui spotted this :bowing_man: I've been thinking this was code just being extracted into a new method, and didn't notice the change in the behaviour.

File: src/dashboard/src/components/helpers.py [hunk 11]
```diff
@@ -300,32 +303,91 @@ def processing_config_path():
...
+    preview_file=False,
+    already_streaming=True,
+):
+    """Prepare the streaming response to return to the caller."""
+    if preview_file:
+        content_disposition = "inline"
+    if not already_streaming:
+        response = StreamingHttpResponse(payload)
+    else:
+        response = payload
```
jraddaoui: In the end it's doing the same on both cases. If we always do `response = StreamingHttpResponse(payload)` here and we pass the `stream` variable directly in https://github.com/artefactual/archivematica/pull/1449/files#diff-254823927ae6fbe4584582f13c9b0c18R386, we could remove the `already_streaming` flag. It also makes a little more sense to me to create the `StreamingHttpResponse` always on this function.

jraddaoui (changes requested):
Thanks a lot @ross-spencer!

Reading again our Slack chat, I completely misunderstood the compromise, I'm sorry about that. While not breaking the stream and not saving the entire download on memory was the most important fix, specially because the same function is used to download full AIPs, I really think that we should improve the new METS file download process (commented [here](https://github.com/artefactual/archivematica/pull/1449#discussion_r370744439)) before merging this PR. METS files can be quite big too and proxying the stream directly is going to be a noticeable enhancement that we should try to address on its first implementation. 

Apart from that, only a minor comment about the `already_streaming` flag.

src/dashboard/src/components/helpers.py
ross-spencer: Thanks @jraddaoui I hope you appreciate the technical aspects here are not lost on me (silly introduced bugs aside). Rather my hesitation (as always is) discussing with both you and @replaceafill were around the other organizational context a new feature finds itself wrapped within within. Thank you for taking the time to review this on your weekend. 

From a CR perspective, I appreciate you taking the time to propose something based on the code that was in AMClient and spotting the short-cut to adding streaming. Where there is room to provide mentorship in our work then that's definitely an approach that works for me, and I'm sure others. I also try and do that work up-front where it can help devs.

For others reading, the PR is [here](https://github.com/artefactual-labs/amclient/pull/13). You'll see a new helper function that enables streaming to the command-line part of AMClient too, plus the changes to make streaming possible. And I've added to ability to `stream` -- ostensibly our streaming flag to be used by the original `extract_file` we touch upon here. 

I feel the changes for AMClient really make it a much stronger component now, and lends itself well to the ADR submitted in support of its use.

src/dashboard/src/components/helpers.py
ross-spencer: That makes sense. Once AMClient is released I'll make this change here. Which again, the new tests make much easier to do while proving the efficacy of the fix. Thanks for the suggestions.

src/dashboard/tests/test_helpers.py
ross-spencer: Oh that's quite nice. Thanks Douglas. I'll put this way in.

src/dashboard/src/components/helpers.py
jraddaoui: Thank you @ross-spencer! I really appreciate the effort.

Of course, I also appreciate the technical aspects of this changes and I like the idea to introduce `amclient`. I hope that, with a few enhancements, it could replace the other helpers across the code, but without the streaming capability I wouldn't have considered it as a good option for this feature.

ross-spencer:
@jraddaoui @replaceafill okay, submitted for re-review. I think I've unwrapped most of the old download work and replaced it with streaming without having to undo too much. The `amclient` mock isn't as clever as perhaps you'd like Douglas, but I've done my best. I'm sure there's a chance I've missed something but my editor's eye is fatiguing so let me know. Commits should run logically on from one-another to help your reading.

jraddaoui (approved):
Thanks @ross-spencer!!

replaceafill (approved):
Thank you both!

ross-spencer:
Merged! @replaceafill @jraddaoui I got a lot out of this work (I can mock better now and understand streaming better!), and I'm sure the changes are useful to the project too. So I appreciate the time you've dedicated to it. Thank you for your patience with me :slightly_smiling_face:

PR #1455, opened 2019-07-16
Ensure structmap isn't included and METS create V2 fails under various error conditions

ross-spencer:
A number of different error conditions were spotted during testing of
the Archivematica 1.10 release candidate. Rather than failing
silently in the system we introduce logic to make the output more
robust. Example failure conditions for custom structMaps include when
files do not exist, or there are empty attributes for the CONTENTID
field.

Additionally, small sections of the code have been made simpler.

Tests have been added, and all sample structmaps created for this feature will still pass. Those are available from [here](https://github.com/artefactual/archivematica-sampledata/blob/master/SampleTransfers/StructMapTransferSamples).

Connected to archivematica/issues#283

### DIFF
--- a/src/MCPClient/lib/clientScripts/create_mets_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_v2.py
@@ -908,22 +908,43 @@ def include_custom_structmap(
                 state.globalStructMapCounter += 1
                 structMap.set("ID", "structMap_{}".format(state.globalStructMapCounter))
             ret.append(structMap)
-            fileids = structMap.xpath(
+            # CONTENTIDS will map to fptrs and area elements where present.
+            fptrs = root.xpath("//mets:fptr", namespaces={"mets": ns.metsNS})
+            area_elements = root.xpath("//mets:area", namespaces={"mets": ns.metsNS})
+            contentids = structMap.xpath(
                 "//*[@CONTENTIDS]", namespaces={"mets:": ns.metsNS}
             )
-            for item in fileids:
+            if not contentids:
+                state.error_accumulator.error_count += 1
+                logger.error(
+                    "No CONTENTIDS found in custom structMap. AIP METS cannot be generated"
+                )
+                return []
+            if len(contentids) < (len(fptrs) + len(area_elements)):
+                logger.error(
+                    "Mismatch of CONTENTID elements to elements we wish to replace. AIP METS cannot be generated"
+                )
+                state.error_accumulator.error_count += 1
+                return []
+            for item in contentids:
                 file_path = item.get("CONTENTIDS")
+                if not file_path:
+                    logger.error(
+                        "Empty file path in custom structMap. AIP METS cannot be generated"
+                    )
+                    state.error_accumulator.error_count += 1
+                    return []
                 normalized_path = _fixup_path_input_by_user(job, file_path)
                 if normalized_path in state.fileNameToFileID:
                     item.set("FILEID", state.fileNameToFileID[normalized_path])
                 else:
-                    job.pyprint(
-                        "Custom structmap error: no fileUUID for",
-                        file_path,
-                        normalized_path,
-                        file=sys.stderr,
-                    )
                     state.error_accumulator.error_count += 1
+                    logger.error(
+                        "No fileUUID for '%s'; original in custom structMap: %s",
+                        normalized_path,
+                        file_path,
+                    )
+                    return []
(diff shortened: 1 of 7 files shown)


src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -908,22 +908,44 @@ def include_custom_structmap(
...
+                    logger.error(
+                        "Empty file path in custom structMap. AIP METS cannot be generated"
+                    )
+                    state.error_accumulator.error_count += 1
+                    return []
                 normalized_path = _fixup_path_input_by_user(job, file_path)
                 if normalized_path in state.fileNameToFileID:
                     item.set("FILEID", state.fileNameToFileID[normalized_path])
                 else:
+                    state.error_accumulator.error_count += 1
```
replaceafill, line 941: Is there a reason to move this line up here?

src/MCPClient/tests/test_create_aip_mets.py
```diff
@@ -883,38 +883,55 @@ def test_create_file_sec(self):
...
+            "custom-structmap-41ab1f1a-34d0-4a83-a2a3-0ad1b1ee1c51",
+            "broken_structmap.xml",
+        )
+        assert os.path.isfile(broken_structmap_path)
+        assert os.path.isfile(self.mets_xsd_path)
+        try:
+            self.validate_mets(self.mets_xsd_path, broken_structmap_path)
+        except etree.DocumentInvalid:
+            assert (
+                True
```
replaceafill, line 904: Nothing to do about this, but I'm never sure how to best represent these cases: `assert True`. A comment may have the same effect.

src/MCPClient/tests/test_create_aip_mets.py
```diff
@@ -944,16 +959,7 @@ def test_get_included_structmap(self):
...
-                try:
-                    self.validate_mets(self.mets_xsd_path, structmap_path)
-                    assert False
-                except etree.DocumentInvalid:
-                    assert True
-                # Break out of the loop, nothing else to test here.
-                continue
-            else:
-                self.validate_mets(self.mets_xsd_path, structmap_path)
+            self.validate_mets(self.mets_xsd_path, structmap_path)
```
replaceafill, line 962: Nice cleanup of these `broken` functionality.

replaceafill (approved):
@ross-spencer This LGTM! :+1:

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -908,22 +908,44 @@ def include_custom_structmap(
...
                     job.pyprint(
                         "Custom structmap error: no fileUUID for",
                         file_path,
                         normalized_path,
                         file=sys.stderr,
                     )
```
replaceafill, line 946: Oh! One thing I just noticed is that part doesn't use the logger like the other  errors:

```
Custom structmap error: no fileUUID for objects/_ferdinand_short_2017_01_27.mp3 objects/_ferdinand_short_2017_01_27.mp3
```

vs

```
createmets_v2.0: ERROR     2019-07-31 14:42:04,990  archivematica.mcp.client.createMETS2.include_custom_structmap:925  Mismatch of CONTENTID elements to elements we wish to replace. AIP METS cannot be generated
```

src/MCPClient/lib/clientScripts/create_mets_v2.py
ross-spencer: Just it looked nicer ^_^;

src/MCPClient/tests/test_create_aip_mets.py
ross-spencer: That's a fair point. I might try a comment style the next time around; though I hope we move away more from test case and use the pytest style you showed me with the checksum validation microservice.

src/MCPClient/lib/clientScripts/create_mets_v2.py
ross-spencer: Hi @replaceafill I've changed it to a logger here: https://github.com/artefactual/archivematica/pull/1455/commits/47af9269e0145ae7ca238d4d72068cb92c579256 can you let me know that's okay, and then I'll merge asap and we can get into regression testing!!

replaceafill: :+1: LGTM!

PR #1457, opened 2019-07-16
Reingest: include all new agents

sevein:
This commit updates `add_events` to ensure that all PREMIS agents are listed
as new PREMIS events are incorporated during reingest. Before, we were only
covering the "software" agent.

Connects to https://github.com/archivematica/Issues/issues/659.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
@@ -343,8 +343,13 @@ def add_rights_elements(job, rights_list, files, state, updated=False):
                         break
 
 
-def _orphan_agents(fsentry):
-    """Find linking agents used in events not listed as agents."""
+def _extract_event_agents(fsentry):
+    """Find linked agents that have not been described yet.
+
+    When altering PREMIS events for a filesystem entry, it is likely to end up
+    with an incomplete set of PREMIS agents. This function returns the linked
+    agents that have no corresponding PREMIS agents yet defined.
+    """
     agents, orphans = set(), set()
 
     for premis_agent in fsentry.get_premis_agents():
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
```diff
@@ -341,32 +343,45 @@ def add_rights_elements(job, rights_list, files, state, updated=False):
                         break
 
 
+def _orphan_agents(fsentry):
+    """Find linking agents used in events not listed as agents."""
+    agents, orphans = set(), set()
```
ross-spencer, line 353: I've been wrestling with this function name and use. Orphan is quite a loaded word. It might suggest something has been lost along the way and we need to do something to fix that? so a possible afterthought? or something that we're making up-for..? But it looks like we haven't orphaned elements at all?

It looks like we're doing the following:

* Retrieving all `fsentry` objects from the METS.
* Collecting all agents from Archivematica's `Agent` model.
* Attaching all events that should belong to an `fsentry` to the `fsentry` itself.
* For each of those `fsentry` objects seen, extract the agent information associated with its events.
* Pair the extracted event agents with an existing agent in Archivematica and attach  it to our data structure (the eventual METS following reingest).

If that's right, or you have a different interpretation then I wonder if you could elaborate in the `docstring` and possibly consider a more descriptive function name to `_orphan_agents` (again, if I read correctly) maybe `_extract_event_agents`? 

Does that work?

ross-spencer (changes requested):
This looks good @sevein. Reingest feels like an incredibly complicated process to think about! :grimacing: The code works well, I've left a comment about the `_orphan_agents` function, which I guess I'd like to see clarified, or modified before we merge. I feel like we can do a little more work for the future reader of this code to explain a bit more about what is happening and why. I hope that's okay.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSReingest.py
sevein: That makes sense. I've made a new attempt, let me know if it's clearer.
Please feel free to suggest a description, happy to use yours instead.

ross-spencer (approved):
Thanks @sevein LGTM! :+1:

PR #1458, opened 2019-07-25
Dashboard: add pathlib2 as base dependency

scollazo:
It was being loaded in dev.txt, but it's needed by the
rebuild_transfer_backlog dashboard command

Connects to https://github.com/archivematica/Issues/issues/367


ross-spencer (approved):
Perfect! :+1:

PR #1459, opened 2019-07-26
Appraisal: fix draggability of arranged files

replaceafill:
This PR fixes the logic to mark nodes in the Backlog pane of the Appraisal tab as not draggable.

Connected to https://github.com/archivematica/Issues/issues/821

### DIFF
--- a/src/dashboard/tests/test_ingest.py
+++ b/src/dashboard/tests/test_ingest.py
@@ -1,3 +1,5 @@
+import os
+
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test.client import Client
@@ -6,6 +8,7 @@ import pytest
 
 from agentarchives.archivesspace import ArchivesSpaceError
 from components import helpers
+from components.ingest.views import _es_results_to_appraisal_tab_format
 from components.ingest.views_as import get_as_system_client
 from main.models import DashboardSetting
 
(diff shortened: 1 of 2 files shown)


src/dashboard/tests/test_ingest.py
```diff
@@ -76,3 +77,187 @@ def test_add_metadata_files_view(self):
             ["<h1>", "  Add metadata files<br />", "  <small>test</small>", "</h1>"]
         )
         assert title in response.content
+
+
+def _assert_file_node_properties_match_record(file_node, record):
```
ross-spencer, line 84: Nice idea to have an umbrella function for a group of validation tests.

src/dashboard/tests/test_ingest.py
```diff
@@ -76,3 +77,187 @@ def test_add_metadata_files_view(self):
...
+    assert file_node["type"] == "file"
+    assert file_node["id"] == record["fileuuid"]
+    # the relative path of the node is encoded
+    assert file_node["relative_path"].decode("base64") == record["relative_path"]
+    # the node title is the encoded file name
+    assert (
+        file_node["title"].decode("base64") == record["relative_path"].rsplit("/", 1)[1]
```
ross-spencer: Would `os.path.sep` be better to replace `/` or even, `os.path.basename()` satisfy what you're doing here?

src/dashboard/tests/test_ingest.py
```diff
@@ -76,3 +77,187 @@ def test_add_metadata_files_view(self):
...
+    transfer_node = nodes[0]
+    data_dir_node = transfer_node["children"][0]
+    # hidden directories can't be dragged either
+    logs_dir_node = data_dir_node["children"][0]
+    assert logs_dir_node["not_draggable"]
+    file_node = logs_dir_node["children"][0]
+    assert file_node["not_draggable"]
+
+
+def test_appraisal_tab_node_draggability_metadata():
```
ross-spencer, line 185: Just a question about tests vs. visibility in the UI - the metadata dirs don't show up in the UI do they? (same for some of the "special" directories below?) 

(I can't see them, either way I do like this test being here for the future)

ross-spencer (approved):
Looks good to me @replaceafill and I tested this following the instructions in the issue and before the fix I can generate an error box at the top of the screen, but after the fix, I cannot even move the file. Nice work.

src/dashboard/tests/test_ingest.py
replaceafill: Ah thanks!

src/dashboard/tests/test_ingest.py
replaceafill: You need to check the `Show metadata & logs directories?` checkbox before pressing the `Search transfer backlog` button in the Appraisal tab:

Captura de pantalla de 2019-07-30 22-47-03

ross-spencer: Oh! Well, I've learned something new! (two things if you include I should click more buttons in the UI :) Thanks @replaceafill.

PR #1460, opened 2019-08-01
Remove need to supply objects dir for checksums

ross-spencer:
Changes made for 1.10 meant the 'objects/' directory needed to be
encoded in a checksum manifest's file listing, e.g. checksum.sha512.
This is reverted here where we were able to find a better way of
achieving the same result without this requirement.

Connected to archivematica/issues#844
Sample data to test against https://github.com/artefactual/archivematica-sampledata/pull/59
Docs changes: https://github.com/artefactual/archivematica-docs/pull/303


jraddaoui (approved):
:+1:

Issue #1199, opened 2018-07-10
Problem: index_aip crashes elasticsearch for large transfers


payten:
While testing large and multiple transfers for the rate limiting investigation, we noticed elasticsearch crash when it hit max-memory.  Transfers with many files produced larger JSON documents (50 to 100MB), and the post to elasticsearch would take longer than the 10 second timeout causing a retry soon after.  As these retries pile up, elasticsearch quickly hits its memory limit and barfs.

We tried increasing the elasticsearch memory allocation to 3x the default and still hit the limit.  However, we think we can avoid this situation by increasing the default timeout from 10 seconds to 5 minutes during AIP indexing.  This will lesson the load the elasticsearch server (by avoiding all those retries) and allow time for those large documents to be indexed.

We'll test this out and prepare a PR.

ross-spencer:
Related (I think) https://github.com/artefactual/archivematica/issues/1136

payten:
Thanks @ross-spencer!  We've spoken to @jhsimpson and he's given us an overview of the greater problem.  We'll investigate reducing the size of this index_aip payload (yet still retain the fields required for searching) and report back.

ross-spencer:
Thanks @payten - is there a link to that greater exposition already on GitHub or the AM Wiki for the community to benefit from? cc. @jhsimpson

marktriggs:
Hi all,

I'm not sure if it's written down anywhere, but @jhsimpson mentioned that some users have seen very large Elasticsearch documents produced when indexing transfers containing lots of files--in the area of 1GB in some cases.

The bulk of this size seems to be due to the `mets` field that gets indexed as a part of the AIP, as this field contains a JSON representation of the full METS document.  I wanted to add a note here because it looks to me that this field isn't actually used for searching, and could potentially be dropped altogether.

The place it *does* show up in the Dashboard is when you search for an AIP and click "View raw".  You'll see a document like this:

```
{
  "_id": "AWSgvw03Sq8t1C0dM5NB", 
  "_index": "aips", 
  "_source": {
    "AICID": null, 
    "countAIPsinAIC": null, 
    "created": 1531705205, 
    "encrypted": false, 
    "filePath": "/var/archivematica/sharedDirectory/www/AIPsStore/0241/211c/ed41/4bff/837b/8400/c589/06aa/test_1-0241211c-ed41-4bff-837b-8400c58906aa.7z", 
    "identifiers": [], 
    "isPartOf": null, 
    "mets": {
[...]

PR #1463, opened 2019-08-09
Search: tune elasticsearch module for AIP indexing speed

sevein:
Indexing times are ~100x faster (index_aip) with some simple changes, broken into small commits. Highest impact seen after moving to lxml, parsing AIP METS document only once and bulk insert (in batches of 50 docs). For a 1000 xml transfer, indexing time goes from 240s to ~2.5s.

This PR also removes METS embedding in "aips" docs. The `view raw` link has been removed as a result. See https://github.com/artefactual/archivematica/issues/1199 for more.


cole (approved):
Approved, this all makes sense and looks good to me.

One thing (for another PR) is do we need to store METS details in the `aipfiles` index? If not we could remove that as well.

sevein:
Thanks @cole. We make METS details available in `aipfiles` so it's easier for users to find files from the archival storage tab via different properties. But it could be cleaned up (e.g. remove XML namespaces) and have a simpler structure/mapping that doesn't follow what's in the METS strictly.

hakamine:
@sevein (not sure if this information is relevant to the PR, just mentioning it because of the switch from ElementTree to lxml; please ignore if unrelated) in the past I had "broken" AIP indexing when trying to replace the library `xml.etree.ElementTree` with `lxml.etree` in `archivematicaCommon/elasticSearchFunctions` (ref. [AM issue 619](https://github.com/archivematica/Issues/issues/619#issuecomment-487190700) ) (had to revert back to ElementTree due to some mapping issues).

sevein:
Ok thanks @hakamine, I'll investigate!

sromkey:
@sevein can you suggest how we should test this for the 1.10.1 release?

jraddaoui:
> @sevein (not sure if this information is relevant to the PR, just mentioning it because of the switch from ElementTree to lxml; please ignore if unrelated) in the past I had "broken" AIP indexing when trying to replace the library `xml.etree.ElementTree` with `lxml.etree` in `archivematicaCommon/elasticSearchFunctions` (ref. [AM issue 619](https://github.com/archivematica/Issues/issues/619#issuecomment-487190700) ) (had to revert back to ElementTree due to some mapping issues).

Thanks for noting this @hakamine and sorry for making you to change it back before. Check https://github.com/artefactual/archivematica/issues/1199#issuecomment-586312684 and https://github.com/artefactual/archivematica/pull/1571 for more info.

PR #1465, opened 2019-08-13
Fix SIP arrange directory creation

replaceafill:
The [old `create_directory_within_arrange` function](https://github.com/artefactual/archivematica/blob/48409e31b72d877b91efefd8c51dfd4c10c70b69/src/dashboard/src/components/filesystem_ajax/views.py#L543) used to take a single `path`. But this became a problem  in https://github.com/archivematica/Issues/issues/610 and the implementation was changed to support bulk creation in https://github.com/artefactual/archivematica/pull/1390.

However, the new `paths` parameter is not being processed as a list preventing the arrange directories to be created and to have levels of description set.

This PR fixes the `paths` parameter handling.

Connected to https://github.com/archivematica/Issues/issues/856

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -626,7 +626,7 @@ def create_directory_within_arrange(request):
     """
     error = None
 
-    paths = map(base64.b64decode, request.POST.getlist("paths", []))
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Thanks @replaceafill!

I've noted a couple of edge cases and possible enhancements in the related issue but I think this is good to go. We may want to address those in different PRs and probably after 1.10.

PR #1468, opened 2019-08-14
Performance: add basic indexes

cole:
Adds indexes to very active tables (Jobs, UnitVariables, and Files) handling many common query cases. These should mostly be helpful in situations where the database has grown too large to fit into memory.

We're using often using `BLOB`/`TEXT` columns as lookups (not a good practice). Adding indexes to those requires setting a key length, which isn't something Django supports, so I've added those indexes via manual SQL. These aren't perfect, they can't be used in `LIKE '%foo'` (or ends with) queries, and in my testing MySQL seems to be ignoring them in `LIKE 'foo%'` (or starts with) queries. The most common case though is equals, and that works great (for values up to 767 bytes at least, the innodb max).

I've also looked at SIP, Transfer and Task, but those models are usually queried by uuid, which has an index.

Also removes subjob from Jobs queries, which was unnecessary and slowing things down.

Connected to archivematica/Issues#907.

### DIFF
--- a/src/MCPClient/lib/clientScripts/email_fail_report.py
+++ b/src/MCPClient/lib/clientScripts/email_fail_report.py
@@ -127,7 +127,7 @@ def get_unit_job_log_html(sip_uuid):
     """
     parser = etree.HTMLParser(remove_blank_text=True)
     rows = (
-        Job.objects.filter(sipuuid=sip_uuid, subjobof="")
+        Job.objects.filter(sipuuid=sip_uuid)
         .exclude(jobtype="Email fail report")
         .order_by("-createdtime", "-createdtimedec")
         .values_list("jobtype", "currentstep", "createdtime")
--- a/src/dashboard/src/main/migrations/0071_index_unitvariables_runsql.py
+++ b/src/dashboard/src/main/migrations/0071_index_unitvariables_runsql.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+"""A MySQL specific index for UnitVariables
+"""
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def create_index(apps, schema_editor):
+    if schema_editor.connection.vendor != "mysql":
+        return
+
+    schema_editor.execute(
+        "CREATE INDEX UnitVariables_ep46xp7f_idx ON UnitVariables (unitUUID, unitType, variable(255));"
+    )
+
+
+def drop_index(apps, schema_editor):
+    if schema_editor.connection.vendor != "mysql":
+        return
+
+    schema_editor.execute("DROP INDEX UnitVariables_ep46xp7f_idx ON UnitVariables;")
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("main", "0070_index_jobs")]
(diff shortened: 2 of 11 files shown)


cole:
Relates to https://github.com/archivematica/Issues/issues/907.

src/dashboard/src/main/migrations/0071_index_unitvariables_runsql.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations
+
+
+def create_index(apps, schema_editor):
+    if schema_editor.connection.vendor != "mysql":
+        return
+
+    schema_editor.execute(
+        "CREATE INDEX UnitVariables_ep46xp7f_idx ON UnitVariables (unitUUID, unitType, variable(255));"
```
replaceafill, line 14: Is the 255 length here determined by the possible values of the field (which I understand come from `workflow.json`) or by the field type (`TextField` in the `variable` case)?

replaceafill (approved):
This looks great to me @cole :+1: . I just left a question about the length of the index with the `variable` field.

src/dashboard/src/main/migrations/0071_index_unitvariables_runsql.py
cole: Good question. The 255 length is driven by the field type (TextField/TEXT column). I think most of the values are in workflow.json, but arbitrary values can be used by client scripts. 255 is the max (767 bytes, at _3 bytes_ per UTF8 character — there's a long digression here about how MySQL uses 3 bytes unless you explictly opt in to 4, which is what UTF8 acutally needs and we should probably do, but that's a whole other issue) that we can use. I don't think there's a downside to going longer — the index will use _up to_ 255 chars.

src/MCPClient/lib/clientScripts/email_fail_report.py
```diff
@@ -127,7 +127,7 @@ def get_unit_job_log_html(sip_uuid):
     """
     parser = etree.HTMLParser(remove_blank_text=True)
     rows = (
-        Job.objects.filter(sipuuid=sip_uuid, subjobof="")
```
cole, line 130: I'm not totally sure of the history of `subjobof`, but as I understand it jobs could previously be nested. It isn't used anymore, but because we have an index on the SIP UUID, but not subjobof, including that clause in the query is much slower.

src/dashboard/src/main/migrations/0073_index_files_runsql.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.db import migrations
+
+
+def create_indexes(apps, schema_editor):
+    if schema_editor.connection.vendor != "mysql":
+        return
+
+    schema_editor.execute(
+        "CREATE INDEX Files_transfer_lvrgv3pn_idx ON Files (transferUUID, currentLocation(767));"
```
cole, line 14: These index names are manually geneated (`{model}_{field}_{random_value}_idx`). They are trying to imitate the naming convention that Django uses, but aren't really the same. They could really be pretty much anything though.

PR #1470, opened 2019-08-20
Workflow: ensure that disabled jobs are no-ops

sevein:
This PR updates transcription and PID binding only. We should do the same in FileID but it's much harder because it uses passVar to pass the choice to all related links.

Connects to https://github.com/archivematica/Issues/issues/866.


cole (approved):
Looks good!

PR #1472, opened 2019-08-21
WIP: MCP Server performance improvements

cole:
This branch includes a number of performance enhancements grouped together for testing.

**It is not intended to be merged as is.**

For performance testing, it's recommended to run multiple MCP clients (ncpus - 1 or 2) and set `ARCHIVEMATICA_MCPSERVER_MCPSERVER_CONCURRENT_PACKAGES` to an equal number.

The main area of work is an update of of MCP server, to create a concept of active packages, and
queue packages when they arrive (via gearman RPC call or watched directory), rather than trying to process them all at once. It also modernizes the project layout, greatly reduces the number of threads used by MCPServer, and watches directories via the much more efficient inotify.

A couple new abstractions have been introduced via classes:

* PackageQueue, intended to track waiting and active packages. This class should be threadsafe.
* Various Job classes have taken over the role previously filled by linkTaskManagers (the execution of workflow links)
* JobChain, producing jobs from a given workflow chain, has a slightly expanded role
* DecisionQueue, to track jobs awaiting decisions. This class should be threadsafe.
* GearmanTaskRequest, to handle batching of file tasks to gearman (replaces taskGroupRunner)
* Package (and SIP, DIP, Transfer), to encapsulate package logic
* PackageContext, to handle some of the logic around command replacement variables

This branch is very much a work in progress. Known shortcomings (hopefully to be addressed shortly):

* ~Running the dashboard UI in a browser seems to greatly slows down or blocks package submission via the API.~ [EDIT: fixed via multiple RPCServer threads]
* ~Interactive workflow decisions (required for non-automated workflows) aren't working right now.~ [EDIT: now working correctly]
* It has dropped support for delayed task execution, used by quarantine
* There are also many new TODOs to capture small problems sprinkled throughout

This branch also includes indexes on very active query paths (and query optimizations),
and performance instrumentation work, which are detailed more in other PRs (https://github.com/artefactual/archivematica/pull/1468 and https://github.com/artefactual/archivematica/pull/1443).


sevein:
I haven't had a chance yet to look at all your changes but it's very exciting, thanks.

Regarding to the new code using inotify: I'm pretty sure that Joseph tried that at some point but we had to give up on it because there's not support in NFS to broadcast remote events. I think that it's pretty common to have the shared directory deployed as a NFS share so we're likely going to hit that limit soon. An alternative could be to fall back to the old mechanism when inotify isn't available, [gunicorn/reloader](https://github.com/benoitc/gunicorn/blob/master/gunicorn/reloader.py) does that but it's driven by config. I guess we could do similarly but looking at `stat -f -L -c %T /var/archivematica/sharedDirectory` which should return `nfs` if it's a NFS mount.

cole:
@sevein ah right, that's a big flaw. I'll have to put a fallback in place.

cole:
@sevein I've added ~fallback to~ polling by default (set `ARCHIVEMATICA_MCPSERVER_MCPSERVER_WATCH_DIRECTORY_METHOD` to `inotify` to use inotify) for users using an NFS mount.

cole:
Superseded by #1485.

PR #1473, opened 2019-08-22
Appraisal: fix bulk extractor results

replaceafill:
This fixes the paths of the bulk extractor reports by adding the
`data` directory created by bags.

Connected to https://github.com/archivematica/Issues/issues/824

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -900,7 +900,7 @@ def _get_file_formats(f):
 
 def _list_bulk_extractor_reports(transfer_path, file_uuid):
     reports = []
-    log_path = os.path.join(transfer_path, "logs", "bulk-" + file_uuid)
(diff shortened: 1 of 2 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -900,7 +900,7 @@ def _get_file_formats(f):
 
 def _list_bulk_extractor_reports(transfer_path, file_uuid):
     reports = []
-    log_path = os.path.join(transfer_path, "logs", "bulk-" + file_uuid)
+    log_path = os.path.join(transfer_path, "data", "logs", "bulk-" + file_uuid)
```
ross-spencer, line 903: Given the opportunity I feel that replacing this style for syntax for something like `"bulk-{}".format(file_uuid)` is a good way to go. But this can be picked up again another time.  I think it's a longer term maintenance issue.

ross-spencer (approved):
LGTM :+1:

PR #1474, opened 2019-08-22
MCPClient: capture std{out,err} after failure

sevein:
This pull request ensures that the the standard error stream of a job is captured (persisted in the database) when the task errors, regardless the value in the output capturing setting.

Connects to https://github.com/archivematica/Issues/issues/873.

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -222,7 +222,10 @@ def execute_command(supported_modules, gearman_worker, gearman_job):
                         end_time = getUTCDate()
 
                         kwargs = {"exitcode": exit_code, "endtime": end_time}
-                        if django_settings.CAPTURE_CLIENT_SCRIPT_OUTPUT:
+                        if (
+                            django_settings.CAPTURE_CLIENT_SCRIPT_OUTPUT
+                            or kwargs["exitcode"] > 0
+                        ):
                             kwargs.update(
                                 {
                                     "stdout": job.get_stdout(),
--- a/src/MCPClient/lib/job.py
+++ b/src/MCPClient/lib/job.py
@@ -57,7 +57,7 @@ class Job:
 
     def set_status(self, int_code, status_code="success"):
         if int_code:
-            self.int_code = int_code
+            self.int_code = int(int_code)
         self.status_code = status_code
 
     def write_output(self, s):


ross-spencer (approved):
Looks good @sevein. :+1:  (I can't find a good example of a microservice, but the simplicity of this makes me wonder if we should capture stdout on error too for a complete trace.)

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -219,6 +219,8 @@ def write_task_results_callback():
                         kwargs.update(
                             {"stdout": job.get_stdout(), "stderror": job.get_stderr()}
                         )
+                    elif kwargs["exitcode"] > 0:
```
replaceafill: Just curious: would this condition be affected by jobs that have multiple success exitcodes, for example some [normalization jobs](https://github.com/artefactual/archivematica/blob/862c4cdd780f1a235f20716d8472d34d9b56094f/src/MCPServer/lib/assets/workflow.json#L2463-L2476)?

sevein: That's a good point. I've been going through all cases: `1` is abused 17 times (from `test_v0.0`, `normalize_v1.0`, `hasPackages_v0.0` and `validateFile_v1.0`) and `2` is abused 13 times (from `normalize_v1.0` and `validateFile_v1.0`). But in all cases, they seem to represent errors to some extent (FPR rule failed or not found mostly), but the workflow doesn't want to mark them as failures.

I think that we have two options, let me know which you think it's preferably!
- We leave it as it is - these error-ish scenarios will be captured but may still turn out useful for some users.
- Use the `status` keyword in the `job.set_status` method somehow to ~force~ disable capturing when the exit code is greater than zero. `status` seems to be unused at the moment, I've just realized.

replaceafill: > abused

:joy: 

Option 2 for me. It'd be nice if we could determine in some cases if a job succeeded or failed based on the `status_code` string instead of the `int_code`. First thing that comes to my mind is [the red/danger exit code](https://github.com/artefactual/archivematica/blob/a67d9d30832a313e49a8242d869739690b35e86c/src/dashboard/src/templates/main/tasks.html#L43-L48) in the `/tasks/<UUID>` page that you get sometimes for successful tasks (like `Determine if transfer contains packages`).

sevein: Okay, that makes sense. I think that I'm going to leave this PR untouched for now until we have a milestone. If this is for 1.10 I think it can be just merged with the known shortcoming (which can be addressed separately) if you're okay with that. If the milestone is 1.11 then I think this PR becomes a venue to address that problem.

ross-spencer:
It seems important to add, that if exit codes are perceived as being used incorrectly, then we should consider what it looks like to fix, or standardize those. So a third option seems to be to tackle in a way that standardizes what exists and also sets guidelines for how to use them in future. Rather than say add another data point for developers to consider. If we do start to leverage `status` then that's another important one to consider holistically. We've developer docs in the works, so perhaps the inclusion of that in a fix would include some developer documentation around writing MCP client scripts.

PR #1476, opened 2019-08-29
MCPClient: Python 3 fixes for microservices

helenst:
A couple of python 3 fixes for https://github.com/archivematica/Issues/issues/808 (was originally just going to include one, hence the name of the branch)

1. Decode the output of rd.replace so that the transcription command runs correctly. 

2. Fix imports for the policy_check module as done in other modules.

3. Get dataverse transfers working


replaceafill (approved):
LGTM :+1:

I tested the policy check and transcribe jobs and they don't fail anymore.

PR #1477, opened 2019-08-30
Fix normalize jobs Python 3

replaceafill:
This fixes the thumbnail command migration and the command failures in the normalize client script.

Connected to https://github.com/archivematica/Issues/issues/808

### DIFF
--- a/src/dashboard/src/fpr/migrations/0031_python_3.py
+++ b/src/dashboard/src/fpr/migrations/0031_python_3.py
@@ -8,6 +8,7 @@ THUMBNAIL_TOOL_UUID = "efa8474a-8526-48c3-8279-e5a76bdc0995"
 OLD_THUMBNAIL_CMD_UUID = "7c2b65c7-6cea-4f81-9f3b-53375efc5bee"
 NEW_THUMBNAIL_CMD_UUID = "95149bc4-0620-4c20-964c-1d6c34b9400e"
 THUMBNAIL_RULES = (
+    "3a19f9a3-c5d5-4934-9286-13b3ad6c24d3",
 )
 NEW_THUMBNAIL_CMD = '''
 import argparse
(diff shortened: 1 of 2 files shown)


src/dashboard/src/fpr/migrations/0031_python_3.py
```diff
@@ -8,6 +8,7 @@
 OLD_THUMBNAIL_CMD_UUID = "7c2b65c7-6cea-4f81-9f3b-53375efc5bee"
 NEW_THUMBNAIL_CMD_UUID = "95149bc4-0620-4c20-964c-1d6c34b9400e"
 THUMBNAIL_RULES = (
+    "3a19f9a3-c5d5-4934-9286-13b3ad6c24d3",
```
helenst, line 11: What does this do?

helenst (commented):
Looks good to me, just one question.

src/dashboard/src/fpr/migrations/0031_python_3.py
replaceafill: It's a fallback rule to create a "default" thumbnail when all the other thumbnail rules can't (for example creating a thumbnail for a `.txt` file). Since we're introducing a new Python 3 compatible command to create the default thumbnail we need to update the existing rule (its current UUID and implementation was introduced in https://github.com/artefactual/archivematica-fpr-admin/pull/50).

mjaddis:
@helenst @replaceafill Apologies for jumping into this thread, but I got hit by a utf-8 issue very similar to the one addressed in this PR whereby the normalization client fails when there are non-latin characters in the original filename of the file to be normalized.  In my case, I'm using a python command rather than a bash script.  I noticed that this PR fixes a similar problem if you are using a bash script, but I'm not sure it fixes the problem for other types of script.   In my case, to_gnu_options() on the replacement_dict fails when the '%originalLocation%' is not ascii.  This is called by transcoder.py when constructing the command line to use when executing the command.  I hacked dicts.py so it ignores the '%originalLocation%'  key completely when creating the gnu style command line options, which avoids the utf-8 filename issue altogether.   Thought I'd at flag this up here rather than opening a new bug.

PR #1479, opened 2019-08-30
Ensure that we write event id for name cleanup

ross-spencer:
An event identifier wasn't being written for name cleanup. Now it is.

Tests have been added to look specifically at this, plus a few other
easier to check details that should also be output in an event.

Connected to archivematica/issues#890


src/MCPClient/lib/clientScripts/sanitize_object_names.py
```diff
@@ -146,6 +147,7 @@ def apply_file_updates(self):
                 event_outcome_detail=self.EVENT_OUTCOME_DETAIL.format(
                     old_location, sanitized_location
                 ),
+                event_id=uuid.uuid4(),
```
sevein: Could this be moved up so it reads as the first keyword since it's listed first in the model?

src/MCPClient/tests/test_sanitize.py
```diff
@@ -161,6 +161,10 @@ def test_sanitize_object_names(self):
         It should handle unicode unit names.
         It should not change a name that is already sanitized.
         """
+        event_detail = (
+            'prohibited characters removed: program="sanitize_names"; version="1.10.'
+        )
+
```
sevein, line 237: Could this be a single line like in the other example below?

src/MCPClient/tests/test_sanitize.py
```diff
@@ -203,10 +207,19 @@ def test_sanitize_object_names(self):
...
+            )
+
+            # Test the event identifier for name cleanup was written for our object.
+            assert event is not None, "Event not returned from database"
+            assert event[0].event_id is not None, "Event ID is None"
+            # Try and parse a type 4 UUID, if it is well-formed, the ID was
+            # written correctly, if it fails, then likely not.
+            uuid.UUID(event[0].event_id, version=4)
+            # Test the event detail was written for every object.
+            assert event_detail in event[0].event_detail, "Event detail written incorrectly"
```
sevein: I think that I prefer conciseness - it's always preferably to have tests that are easier to scan quickly.

An alternative that I find much more readable and in the spirit of how this test module was originally written:

```python
            event = Event.objects.get(
                file_uuid="47813453-6872-442b-9d65-6515be3c5aa1",
                event_type="name cleanup",
            )

            assert is_uuid(event[0].event_id)
            assert event_detail in event[0].event_detail, "Event detail written incorrectly"
```

Notice that `is_uuid` needs to be implemented which makes sense since you're doing this twice.

I don't think that what I wrote is cryptic, it's still pretty manageable for the average Python/Django dev.

src/MCPClient/tests/test_sanitize.py
```diff
@@ -300,6 +316,16 @@ def test_sanitize_transfer_with_multiple_files(
...
 
+        # Test the event identifier for name cleanup was written for our object.
+        event = Event.objects.filter(file_uuid=file_obj.uuid, event_type="name cleanup")
+        assert event is not None, "Event not returned from database"
+        assert event[0].event_id is not None, "Event ID is None"
+        # Try and parse a type 4 UUID, if it is well-formed, the ID was
+        # written correctly, if it fails, then likely not.
+        uuid.UUID(event[0].event_id, version=4)
+        # Test the event detail was written for every object.
+        assert event_detail in event[0].event_detail, "Event detail written incorrectly"
```
sevein: Same here!

sevein (changes requested):
Thanks for doing this! I've suggested some changes.

src/MCPClient/tests/test_sanitize.py
ross-spencer: This is `black`. I think the previous example was while black was failing.

src/MCPClient/tests/test_sanitize.py
ross-spencer: Done.

src/MCPClient/tests/test_sanitize.py
ross-spencer: As above.

src/MCPClient/tests/test_sanitize.py
ross-spencer: The original suggestion is good, but it is still ambiguous and won't work in its own right. To work it needs to become:
```python
            event = Event.objects.get(
                file_uuid="47813453-6872-442b-9d65-6515be3c5aa1",
                event_type="name cleanup",
            )

            assert is_uuid(event.event_id)
            assert event_detail in event.event_detail, "Event detail written incorrectly"
```
We no longer need to use indexing on the returned object which I believe is the goal of the suggestion.

The change has been made now too.

src/MCPClient/tests/test_sanitize.py
```diff
@@ -300,6 +327,14 @@ def test_sanitize_transfer_with_multiple_files(
         )
         assert u"bulk-file" in file_obj.currentlocation
 
+        # Test the event identifier for name cleanup was written for our object.
+        event = Event.objects.get(file_uuid=file_obj.uuid, event_type="name cleanup")
+        assert event is not None, "Event not returned from database"
```
sevein: This assertion is always going to succeed because `get` raises when the object does not exist.

src/MCPClient/tests/test_sanitize.py
```diff
@@ -300,6 +327,14 @@ def test_sanitize_transfer_with_multiple_files(
...
+        assert event.event_id is not None, "Event ID is None"
```
sevein: We don't need this line, `uuid.UUID(None, version=4)` raises a `TypeError`.

src/MCPClient/tests/test_sanitize.py
```diff
@@ -203,10 +216,19 @@ def test_sanitize_object_names(self):
...
             )
-            assert Event.objects.filter(
+            event = Event.objects.filter(
                 file_uuid="47813453-6872-442b-9d65-6515be3c5aa1",
                 event_type="name cleanup",
-            ).exists()
+            )
+
+            # Test the event identifier for name cleanup was written for our object.
+            assert event is not None, "Event not returned from database"
```
sevein: Above you're doing:

```
event = Event.objects.filter(
    file_uuid="47813453-6872-442b-9d65-6515be3c5aa1",
    event_type="name cleanup",
)
```

I think that it's preferably to use `get` in L219.
The assertion in L225 becomes unnecessary.

src/MCPClient/tests/test_sanitize.py
```diff
@@ -203,10 +216,19 @@ def test_sanitize_object_names(self):
...
+            assert event[0].event_id is not None, "Event ID is None"
```
sevein: We don't need this line, `uuid.UUID(None, version=4)` raises a TypeError.

sevein (changes requested):

src/MCPClient/tests/test_sanitize.py
```diff
@@ -142,6 +142,15 @@ def multiple_transfer_file_objs(db, transfer, tmp_path, multiple_file_paths):
     return File.objects.bulk_create(file_objs)
 
 
+def is_uuid(uuid_):
+    """Test for a well-formed UUID string."""
+    try:
+        uuid.UUID(uuid_, version=4)
+        return True
+    except ValueError:
+        return False
```
sevein: Please align the happy path to the left edge.

```
try:
    uuid.UUID(uuid_, version=4)
except ValueError:
    return False
return True
```

sevein (approved):
Hey sorry @ross-spencer I noted a few nitpicks yesterday but I'm ok if you choose to merge.

sevein (approved):

PR #1480, opened 2019-09-16
Add job for importing PREMIS events XML

replaceafill:
This PR adds a job to the **Characterize and extract metadata** microservice to import events from PREMIS 3 compliant XML files.

The new clientScript may seem a bit... verbose (I split most things in small functions), but the idea of this implementation is:

1. Validate and parse the XML file

2. For each `premis:object` (representing a file), `premis:event` and `premis:agent` element in the XML, [build a dictionary with attributes](https://github.com/artefactual/archivematica/blob/cc691001693ad97e0e189ea04891b083eb72d4bc/src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py#L369-L435) needed to build the events and agents in the database. An identifier is built for each element based on `(PREMIS identifier type, PREMIS identifier value)` which will be used to reference the element from other elements.

3. Go through the files collection looking for elements that don't have a filename or which filename is not included in the transfer. These are considered "invalid".

4. In the PREMIS specification `premis:object` elements and `premis:agent` elements may have multiple `premis:linkingEventIdentifier` elements referencing `premis:events`.

   `premis:event` elements in turn may have multiple `premis:linkingObjectIdentifier` elements or `premis:linkingAgentIdentifier` elements.

   These relationships are tracked in this implementation using sets of element identifiers (tuples). The dictionaries that represent files and agents have an `events` set and the ones that represent events have two sets: `files` and `agents`.

   So [the next step is to build these relationships between the three types of elements](https://github.com/artefactual/archivematica/blob/cc691001693ad97e0e189ea04891b083eb72d4bc/src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py#L723-L727).

5. Then [some logging of unusable elements is done](https://github.com/artefactual/archivematica/blob/cc691001693ad97e0e189ea04891b083eb72d4bc/src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py#L729-L738).

6. Finally events with valid file and agent information are saved to the database with the unusable elements mentioned previously being ignored. 

Connected to https://github.com/archivematica/Issues/issues/710


src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
```diff
@@ -0,0 +1,752 @@
...
+            formatted_object_identifier=format_identifier(
+                file_identifier, element_type="object"
+            ),
+            original_name=original_name,
+        ),
+        file=sys.stderr,
+    )
+
+
+def log_error_missing_xsd(path, printfn=print):
```
cole: This is a bit of a style thing but some of these logging functions are single statement and only called once; it seems like we could simply remove them and do the logging inline?

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
```diff
@@ -0,0 +1,752 @@
...
+    """
+    result = {}
+    root = tree.getroot()
+    premis_version = root.get("version", metsrw.plugins.premisrw.PREMIS_VERSION)
+    elements = tree.findall(
+        selector,
+        metsrw.plugins.premisrw.PREMIS_VERSIONS_MAP[premis_version]["namespaces"],
+    )
+    # metsrw.plugins.premisrw.premis_to_data (used later) expects the PREMISElements
+    # to have their own version attribute, so we set it if they don't
```
cole, line 305: We should probably update metsrw to allow this case.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
```diff
@@ -0,0 +1,752 @@
...
+        pass
+
+    # iterate files and relate them to their events
+    relate_files_to_events(files, events)
+
+    # iterate agents and relate them to their events
+    relate_agents_to_events(agents, events)
+
+    # log elements that are not related to any other element
+    print_unrelated_files(files, events, job.pyprint)
```
cole: What's the reasoning behind using a print function instead `logger` and then running main inside `with job.JobContext(logger=logger)` here? This is a very common pattern in existing scripts, but I think logger objects have a bit more flexibility and that's probably the way we should be going.

I think it's probably not worth changing this PR (printfn is passed in many places and used in tests), I'd just like to understand if there are barriers to using loggers.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
```diff
@@ -0,0 +1,752 @@
...
+    file_identifiers = set(files)
+    agent_identifiers = set(agents)
+
+    for event in six.itervalues(events):
+
+        # get the agents for this event
+        event_agents = get_event_agents(event, agents, agent_identifiers, printfn)
+        if not event_agents:
+            log_error_event_without_agents(event["identifier"], printfn)
+            # events shouldn't be crated without any agents
```
cole: *created

src/MCPServer/lib/assets/workflow.json
```diff
@@ -6355,6 +6355,38 @@
...
+                "en": "Load PREMIS events from metadata/premis.xml"
+            },
+            "exit_codes": {
+                "0": {
+                    "job_status": "Completed successfully",
+                    "link_id": "a536828c-be65-4088-80bd-eb511a0a063d"
+                }
+            },
+            "fallback_job_status": "Failed",
+            "fallback_link_id": "a536828c-be65-4088-80bd-eb511a0a063d",
```
cole, line 6380: It sounds from discussion like the failure case should fail the transfer — if so we should probably link to `61c316a6-0a50-4f65-8767-1f44b1eeb6dd` here.

cole (approved):
This looks awesome, I'm really happy with the level of testing and it looks Python 3 compatible as well.

The only thing that is required before this is merged I think is there may be some issues with how this integrates into the workflow. I'd say that needs to be fleshed out in the issue descriptions first.

Otherwise, I have a few questions/style comments inline, nothing important. I'd be fine with this being merged as is and updated later if necessary.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
replaceafill: I get your point about some of them being called just once, but I had logging inline at the beginning and it made the other functions so long. I think it was a combination of the long variable and keyword names I picked (i.e. `formatted_object_identifier`) and how black breaks them. For example `log_error_missing_original_name` becomes 8 lines long and blocks like this seems cleaner this way:

```
   for file_identifier, file_ in six.iteritems(files):
        original_name = file_["original_name"]
        if not original_name:
            log_error_missing_original_name(file_identifier, printfn)
            result.add(file_identifier)
            continue
```

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
replaceafill: I agree. We could also consider adding the `PREMISFile` (or `PREMISObjectFile`) with the `originalName` child element.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
replaceafill: To be honest I'm not sure what to use under what conditions. I found https://github.com/archivematica/Issues/issues/26 and the conversation linked there made look
 `pyprint` like the flexible choice.

src/MCPServer/lib/assets/workflow.json
replaceafill: I just asked Evelyn and Peter about this and they think it's OK to fail only the job and keep the transfer workflow going. I'm also going to change the script to error under some other conditions like duplicated event UUIDs being used.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
cole: Makes sense. Looks like it's been an issue for a while 😁

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
cole: Yeah, I guess so. I'm not sure that's so bad if it's a few more lines though? The benefit of inline is you can see right away what is happening. The functions are pretty clearly named though. I'm fine with it as is.

src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
cole: Can you add that to https://github.com/artefactual-labs/mets-reader-writer/issues/75? I don't fully understand how PREMISObjectFile would work.

replaceafill:
@cole Do you mind giving these last commits a quick look? They basically allow the script to error until the end after any irregularity.

cole (approved):
The only thing I'm not sure of is avoiding failure when a duplicate event id is submitted; surely that would be invalid PREMIS XML? The code is fine, but the behaviour seems off to me. It's not a big deal, just wanted to check that it's Archivist Approved™.

Otherwise all looks good.

replaceafill:
Hmm, good point. I added the check because after the demo today I imported the same file twice and saw the integrity error because of the duplicated UUID. I think it's a bit less of a concern now that the script doesn't import things incompletely.

I'm going to merge this as it is and ask the analysts about it. I can address the loose end when I create the AMAUAT for this.

PR #1482, opened 2019-09-18
Don't move empty transfer dir to watched dir after creating SIP

cole:
Connects to https://github.com/archivematica/Issues/issues/782, https://github.com/archivematica/Issues/issues/770.

The "Create SIP from transfer objects" microservice is already doing the move, but leaves a directory containing only metadata. Because of the extra move to the watched dir, the
"Check transfer directory for objects" chain is triggered for completed transfers on restart.

I made this change as part of package queing work because the empty transfers were clogging up the queue. It seems like an improvement on the current state of affairs to me, but **It still leaves useless mostly empty directories in the currentlyProcessing folder**. Simply trying to remove the remains at the end of the "Create SIP from transfer objects" microservice fails, and I'm not sure yet why that's the case.

### DIFF
--- a/src/MCPClient/lib/clientScripts/create_sip_from_transfer_objects.py
+++ b/src/MCPClient/lib/clientScripts/create_sip_from_transfer_objects.py
@@ -49,8 +49,9 @@ def call(jobs):
                 sharedPath = job.args[6]
                 sipName = transferName
 
-                tmpSIPDir = os.path.join(processingDirectory, sipName) + "/"
-                destSIPDir = os.path.join(autoProcessSIPDirectory, sipName) + "/"
+                # tmpSIPDir has a trailing slash, destSIPDir does not
+                tmpSIPDir = os.path.join(processingDirectory, sipName, "")
+                destSIPDir = os.path.join(autoProcessSIPDirectory, sipName)
                 archivematicaFunctions.create_structured_directory(
                     tmpSIPDir, manual_normalization=False
                 )
@@ -86,7 +87,7 @@ def call(jobs):
                 diruuids = len(dir_mdls) > 0
 
                 # Create row in SIPs table if one doesn't already exist
-                lookup_path = destSIPDir.replace(sharedPath, "%sharedPath%")
+                lookup_path = destSIPDir.replace(sharedPath, "%sharedPath%") + os.sep
                 try:
                     sip = SIP.objects.get(currentpath=lookup_path)
                     if diruuids:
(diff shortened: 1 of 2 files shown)


ross-spencer:
Hi @cole, on this: 

>  Simply trying to remove the remains at the end of the "Create SIP from transfer objects" microservice fails, and I'm not sure yet why that's the case.

Just out of interest how does that failure manifest itself? e.g. something like permissions at the OS level? Or something you're see in the workflow/dashboard?

cole:
@ross-spencer I should have specified, it seems like there's a path mismatch (the path in the microservice is without the UUID appended, while the empty dir has the UUID). Makes me think I'm missing something or looking in the wrong place.

ross-spencer:
Oh yeah, seeing the same @cole. I haven't looked at it much, but can see today that microservice/script needs some work. There's a lot going on, but I don't think there's a lot we're trying to achieve? 

I think you'd be fine to add `\"%SIPDirectory%\""` to the args, and `shutil.rmtree(..)` that at the end of the script as-is. 
```json
"39a128e3-c35d-40b7-9363-87f75091e1ff":{ 
   "config":{ 
      "@manager":"linkTaskManagerDirectories",
      "@model":"StandardTaskConfig",
      "arguments":"\"%SIPObjectsDirectory%\" \"%SIPName%\" \"%SIPUUID%\" \"%processingDirectory%\" \"%sharedPath%watchedDirectories/system/autoProcessSIP/\" \"%sharedPath%\" \"%SIPDirectory%\"",
      "execute":"createSIPfromTransferObjects_v0.0",
      "filter_file_end":null,
      "filter_file_start":null,
      "filter_subdir":null,
      "stderr_file":null,
      "stdout_file"
}, 
```
But would love to just sit down with it in more detail and tighten it up a bit.

ross-spencer:
> I think you'd be fine to add \"%SIPDirectory%\"" to the args, and shutil.rmtree(..) that at the end of the script as-is.

NB. Turns out there are plenty of unintended consequences to this, including, it seems, the transfer mets being lost and reingest failing further down the line... *sigh* ...one to rethink.

cole:
@ross-spencer oh, that's no good. Does this PR as is have those consequences, or did you test with the addition of `shutil.rmtree`?

ross-spencer:
Good question @cole and apologies for muddying the waters. I quite forgot I was running this version of `workflow.json` as well. I tried a transfer and repeated the same reingest method, and all looks good.

cole:
No problem, thanks for testing! Huh, that's too bad about rmtree though. I'm concerned that as this PR is the processing directory contents will continue to grow forever.

cole:
OK, if I change the move slightly (_shakes fist at trailing slashes_) we can move the whole transfer directory without having to rmtree. This change works for me with reingest. @ross-spencer if you have time can you give it another shot?

ross-spencer (approved):
I did have time, works like a charm! 

I'd consider, `os.path.join("a", "path", "")` for adding trailing slashes, and `os.sep` for stripping them. 

Otherwise, looks great, and it's nice to to leave that data behind any more!

PR #1484, opened 2019-09-20
Add inotify support to watched dir tracking

cole:
Connects to: https://github.com/archivematica/Issues/issues/916.

Switch to a single thread for watching directories, and add an option (WATCHED_DIR_METHOD) to use inotify. Defaults to poll for backwards compatibility, but users who are concerned with performance will likely want to use a local processing directory and inotify.

Performance testing on my laptop isn't very accurate, but after running a few tests with a trivial transfer (two photos), it is  ~~10%~~ up to 11s faster with inotify.


src/MCPServer/install/README.md
```diff
@@ -153,11 +155,18 @@ This is the full list of variables supported by MCPServer:
     - **Type:** `int`
     - **Default:** `""`
 
-- **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_WATCHDIRECTORIESPOLLINTERVAL`**:
-    - **Description:** time in seconds between filesystem poll intervals.
-    - **Config file example:** `MCPServer.watchDirectoriesPollInterval`
```
cole, line 158: I've changed this for consistency, but it may be worth reverting. As far as I can tell it's not set by any ansible scripts, but I may have missed some.

src/MCPServer/install/README.md
```diff
@@ -2,11 +2,13 @@
 
 ## Table of contents
 
-- [Introduction](#introduction)
-- [Environment variables](#environment-variables)
-- [Configuration file](#configuration-file)
-- [Parameter list](#parameter-list)
-- [Logging configuration](#logging-configuration)
+- [MCPServer Configuration](#mcpserver-configuration)
+  - [Table of contents](#table-of-contents)
```
sevein: There's something going on here with your TOC generator.

src/MCPServer/install/README.md
```diff
@@ -153,11 +155,18 @@ This is the full list of variables supported by MCPServer:
...
+    when using a local filesystem on Linux.
+    - **Config file example:** `MCPServer.watch_directory_method`
+    - **Type:** `string`
+    - **Default:** `"poll"`
+
+- **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_WATCH_DIRECTORY_INTERVAL`**:
+     - **Description:** time in seconds between filesystem poll intervals.
+     - **Config file example:** `MCPServer.watch_directory_interval`
+     - **Type:** `int`
+     - **Default:** `"1"`
```
sevein, line 167: We have a mechanism in place to handle backward-incompatible changes that you may want to use: https://github.com/artefactual/archivematica/blob/862c4cdd780f1a235f20716d8472d34d9b56094f/src/MCPClient/lib/settings/common.py#L81.

sevein (changes requested):

src/MCPServer/install/README.md
cole: Oops, weird. Fixed now.

src/MCPServer/install/README.md
cole: Added a fallback to the old value in 1eccc979841297f2e635b652a10f144aa3fe384f.

sevein (approved):

PR #1485, opened 2019-09-23
Add package queueing to MCPServer

cole:
Connects to https://github.com/archivematica/Issues/issues/911.

About these changes
===============

These changes introduce a number of new abstractions:

  * `Package` (previously `unit`) subclasses `SIP`, `DIP`, and `Transfer` handle package related logic
  * `Job` (previously `linkTaskManager`) and subclasses handle execution of a single link of the workflow
  * `JobChain` (previously `jobChainLink`) handles passing context between jobs, and determining the next `Job` to be executed based on the workflow chain
  * `Task` (previously `Task`, just rewritten a bit) corresponds to a task executed by MCPClient
  * `GearmanTaskBackend` (previously `taskGroupRunner`) handles passing tasks to MCPClient via gearman (in batches)
  * `ThreadPoolExecutor` (from the `concurrent.futures` library, previously `Executor`) handles out of process execution
  * `PackageQueue` handles scheduling of `Job` objects for execution (throttled by package). The package queue is thread-safe; it can be used to schedule work from any thread safely. It _also_ handles the list of jobs awaiting decisions (previously handled by a shared dictionary, `choicesAvailableForUnits`).

Basically unchanged are the  `RPCServer` and workflow related abstractions.

All of this could still use more cleanup! There are a lot of things with mixed responsibilities left. This PR should be thought of as a starting point for future work.

Backwards incompatibilities
====================

This PR breaks delayed job execution, required by quarantine. Quarantine already doesn't function properly, though (see https://github.com/artefactual/archivematica/issues/598).

What do I need to change about my development process after this is merged?
=========================================================

Hopefully nothing. This change is meant to not change any of the interfaces to MCPServer. It doesn't touch the format data is sent in via gearman, or the workflow or workflow schema.

These changes also shouldn't change behaviour for users of Archivematica; there are some new settings, but the defaults should be fine.

Why is this PR so big?
================

😅  Package queuing itself isn't quite this big, but it required some concepts around packages that were kind of lacking in MCPServer. The linkTaskManager abstractions also needed some refactoring. Those changes are responsible for a lot of this PR.

How MCPServer works now
====================

Essentially MCPServer starts a number of threads and waits for activity to be triggered via the watched directories or an API call. Probably the best way to go through the changes in depth is to walk through app startup and job processing.

MCPServer startup
-------------------

Startup is useful for understanding what is running in MCPServer.

`main` goes through the following steps:
1. A signal listener is setup to handle shutdown on SIGINT/SIGTERM events.
2. The default workflow is loaded (from workflow.json).
3. The configured SHARED_DIRECTORY is populated with the expected directory structure, and default processing configs added.
4. Any in progress Job and Task entries in the database are marked as errors, as they are presumed to have been the result of a shutdown while processing.
5. If Prometheus metrics are enabled, an thread is started to serve metrics for scraping.
6. A `ThreadPoolExecutor` is initialized with a configurable number of threads (default ncpus).
7. A `PackageQueue` (see the `queues` module) is initialized.
8. A configured number (default 4) of RPCServer (see the `rpc_server` module) threads are started to handle gearman "RPC" requests from the dashboard.
9. A watched directory thread is started to observe changes in any of the watched dirs as set in the workflow.
10. The `PackageQueue.work` processing loop is started on the main thread.

The PackageQueue and Job processing
---------------------------------------

This process happens when a `Job` is scheduled via `PackageQueue.schedule_job`.

1. If there are too many active packages, the `Job` is placed on a deferred package queue and held until a package stops processing  (packages stop processing when they reach a decision point in the workflow that hasn't been pre configured, or when they hit specific workflow links denoting SIP/Transfer completion or failure). If there is room to process the package, the job is placed on an active job queue, which is the consumed by the processing loop running on the main thread.
2. The processing loop retrieves any jobs from the active queue, and schedules their `run` method for execution in a worker thread (via `ThreadPoolExecutor.submit`).
3. The `Job.run` method executes. If it is a `ClientScriptJob` (executing on MCPClient), it generates the `Task` objects required, and sends them to `MCPClient` via `GearmanTaskBackend`, and waits for the results. All of this happens on the one worker thread.
4. On the completion of tasks (i.e. results are returned by Gearman), `Job.run` returns the _next_ job to schedule, if any. In practice this is usually retrieved from the `JobChain` via `next(self.job_chain)`.
5. Back in the main thread, a callback attached to the result of `Job.run` triggers adding the next job to the active job queue. This cycle continues until the workflow chain ends.

Create Package API call
------------------------

The create package API call is triggered by the dashboard (either the UI, or directly via the API).

1. A `packageCreate` gearman job is passed from the dashboard to MCPServer, via gearman.
2. One of the running RPCServer threads picks up the job, and runs the `create_package` function. 
3. `create_package` returns a transfer UUID to the dashboard via gearman, and schedules package start to occur on a worker thread (i.e. via the `ThreadPoolExecutor`). Package start consists of copying the package to a tempdir, and then either placing it into a watched dir (see Watched Directory activity below), or creating and scheduling a `JobChain` (with a starting point defined by the transfer type).
4. The first `Job` of the `JobChain` is scheduled via `PackageQueue.schedule_job`.

Watched Directory activity
--------------------------

1. The thread running `watched_dirs.watch_directories` (either via inotify or poll) observes a new folder or file, and triggers it's callback function, `mcp.watched_dir_handler`.
2. `mcp.watched_dir_handler` determines the package type from the WatchedDir object passed in, and gets the package UUID from the path.
3. Also based on the WatchedDir object it determines the workflow chain to start, initializes a `JobChain`, and schedules the first `Job`.

TL;DR
====

This PR limits the number of concurrent packages Archivematica will process at one time, so that packages can be continually submitted via the API without Archivematica grinding to a halt.

In order to do that effectively, some refactoring was necessary due to the state of MCPServer.


cole:
~Note for testing that until https://github.com/artefactual/archivematica/pull/1482 is addressed (issue https://github.com/archivematica/Issues/issues/782), you'll have to either increase the `CONCURRENT_PACKAGES` option or clear your shared directory after every few transfers.~ Update: merged now.

sevein (commented):
I've been having success in my first tests. The number of db conns is pretty low.

This PR does not install the inotify_simple package but I guess you're planning to take care of that once #1484 is merged. I've installed it locally for the sake of testing.

cole:
@sevein oops, right, sorry about that. I've merged #1484 and rebased, so inotify is now included.

src/MCPServer/lib/server/rpc_server.py
```diff
@@ -460,23 +475,26 @@ def _pull_choices(job_id, lang, jobs_awaiting_for_approval):
...
+    for item in choices.items():
+        id_, label = item
         ret[id_] = label[lang]
     return ret
 
 
-def start(workflow):
-    worker = RPCServer(workflow)
+def start(workflow, shutdown_event, package_queue):
+    worker = RPCServer(workflow, shutdown_event, package_queue)
```
sevein: Hey @cole, I've seen a `MySQL server has gone away` today after 8+ hours running your branch idle (see https://gist.github.com/sevein/266f8a2e71af2a1f8a527c8900832f52). Using the application defaults. MySQL's wait_timeout equals 28800. I can count up to two places in the same thread where we close old connections in that code path: the RPC handler wrapper plus create_package - not sure what's going on!

src/MCPServer/lib/server/shared_dirs.py
```diff
@@ -0,0 +1,329 @@
...
+    <!-- Delete packages after extraction -->
+    <preconfiguredChoice>
+      <appliesTo>f19926dd-8fb5-4c79-8ade-c83f61f55b40</appliesTo>
+      <goToChain>85b1e45d-8f98-4cae-8336-72f40e12cbef</goToChain>
+    </preconfiguredChoice>
+    <!-- Transcribe files (OCR) -->
+    <preconfiguredChoice>
+      <appliesTo>7079be6d-3a25-41e6-a481-cee5f352fe6e</appliesTo>
+      <goToChain>1170e555-cd4e-4b2f-a3d6-bfb09e8fcc53</goToChain>
+    </preconfiguredChoice>
```
sevein, line 145: This should be:

```xml
    <!-- Transcribe files (OCR) -->
    <preconfiguredChoice>
      <appliesTo>82ee9ad2-2c74-4c7c-853e-e4eaf68fc8b6</appliesTo>
      <goToChain>0a24787c-00e3-4710-b324-90e792bfb484</goToChain>
    </preconfiguredChoice>
```

src/MCPServer/lib/server/shared_dirs.py
```diff
@@ -0,0 +1,329 @@
...
+    <!-- Perform file format identification (Submission documentation & metadata) -->
+    <preconfiguredChoice>
+      <appliesTo>087d27be-c719-47d8-9bbb-9a7d8b609c44</appliesTo>
+      <goToChain>4dec164b-79b0-4459-8505-8095af9655b5</goToChain>
+    </preconfiguredChoice>
+    <!-- Bind PIDs -->
+    <preconfiguredChoice>
+      <appliesTo>05357876-a095-4c11-86b5-a7fff01af668</appliesTo>
+      <goToChain>fcfea449-158c-452c-a8ad-4ae009c4eaba</goToChain>
+    </preconfiguredChoice>
```
sevein, line 39: This should be:

```xml
    <!-- Bind PIDs -->
    <preconfiguredChoice>
      <appliesTo>a2ba5278-459a-4638-92d9-38eb1588717d</appliesTo>
      <goToChain>44a7c397-8187-4fd2-b8f7-c61737c4df49</goToChain>
    </preconfiguredChoice>
```

Same in L126.

src/MCPServer/lib/server/shared_dirs.py
cole: Doh, I guess the auto merge when rebasing missed a few things. Changed now.

src/MCPServer/lib/server/shared_dirs.py
cole: Thanks, fixed.

src/MCPServer/lib/server/rpc_server.py
cole: Looking at the traceback I think I know what is happening here — the key is that db connections in the Django ORM are per thread, and the RPC handler wrapper and create pacakge will be executed in one of the RPCServer threads, but then create package runs `_start_package_transfer_with_auto_approval` using the executor, meaning it runs on a different thread. It looks like in that chain, there is no close old connections. I'll add some more.

cole: OK, hopefully fixed now!

src/MCPServer/lib/server/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+MCPServer (Master Control Program Server) determines the jobs/tasks/microservices
+run by Archivematica and arranges for their execution.
+
+It makes use of the following major abstractions:
+    * `workflow.Workflow` (and related classes `Link` and `Chain`) handle the workflow
+      logic for Archivematica, described in workflow.json.
+    * `jobs.base.Job` and subclasses handle execution of a single link of the workflow.
+        * `jobs.client.ClientScriptJob` for jobs to be executed via MCPClient script
+        * `jobs.client.DecisionJob` for workflow decision points
+        * `jobs.client.LocalJob` for jobs that are exceuted on MCPServer directly
```
sevein: Typo.

sevein (approved):
Thanks @cole. I'd be comfortable merging. This brings a lot of clarity to MCPServer and the sensible use of resources is very noticeable!

src/MCPServer/lib/server/mcp.py
```diff
@@ -0,0 +1,147 @@
...
+import threading
+
+import django
+
+django.setup()
+
+from django.conf import settings
+
+from server import metrics, rpc_server, shared_dirs
+from server.executor import executor
```
sevein: This needs to go!

cole: 😱 how did that pass on TravisCI?

sevein: I bet that wouldn't happen if we stopped doing PYTHONPATH magic!

sevein:
Testing today, working great.

sevein:
`make test-at-black-box` in this branch taking 7m vs 11m in stable/1.10.x!
A big chunk of the time was also spent in AIP indexing.

PR #1496, opened 2019-10-16
Fix DIP database access

cole:
Connects to: https://github.com/archivematica/Issues/issues/911

* DIPs use the SIP model (there is no DIP model, whoops)
* Package subclasses need a reload method
* Make uuid_from_path not care about trailing slashes
  (eases testing)
* Add tests to make sure we don't have a later regression


replaceafill (approved):
I tested this with a standard workflow and AIPs and DIPs are stored now. I also tested DIP upload to atom `stable/2.5.x` and that also works. I think this just needs that `self.self` typo and the tests fixed and it's good to go! :+1:

PR #1498, opened 2019-10-18
Fix move SIP service error

cole:
Relates to: https://github.com/archivematica/Issues/issues/951

The return value of `rename` is used in many places for job status, so it shouldn't be returning
None. Also handle the case where the source and destination are the same, as that's a non-fatal error.

Edit: Connected to https://github.com/archivematica/Issues/issues/770

### DIFF
--- a/src/MCPClient/lib/job.py
+++ b/src/MCPClient/lib/job.py
@@ -28,21 +28,21 @@ class Job:
 
     def dump(self):
         return (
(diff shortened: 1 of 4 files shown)


cole:
Also, non-ascii chars shouldn't cause failures, still working on that.

replaceafill (approved):
Works like a charm! :+1: I even tested the commits separately just to make sure the decoding problem is gone.

PR #1499, opened 2019-10-21
WIP: Update MCPClient forking model

cole:
Relates to: https://github.com/archivematica/Issues/issues/943

Initial attempt at making MCPClient process and resource usage more consistent.

### Background ###

We have two somewhat conflicting targets for MCPClient performance: a) processing time for a single large transfer and b) total processing time for many simultaneous transfers. This PR attempts to improve b), without having a signifigant impact on a).

This approach makes no changes to the interface between MCPClient and Server (gearman, or the format of data used by gearman. It makes minor changes to the `Job` API but it should not require any major rework.

### Previous forking model ###

Detailed in https://github.com/archivematica/Issues/issues/943, the previous model would start a process pool only when processing a client script that required it. A single gearman worker relayed results to MCPServer.

### Updated forking model ###

This PR creates a process pool of on initial startup. Each process is essentially an independent MCPClient; it receives work via gearman, and communicates the results that way as well. However in this case, unless a client script is marked as safe for concurrent instances, only one process in the pool will handle tasks of that type.

Processes are restarted on exit, and will be restarted automatically after processing a certain number of jobs, to try to mitigate any memory leaks.

Note that tasks are still broken up into batches, and this PR will probably work better with a lower BATCH_SIZE setting than the current default of 128.

### TODOs ###

- [ ] More testing
- [ ] Unit test coverage
- [ ] Document how things work
- [ ] Add config option for number of client processes
- [ ] Add config option for reducing the number of processes used by bagit (on top of everything else)


cole:
(I've also looked a bit at having one gearman listener for the MCP Client, and passing work to a process pool, but that appoach resulted in a lot of work handling message passing at a layer below gearman, which seemed kind of pointless.)

cole:
Unfortunately this PR still needs a lot of work, and I haven't been able to progress on it. I'll leave it open for visibility, in case someone else wants to pick things up.

PR #1500, opened 2019-10-24
Remove mutable default argument

daaang:
When reindexing, `_list_files_in_dir()` is run once for each item. If the `filepaths` argument is left set to `[]`, then each item will erroneously include all previously indexed files.

By setting it to `None` and initing the `[]` default inside the function body, we avoid this undesirable behavior because the list is inited anew each time the function runs.

To demonstrate what was happening before this change, you can see what I mean by running the following in the python console:

```python
def f(x=[]):
    x.append(5)
    print x

f()
f()
f()
f(["hi"])
f()
```

[more information](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

Thanks!

### DIFF
--- a/src/MCPClient/lib/clientScripts/index_aip.py
+++ b/src/MCPClient/lib/clientScripts/index_aip.py
@@ -19,6 +19,7 @@ django.setup()
 
 from django.conf import settings as mcpclient_settings
 
+
 logger = get_script_logger("archivematica.mcp.client.indexAIP")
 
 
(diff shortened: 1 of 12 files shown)


ross-spencer:
Hi @daaang thank you for submitting this! Is it possible for you to follow through the contributing steps outlined in [CONTRIBUTING.md](https://github.com/artefactual/archivematica/blob/stable/1.10.x/CONTRIBUTING.md#submitting-code-changes) to help us identify the extent of the problem and the most appropriate place to make the change? At this point in time, it is likely this will need to be fixed in the `qa/1.x` branch and then there is a chance it will appear in `1.11` or `1.12` releases of Archivematica if the problem is still present. Looking ahead to step 8 on that list, will we have a contributor's agreement for you do you know? Cheers!

daaang:
I've added issue [archivematica/Issues#962](https://github.com/archivematica/Issues/issues/962) to describe the problem this PR fixes. As for the CLA, I'll check to see if we have one already on file.

daaang:
I've signed and emailed you a CLA.

ross-spencer (commented):
@daaang thank you for catching this and fixing it. Are you able to rebase this against `qa/1.x` and we'll seek to include it in `1.11` asap?

ross-spencer (approved):
Thanks again @daaang LGTM! :+1:

ross-spencer:
I couldn't figure out how to rebase this via this PR so created https://github.com/artefactual/archivematica/pull/1521 through cherry-picking and we just need to approve and merge that.

PR #1501, opened 2019-10-25
Instrumentation improvements

cole:
Connected to: https://github.com/archivematica/Issues/issues/964, 
Connected to: https://github.com/archivematica/Issues/issues/906

* update metric descriptions
* add new metrics for file format and modified timestamps
* zero mcpserver counters on startup
* switch timing metrics to histograms with defined buckets

This needs a few dashboard changes before it can be reviewed.


src/archivematicaCommon/lib/common_metrics.py
```diff
@@ -0,0 +1,96 @@
...
+    1000000000.0,  # 1 GB
+    5000000000.0,  # 5 GB
+    10000000000.0,  # 10 GB
+    float("inf"),
+)
+
+
+db_retry_time_counter = Counter(
+    "common_db_retry_time_seconds",
+    "Total time waiting to retry database transactions in seconds",
```
replaceafill: I've noticed you've been adding "labeled by" to some labels. Is it worth adding that here (and in the counter below) too?

replaceafill (approved):
LGTM! :+1:

Issue #891, opened 2018-01-22
MCP config drop-down orders need review


sallain:
Several of the drop-down options appear in illogical or inconsistent order. For example:

```
Select compression level:

#. None
#. 5 - normal compression mode
#. 7 - maximum compression
#. 9 - ultra compression
#. 3 - fast compression mode
#. 1 - fastest mode
```

Also, there are 3 different file identification drop-downs (transfer, ingest, submission docs/metadata) and the order of choices in each of those is different.

Reviewing the order of options (as well as, perhaps, the wording, if possible) and updating them for consistency would be helpful for understanding the processing configuration options.

ross-spencer:
Hi @sallain we managed to fix the order of the compression methods (7z, gzip) etc. in https://github.com/wellcometrust/archivematica/commit/7fdcb35f26b2cc39a244cecb0390a19e7bf30369. If you are interested, I believe they should largely be `workflow.json` changes. So if you look at this section of code [here](https://github.com/wellcometrust/archivematica/blob/7fdcb35f26b2cc39a244cecb0390a19e7bf30369/src/MCPServer/lib/assets/workflow.json#L896-L972) the order is defined by the order that the different JSON objects appear in the square brackets. 

Taking a smaller segment:

```json
        "01c651cb-c174-4ba4-b985-1d87a44d6754": {
            "config": {
                "@manager": "linkTaskManagerReplacementDicFromChoice",
                "@model": "MicroServiceChoiceReplacementDic",
                "replacements": [
                    {
                        "description": {
                            "en": "5 - normal compression mode",
                            "es": "5 - modo de compresión normal",
                            "fr": "5 - mode de compression normal",
                            "ja": "5 - 通常の圧縮モード",
                            "pt_BR": "5 - modo de compressão normal",
                            "sv": "5 - normal komprimeringsläge"
                        },
                        "id": "414da421-b83f-4648-895f-a34840e3c3f5",
                        "items": {
                            "AIPCompressionLevel": "5"
                        }
                    },
                    {
                        "description": {
                            "en": "7 - maximum compression",
                            "es": "7 - máxima compresión",
                            "fr": "7 - compression maximum",
                            "ja": "7 - 最大圧縮",
                            "pt_BR": "7 - compressão máxima",
                            "sv": "7 - maximal komprimering"
                        },
                        "id": "4e31f579-68bd-4be1-a10e-ec5411897121",
                        "items": {
                            "AIPCompressionLevel": "7"
                        }
                    }
[...]

PR #1502, opened 2019-10-29
Reorder processing configuration choices

sallain:
I've gone through each of the processing configuration choices and reordered where necessary. Many of the changes are just for consistency - I've changed the order to `None`, `Yes`, `No` where those were the only options. For some of the jobs with more complex options I've tried to apply a more logical order:

```
Normalize:
* Normalize for preservation and access
* Normalize for preservation
* Normalize for access
* Normalize service files for access
* Normalize manually
* Do not normalize
```

```
Select compression level:
* 1 - fastest compression
* 3 - fast compression
* 5 - normal compression
* 7 - maximum compression
* 9 - ultra compression
```

```
Upload DIP:
* Upload DIP to AtoM/Binder
* Upload DIP to ArchivesSpace
* Upload DIP to CONTENTdm
* Do not upload DIP
```

One other change I've made here is in the `Generate thumbnails` option, where I've clarified the options:

```
Generate thumbnails:
* Yes
* Yes, without default icons
* No
```

The reason that this is still a WIP is because while the changes appear on the processing configuration page (Administration > Processing configuration > Default), they're not all reflected on the Transfer/Ingest tab drop-down menus. I'm not sure if this is a not-knowing-enough-about-docker problem or if there's something else in the code that I need to change.

Connected to #891
Requires https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/154

### DIFF
--- a/src/MCPServer/lib/assets/workflow.json
+++ b/src/MCPServer/lib/assets/workflow.json
@@ -826,9 +826,37 @@
                 "@manager": "linkTaskManagerReplacementDicFromChoice",
                 "@model": "MicroServiceChoiceReplacementDic",
                 "replacements": [
+                  {
+                      "description": {
+                          "en": "1 - fastest compression",
+                          "es": "1 - el modo más rápido",
+                          "fr": "1 - mode rapide",
+                          "ja": "1 - 最高速モード",
+                          "pt_BR": "1 - modo mais rápido",
+                          "sv": "1 - snabbaste läget"
+                      },
+                      "id": "ecfad581-b007-4612-a0e0-fcc551f4057f",
+                      "items": {
+                          "AIPCompressionLevel": "1"
+                      }
+                  },
+                  {
+                      "description": {
+                          "en": "3 - fast compression",
+                          "es": "3 - mode de compresión rápido",
+                          "fr": "3 - mode de compression rapide",
+                          "ja": "3 - 高速圧縮モード",
+                          "pt_BR": "3 - modo de compressão rápida",
+                          "sv": "3 - snabbkomprimeringsläge"
+                      },
+                      "id": "85b2243e-ff97-4ca8-80e8-3c6b0842b360",
+                      "items": {
+                          "AIPCompressionLevel": "3"
+                      }
+                  },
                     {
                         "description": {
-                            "en": "5 - normal compression mode",
+                            "en": "5 - normal compression",
                             "es": "5 - modo de compresión normal",
                             "fr": "5 - mode de compression normal",
                             "ja": "5 - 通常の圧縮モード",
@@ -866,34 +894,6 @@
                         "id": "6d52fd24-8c06-4c8e-997a-e427ba0acc36",
                         "items": {
                             "AIPCompressionLevel": "9"
-                        }
-                    },
-                    {
-                        "description": {
-                            "en": "3 - fast compression mode",
-                            "es": "3 - mode de compresión rápido",
-                            "fr": "3 - mode de compression rapide",
-                            "ja": "3 - 高速圧縮モード",
-                            "pt_BR": "3 - modo de compressão rápida",
-                            "sv": "3 - snabbkomprimeringsläge"
-                        },
-                        "id": "85b2243e-ff97-4ca8-80e8-3c6b0842b360",
-                        "items": {
-                            "AIPCompressionLevel": "3"
-                        }
-                    },
-                    {
-                        "description": {
-                            "en": "1 - fastest mode",
-                            "es": "1 - el modo más rápido",
-                            "fr": "1 - mode rapide",
-                            "ja": "1 - 最高速モード",
-                            "pt_BR": "1 - modo mais rápido",
-                            "sv": "1 - snabbaste läget"
-                        },
-                        "id": "ecfad581-b007-4612-a0e0-fcc551f4057f",
-                        "items": {
-                            "AIPCompressionLevel": "1"
                         }
                     }
                 ]
@@ -3187,8 +3187,8 @@
                 "@manager": "linkTaskManagerChoice",
                 "@model": "MicroServiceChainChoice",
                 "chain_choices": [
-                    "433f4e6b-1ef4-49f8-b1e4-49693791a806",
-                    "9efab23c-31dc-4cbd-a39d-bb1665460cbe"
+                    "9efab23c-31dc-4cbd-a39d-bb1665460cbe",
+                    "433f4e6b-1ef4-49f8-b1e4-49693791a806"
                 ]
             },
             "description": {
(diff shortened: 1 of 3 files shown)


ablwr:
TTTTHHHHHHAAAANNNKKK YOOUUUUUUU

replaceafill:
@sallain I think it's a problem in our code. From what I understand [the `_pull_choices` helper](https://github.com/artefactual/archivematica/blob/862c4cdd780f1a235f20716d8472d34d9b56094f/src/MCPServer/lib/RPCServer.py#L445) loses the sorting of the choices because it's returning a Python dictionary (presumably for filtering purposes?) which order is arbitrary and doesn't correspond to the chain in the workflow.

That dictionary is then [taken by the JS code](https://github.com/artefactual/archivematica/blob/4783c93467c14a95d0e4e23def141e858ac1e474/src/dashboard/src/media/js/transfer.js#L150-L168) and iterated (in the said arbitrary order) and used to build the dropdown in the UI.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -12225,6 +12213,18 @@
                         "id": "85b1e45d-8f98-4cae-8336-72f40e12cbef",
                         "items": {
                             "DeletePackage": "True"
+                      }
```
sevein: Indentation should use two extra spaces.

src/MCPServer/lib/server/rpc_server.py
```diff
@@ -27,10 +27,10 @@
 
 
 from main.models import Job, SIP, Transfer
-
 from server.db import auto_close_old_connections
 from server.packages import create_package, get_approve_transfer_chain_id
 from server.processing_config import get_processing_fields
+from collections import OrderedDict
```
sevein: This import should be moved the first import block (standard library imports).

src/MCPServer/lib/server/rpc_server.py
```diff
@@ -475,7 +475,9 @@ def _pull_choices(job_id, lang, jobs_awaiting_for_approval):
     The caller should expect ``JobNotWaitingForApprovalError`` to be raised
     when the job does not need a decision to be made.
     """
-    ret = {}
+
+    ret = OrderedDict()
+
```
sevein: I don't think we need the extra blank lines in L478 and L480.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -902,7 +902,35 @@
...
+                            "es": "1 - el modo más rápido",
+                            "fr": "1 - mode rapide",
+                            "ja": "1 - 最高速モード",
+                            "pt_BR": "1 - modo mais rápido",
+                            "sv": "1 - snabbaste läget"
+                        },
+                        "id": "ecfad581-b007-4612-a0e0-fcc551f4057f",
+                        "items": {
+                            "AIPCompressionLevel": "1"
+                      	}
```
sevein: Use of tab here instead of spaces, right before the closing bracket.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -902,7 +902,35 @@
...
+	                          "fr": "3 - mode de compression rapide",
+	                          "ja": "3 - 高速圧縮モード",
+	                          "pt_BR": "3 - modo de compressão rápida",
+	                          "sv": "3 - snabbkomprimeringsläge"
+                      	},
+	                      "id": "85b2243e-ff97-4ca8-80e8-3c6b0842b360",
+	                      "items": {
+	                          "AIPCompressionLevel": "3"
+	                      }
+	                  },
```
sevein: This block has indentation issues, e.g. L918 has a tab, other lines seem to be off by some spaces.

sevein (approved):
I've made some comments on cosmetics but the changes look good.

ablwr:
@replaceafill did all the work, directing me where we could simply add an `OrderedDict` to resolve this. I've tested this locally and seems well (although not extensive testing). I think this is ready for review again.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -902,7 +902,35 @@
...
+                            "sv": "3 - snabbkomprimeringsläge"
+                        },
+                        "id": "85b2243e-ff97-4ca8-80e8-3c6b0842b360",
+                        "items": {
+                            "AIPCompressionLevel": "3"
+                        }
+                    },
+                    {
+                        "description": {
+                            "en": "5 - normal compression",
```
ross-spencer, line 859: Unfortunately changing the option name will break AMAUAT so I feel there are two options before merging, 

* We revert the name to: `5 - normal compression mode`.
* Create a PR submitted against AMAUAT to look for the new string [here](https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/14bcaef70793668d43071083ee00521077b9b862/amuser/am_browser_ability.py#L477-L480).

I have no opinion about which is the correct approach, but I'd be tempted to keep the scope of this issue to re-ordering which doesn't impact AMAUAT, and then we can address renaming another time. If there's the capacity, a PR against AMAUAT would be equally welcome now.

ross-spencer (commented):
UUIDs are tricky! This is nice, I've left one comment about AMAUAT which should be addressed first. Once that's done, we can get this approved and merged. Nice work all figuring it out and making the changes.

src/MCPServer/lib/assets/workflow.json
sallain: @ross-spencer Is that the only one? I've renamed other workflow choices as well, so I'd prefer to keep those in and change the AMAUAT.

ross-spencer: I believe it might be having a look today. So that's not so bad. The change needed to be made [here](https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/14bcaef70793668d43071083ee00521077b9b862/amuser/am_browser_ability.py#L477-L480) and I've created a PR while I was in there so you can have a look [here](https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/154). If y'all rebase and merge this during the next sprint, and then Douglas or yourself can review that change, and it will quickly show up if that's not the only problem when we run the integration tests in Jenkins again and we can fix it pretty easily. It's just something to consider for future workflow changes.

sallain:
@ross-spencer Coming back to this after ages and ages - your PR fixes AMAUAT, so I think this could be merged? I have just rebased.

ross-spencer:
@sallain nice! go for it!

@scollazo FYI! So I think we'll see the tests break, and then I can follow up and merge AMAUAT as soon as I can, and then we'll see the tests pass. And all will be well with the world again!

ross-spencer:
Although also, just checking @sallain right now we've seven commits associated with this PR we'd probably normally squish if possible; what approach will you use to create a single feature commit?

PR #1503, opened 2019-10-31
Workflow: move pid declaration job

replaceafill:
This PR moves the `Persistent identifier (PID) declaration` job from the `Bind PIDs` microservice to the `Process metadata directory` microservice in the Ingest tab so it can be executed when the `Bind PIDs` config option is set to `No`.

The description of the job has also been updated to `Load persistent identifiers from metadata/identifiers.json` for more transparency.

**Here's the position of the job (highlighted in grey) in the workflow before the change:**

before

**And this is its position (and label) now:**

after

Connected to https://github.com/archivematica/Issues/issues/963

### DIFF
--- a/src/MCPServer/lib/assets/workflow.json
+++ b/src/MCPServer/lib/assets/workflow.json
@@ -507,7 +507,7 @@
                 "pt_BR": "Sim",
                 "sv": "Ja"
             },
-            "link_id": "873b428f-2c86-42b6-b463-aeda925bf559"
+            "link_id": "87e93d08-36e4-4c81-99a8-beea00b18400"
         },
         "94f764ad-805a-4d4e-8a2b-a6f2515b30c7": {
             "description": {
@@ -1224,11 +1224,11 @@
             "exit_codes": {
                 "0": {
                     "job_status": "Completed successfully",
-                    "link_id": "75fb5d67-5efa-4232-b00b-d85236de0d3f"
+                    "link_id": "873b428f-2c86-42b6-b463-aeda925bf559"
                 }
             },
             "fallback_job_status": "Failed",
-            "fallback_link_id": "75fb5d67-5efa-4232-b00b-d85236de0d3f",
+            "fallback_link_id": "873b428f-2c86-42b6-b463-aeda925bf559",
             "group": {
                 "en": "Process metadata directory",
                 "es": "Procesar directorio de metadatos",
@@ -7526,21 +7526,21 @@
                 "stdout_file": null
             },
             "description": {
-                "en": "Persistent identifier (PID) declaration"
+                "en": "Load persistent identifiers from external file"
             },
             "exit_codes": {
                 "0": {
                     "job_status": "Completed successfully",
-                    "link_id": "87e93d08-36e4-4c81-99a8-beea00b18400"
+                    "link_id": "75fb5d67-5efa-4232-b00b-d85236de0d3f"
                 }
             },
             "fallback_job_status": "Failed",
-            "fallback_link_id": "7d728c39-395f-4892-8193-92f086c0546f",
+            "fallback_link_id": "75fb5d67-5efa-4232-b00b-d85236de0d3f",
             "group": {
-                "en": "Bind PIDs",
-                "fr": "Lier les PIDs",
-                "pt_BR": "Vincular PIDs",
-                "sv": "Lås PID:er."
+                "en": "Process metadata directory",
+                "es": "Procesar directorio de metadatos",
+                "pt_BR": "Processar diretório de metadados",
+                "sv": "Bearbeta metadatamapp"
             }
         },
         "87e7659c-d5de-4541-a09c-6deec966a0c0": {


src/MCPServer/lib/assets/workflow.json
```diff
@@ -7526,21 +7526,21 @@
                 "stdout_file": null
             },
             "description": {
-                "en": "Persistent identifier (PID) declaration"
+                "en": "Load persistent identifiers from metadata/identifiers.json"
```
ross-spencer: > Persistent identifier (PID) declaration

I feel like this warrants more discussion in this GitHub issue as you have done with the decision to move this to a different microservice group. This is the original title for the microservice we came up with for the client in `1.9` so it's good to make that open in there too. 

The docs are a little more ambiguous about the title of the service [here](https://www.archivematica.org/en/docs/archivematica-1.10/user-manual/transfer/transfer/#transfers-with-existing-persistent-identifiers). This PR does introduce a need to just fix those up so the new microservice group is reflected, and the new microservice name if we change it.

src/MCPServer/lib/assets/workflow.json
```diff
@@ -1224,11 +1224,11 @@
             "exit_codes": {
                 "0": {
                     "job_status": "Completed successfully",
-                    "link_id": "75fb5d67-5efa-4232-b00b-d85236de0d3f"
+                    "link_id": "873b428f-2c86-42b6-b463-aeda925bf559"
```
ross-spencer, line 1227: This looks good and is a sensible change. It's interesting to think about the SIP more completely now with the 1.9 changes for SFU, i.e. ensuring persistent identifiers arrive there does seem important.

ross-spencer (changes requested):
This looks good but requesting changes just so the decision and discussion around changing the title of this microservice job can be surfaced in the Github issue.

Also, I was thinking about how this issue was introduced into the QA branch. Do you think it would be helpful for devs to consider including snippets from [amflow](https://github.com/artefactual-labs/amflow) or the dashboard workflow itself when workflow changes are made to make it more explicit when a pathway in Archivematica is changing? Just considering this from a CR/development guide perspective where we can perhaps avoid changes like were introduced. 

FWIW: I don't think it's needed here. Just more a more general consideration.

ross-spencer:
@replaceafill sorry, I just noticed this PR is tagged as a draft. I missed that. I would have waited before CRing otherwise. My bad!

src/MCPServer/lib/assets/workflow.json
replaceafill: My bad. I discussed the title change in Slack with @peterVG. I should have raised it in the Github issue. And yes, I was thinking of adding one of those tasks checkboxes to the PR description regarding the documentation change but I need to figure out how to do that.

replaceafill:
@ross-spencer NP! Thanks for taking a look. I kept this as a draft because I was waiting on @peterVG's feeback on https://github.com/archivematica/Issues/issues/963#issuecomment-548149506.

> Do you think it would be helpful for devs to consider including snippets from amflow or the dashboard workflow itself when workflow changes are made to make it more explicit when a pathway in Archivematica is changing?

Absolutely! When I was creating the PR I was like "how is the reviewer going to understand all these UUID updates?" I thought about updating the description explaining things like "Now the Yes choice (uuid-...-...) will point to the Bind PIDs job (uuid-...-...)". I think it's worth representing the changes in this PR using amflow screenshots. Thanks for the idea!

sevein:
> Absolutely! When I was creating the PR I was like "how is the reviewer going to understand all these UUID updates?" I thought about updating the description explaining things like "Now the Yes choice (uuid-...-...) will point to the Bind PIDs job (uuid-...-...)". I think it's worth representing the changes in this PR using amflow screenshots. Thanks for the idea!

We could have an alternative representation of the workflow written as code. I remember that Mark had to do that at some point to wrap his head around the changes that he was making around normalization. See https://github.com/hudmol/am-generate-microservice-chains. So he ended up comparing code diffs, aka workflow code, as opposed to json diffs. We could have amflow generate similar output.

It would help to reduce the size of the workflow too of course. It'd be helpful to have a tool to communicate changes that shrink the workflow.

sevein (approved):

ross-spencer (approved):
Thanks @replaceafill this works as anticipated. I can see the job trigger in the correct group:

image

Great to have this functionality back! :slightly_smiling_face: 

I'd still like to ask about the job name while there is a chance today, `Load persistent identifiers from metadata/identifiers.json`.

If you consider ways this service could change:

* `identifiers.json` could be loaded from a different folder.
* `identifiers.json` could also be `identifiers.xml` `identifiers.csv`, 
* We could build redundancy into this service by loading it from `a-different-name.json`.
* The job could also now simplify workflow elsewhere by wrapping a handful of others, bringing in the `mods` persistent identifier work, or "a new source yet to be identified" as the system grows. 

Each of those changes would also require writing a new microservice name in `workflow.json` and as we have seen, updating documentation. Whereas the right abstraction here just means the code underneath can be worked on without impacting what the user sees or understands the service to do. Could we consider `Load persistent identifiers from external file` or (`Load persistent identifiers from external source`) perhaps, or something else that doesn't show the implementation as literally?

peterVG:
We were literal about the recent PREMIS events import using premis.xml in the Job name which is maybe where this convention came from. However, Ross' arguments make sense. I would go with `Load persistent identifiers from external file`.

PR #1505, opened 2019-11-01
Fix constant gearman polling in MCPServer

cole:
A tight loop in GearmanTaskBackend resulted in constant polling of gearman for status updates while waiting for a job to complete, which burns a lot of CPU for no reason 😭 .  This PR uses a lower level client method to wait for job activity.

Also removes possible naming confusion by renaming `gearman` to `gearman_backend`, so it doesn't have the same name as the library.

This issue was the worst when running more packages in parallel than there are MCPClients on the system, or with longer gearman jobs.

Test results with 3 transfers, 2 mcp clients:

Original (CPU seconds by process, orange is MCPServer):
Screen Shot 2019-11-01 at 10 59 59 AM

Revised (CPU seconds by process, orange is MCPServer):
Screen Shot 2019-11-01 at 11 26 27 AM

Note that MCPServer still uses the most cpu! 😞 However, this is a huge improvement.

Connects to archivematica/Issues#911.

### DIFF
--- a/src/MCPServer/lib/server/tasks/backends/__init__.py
+++ b/src/MCPServer/lib/server/tasks/backends/__init__.py
@@ -6,7 +6,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import threading
 
 from server.tasks.backends.base import TaskBackend
-from server.tasks.backends.gearman import GearmanTaskBackend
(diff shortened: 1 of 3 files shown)


cole:
Relates to: https://github.com/archivematica/Issues/issues/911.

sevein (approved):
Nice, much better than a random time.sleep!

PR #1506, opened 2019-11-06
Instrumentation: Zero file format counters on startup

cole:
Relates to: https://github.com/archivematica/Issues/issues/964

I thought we could get away with not zeroing these, but we need to start at zero to calculate the rate of change properly.

Cardinality is very high here, but the recommendation for fewer than 100 labels in Prometheus is for use with 1000 nodes (MCPClient instances in this case), which I think is many more than anyone will be running.


replaceafill (approved):
LGTM :+1: Tested it with the latest Grafana dashboard in am/compose and the file format charts are now populated from the beginning.

PR #1509, opened 2019-11-15
MCPServer: fix expiring batches issue

sevein:
In Gearman, clients can submit background or non-background jobs. We use the
latter option in which Gearman automatically forwards work updates from workers
to clients, i.e. from MCPClient to MCPServer.

In 5969edae8, MCPServer learned how to process these worker updates efficiently
by monitoring connection activity. It removed the need for sending `GET_STATUS`
requests to the server which would only make sense when using background jobs.
The problem with that commit was that `wait_until_any_job_completed` was
inadvertently marking pending jobs as timed out.

This commit updates `wait_until_any_job_completed` so we don't give up on
pending jobs unless the server delivers an status update for them.

For context, see http://gearman.org/protocol/.

> WORK_DATA, WORK_WARNING, WORK_STATUS, WORK_COMPLETE,
> WORK_FAIL, WORK_EXCEPTION
>
>     For non-background jobs, the server forwards these packets from
>     the worker to clients. See "Worker Requests" for more information
>     and arguments.

Connects to https://github.com/archivematica/Issues/issues/911.
Follow up on https://github.com/artefactual/archivematica/pull/1505.

### DIFF
--- a/src/MCPServer/lib/server/tasks/backends/gearman_backend.py
+++ b/src/MCPServer/lib/server/tasks/backends/gearman_backend.py
@@ -49,9 +49,7 @@ class MCPGearmanClient(GearmanClient):
 
         # Mark any job still in the queued state to poll_timeout
         for current_request in job_requests:
-            current_request.timed_out = not current_request.complete
-
-            if not current_request.timed_out:
+            if current_request.complete:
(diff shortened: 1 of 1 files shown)


replaceafill (approved):
LGTM :+1: and works great!

PR #1513, opened 2019-11-26
Search: index SIP and File identifiers

replaceafill:
This PR adds SIP and File identifiers (set by the `Load persistent identifiers from external file` job) to the existing `identifiers` field which is already indexed in ElasticSearch. This allows to search these values in the Archival storage.

I also updated a couple of docstrings, fixed the string representation of the Identifier model and changed a couple of mutable default arguments.

Connected to https://github.com/archivematica/Issues/issues/1006


sevein (approved):
LGTM

PR #1514, opened 2019-12-02
Make AIP and directory level metadata searchable

replaceafill:
This PR extracts descriptive and administrative metadata attached to directories through the `metadata.csv` file, the UI metadata forms or the `bag-info.txt` file and makes it searchable through the `transferMetadata` option in the Archival storage tab (the Any option will work too).

So far searching file metadata has relied on us indexing their `dmdSec` sections from the METS. But this presents several (potential) problems:

* See enhacement 3. in https://github.com/archivematica/Issues/issues/404
* https://github.com/artefactual/archivematica/pull/1463#issuecomment-536711483
* https://github.com/archivematica/Issues/issues/619#issuecomment-487190700

This PR also represents a step forward on deprecating the `METS` entry in the `aipfiles` index (the `METS` entry of the `aips` index was already removed in https://github.com/artefactual/archivematica/pull/1463) by:

* Creating a simple data structure with the descriptive metadata of the file. We are still missing a structure to index PREMIS data like events.
* Adding AIP metadata information to each entry in the `aipfiles` index. This may increase the size of the index but would allow us to avoid making separate queries when searching for AIPs. See enhancement 9. in https://github.com/archivematica/Issues/issues/404

Connected to https://github.com/archivematica/Issues/issues/547
Connected to https://github.com/archivematica/Issues/issues/888

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -24,6 +24,7 @@ from __future__ import absolute_import
 from __future__ import division
 
 import calendar
+import copy
 import datetime
 import json
 import logging
@@ -446,6 +447,8 @@ def index_aip_and_files(
         identifiers = []
     identifiers += _get_sip_identifiers(uuid)
 
+    aip_metadata = _get_aip_metadata(root)
+
     aip_data = {
         "uuid": uuid,
         "name": name,
@@ -457,7 +460,7 @@ def index_aip_and_files(
         "isPartOf": is_part_of,
         "countAIPsinAIC": aips_in_aic,
         "identifiers": identifiers,
-        "transferMetadata": _extract_transfer_metadata(root),
+        "transferMetadata": aip_metadata,
         "encrypted": encrypted,
     }
     _wait_for_cluster_yellow_status(client)
@@ -465,12 +468,17 @@ def index_aip_and_files(
     printfn("Done.")
     printfn("Indexing AIP files ...")
     _index_aip_files(
-        client=client, uuid=uuid, mets=root, name=name, identifiers=identifiers
+        client=client,
+        uuid=uuid,
+        mets=root,
+        name=name,
+        identifiers=identifiers,
+        aip_metadata=aip_metadata,
     )
     return 0
 
 
-def _index_aip_files(client, uuid, mets, name, identifiers=None):
+def _index_aip_files(client, uuid, mets, name, identifiers=None, aip_metadata=None):
     """Index AIP files from AIP with UUID `uuid` and METS at path `mets_path`.
 
     :param client: The ElasticSearch client.
@@ -478,6 +486,8 @@ def _index_aip_files(client, uuid, mets, name, identifiers=None):
     :param mets: root Element of the METS document.
     :param name: AIP name.
     :param identifiers: optional additional identifiers (MODS, Islandora, etc.).
+    :param aip_metadata: list with the descriptive and administrative metadata
+                         of each directory in the AIP
     """
     # Extract isPartOf (for AIPs) or identifier (for AICs) from DublinCore
     dublincore = ns.xml_find_premis(
@@ -499,6 +509,9 @@ def _index_aip_files(client, uuid, mets, name, identifiers=None):
     if identifiers is None:
         identifiers = []
 
+    if aip_metadata is None:
+        aip_metadata = []
+
     # Establish structure to be indexed for each file item
     fileData = {
         "archivematicaVersion": version.get_version(),
@@ -512,7 +525,6 @@ def _index_aip_files(client, uuid, mets, name, identifiers=None):
         "AICID": aic_identifier,
         "METS": {"dmdSec": {}, "amdSec": {}},
         "origin": get_dashboard_uuid(),
-        "transferMetadata": _extract_transfer_metadata(mets),
     }
 
     # Index all files in a fileGrup with USE='original' or USE='metadata'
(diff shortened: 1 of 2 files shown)


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -457,20 +460,25 @@ def index_aip_and_files(
...
+        mets=root,
+        name=name,
+        identifiers=identifiers,
+        aip_metadata=aip_metadata,
     )
     return 0
 
 
-def _index_aip_files(client, uuid, mets, name, identifiers=None):
+def _index_aip_files(client, uuid, mets, name, identifiers=None, aip_metadata=None):
```
sevein, line 481: Since we have a docstring describing all the existing parameters, could you add the new one to the list?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -824,13 +841,131 @@ def _remove_tool_output_from_mets(doc):
...
 
 
-def _extract_transfer_metadata(doc):
-    return [
-        xmltodict.parse(etree.tostring(el))["transfer_metadata"]
-        for el in ns.xml_findall_premis(
-            doc, "mets:amdSec/mets:sourceMD/mets:mdWrap/mets:xmlData/transfer_metadata"
-        )
-    ]
+def get_directories_with_metadata(container):
```
sevein: I see a pattern in your PR where internal functions are not using the underscore convention, are you doing that on purpose because you think they should be public?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -824,13 +841,131 @@ def _remove_tool_output_from_mets(doc):
...
+    """
+    Look for Directory entries that have descriptive or administrative
+    metadata sections.
+    """
+    result = []
+    for directory in ns.xml_findall_premis(container, './/mets:div[@TYPE="Directory"]'):
+        if "DMDID" in directory.attrib or "ADMID" in directory.attrib:
```
sevein: Not sure how expensive `directory.attrib` is, but if it was, should be avoided?
Using an extra var? Or `directory.attrib in ('DMDID', 'AMDID')`?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -824,13 +841,131 @@ def _remove_tool_output_from_mets(doc):
...
+    each directory.
+    """
+    result = []
+    physical_struct_map = ns.xml_find_premis(doc, 'mets:structMap[@TYPE="physical"]')
+    if physical_struct_map is not None:
+        for directory in get_directories_with_metadata(physical_struct_map):
+            directory_metadata = get_directory_metadata(directory, doc)
+            if directory_metadata:
+                result.append(directory_metadata)
+    return result
```
sevein, line 971: I'm wondering if this would be a good case for generators. Say you're looking at 10k directories, would generators help noticeably? If it's not much of an effort (yields and yields), perhaps we should?

sevein (commented):
I've submitted a few comments. I haven't tested yet, will do that next.

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: No. Forgot to make them "private". Will fix.

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: `directory.attrib` is a dictionary in this case. I really wanted to do it through a single XPath query but couldn't make it work. For example you could do:

```
ns.xml_findall_premis(container, './/mets:div[@TYPE="Directory"][@DMDID]')
```

and get all directories with descriptive metadata sections, but I couldn't make an `OR` work there, like:

```
ns.xml_findall_premis(container, './/mets:div[@TYPE="Directory"][@DMDID or @AMDID ]')
```

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: Thanks for the idea. `_get_directories_with_metadata` can be converted to a generator :+1:

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: Thanks to @jraddaoui's help I made the `OR` work in https://github.com/artefactual/archivematica/pull/1514/commits/220c93c5e0a34943449c6ff4411ed718622c8fed :bowing_man:

src/archivematicaCommon/lib/elasticSearchFunctions.py
replaceafill: `_get_directories_with_metadata` ended as a single call to lxml's xpath.

replaceafill:
@sevein I think I addressed your observations. Let me know how testing goes.

sevein (approved):
Thanks, Douglas. I've tested this and the limitations described around searches now seem to be solved. It also appears that you've made an effort to not introduce backward-incompatible changes in the searches, i.e. users won't need to rebuild the search indices necessarily while they may still do if they wish.

It'd be interesting to identify what's missing in metsrw that it didn't make it appealing in this particular use case. If you have a clear idea, I'd suggest to file [a new issue](https://github.com/artefactual-labs/mets-reader-writer/issues/new) in the metsrw repo. We could try to complete metsrw for this particular use case at some point.

:clap: :clap: :clap:

ross-spencer:
Hi @sevein @replaceafill I feel the issue should be https://github.com/archivematica/issues still, but the potential use case does sound good. We haven't a label in issues yet for `metsrw` I believe, but we can add one if it helps tease apart AM core from other libs.

PR #1515, opened 2019-12-05
Correct behavior of move or merge

ross-spencer:
archivematica/issues#1018 documents that Archivematica previously
assumed a `metadata` folder existed, or would be created in the
correct location in a potential AIP during the ingest workflow.

This commit fixes a regression that caused this directory not to be
created when a 'src' location was otherwise empty. If a top-level
directory doesn't exist it will now get created.

NB. We create some additional logging in the hope it will help debug in future. A character encoding is added to the head of the Python script as a matter of course.

Connected to archivematica/issues#1018

### DIFF
--- a/src/MCPClient/lib/clientScripts/move_or_merge.py
+++ b/src/MCPClient/lib/clientScripts/move_or_merge.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python2
+# -*- encoding: utf-8
 
 import errno
 import filecmp
@@ -6,6 +7,10 @@ import os
 import shutil
 
 import scandir
+
+from custom_handlers import get_script_logger
+
+logger = get_script_logger("archivematica.mcp.client.move_or_merge")
 
 
 def mkdir_p(path):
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -51,20 +56,32 @@ def move_or_merge(src, dst):
...
         # it finds the path relative to the top of ``src``, then moves it
-        # to the corresponding path in ``dst``.
+        # to the corresponding path in ``dst``. If the top-level ``dst``
+        # directory doesn't already exist, we guarantee its creation even if
+        # ``src`` lower-levels are empty.
+        logger.info("src: %s is a directory", src)
+
+        if not os.path.exists(dst):
+            logger.info("Creating top-level dst folder: %s", dst)
+            mkdir_p(dst)
```
replaceafill: I wonder if we could `shutil.move(src, dst)` here (instead of `mkdir_p`) and avoid moving files individually (the merging logic of the function) with an `else`?

src/MCPClient/lib/clientScripts/move_or_merge.py
```diff
@@ -82,7 +99,7 @@ def main(src, dst):
 
 def call(jobs):
     for job in jobs:
-        with job.JobContext():
+        with job.JobContext(logger=logger):
```
replaceafill, line 100: Thank you! I still wonder what should be our logging strategy in client scripts: stderr, logger, job.pyprint, etc etc etc.

src/MCPClient/tests/test_move_or_merge.py
```diff
@@ -1,5 +1,6 @@
 # -*- encoding: utf-8
 
+from os import listdir
```
replaceafill: I think you can use pytest for this:

https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture
https://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.listdir

replaceafill (changes requested):
Looks good overall @ross-spencer . Left a question about moving instead of keep merging files individually and suggested using the pytest fixture for listing directory contents.

src/MCPClient/lib/clientScripts/move_or_merge.py
ross-spencer: I know I know :D I like logging here because it's like a utility script, but really, I haven't found a pattern I'm happy with yet either. Like you too, I'm just happy to have a bit more logging here. Do you think the granularity is okay?

replaceafill: I think it's a matter of taste, but a couple seem to verbose to me: you could infer if `src` is a file or a directory from the initial `info`s?

src/MCPClient/tests/test_move_or_merge.py
ross-spencer: That's a nice tip, thank you!

ross-spencer:
Hi @replaceafill I'm pretty comfortable archivematica/issues#955 is a separate issue, possibly the same for the other possibly related issue we've been speaking about. As such, I've made the CR changes and rebased, so we can move forward with this as an incremental improvement and then I'll move back to 955 and other tasks. Hopefully it's looking okay. I've tightened up my use of newlines a little, and added another log line too.

src/MCPClient/lib/clientScripts/move_or_merge.py
ross-spencer: True, I can remove those two for sure. Will make another commit.

replaceafill (approved):
LGTM :+1:

PR #1516, opened 2019-12-06
Fix Upload to ArchivesSpace job status

replaceafill:
This PR makes the `Upload to ArchivesSpace` job to fail if any file cannot be paired or uploaded. It also cleans some path handling and adds a unit test for the client script.

Connected to https://github.com/archivematica/Issues/issues/258

### DIFF
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -312,24 +312,26 @@ def call(jobs):
 
                 try:
                     files = get_files_from_dip(args.dip_location)
+                    if not upload_to_archivesspace(
+                        files,
+                        client,
+                        args.xlink_show,
+                        args.xlink_actuate,
+                        args.object_type,
+                        args.use_statement,
+                        args.uri_prefix,
+                        args.dip_uuid,
+                        args.access_conditions,
+                        args.use_conditions,
+                        args.restrictions,
+                        args.dip_location,
+                        args.inherit_notes,
+                    ):
+                        raise ValueError("not all files could be paired")
                 except ValueError:
                     job.set_status(2)
                     continue
                 except Exception:
                     job.set_status(3)
                     continue
-                upload_to_archivesspace(
-                    files,
-                    client,
-                    args.xlink_show,
-                    args.xlink_actuate,
-                    args.object_type,
-                    args.use_statement,
-                    args.uri_prefix,
-                    args.dip_uuid,
-                    args.access_conditions,
-                    args.use_conditions,
-                    args.restrictions,
-                    args.dip_location,
-                    args.inherit_notes,
-                )
+                job.set_status(0)
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -104,7 +104,7 @@ def upload_to_archivesspace(
         uuid = file_name[0:36]
 
         if uuid not in pairs:
-            logger.warning("Skipping file {} ({}) - no pairing found".format(f, uuid))
+            logger.error("Skipping file {} ({}) - no pairing found".format(f, uuid))
             all_files_paired_successfully = False
             continue
 
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -137,6 +137,7 @@ def upload_to_archivesspace(
                     ]
 
         # determine restrictions
+        restrictions_apply = False
         if restrictions == "no":
             restrictions_apply = False
         elif restrictions == "yes":
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -34,11 +34,14 @@ def recursive_file_gen(mydir):
 
 def get_files_from_dip(dip_location):
     # need to find files in objects dir of dip:
-    # go to dipLocation/dipName/objects
+    # go to dipLocation/objects
     # get a directory listing
     # for each item, set fileName and go
     try:
-        mydir = dip_location + "objects/"
+        # remove trailing slash
+        if dip_location != os.path.sep:
+            dip_location = dip_location.rstrip(os.path.sep)
+        mydir = os.path.join(dip_location, "objects")
         mylist = list(recursive_file_gen(mydir))
 
         if len(mylist) > 0:
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -235,12 +235,7 @@ def upload_to_archivesspace(
         delete_pairs(dip_uuid)
 
 
-def call(jobs):
-    RESTRICTIONS_CHOICES = ["yes", "no", "premis"]
-    EAD_SHOW_CHOICES = ["embed", "new", "none", "other", "replace"]
-    EAD_ACTUATE_CHOICES = ["none", "onLoad", "other", "onRequest"]
-    INHERIT_NOTES_CHOICES = ["yes", "y", "true", "1"]
-
+def get_parser(RESTRICTIONS_CHOICES, EAD_ACTUATE_CHOICES, EAD_SHOW_CHOICES):
     parser = argparse.ArgumentParser(
         description="A program to take digital objects from a DIP and upload them to an ArchivesSpace db"
     )
@@ -285,6 +280,16 @@ def call(jobs):
         type=str,
     )
     parser.add_argument("--version", action="version", version="%(prog)s 0.1.0")
+    return parser
+
+
+def call(jobs):
+    RESTRICTIONS_CHOICES = ["yes", "no", "premis"]
+    EAD_SHOW_CHOICES = ["embed", "new", "none", "other", "replace"]
+    EAD_ACTUATE_CHOICES = ["none", "onLoad", "other", "onRequest"]
+    INHERIT_NOTES_CHOICES = ["yes", "y", "true", "1"]
+
+    parser = get_parser(RESTRICTIONS_CHOICES, EAD_ACTUATE_CHOICES, EAD_SHOW_CHOICES)
 
     with transaction.atomic():
         for job in jobs:
--- a/src/MCPClient/lib/clientScripts/upload_archivesspace.py
+++ b/src/MCPClient/lib/clientScripts/upload_archivesspace.py
@@ -32,7 +32,7 @@ def recursive_file_gen(mydir):
             yield os.path.join(root, file)
 
 
-def get_files_from_dip(dip_location, dip_name, dip_uuid):
+def get_files_from_dip(dip_location):
     # need to find files in objects dir of dip:
     # go to dipLocation/dipName/objects
     # get a directory listing
@@ -303,9 +303,7 @@ def call(jobs):
                 )
 
                 try:
-                    files = get_files_from_dip(
-                        args.dip_location, args.dip_name, args.dip_uuid
-                    )
+                    files = get_files_from_dip(args.dip_location)
                 except ValueError:
                     job.set_status(2)
                     continue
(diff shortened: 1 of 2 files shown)


src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -91,16 +95,18 @@ def upload_to_archivesspace(
         or len(use_conditions) == 0
     ):
         logger.debug("Looking for mets: {}".format(dip_uuid))
-        mets_source = dip_location + "METS." + dip_uuid + ".xml"
+        mets_source = os.path.join(dip_location, "METS." + dip_uuid + ".xml")
```
ross-spencer: Good to see your use of the `os` path functions throughout. Making things more robust as we go, I wonder if we can use `"{}".format()` syntax for the filename too?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -134,6 +140,7 @@ def upload_to_archivesspace(
                     ]
 
         # determine restrictions
+        restrictions_apply = False
         if restrictions == "no":
             restrictions_apply = False
```
ross-spencer: We can probably remove `144` and `145`  as it won't result in a change to the `restrictions_apply` following the addition of 143.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -211,36 +218,42 @@ def upload_to_archivesspace(
...
+                use_statement=use_statement,
+                xlink_show=xlink_show,
+                xlink_actuate=xlink_actuate,
+                restricted=restrictions_apply,
+                use_conditions=use_conditions,
+                access_conditions=access_conditions,
+                size=size,
+                format_name=format_name,
+                format_version=format_version,
+                inherit_notes=inherit_notes,
```
ross-spencer, line 238: Yikes! This is a lot of parameters! :sweat_smile:

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -211,36 +218,42 @@ def upload_to_archivesspace(
...
+            )
+        except ArchivesSpaceError as e:
```
ross-spencer: Wherever possible I think we should try to avoid single character vars - `e` makes perfect sense I can see why it's used in this line, but `err` is self-descriptive and makes sense on any line.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -211,36 +218,42 @@ def upload_to_archivesspace(
...
+
+            logger.error(
+                "Could not upload {} to ArchivesSpace record {}. Error: {}".format(
+                    file_name, as_resource, str(e)
```
ross-spencer: Is there a reason not to use `%s` syntax here for the logger?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -211,36 +218,42 @@ def upload_to_archivesspace(
...
+                )
+            )
+            all_files_paired_successfully = False
```
ross-spencer, line 247: What do you think about raising the ` raise ValueError("not all files could be paired")` from `main` here instead? Rather than evaluating the return value there. Perhaps with your enhanced error string?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -298,28 +321,27 @@ def call(jobs):
                 )
 
                 try:
-                    files = get_files_from_dip(
-                        args.dip_location, args.dip_name, args.dip_uuid
-                    )
+                    files = get_files_from_dip(args.dip_location)
+                    if not upload_to_archivesspace(
```
ross-spencer: I see why you might have done this, but as we evaluate the result of upload to ArchivesSpace independently, we could keep that block separate and let this `try/except` manage the result from get_files_from_dip. IDK. What do you think?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
```diff
@@ -298,28 +321,27 @@ def call(jobs):
...
+                        args.restrictions,
+                        args.dip_location,
+                        args.inherit_notes,
+                    ):
+                        raise ValueError("not all files could be paired")
                 except ValueError:
                     job.set_status(2)
                     continue
                 except Exception:
                     job.set_status(3)
```
ross-spencer, line 327: I wish I knew historically what exceptions we couldn't anticipate here. Maybe with the range of your tests we could consider removing `except Exception`?

src/MCPClient/tests/test_upload_archivesspace.py
```diff
@@ -0,0 +1,309 @@
...
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.abspath(os.path.join(THIS_DIR, "../lib/clientScripts")))
+
+import upload_archivesspace
+
+
+def test_recursive_file_gen(tmpdir):
+    mydir = tmpdir.mkdir("mydir")
+    hello = mydir.join("hello.txt")
+    hello.write("hello!")
```
ross-spencer: `py.path` function signatures are strange, they have an `ensure=bool` parameter but I can't figure out if it will help drop this into two lines? `hello = tmpdir.join("mydir", "hello.txt")` and then write with ensure=True?

* py.path.write: https://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.write

src/MCPClient/tests/test_upload_archivesspace.py
```diff
@@ -0,0 +1,309 @@
...
+        params["use_conditions"],
+        params["restrictions"],
+        "/dip/location/path",
+        "",
+    )
+    mets_file_mock.assert_called_once_with("/dip/location/path/METS.dipuuid.xml")
+    logger.debug.assert_has_calls(
+        [
+            mocker.call("Looking for mets: dipuuid"),
+            mocker.call("Found mets file at path: /dip/location/path/METS.dipuuid.xml"),
```
ross-spencer, line 148: I didn't know about this use of `assert_has_calls`. useful!

ross-spencer (commented):
December 6 -> Jan 30 -- Not a bad turn-around for CR is it?! 

Looks great @replaceafill Your commitment to good/extensive tests is always appreciated. I've left some comments that would be good to get some feedback/responses from.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
replaceafill: Not really. I'm just not used to it. Does it provide more value than `.format` like in this case?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
replaceafill: You're in a per file iteration block here. Suppose you're pairing 5 files and the call to the API fails in the 3rd one. You still want to try to pair the rest, right?

src/MCPClient/lib/clientScripts/upload_archivesspace.py
replaceafill: Do you mean something like this?

```python
                try:
                    files = get_files_from_dip(args.dip_location)
                except ValueError:
                    job.set_status(2)
                    continue
                except Exception:
                    job.set_status(3)
                    continue
                if upload_to_archivesspace(
                    files,
                    client,
                    args.xlink_show,
                    args.xlink_actuate,
                    args.object_type,
                    args.use_statement,
                    args.uri_prefix,
                    args.dip_uuid,
                    args.access_conditions,
                    args.use_conditions,
                    args.restrictions,
                    args.dip_location,
                    args.inherit_notes,
                ):
                    job.set_status(0)
                else:
                    job.set_status(2)
```

src/MCPClient/lib/clientScripts/upload_archivesspace.py
replaceafill: Yeah, I wasn't sure either and tried to extend the coverage changing as little as possible.

src/MCPClient/tests/test_upload_archivesspace.py
replaceafill: Ah! Thanks for the tip!

replaceafill:
> December 6 -> Jan 30 -- Not a bad turn-around for CR is it?!

We've been busy! :rofl: 

Thank you @ross-spencer! I think I've addressed most of your observations and left a couple of questions too.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
ross-spencer: It probably doesn't need changing, other than one is native to the function, and the other is a second explicit function call. I'd be interested to see the behavior of the cast with `%s` here too. This is not a huge deal. I was asking to learn as much as anything :slightly_smiling_face:

src/MCPClient/lib/clientScripts/upload_archivesspace.py
ross-spencer: Right, and so we want to report on each record that doesn't pair so that they can be targeted more specifically if the user tries again? That makes sense to me.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
ross-spencer: I think that will be fine.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
ross-spencer: To risky to remove?

ross-spencer:
Thanks @replaceafill -- I've responded! Feel free to click re-review once you're satisfied there's nothing else, and then I'll give it a quick once over and approve.

src/MCPClient/lib/clientScripts/upload_archivesspace.py
replaceafill: As far as I can tell [`workflow.json` only cares about `0` or `not 0` for success and failure in this job](https://github.com/artefactual/archivematica/blob/81e3ebbd210cb9c98120146951569bfd7e02ad39/src/MCPServer/lib/assets/workflow.json#L7909-L7914) and we could probably get away using `0` and `1` exit codes. But that goes for a lot of other client scripts.

> I wish I knew historically what exceptions we couldn't anticipate here

Digging a bit on the history:

- [The original version of this script came after the ArchivesToolkit one](https://github.com/artefactual/archivematica/pull/185#discussion_r29613336)
- [Most of the exit codes, which just seem sequential, come from there too](https://github.com/artefactual/archivematica/commit/8d6bfa176c0ffe5c8f7ea2383aec630899f208d2#diff-f4b67121356eadc825560fe7f5af5c34R98)
- [Why `except Exception` was used](https://github.com/artefactual/archivematica/pull/185#discussion_r29622574)

One possibility I can think of is `scandir.walk` in `recursive_file_gen` raising a permission error like we saw in the Storage Service recently. The almost bare `except` would catch that :man_shrugging:

I'd prefer to leave it untouched if it's OK with you.

ross-spencer: No problem with me! 

Nice research! Thank you :slightly_smiling_face:

src/MCPClient/tests/test_upload_archivesspace.py
```diff
@@ -13,12 +13,11 @@
...
-    hello = mydir.join("hello.txt")
-    hello.write("hello!")
-    bye = mydir.mkdir("sub").join("bye.txt")
-    bye.write("bye!")
-    result = list(upload_archivesspace.recursive_file_gen(str(mydir)))
+    hello = tmpdir.join("mydir", "hello.txt")
+    hello.write("hello!", ensure=True)
+    bye = tmpdir.join("mydir", "sub", "bye.txt")
+    bye.write("bye!", ensure=True)
+    result = list(upload_archivesspace.recursive_file_gen(str(tmpdir / "mydir")))
```
ross-spencer, line 20: It looks like this approach will make sense for `test_get_files_from_dip_finds_files` as well?

ross-spencer (approved):
One comment left about `test_get_files_from_dip_finds_files` but you can choose to ignore it! :p or if you make the change and the tests pass, then it's also good to go! :+1:

PR #1517, opened 2019-12-06
Restructure bag: check for pre-existing metadata dir

hakamine:
Check for pre-existing metadata directory at the top level of the
transfer before moving bag metadata, merge contents if so.

Connected to archivematica/Issues#1022

### DIFF
--- a/src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
+++ b/src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
@@ -36,6 +36,8 @@ import bag
 import fileOperations
 from databaseFunctions import insertIntoEvents
 
+from move_or_merge import move_or_merge
+
 
 def restructureBagForComplianceFileUUIDsAssigned(
     job,
@@ -55,6 +57,15 @@ def restructureBagForComplianceFileUUIDsAssigned(
         dirPath = os.path.join(unitPath, dir)
         dirDataPath = os.path.join(unitPath, "data", dir)
         if os.path.isdir(dirDataPath):
+            if dir == "metadata" and os.path.isdir(dirPath):
+                # We move the existing top-level metadata folder, or merge it
+                # with what is currently there, before the next set of
+                # directory operations to move everything up a level below.
+                job.pyprint(
+                    "{}: moving/merging {} to {}".format(dir, dirPath, dirDataPath)
+                )
+                move_or_merge(dirPath, dirDataPath)
+
             # move to the top level
             src = dirDataPath
             dst = dirPath
@@ -67,6 +78,7 @@ def restructureBagForComplianceFileUUIDsAssigned(
                 unitPathReplaceWith,
             )
             job.pyprint("moving directory ", dir)
+
         else:
             if not os.path.isdir(dirPath):
                 job.pyprint("creating: ", dir)


ross-spencer:
Hi @hakamine prior to CR do you have an indication when you'd need this in AM core, with the scope for 1.11 set it might need to wait for 1.12, but we have a backlog grooming meeting later today, and same next week, or perhaps a quick note to @sromkey in #am-releases would work too.

hakamine:
Hi @ross-spencer @sromkey, for the immediate term this fix is needed on a client branch (which is 1.8.x based). However, the issue affects all AM versions. The fix is the same for both 1.x and 1.8.x (thus I only made a PR for 1.x so far). If preferable, I will make a separate PR for the fix applied client branch (btw I tested the fix on the client branch and looks to be working as expected)

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
```diff
@@ -55,6 +56,21 @@ def restructureBagForComplianceFileUUIDsAssigned(
...
         if os.path.isdir(dirDataPath):
+            # When moving metadata directory from the bag, check if
+            # there is already a metadata dir at the top level
+            # (it could have been created by automation tools)
+            # (ref. AM issue 1022 )
+            topleveldir_exists = False
+            if dir == "metadata" and os.path.isdir(dirPath):
+                # temporarily rename top level metadata directory
+                job.pyprint(
+                    "{}: temporarily renaming pre-existing directory".format(dir)
```
ross-spencer: If we've reached this part of the code we know `metadata` is going to be renamed. I'd be tempted to add the path as well so that we can get a complete rendering of exactly where in the structure the folder sits.

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
```diff
@@ -55,6 +56,21 @@ def restructureBagForComplianceFileUUIDsAssigned(
...
+                )
+                topleveldir_exists = True
+                tempdir = "{}_{}".format(dir, os.getpid())
```
ross-spencer: I like this use of `getpid()` to create something unique and simple! I think given the opportunity to refactor this, I'd avoid creating `tempdir` and call `"...".format(...)` directly on line 71 to tighten up on lines and avoid single use vars.

ross-spencer (commented):
Hi @hakamine I've updated the issue with some sample data that I've been using to recreate this without automation tools. The fix works nicely. 

Understanding the problem a little more and looking through your code, my instinct is to see if we can reuse this recent addition to `qa/1.x`: https://github.com/artefactual/archivematica/blob/82c624dd7ce3d0859d753e5774a9818f476d8d22/src/MCPClient/lib/clientScripts/move_or_merge.py#L49-L83 

Can you see a way to import the function from `move_or_merge` and use it to replace your move code here? It has the benefit too of having unit tests available for it, and ensures that we don't overwrite any existing material with a duplicate name in the pre-existing metadata folder. 

What do you think?

Just leaving this as a comment, and then, perhaps can re-review once done. Or let me know if there are any difficulties. (FWIW, I don't see any particular issue cherry-picking both of these client script fixes back into `1.8` and whatever next branches the client needs to be on.

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
hakamine: Good idea. Changed the pyprint message to show the complete path instead.

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
hakamine: done!

hakamine:
@ross-spencer Thank you very much for the review. Also thank you for letting me know about the move_or_merge() function (didn't know we had this helpful function in the code). I believe I addressed all your suggestions (I added new commits instead of squashing so that it is easier to review, please let me know if all looks good and I'll squash). I tried the fix (adapted to 1.8.x) in the client branch [here](https://github.com/artefactual/archivematica/tree/dev/am-issue-1022-bag-metadata-moma) and it is working fine.

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
```diff
@@ -55,6 +57,20 @@ def restructureBagForComplianceFileUUIDsAssigned(
...
+            # (ref. AM issue 1022 )
+            topleveldir_exists = False
+            if dir == "metadata" and os.path.isdir(dirPath):
+                # temporarily rename top level metadata directory
+                job.pyprint(
+                    "temporarily renaming pre-existing directory {}".format(dirPath)
+                )
+                topleveldir_exists = True
+                tempdirPath = os.path.join(unitPath, "{}_{}".format(dir, os.getpid()))
+                os.rename(dirPath, tempdirPath)
```
ross-spencer:
```suggestion
                move_or_merge(dirPath, dirDataPath)
```
I wonder what you think of this? (remove lines 67->72 and add this one line here instead). We can then avoid the rename here and then remove lines `87->94` too. Do the results match what you expect still? (for me they look good).

**NB.** _you might need to rebase to use the updated `move_or_merge` there is some additional robustness in that script._ 

This microservice is quite old and pretty complex, I'm not a fan of the `fileOperations.updateDirectoryLocation(...)` in the next few lines, there's another move going on there (as you know). I'd prefer to keep that all within a single microservice script, and maybe only make calls out to update the DB. But this isn't really in scope to fix here. What I think we can do is avoid adding too much more complexity, so if we can reduce the operations involved from a handful of lines to one, and possibly include some logging, then as a fix, that probably works quite well for the client here.

I've shown an example in a gist [here](https://gist.github.com/ross-spencer/5e4bf0290990b44bed0f4243a2ce7fc3/revisions) to make it clearer. Let me know if that change would be any good.

ross-spencer (commented):
Almost there @hakamine but I wonder if we can simplify it further. Comments inline. 

Yeah, it's so tough with this code base as you know, I wouldn't have known about `move_or_merge` either if I hadn't CR'ed it a week or so ago. But I'm glad it's helpful to you too.

src/MCPClient/lib/clientScripts/verify_and_restructure_transfer_bag.py
hakamine: @ross-spencer thank you very much! Using `move_or_merge()` from the beginning makes the logic simpler (like you, also not a fan of `fileOperations.updateDirectoryLocation(...)`, I was afraid that after the change something in the code inside it could produce some issue/error, but a quick test of the fix in the client branch worked good)

hakamine:
@ross-spencer Thank you for all the suggestions, this looks much better now. Please let me know if all is good and I'll squash in a single commit before merging (I'll also add you as co-author, if it's ok for you)

ross-spencer (approved):
Great stuff! Please do feel free to add me as co-. Has been good working through things - thanks for your patience!! 
 
LGTM! :+1:

PR #1520, opened 2019-12-11
Update METS version

replaceafill:
This updates the METS XML schema to the latest 1.12.1 version and the schema URL used for constructing the root element of METS files.

The version of the `mets-reader-writer` library has also been increased.

Connected to https://github.com/archivematica/Issues/issues/949


ross-spencer (approved):
:+1:

PR #1521, opened 2019-12-11
Remove mutable default argument

ross-spencer:
When reindexing, `_list_files_in_dir()` is run once for each item. If
the `filepaths` argument is left set to `[]`, then each item will
erroneously include all previously indexed files.

By setting it to `None` and initing the `[]` default inside the function
body, we avoid this undesirable behavior because the list is inited anew
each time the function runs.

Duplicate of https://github.com/artefactual/archivematica/pull/1500 to avoid more complicated rebase/merge effort (and because I thought it was going to be possible via the GH UI but was wrong!).


replaceafill (approved):
LGTM :+1:

ross-spencer:
Thanks @replaceafill. Thanks too @daaang for the contribution!

PR #1522, opened 2019-12-11
Fix user attributes encoding on agent creation

replaceafill:
This fixes an `UnicodeEncodeError` thrown by the add and edit user forms when non-ascii values are used in the First name and Last name fields.

Connected to https://github.com/archivematica/Issues/issues/261

### DIFF
--- a/src/dashboard/tests/test_models.py
+++ b/src/dashboard/tests/test_models.py
@@ -1,16 +1,19 @@
-from django.test import TestCase
+# -*- coding: utf-8 -*-
 
-from main.models import SIP, Transfer, UnitVariable, User
+from django.test import TestCase
+import mock
+
+from main import models
 
 
 class TestActiveAgent(TestCase):
     fixtures = ["test_user"]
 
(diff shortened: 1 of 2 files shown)


src/dashboard/tests/test_models.py
```diff
@@ -1,56 +1,77 @@
...
+    user_mock = mock.Mock(
+        id=1234, username=u"maría", first_name=u"María", last_name=u"Martínez"
+    )
+    models.create_user_agent(None, user_mock)
+    agent_mock.objects.update_or_create.assert_called_once_with(
+        userprofile__user=user_mock,
+        defaults={
+            "identifiertype": "Archivematica user pk",
+            "identifiervalue": "1234",
+            "name": 'username="maría", first_name="María", last_name="Martínez"',
```
ross-spencer, line 74: Subtle i-acute!! :D Nice choice though.

ross-spencer (approved):
I think this looks great @replaceafill I didn't realize it would be as simple. Nice work. :+1:

Issue #684, opened 2017-07-25
Problem: unused client scripts


sevein:
They seem to be unused at least in qa/1.x today (`819e074`).
```
backupDIP_v0.0 = %clientScriptsDirectory%backupDIP.sh
generateDIPFromAIPGenerateDIP_v0.0 = %clientScriptsDirectory%generateDIPFromAIPGenerateDIP.py
moveDspaceLicenseFilesToDSpaceLicenses_v0.0 = %clientScriptsDirectory%moveDspaceLicenseFilesToDSpaceLicenses.py
moveDspaceMetsFilesToDSpaceMETS_v0.0 = %clientScriptsDirectory%moveDspaceMetsFilesToDSpaceMETS.py
removeAIPFilesFromIndex_v0.0 = %clientScriptsDirectory%removeAIPFilesFromIndex.py
setSIPQuarantine_v0.0 = %clientScriptsDirectory%quarantineSIP.sh
verifyBAG_v0.0 = %clientScriptsDirectory%verifyBAG.py
```

Related: #607.

PR #1523, opened 2019-12-16
Use removeDirectories for removing bagged files

replaceafill:
See this https://github.com/archivematica/Issues/issues/651#issuecomment-566253022 with an explanation and diagram of the current situation (but ignore that failed idea at the end!).

This PR replaces the use of the `remove_v0.0` script with the `removeDirectories_v0.0` one when removing temporary bag directories in an uncompressed AIP. If one of the expected directories (logs, objects, thumbnails) is not there, the job won't fail anymore and will just print a warning instead (same logic for compressed AIPs).

The job titles (descriptions in the workflow) of the two jobs to delete temporary bag directories have been synchronized also.

Connected to https://github.com/archivematica/Issues/issues/651


ross-spencer (approved):
This is a pretty sound fix. Nice to see the microservice jobs configuration working so nicely in this instance.

Thanks @replaceafill.

Just an additional question, I notice `remove_v0.0` is now only used once in `Remove the processing directory`. This is one of the final tasks in the AIP creation in Ingest. Is it worth us taking this opportunity to consider its removal, replacing it with the remove directories script?

replaceafill:
Thanks @ross-spencer.

> Just an additional question, I notice `remove_v0.0` is now only used once in `Remove the processing directory`. This is one of the final tasks in the AIP creation in Ingest. Is it worth us taking this opportunity to consider its removal, replacing it with the remove directories script? 

I remember I noticed it (I even found https://github.com/artefactual/archivematica/issues/684 because of it) , but there was something (that I can't remember off the top of my head) that made me doubt. I'll check again and report back.

replaceafill:
@ross-spencer I tested back and everything seems fine replacing the script. I even found a string issue in the output of the `removeDirectories` script (extra `%s`) that I hadn't noticed. Could you please take a quick look at the new commits in case I missed something? Thanks!

src/MCPClient/lib/clientScripts/remove_directories.py
```diff
@@ -29,7 +29,7 @@ def call(jobs):
         with job.JobContext():
             for directory in job.args[1:]:
                 if os.path.isdir(directory):
-                    job.pyprint("Removing directory: %s", directory)
+                    job.pyprint("Removing directory:", directory)
```
ross-spencer, line 32: Good :eye: !!

src/MCPServer/lib/assets/workflow.json
```diff
@@ -10841,8 +10841,8 @@
             "config": {
                 "@manager": "linkTaskManagerDirectories",
                 "@model": "StandardTaskConfig",
-                "arguments": "-R \"%SIPDirectory%\"",
-                "execute": "remove_v0.0",
+                "arguments": "\"%SIPDirectory%\"",
+                "execute": "removeDirectories_v0.0",
```
ross-spencer, line 10845: Wunderbar!! :bear:

ross-spencer (approved):
Looks pretty great to me! Thanks @replaceafill (the deletions have been strong with this release!) :+1:

PR #1526, opened 2019-12-20
Create or reuse database transfers for watched directories

replaceafill:
For a bit of more context see my https://github.com/archivematica/Issues/issues/1032#issuecomment-567768052 explaining the scenarios where this issue appears.

When a watched directory containing `Transfer` units is polled, its
contents must have `models.Transfer` objects associated. Otherwise,
the package queue machinery fails.

This checks if the transfer exists based on the directory or file name and
reuses it, or creates it when needed.

It might be useful to check how [the previous logic](https://github.com/artefactual/archivematica/blob/f4cae3165434d92af6a0bb8fdddf2e0656fb2231/src/MCPServer/lib/unitTransfer.py#L46-L64) for `finding or creating` transfers worked in 1.10.1.

Connected to https://github.com/archivematica/Issues/issues/1032

### DIFF
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -722,15 +722,21 @@ class DIP(Package):
         path = path.replace(settings.SHARED_DIRECTORY, r"%sharedPath%", 1)
 
         sip_uuid = uuid_from_path(path)
+        created = True
         if sip_uuid:
-            sip_obj, _ = models.SIP.objects.get_or_create(
+            sip_obj, created = models.SIP.objects.get_or_create(
                 uuid=sip_uuid, defaults={"currentpath": path, "diruuids": False}
             )
         else:
             sip_obj = models.SIP.objects.create(
                 uuid=uuid4(), currentpath=path, diruuids=False
             )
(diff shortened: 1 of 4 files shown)


src/MCPServer/lib/server/packages.py
```diff
@@ -758,6 +758,29 @@ class Transfer(Package):
...
+            )
+            logger.info("Updating Transfer %s path: %s", transfer_obj.uuid, path)
+            if not created and transfer_obj.currentlocation != path:
+                transfer_obj.currentpath = path
+                transfer_obj.save()
+        else:
+            transfer_obj = models.Transfer.objects.create(
+                uuid=uuid4(), currentlocation=path
+            )
+            logger.info("Creating Transfer %s at %s", transfer_obj.uuid, path)
```
sevein: Would't the log entry be inaccurate if uuid_from_path returns a truthy value and `created=True`? Maybe a single log call after the if-else, having `created = True` declared earlier would fix it?

```python
logger.info(
    "Transfer %s %s (%s)",
    transfer_obj.uuid, "created" if created else "updated", path)
```

sevein (approved):
Thanks! LGTM.

PR #1531, opened 2019-12-25
Update pip-tools in dev.in

ross-spencer:
This updates pip-tools in dev.in and sees pip-compile run for each
of Archivematica's components requirements folders.

Connected to archivematica/issues#1039
Also connects itself to archivematica/Issues#1025


replaceafill (approved):
LGTM :+1:

PR #1534, opened 2020-01-03
Include UUID in SIP path created by dashboard

helenst:
This ensures the SIP UUID generated by the dashboard sip creation process and linked to the files in the SIP, is preserved when MCP Server picks it up - otherwise MCP Server will try to create its own SIP object with a new UUID and the files are not linked to it.

Fixes https://github.com/archivematica/Issues/issues/1031


ross-spencer:
Closing this one for now, as we have tackled it in https://github.com/artefactual/archivematica/pull/1572 which also puts back into place some additional logic lost during the MCP Server refactor.

Issue #955, opened 2018-03-06
Creation of archivematica user in postinstall scripts for CentOS 7 needs more robust checks


jpellman:
During an attempt at installing Archivematica 1.7 RC on a fresh CentOS 7 host, I noticed the following line pop up while I was installing _archivematica-storage-service_:

```
creating archivematica user
id: ‘archivematica’: no such user
Adding system user `archivematica' (UID 333) ...
```
The _no such user_ error piqued my curiosity so I checked what the pre-/post-install scripts were doing.

```
[centos@ip-172-31-17-204 tmp]$  sudo rpm -q --scripts archivematica-storage-service
postinstall scriptlet (using /bin/sh):

# Create archivematica user
userID=`id -u archivematica`
if [ "${userID}" != 333 ]; then
  useradd --uid 333 --user-group --home /var/lib/archivematica/ archivematica
fi
```

As-is, this user creation code is problematic for a variety of reasons.  For instance, if there is already an Archivematica account with username _archivematica_ and a UID other than 333 (say, 1001), this will create another _archivematica_ account with UID 333. If the _archivematica_ with 1001 has an entry in _/etc/shadow_ already with a hashed password, this same password should also work with the _archivematica_ with UID 333, since the join between _/etc/passwd_ and _/etc/shadow_ is based on username rather than UID. This means that both accounts named _archivematica_ could be logged into with the same password, which seems like a scenario we wouldn't want (not to mention that if either _archivematica_ account changed their password, it would also be changed for the other account as well). Also, this code doesn't check whether or not there is already a user out there with UID 333 on the host before running useradd.

The Fedora project has some suggestions for [better ways to do this](http://fedoraproject.org/wiki/Packaging%3aUsersAndGroups#Soft_static_allocation), but at a minimum the user creation code should ideally a) check that a user with UID 333 doesn't already exist and b) ensure that only one user account exists with the name _archivematica_.

(Leaving this untagged since it's out of scope for Phase 1.)

sevein:
One more PR needs to be merged before this can considered fixed!
https://github.com/artefactual-labs/am-packbuild/pull/161

sevein:
https://github.com/artefactual-labs/am-packbuild/pull/161 merged.

PR #1535, opened 2020-01-07
Re-enable package.py to update its file list

ross-spencer:
Before c1e7468 Archivematica package objects were able to be updated
with new file information. c1e7468 removed this asking the files
associated with a package to be returned from either the database
when objects were known to be in the database, or the file system
when no database objects existed.

Archivematica currently relies on being able to update a package
object's associated files with those on disk as well as in the
database where microservices are interacting e.g. creating or
updating file information in contrasting and potentially
inconsistent ways.

Black box testing results (**nb.** my test server was running slow for a few things):
```
6 features passed, 0 failed, 20 skipped
13 scenarios passed, 0 failed, 58 skipped
76 steps passed, 0 failed, 607 skipped, 0 undefined
Took 16m5.002s
```

Connected to archivematica/issues#955
Connected to archivematica/issues#973
Connected to archivematica/issues#1060

Once merged we will need to create a follow-up ticket for the suggestions in #955 [comment](https://github.com/archivematica/Issues/issues/955#issuecomment-571126909)

### DIFF
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -635,7 +635,8 @@ class Package(object):
     def files(
         self, filter_filename_start=None, filter_filename_end=None, filter_subdir=None
     ):
-        """Generator that yields all files associated with the package.
+        """Generator that yields all files associated with the package or that
+        should be associated with a package.
         """
         with auto_close_old_connections():
             queryset = self.base_queryset
@@ -651,37 +652,38 @@ class Package(object):
                 filter_path = "".join([self.REPLACEMENT_PATH_STRING, filter_subdir])
                 queryset = queryset.filter(currentlocation__startswith=filter_path)
 
-            # If we don't have any matching files in the database, we're in the process of
-            # generating file UUIDs, so walk the filesystem.
-            # TODO: restructure workflow to remove these cases.
-            if not queryset.exists():
-                start_path = self.current_path.encode(
-                    "utf-8"
-                )  # use bytes to return bytes
-                if filter_subdir:
-                    start_path = start_path + filter_subdir.encode("utf-8")
+            start_path = self.current_path.encode("utf-8")  # use bytes to return bytes
+            if filter_subdir:
+                start_path = start_path + filter_subdir.encode("utf-8")
 
-                for basedir, subdirs, files in scandir.walk(start_path):
-                    for file_name in files:
-                        if (
-                            filter_filename_start
-                            and not file_name.startswith(filter_filename_start)
-                        ) or (
-                            filter_filename_end
-                            and not file_name.endswith(filter_filename_end)
-                        ):
-                            continue
+            files_returned_already = set()
+            if queryset.exists():
+                for file_obj in queryset.iterator():
+                    file_obj_mapped = get_file_replacement_mapping(
+                        file_obj, self.current_path
+                    )
+                    if not os.path.exists(file_obj_mapped.get("%inputFile%")):
+                        continue
+                    files_returned_already.add(file_obj_mapped.get("%inputFile%"))
+                    yield file_obj_mapped
 
-                        file_path = os.path.join(basedir, file_name)
+            for basedir, subdirs, files in scandir.walk(start_path):
+                for file_name in files:
+                    if (
+                        filter_filename_start
+                        and not file_name.startswith(filter_filename_start)
+                    ) or (
+                        filter_filename_end
+                        and not file_name.endswith(filter_filename_end)
+                    ):
+                        continue
+                    file_path = os.path.join(basedir, file_name)
+                    if file_path not in files_returned_already:
                         yield {
                             r"%relativeLocation%": file_path,
-                            # empty uuid expects the string 'None'
                             r"%fileUUID%": "None",
                             r"%fileGrpUse%": "",
                         }
-            else:
-                for file_obj in queryset.iterator():
-                    yield get_file_replacement_mapping(file_obj, self.current_path)
 
     @auto_close_old_connections()
     def set_variable(self, key, value, chain_link_id):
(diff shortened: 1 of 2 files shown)


src/MCPServer/lib/server/packages.py
```diff
@@ -651,37 +652,50 @@ def files(
...
+                        file_obj, self.current_path
+                    )
+                    if file_obj_mapped.get("%inputFile%") in files_on_disk:
+                        package_file_list.append(file_obj_mapped)
+                        files_on_disk.remove(file_obj_mapped.get("%inputFile%"))
+
+            # We have objects on disk that aren't yet associated with the
+            # package but need to be processed, e.g. tasks created for to be
+            # associated with the current unit.
+            if files_on_disk:
```
replaceafill: The `for` loop below makes this `if` unnecessary, right?

src/MCPServer/lib/server/packages.py
```diff
@@ -651,37 +652,50 @@ def files(
...
-                            # empty uuid expects the string 'None'
                             r"%fileUUID%": "None",
                             r"%fileGrpUse%": "",
                         }
-            else:
-                for file_obj in queryset.iterator():
-                    yield get_file_replacement_mapping(file_obj, self.current_path)
+                    )
+
+            for file in package_file_list:
```
replaceafill: I've learned using `file` as a variable name can be a bit dangerous since it's also a type in Python.

src/MCPServer/tests/test_package.py
```diff
@@ -123,6 +123,107 @@ def test_transfer_get_or_create_from_db_path_with_uuid(tmp_path):
...
+    transfer_path.mkdir()
+    objects_path = transfer_path / "objects"
+    objects_path.mkdir()
+    first_file = objects_path / "file.txt"
+    first_file.touch()
+
+    kwargs = {
+        "uuid": uuid.uuid4(),
+        "currentlocation": "".join(
+            [transfer.REPLACEMENT_PATH_STRING, "/objects/file.txt"]
```
replaceafill: Would `os.path.join` fit here (and below) or maybe take advantage of the `Path` objects to avoid using `/` directly?

src/MCPServer/tests/test_package.py
```diff
@@ -123,6 +123,107 @@ def test_transfer_get_or_create_from_db_path_with_uuid(tmp_path):
...
+    # all objects will be returned from the database and we will have uuids.
+    for file_count, file_info in enumerate(transfer.files(None, None, "/objects"), 1):
+        if file_info.get("%fileUUID%") == "None":
+            print(file_info)
+            assert False, "Non-database entries returned from package.files()"
+    assert file_count == 3
+
+    # Clean up state and ensure test doesn't interfere with other transfers
+    # expected to be in the database, e.g. in test_queues.py.
+    models.File.objects.filter(transfer_id=str(transfer_uuid)).delete()
```
replaceafill, line 240: I thought the model instances you create in a test were removed (or the database reinitialized) at the end of the run? :astonished:

replaceafill (commented):
This looks good to me overall to me and I tested that both issues are fixed by it :+1: I left a few non significant comments.

My only concern is that we're building a list and returning each item in a generator now (losing some value from scandir and the queryset iterator), but I couldn't find an alternative approach to improve that situation.

@sevein has helped me with generators in the past, maybe he can see a better way on this?

src/MCPServer/tests/test_package.py
```diff
@@ -123,6 +123,107 @@ def test_transfer_get_or_create_from_db_path_with_uuid(tmp_path):
...
+    assert (
+        file_count == 3
+    ), "Database and file objects were not returned by the generator"
+    assert models.File.objects.filter(transfer_id=str(transfer_uuid)).count() == 3
+
+    # Now the database is updated, we will still have the same file count, but
+    # all objects will be returned from the database and we will have uuids.
+    for file_count, file_info in enumerate(transfer.files(None, None, "/objects"), 1):
+        if file_info.get("%fileUUID%") == "None":
+            print(file_info)
```
replaceafill: Forgot to add in the review: this `print` could be avoided by adding the `file_info` in the assert message of the next line.

src/MCPServer/tests/test_package.py
ross-spencer: Good point, `str(Path(...))` works okay here and achieves the same.

src/MCPServer/tests/test_package.py
ross-spencer: Oh, this is just testing detritus, but it's a good point, some debug info here is always useful, so have added it to the assert message.

src/MCPServer/tests/test_package.py
ross-spencer: It doesn't look like it `¯\_(ツ)_/¯` I feel this could be a larger audit task where tests aren't currently being properly isolated. We should play with tox/pytest some more to see what we can get out of it to help do this. 

Providing you're comfortable with this line, I'm happy to keep it. Otherwise, perhaps there's another way to tear down the test I don't know about?

ross-spencer:
@replaceafill I am happy with the generator changes (really nice change). And I've made the change to the test as well as added one final fairly arbitrary one that just adds a greater number of files to disk to observe the behavior for something that is an increment greater than `+=1`. I'm happy with the AM behavior too. Once you sign off I can rebase, add you as co-author and then we can get this tested on the bigger servers. Let me know though!

replaceafill (approved):
Thanks for the changes @ross-spencer! :+1:

src/MCPServer/tests/test_package.py
replaceafill: I'm OK with the explicit clean up :blush:

PR #1537, opened 2020-01-15
Ensure workflow endpoints are captured to reduce active package count as reliably as possible

ross-spencer:
This commit ensures that workflow endpoints are identified that when
reached should reduce the number of active packages in the
Archivematica workflow. This becomes critical when Archivematica
only recognizes one core available to be used on a machine. If the
count is 1 and the cores are 1 then no new packages can be generated.
We need to increment and decrement our internal counters as reliably
as possible.

Connected to archivematica/issues#1052
Connected to archivematica/issues#1055

See analysis [here](https://github.com/archivematica/Issues/issues/1055#issuecomment-574429288) for more information.


replaceafill (approved):
LGTM :+1: Tested the send to backlog workflow, rejected transfers and SIPs and hit https://github.com/archivematica/Issues/issues/1031 when I tried to create a SIP from the backlog but I guess that's expected and not part of this fix.

PR #1538, opened 2020-01-16
rebuild_es_aip_index_from_files: add tmp_dir, delete_only options

hakamine:
Add the following options to the management command
rebuild_elasticsearch_aip_index_from_files:

- --tmp-dir: to specify the temporary directory use for processing
  (instead of using the system default)

- --delete-only: to delete AIP data from the indexes without
  re-indexing

Connected to https://github.com/archivematica/Issues/issues/1064


sevein (approved):
Rebased and tested, :+1:

hakamine:
Thank you @sevein for the rebase and the fixes! 👍

PR #1539, opened 2020-01-20
Makes user email a required field in UserCreationForm

helrond:
Flips the `required` boolean on `email` in the `UserChangeForm` to `False` so that this field is never required.

Fixes archivematica/Issues#1067


sevein:
@helrond make sure that you use qa/1.x in the base branch!

PR #1546, opened 2020-01-25
Appraisal: fix directories draggability

replaceafill:
This PR ensures directories in the Backlog pane of the Appraisal tab can be dragged into Arrangement if they still contain any draggable children. It also handles files that have been already arranged and are dragged again.

Connected to https://github.com/archivematica/Issues/issues/822

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -29,6 +29,7 @@ import django.http
 import django.template.defaultfilters
 from django.utils.translation import ugettext as _, ungettext
 from django.utils import six
+from django.db import IntegrityError
 
 from components import helpers
 import components.filesystem_ajax.helpers as filesystem_ajax_helpers
@@ -617,7 +618,7 @@ def create_arrange_directories(paths):
         )
         for path in paths
     ]
-    models.SIPArrange.objects.bulk_create(arranges, BULK_CREATE_BATCH_SIZE)
+    _save_sip_arranges(arranges)
 
 
 def create_directory_within_arrange(request):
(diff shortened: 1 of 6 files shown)


src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -831,6 +832,23 @@ def _copy_files_to_arrange(
...
+    individually ignoring the existing ones.
+    """
+    try:
+        models.SIPArrange.objects.bulk_create(arranges, BULK_CREATE_BATCH_SIZE)
+    except IntegrityError:
+        for arrange in arranges:
+            try:
+                arrange.save()
+            except IntegrityError:
+                continue
```
sevein: Would it be correct to say that this is only needed because we don't have Django v2.2's `ignore_conflicts`?

Btw we have a `Directory.create_many` which I like because of the separation of duties, do you think we should create a similar `SIPArrange.create_many`?

src/dashboard/src/components/ingest/views.py
```diff
@@ -680,11 +685,7 @@ def transfer_backlog(request, ui):
     if ui == "legacy":
         response = return_list
     else:
-        if not request.GET.get("hidemetadatalogs"):
-            # if metadata and log file are shown in the appraisal tab
-            # directories should not be draggable if they contain
-            # non draggable children
-            adjust_non_draggable_nodes(return_list)
+        _adjust_directories_draggability(return_list)
```
sevein, line 688: `hidemetadatalogs` goes away, but don't we still need it? backlog.js uses it.

sevein (changes requested):
Looking good. I haven't tested yet but I have a couple of questions.

src/dashboard/src/components/filesystem_ajax/views.py
replaceafill: > Would it be correct to say that this is only needed because we don't have Django v2.2's `ignore_conflicts`?

That's correct. Cole pointed me to this strategy for `1.8` back when I first noticed that `bulk_create` could raise the `IntegrityError`s.

> Btw we have a `Directory.create_many` which I like because of the separation of duties, do you think we should create a similar `SIPArrange.create_many`?

Ah, nice! :+1:

src/dashboard/src/components/ingest/views.py
replaceafill: It's not going away. This is only adjusting directories that can be dragged without considering if logs are being displayed or not.

sevein (approved):
I've tested this and it's working as described. I've noticed that directories are not grayed out unless I reload the page, but that's likely not in the scope of this fix. Thanks!

replaceafill:
> I've noticed that directories are not grayed out unless I reload the page, but that's likely not in the scope of this fix.

Yeah. We need to teach both of those panes (Backlog and Arrangement) to refresh/reload and keep their folders open and selections highlighted.

Thank you @sevein!

PR #1549, opened 2020-01-30
ArchivesSpace: relation: use parent_obj display_string when there is no title

djpillen:
This PR checks if a `parent_obj` has a `title` when constructing the `relation` element to ensure that SIPs can be created with an intellectual arrangement consisting of `titles` and/or `dates`, not just `titles`. If a `title` exists, it is used as the parent_title and, if a `title` does not exist, then the `parent_obj` `display_string` is used instead.

Fixes archivematica/Issues#799
Replaces #1456

### DIFF
--- a/src/dashboard/src/components/access/views.py
+++ b/src/dashboard/src/components/access/views.py
@@ -464,7 +464,8 @@ def access_arrange_start_sip(client, request, mapping):
     parent = archival_object.get("parent", {}).get("ref")
     while parent:
         parent_obj = client.get_record(parent)
-        relation = [parent_obj["title"]] + relation
+        parent_title = parent_obj["title"] if parent_obj.get("title") else parent_obj["display_string"]
+        relation = [parent_title] + relation
         parent = parent_obj.get("parent", {}).get("ref")
     relation = " - ".join(relation)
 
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/access/views.py
```diff
@@ -464,7 +464,8 @@ def access_arrange_start_sip(client, request, mapping):
     parent = archival_object.get("parent", {}).get("ref")
     while parent:
         parent_obj = client.get_record(parent)
-        relation = [parent_obj["title"]] + relation
+        parent_title = parent_obj["title"] if parent_obj.get("title") else parent_obj["display_string"]
+        relation = [parent_title] + relation
```
ross-spencer, line 472: Thanks @djpillen this is neat an tidy. Only thing is we'll need you to run:

* https://github.com/psf/black

As you'll see CI currently fails on this snippet. 

You might also just be able to match the syntax which I think might be:
```
    parent_title = (
        parent_obj["title"] if parent_obj.get("title") else parent_obj["display_string"]
    )
    relation = [parent_title] + relation
```

But `black` is a good tool to have in your environment for future PR work and is in our `CONTRIBUTING.md`.

Once the CI passes I'll approve and we can merge.

ross-spencer (changes requested):
One requested change inline.

src/dashboard/src/components/access/views.py
djpillen: Just ran black and submitted a new commit.

ross-spencer (approved):
LGTM :+1:

ross-spencer:
Awesome! Merged and ready to go when `1.11` is shipped! Thanks for your contribution and patience @djpillen.

PR #1551, opened 2020-01-31
Create dmdSec for non-ascii filenames

replaceafill:
This PR handles non-ascii filenames in a `metadata/metadata.csv` import. The specific fix for that problem is in https://github.com/artefactual/archivematica/commit/9ec86dc8b7ac9dd2f63d18d29ccd3329ddd086e8

However, two fixes were required first (apologies for mixing them here):

1. Currently, if a transfer contains files with non-ascii names, the MCPServer will crash without starting the transfer. The problem was that the generator introduced in https://github.com/artefactual/archivematica/pull/1535 considered non-ascii filenames retrieved from the filesystem to be different from the same name retrieved from the database. This is the same problem in Helen's https://github.com/wellcometrust/platform/issues/4175#issuecomment-578136053. This has been fixed and tested in https://github.com/artefactual/archivematica/commit/b6c26eb987e83085f03541a1902c39a7d33a2c5d

1.  Then the helper that creates MCPClient jobs needed to be updated to accept the decoded non-ascii arguments. This has been fixed and tested in https://github.com/artefactual/archivematica/commit/deeb9fe9d1c9a00e597c2f0343a5762982700de5

Finally I realized `tox` is using the `settings.test` module from the dashboard for the MCPClient tests. I worked around this in https://github.com/artefactual/archivematica/commit/2acad6baaeb9dce13ac8e930555c4d196d478e76 by adding the MCPClient specific settings there. Maybe a way to handle this would be to create a separate tox environment only for the MCPClient.

Connected to https://github.com/archivematica/Issues/issues/1073

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -68,6 +68,7 @@ from django.conf import settings as django_settings
 import gearman
 
 from main.models import Task
+from archivematicaFunctions import unicodeToStr
 from databaseFunctions import getUTCDate, retryOnFailure
 
 from django.db import transaction
@@ -83,9 +84,9 @@ from job import Job
 logger = logging.getLogger("archivematica.mcp.client")
 
 replacement_dict = {
-    "%sharedPath%": django_settings.SHARED_DIRECTORY,
-    "%clientScriptsDirectory%": django_settings.CLIENT_SCRIPTS_DIRECTORY,
-    "%clientAssetsDirectory%": django_settings.CLIENT_ASSETS_DIRECTORY,
+    r"%sharedPath%": django_settings.SHARED_DIRECTORY,
+    r"%clientScriptsDirectory%": django_settings.CLIENT_SCRIPTS_DIRECTORY,
+    r"%clientAssetsDirectory%": django_settings.CLIENT_ASSETS_DIRECTORY,
 }
 
 
@@ -119,14 +120,14 @@ def handle_batch_task(gearman_job, supported_modules):
         replacements = (
             replacement_dict.items()
             + {
-                "%date%": utc_date.isoformat(),
-                "%taskUUID%": task_uuid,
-                "%jobCreatedDate%": task_data["createdDate"],
+                r"%date%": utc_date.isoformat(),
+                r"%taskUUID%": task_uuid,
+                r"%jobCreatedDate%": task_data["createdDate"],
             }.items()
         )
 
         for var, val in replacements:
-            arguments = arguments.replace(var, val)
+            arguments = arguments.replace(var, unicodeToStr(val))
 
         job = Job(
             gearman_job.task,
--- a/src/MCPClient/lib/clientScripts/create_mets_v2.py
+++ b/src/MCPClient/lib/clientScripts/create_mets_v2.py
@@ -66,8 +66,9 @@ from archivematicaCreateMETSTrim import getTrimFileAmdSec
 
 # archivematicaCommon
 from archivematicaFunctions import escape
-from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import normalizeNonDcElementName
+from archivematicaFunctions import strToUnicode
+from archivematicaFunctions import unicodeToStr
 from create_mets_dataverse_v2 import (
     create_dataverse_sip_dmdsec,
     create_dataverse_tabfile_dmdsec,
@@ -273,7 +274,7 @@ def createDMDIDsFromCSVMetadata(job, path, state):
     :param path: Path relative to the SIP to find CSV metadata on
     :return: Space-separated list of DMDIDs or empty string
     """
-    metadata = state.CSV_METADATA.get(path, {})
+    metadata = state.CSV_METADATA.get(unicodeToStr(path), {})
     dmdsecs = createDmdSecsFromCSVParsedMetadata(job, metadata, state)
     return " ".join([d.get("ID") for d in dmdsecs])
 
--- a/src/MCPClient/tests/test_archivematicaClient.py
+++ b/src/MCPClient/tests/test_archivematicaClient.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from archivematicaClient import handle_batch_task
+
+
+def test_handle_batch_task_replaces_non_ascii_arguments(mocker):
+    # We are only interested in verifying the string replacement logic
+    # for task arguments and mock the remaining functionality
+    mocker.patch("archivematicaClient.auto_close_db")
+    mocker.patch("archivematicaClient.Job")
+    mocker.patch("archivematicaClient.Task")
+    mocker.patch("archivematicaClient.retryOnFailure")
+    mocker.patch(
+        "cPickle.loads",
+        return_value={
+            "tasks": {
+                "some_task_uuid": {
+                    "uuid": "some_task_uuid",
+                    "arguments": "montréal %taskUUID% %jobCreatedDate%",
+                    "createdDate": "some montréal datetime",
+                    "wants_output": False,
+                }
+            }
+        },
+    )
+
+    # The mocked module will not have a `concurrent_instances` attribute
+    mocker.patch(
+        "importlib.import_module", return_value=mocker.MagicMock(spec=["call"])
+    )
+
+    # This is the only function that uses the arguments after the replacements
+    _parse_command_line = mocker.patch("archivematicaClient._parse_command_line")
+
+    # Mock the two parameters sent to handle_batch_task
+    gearman_job_mock = mocker.Mock()
+    supported_modules_mock = mocker.Mock(**{"get.side_effect": "some_module_name"})
+    handle_batch_task(gearman_job_mock, supported_modules_mock)
+
+    # Check that string replacement were successful
+    _parse_command_line.assert_called_once_with(
+        "montréal some_task_uuid some montréal datetime".encode("utf8")
+    )
--- a/src/MCPClient/tests/test_archivematicaClient.py
+++ b/src/MCPClient/tests/test_archivematicaClient.py
@@ -2,13 +2,15 @@
 
 from __future__ import unicode_literals
 
+import pytest
+
 from archivematicaClient import handle_batch_task
 
 
+@pytest.mark.django_db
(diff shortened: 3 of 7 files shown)


src/MCPServer/tests/test_package.py
```diff
@@ -240,6 +242,36 @@ def test_reload_file_list(tmp_path):
...
+        "currentlocation": "".join(
+            [transfer.REPLACEMENT_PATH_STRING, "/objects/montréal.txt"]
+        ),
+        "filegrpuse": "original",
+        "transfer_id": transfer_uuid,
+    }
+    models.File.objects.create(**kwargs)
+
+    # Assert only one file is returned
+    assert len(list(transfer.files(None, None, "/objects"))) == 1
```
ross-spencer: I think this might be a bit pedantic of me, because the test is really clear. What I tend to like to do here is also make an assertion that the filename/filepath is the one we're expecting and not `a.n.other file` from the model or that the input hasn't been mutated. This has the benefit of mitigating test-pollution too but provides a nice belt and braces guaranteeing our input/output.

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -83,9 +84,9 @@
 logger = logging.getLogger("archivematica.mcp.client")
 
 replacement_dict = {
-    "%sharedPath%": django_settings.SHARED_DIRECTORY,
-    "%clientScriptsDirectory%": django_settings.CLIENT_SCRIPTS_DIRECTORY,
-    "%clientAssetsDirectory%": django_settings.CLIENT_ASSETS_DIRECTORY,
+    r"%sharedPath%": django_settings.SHARED_DIRECTORY,
+    r"%clientScriptsDirectory%": django_settings.CLIENT_SCRIPTS_DIRECTORY,
+    r"%clientAssetsDirectory%": django_settings.CLIENT_ASSETS_DIRECTORY,
```
ross-spencer, line 89: I wonder if a specific use-case for these raw-strings can be encoded in the unit tests as well? I have been trying them out locally but the test works with or without, but I could be missing something. Tests have not been friendly today!

src/MCPClient/tests/test_archivematicaClient.py
```diff
@@ -0,0 +1,45 @@
...
+        "importlib.import_module", return_value=mocker.MagicMock(spec=["call"])
+    )
+
+    # This is the only function that uses the arguments after the replacements
+    _parse_command_line = mocker.patch("archivematicaClient._parse_command_line")
+
+    # Mock the two parameters sent to handle_batch_task
+    gearman_job_mock = mocker.Mock()
+    supported_modules_mock = mocker.Mock(**{"get.side_effect": "some_module_name"})
+    handle_batch_task(gearman_job_mock, supported_modules_mock)
```
ross-spencer, line 42: I think I understand why this might fail, but not why the inconsistency across environments. What I think I might be able to do is show that the mock on line `11` doesn't work as anticipated though. If you navigate to:

* https://github.com/artefactual/archivematica/blob/6946cb55158ca3fe33e2504c358ace79b1be1ed5/src/archivematicaCommon/lib/databaseFunctions.py#L39-L49 

And stick `assert False` before `close_old_connections()` you'll see that when the test fails, it will have called `handle_batch_task` and the `finally` clause is still hit. Whereas, I think we're trying to avoid any interaction with decorator entirely. We'd like the `handle_batch_task` function unwrapped to test it and not hit `close_old_connections()`. So I think that this is something we might look at fixing? But it's complicated. 

For completeness, this is what I see, when this happens is the transaction error as mentioned in Slack:
```
self = <django.db.backends.mysql.base.DatabaseWrapper object at 0x7f8309366d10>

    def get_autocommit(self):
        """
        Check the autocommit state.
        """
>       self.ensure_connection()
E       RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```

But only when I run all of the MCP Client tests. Not when I run this one test on its own. 

When I run tox in a clean virtual environment, I can see the finally being called, but I don't receive a tox/pytest warning at all. It seems to interact with Django differently? (NB. this is the same as if I run `sudo docker-compose run --workdir /src/MCPClient --rm --entrypoint=py.test archivematica-mcp-client -p no:cacheprovider -p no:warnings --reuse-db -v tests/test_archivematicaClient.py`) -- I've tried looking at `pytest` versions briefly here, and can see there may be platform differences. I don't know if these are affecting the output.
```
>       handle_batch_task(gearman_job_mock, supported_modules_mock)

src/MCPClient/tests/test_archivematicaClient.py:57: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<Mock id='140345534660560'>, <Mock id='140345534660496'>), kwargs = {}

    @wraps(f)
    def wrapper(*args, **kwargs):
        try:
            return f(*args, **kwargs)
        finally:
>           assert False
E           AssertionError
```

I don't know what to do to fix it. We can use the `@pytest.mark.django_db` decorator on the test function. It's technically not incorrect, even though we'd rather mock out any calls to the db. 

Otherwise, I've tried:

* https://stackoverflow.com/a/43213188
* Undecorated: https://pypi.org/project/undecorated/

And neither work.

And this person has some mega insights into decorators and doesn't seem pleased by them:

* https://hynek.me/articles/decorators/

In Python 3 we get a `handle_batch_task.__wrapped__` helper which would solve all of this and we wouldn't need the db mocks at all.

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -66,8 +66,9 @@
 
 # archivematicaCommon
 from archivematicaFunctions import escape
-from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import normalizeNonDcElementName
+from archivematicaFunctions import strToUnicode
+from archivematicaFunctions import unicodeToStr
```
ross-spencer, line 71: As the no. of imports starts to increase, we can adopt a different style, which for me is a little nicer to read, but ymmv. 
```
# archivematicaCommon
from archivematicaFunctions import (
    escape,
    normalizeNonDcElementName,
    strToUnicode,
    unicodeToStr,
)
```
But I can see the pattern adopted in the file, so it doesn't seem that important to add another line. Up to you.

src/MCPClient/tests/test_create_mets_v2.py
```diff
@@ -0,0 +1,35 @@
...
+        "create_mets_v2.createDmdSecsFromCSVParsedMetadata", return_value=[]
+    )
+    state_mock = mocker.Mock(
+        **{
+            "CSV_METADATA": {
+                "montréal".encode("utf8"): "montreal metadata",
+                "dvořák".encode("utf8"): "dvorak metadata",
+            }
+        }
+    )
```
ross-spencer, line 23: As ever your tests are great to read and I learn a lot from CR-ing this type of thing. Nice work.

ross-spencer (commented):
@replaceafill this is good, I've left comments. Specifically, just a bit more of a follow up to the database issues. The other comments are small observations that might spark your inspiration. Let me know about the decorator, or anything else, and we can work to finalize this PR/CR first thing tomorrow and get it merged.

src/MCPClient/lib/archivematicaClient.py
replaceafill: Sorry about this. This change is not related to the functionality I'm modifying or testing and maybe I should have not done it as part of this PR. I can revert it if you think it's convenient.

I just noticed that our replacement strings usage has not been consistent and wanted to clean a bit:

https://github.com/artefactual/archivematica/blob/29a0423a9bc1ab77ebe4ae7bfdc2bdde2099ef05/src/MCPServer/lib/server/packages.py#L100-L105

https://github.com/artefactual/archivematica/blob/29a0423a9bc1ab77ebe4ae7bfdc2bdde2099ef05/src/MCPServer/lib/server/packages.py#L232

https://github.com/artefactual/archivematica/blob/29a0423a9bc1ab77ebe4ae7bfdc2bdde2099ef05/src/MCPServer/lib/server/packages.py#L533-L534

https://github.com/artefactual/archivematica/blob/29a0423a9bc1ab77ebe4ae7bfdc2bdde2099ef05/src/MCPServer/lib/server/packages.py#L590

src/MCPClient/lib/clientScripts/create_mets_v2.py
replaceafill: Cool. Thanks for pointing it out. I'll pay more attention.

I love that we use `black` now because it removes a lot of "taste based" decisions from our workflow. We could include `isort` in our tool chain too to handle our imports. Check its [Multi line output modes](https://github.com/timothycrosley/isort#multi-line-output-modes)

sevein: Cole suggested the same! https://adr.archivematica.org/0004-isort-import-ordering.html

src/MCPClient/tests/test_archivematicaClient.py
replaceafill: Thanks for the investigation on the decorator!

At this point I'm tempted to start over the change to `lib.archivematicaClient.py`. As I wrote in the test I'm only interested in verifying that the string replacement logic doesn't break with non-ascii, and I could achieve the same by extracting only this block to a separate function:

https://github.com/artefactual/archivematica/blob/29a0423a9bc1ab77ebe4ae7bfdc2bdde2099ef05/src/MCPClient/lib/archivematicaClient.py#L128-L129

and just testing that. No decorator, no database, no nothing!

That would still leave us with the discrepancies between environments, etc etc etc. What do you think?

replaceafill: I went ahead and just marked the test to have interaction with the database and removed the mocking :disappointed:

src/MCPClient/lib/clientScripts/create_mets_v2.py
replaceafill: Thanks! I didn't know about that ADR :grimacing:

ross-spencer (approved):
Good to go! :+1:

PR #1553, opened 2020-02-04
Fix non-ascii transfer names

replaceafill:
For more context on what caused this see this https://github.com/archivematica/Issues/issues/1051#issuecomment-578623835

This PR handles non-ascii transfer names by using unicode paths in:

- The `packages.py` module of the MCPServer
- The `Verify mets_structmap.xml compliance` job
- The `Verify metadata directory checksums` job

Connected to https://github.com/archivematica/Issues/issues/1051

### DIFF
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -16,6 +16,7 @@ from django.conf import settings
 from django.utils import six
 
 import storageService as storage_service
+from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import unicodeToStr
 from fileOperations import get_extract_dir_name
 from main import models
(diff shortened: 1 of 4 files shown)


src/MCPServer/lib/server/packages.py
```diff
@@ -36,6 +37,15 @@
 
 StartingPoint = collections.namedtuple("StartingPoint", "watched_dir chain link")
 
+
+def _get_setting(name):
+    """Retrieve a Django setting decoded as a unicode string."""
+    try:
+        return strToUnicode(getattr(settings, name))
+    except:
```
ross-spencer: I don't know if I read this entirely correctly, but we could potentially receive a `UnicodeDecodeError` or `AttributeError` here, but for any exception, the function will return `None`. This will cause the joins below to output: `AttributeError: 'NoneType' object has no attribute 'endswith'`? If that's right, is there a reason we don't just let the failure occur here?

ross-spencer (approved):
Today is my day for asking 'just one question'.

image

But otherwise, LGTM :+1: and works!

src/MCPServer/lib/server/packages.py
replaceafill: Ah, good point! I remember thinking of returning `""` instead of `pass` but I think we could just let it fail :+1:

PR #1560, opened 2020-02-11
Fix .tar handling in get_extract_dir_name helper

replaceafill:
This PR fixes the `get_extract_dir_name` helper by returning an absolute path when the file path contains a `.tar` suffix.

Connected to https://github.com/archivematica/Issues/issues/1091


ross-spencer (approved):
Nice :+1: 

image

PR #1561, opened 2020-02-12
Remove base directory creation in Dspace transfers

replaceafill:
For context on the change see this https://github.com/archivematica/Issues/issues/1071#issuecomment-583586743

This PR removes the creation (and path check) of a base directory for a Dspace transfer zip file. By the point the `fileToFolder_v1.0` client script is run this directory has already been created by the MCPServer during transfer approval so this is not necessary anymore.

An associated AMAUAT PR for testing this is available here https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/157

Connected to https://github.com/archivematica/Issues/issues/1071

### DIFF
--- a/src/MCPClient/lib/clientScripts/file_to_folder.py
+++ b/src/MCPClient/lib/clientScripts/file_to_folder.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import os
-import sys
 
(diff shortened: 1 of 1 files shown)


ross-spencer (approved):
LGTM! :+1:

PR #1568, opened 2020-02-13
Reorder reingest dropdown options

sallain:
This was missed in the original PR, since I was using the processing
config for guidance. I've just switched the order of these two options
so that "Approve" is first, followed by "Reject", to be consistent with
all of the other dropdowns.

Connected to archivematica/Issues#819

### DIFF
--- a/src/MCPServer/lib/assets/workflow.json
+++ b/src/MCPServer/lib/assets/workflow.json
@@ -8345,8 +8345,8 @@
                 "@manager": "linkTaskManagerChoice",
                 "@model": "MicroServiceChainChoice",
                 "chain_choices": [
-                    "a6ed697e-6189-4b4e-9f80-29209abc7937",
-                    "2eae85d6-da2f-4f1c-8c33-3810b55e23aa"
+                    "2eae85d6-da2f-4f1c-8c33-3810b55e23aa",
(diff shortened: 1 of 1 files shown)


replaceafill (approved):
LGTM :+1: 

I also noticed `a6ed697e-6189-4b4e-9f80-29209abc7937` appears in https://github.com/artefactual/archivematica/blob/dddf0a6a74733463209bc5614a136b60c3f418a9/src/MCPServer/lib/assets/workflow.json#L8348-L8349 but I'm not sure how the user sees it?

replaceafill:
Ah! Maybe it's related to the watched directories functionality :man_shrugging: :

https://github.com/artefactual/archivematica/blob/dddf0a6a74733463209bc5614a136b60c3f418a9/src/MCPServer/lib/assets/workflow.json#L12268-L12273

sallain:
Good catch @replaceafill. That looks to me like a SIP approval step - it might be the one when you create a SIP from the Appraisal tab. I'll check!

sallain:
@replaceafill It's the create-SIP-from-backlog approval dropdown, and it's now in yes/no order :smile:

replaceafill (approved):
Gracias!

PR #1570, opened 2020-02-14
Use unit variable for copying processing config

replaceafill:
This PR saves the processing configuration name of the transfer in a unit variable (also accessible from the `Transfer` model) in the database and retrieve later to copy it to the processing directory. More context for the change can be found in this https://github.com/archivematica/Issues/issues/1092#issuecomment-585313063 

Additionally the changes made in https://github.com/artefactual/archivematica/pull/1561 have been reverted.

Connected to https://github.com/archivematica/Issues/issues/1092
Connected to https://github.com/archivematica/Issues/issues/1063

### DIFF
--- a/src/MCPClient/lib/clientScripts/file_to_folder.py
+++ b/src/MCPClient/lib/clientScripts/file_to_folder.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import sys
 
 import django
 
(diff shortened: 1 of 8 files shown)


ross-spencer (approved):
Excellent fix, and works well. Nice work @replaceafill :+1: 

I wonder, given you've had to delve into this so deeply, have you any thoughts on any additional documentation that could be generated to help future devs? One of the more obscure parts, as an example, is the fact we rely on no-clobber in `copy_v0.0` to try to write a copy of a shared processing config but only if one doesn't already exist in a transfer. Which you logically get to after working through the other parts of the process of supplying a custom processing configuration. It's not a self-documenting pattern by any means. I don't know if we have a platform, or where this would fit. But thought I'd ask. Happy to follow up in Slack.

replaceafill:
> have you any thoughts on any additional documentation that could be generated to help future devs?

A developer handbook generated from the code base :pray: I think in general `workflow.json` could use its own doc.

> One of the more obscure parts, as an example, is the fact we rely on no-clobber in `copy_v0.0` to try to write a copy of a shared processing config but only if one doesn't already exist in a transfer.

I first realized about the flag this https://github.com/archivematica/Issues/issues/771#issuecomment-506390516 from Helen. Then I can track it back to [2011](https://github.com/artefactual/archivematica-history/commit/310b213d6#diff-dae8ffd0b9f8feddea324f9e409ee080R1809). I have to admit that I first removed it trying to get rid of it :blush: but then realized it was necessary for the logic of the job.

ross-spencer:
> I have to admit that I first removed it trying to get rid of it blush but then realized it was necessary for the logic of the job. 

Easily done! This PR and related issues since last year have been a useful tool (for me at least) in demonstrating and reminding me how distributed the logic can be across AM sometimes.

image

PR #1571, opened 2020-02-14
Remove METS mappings declaration from `aipfiles`

jraddaoui:
After the changes in 541cfef, the `aipfiles_mets_mapping.json` is no
longer a representation of the final mapping. Using `lxml.etree`
modifies the namespaces included on the field names. E.g.:
`METS.ns0:xmlData_dict_list` is now `METS.mets:xmlData_dict_list`.

Considering the speed improvement using `lxml.etree` and that we don't
have a strict mapping yet and don't target directly any of those fields
(at least internally), we'll leave the mapping changes but we'll remove
the old declaration to avoid confusion.

Refs #1199.


jraddaoui:
See https://github.com/artefactual/archivematica/issues/1199#issuecomment-586312684.

ross-spencer (approved):
I left a [comment](https://github.com/artefactual/archivematica/issues/1199#issuecomment-589464893) about this change in context which looks good and functioned as anticipated when I gave it a whirl. Nice addition of function comments. LGTM :+1:

Issue #1108, opened 2018-05-28
Problem: Assign file UUIDs to objects fails with win_1252, big5, and shiftjis encodings


ross-spencer:
In `qa/1.x` we are seeing the following failures at this stage in the transfer process for standard files:

**win_1252:**
```
archivematicaAssignFileUUID.py: INFO      2018-05-28 11:04:37,594  archivematica.mcp.client.assignFileUUID:main:142:  Generated UUID for this file: ce0d9432-3cfe-4580-8233-f871c1391da8.
archivematicaAssignFileUUID.py: WARNING   2018-05-28 11:04:37,598  py.warnings:_warning_check:155:  /usr/local/lib/python2.7/dist-packages/django/db/backends/mysql/base.py:124: Warning: (1300L, u"Invalid utf8 character string: 'F87374'")
  return self.cursor.execute(query, args)

Traceback (most recent call last):
  File "/src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py", line 174, in <module>
    sys.exit(main(**vars(args)))
  File "/src/MCPClient/lib/clientScripts/archivematicaAssignFileUUID.py", line 143, in main
    addFileToTransfer(file_path_relative_to_sip, file_uuid, transfer_uuid, event_uuid, date, use=use, sourceType=event_type)
  File "/src/archivematicaCommon/lib/fileOperations.py", line 72, in addFileToTransfer
    eventOutcomeDetailNote="")
  File "/src/archivematicaCommon/lib/databaseFunctions.py", line 172, in insertIntoEvents
    agents = getAMAgentsForFile(fileUUID)
  File "/src/archivematicaCommon/lib/databaseFunctions.py", line 119, in getAMAgentsForFile
    f = File.objects.get(uuid=fileUUID)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 328, in get
    num = len(clone)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 144, in __len__
    self._fetch_all()
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 965, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/query.py", line 254, in iterator
[...]

PR #1576, opened 2020-02-22
Update error path for "Check transfer directory for objects"

jorikvankemenade:
This PR solves the two problems that are discussed in [#1108](https://github.com/archivematica/Issues/issues/1108):
- e51ca85 removes the "Check transfer directory for objects" as a terminal link. This should resolve the reported `RuntimeError` from showing up in the logs. 
- 2745bf8 configures links the "Check transfer directory for objects" to the failure path for the Transfer. This means that in case of failure an e-mail is sent to the user and the transfer is cleaned. Besides that, I have changed the error codes and code style. For a full explanation please check the commit message.

Using the Images transfer from the samples I have verified that I can Transfer/Ingest without the `RuntimeError` showing up in the logs.

image

When I run the same transfer, but make sure to clear the objects directory before hitting "Check transfer directory for objects", I get a failed transfer and a failure report:

image

image

### DIFF
--- a/src/MCPClient/lib/clientScripts/check_transfer_directory_for_objects.py
+++ b/src/MCPClient/lib/clientScripts/check_transfer_directory_for_objects.py
@@ -25,17 +25,17 @@ import os
 import scandir
 
 
-exitInidcatingThereAreObjects = 179
-
-
 def call(jobs):
+    """
+    Check the given directory and it's subdirectories for files.
+    Returns job status 0 if there are files.
+    Returns job status 1 if the directories are empty.
+    """
     for job in jobs:
         with job.JobContext():
-            objectsDir = job.args[1]
-            os.path.isdir(objectsDir)
-            ret = 0
-            for dirs, subDirs, files in scandir.walk(objectsDir):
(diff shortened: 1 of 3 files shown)


sevein:
Thanks. Listing `1` feels redundant, could you try the following?

```json
{
    "exit_codes": {
        "0": {
            "job_status": "Completed successfully",
            "link_id": "b04e9232-2aea-49fc-9560-27349c8eba4e"
        }
    },
    "fallback_job_status": "Failed",
    "fallback_link_id": "61c316a6-0a50-4f65-8767-1f44b1eeb6dd"
}
```

This could be a simplification of the script:

```python
def call(jobs):
    for job in jobs:
        with job.JobContext():
            objects_dir = job.args[1]
            for _, _, files in scandir.walk(objects_dir):
                if files:
                    return
            job.set_status(1)
```

jorikvankemenade:
I agree that listing 1 is a bit redundant. I have removed it and it works without problems:

image

image

image

image

I have also applied your simplification and added a docstring to the script.

sevein (approved):
Thanks!

Issue #1055, opened 2018-04-16
Bug: Reingest fails if AIP originated from DSpace


sromkey:
If the DSpace workflow is used to create AIPs, when doing a full reingest on those AIPs they will fail during Generate METS.xml document. Full reingest should be supported for AIPs that originated from DSpace in the same way other AIPs can be reingested.

sromkey:
Duplicates #1083

PR #1577, opened 2020-02-23
Add terminal links to the Archivematica workflow

jorikvankemenade:
This PR includes three commits. We might want to squash these into a single commit before merging, but I think that at this point it makes sense to separate them. 

2e1f000 addresses the comment I made [here](https://github.com/archivematica/Issues/issues/1055#issuecomment-589644118). My jobs were not scheduled correctly because the `queues` class missed one of the links that indicate that a package is done. This sparked the discussion in [#1055](https://github.com/archivematica/Issues/issues/1055) about replacing the `PACKAGE_COMPLETED_LINK_IDS` altogether. This also removes `27fd07e-d3ed-4767-96a5-44a2251c6d0a` as a terminal link. Having this link marked as terminal causes some of the problems as discussed in [#1010](https://github.com/archivematica/Issues/issues/1010). 

efcc259 introduces the discussed solution. This commit basically adds an `is_terminal` function to each link in the workflow. This checks for the `end: true` value that can be added to the `workflow.json`.

Finally, bb6d3b4 applies this concept to the currently existing queueing logic. I have removed all the `PACKAGE_COMPLETED_LINK_IDS` and added the `end` field to each of these links.

I have tested the changes by running a few transfers, including some failing ones. I set `ARCHIVEMATICA_MCPSERVER_CONCURRENT_PACKAGES` to 1 to see if I could get the queueing of packages to block. So far I haven't, but maybe you guys have a more efficient way to test this.

Connects to https://github.com/archivematica/Issues/issues/1108.

### DIFF
--- a/src/MCPServer/lib/assets/workflow-schema-v1.1.json
+++ b/src/MCPServer/lib/assets/workflow-schema-v1.1.json
@@ -1,5 +1,5 @@
 {
-  "$id": "https://www.archivematica.org/labs/workflow/schema/v1.json",
+  "$id": "https://www.archivematica.org/labs/workflow/schema/v1.1.json",
   "$schema": "http://json-schema.org/draft-06/schema#",
   "title": "Archivematica Workflow JSON Schema",
   "description": "In-progress JSON Schema document for Archivematica JSON-encoded workflows.",
@@ -97,6 +97,9 @@
         },
         "group": {
           "$ref": "#/definitions/translations"
+        },
+        "end": {
+          "type": "boolean"
         }
       },
       "additionalProperties": false,
--- a/src/MCPServer/lib/server/queues.py
+++ b/src/MCPServer/lib/server/queues.py
@@ -57,33 +57,6 @@ class PackageQueue(object):
        continues until the workflow chain ends.
     """
 
-    # TODO: make this better by signifying the end of a package in the workflow
-    PACKAGE_COMPLETED_LINK_IDS = set(
-        [
-            # Move to SIP creation directory for completed transfers
-            "d27fd07e-d3ed-4767-96a5-44a2251c6d0a",
-            # Move to SIP Creation
-            "39a128e3-c35d-40b7-9363-87f75091e1ff",
-            # AIP completed
-            "d5a2ef60-a757-483c-a71a-ccbffe6b80da",
-            # Move SIP to failed directory
-            "828528c2-2eb9-4514-b5ca-dfd1f7cb5b8c",
-            # Move Transfer to failed directory
-            "377f8ebb-7989-4a68-9361-658079ff8138",
-            # Move transfer to backlog
-            "abd6d60c-d50f-4660-a189-ac1b34fafe85",
-            # Move to the rejected directory
-            "0d7f5dc2-b9af-43bf-b698-10fdcc5b014d",
-            "333532b9-b7c2-4478-9415-28a3056d58df",
-            "3467d003-1603-49e3-b085-e58aa693afed",
-            # Failed compliance. See output in dashboard. SIP moved back to SIPsUnderConstruction
-            "f025f58c-d48c-4ba1-8904-a56d2a67b42f",
-            # Failed compliance. See output in dashboard. Transfer moved back to activeTransfers
-            "61af079f-46a2-48ff-9b8a-0c78ba3a456d",
-            # Failed compliance
-            'bbfbecde-370c-4e26-8087-cfa751e72e6a',
-        ]
-    )
     # An arbitrary, large value, so we don't accept infinite packages.
     MAX_QUEUED_PACKAGES = 4096
 
@@ -193,9 +166,12 @@ class PackageQueue(object):
     def process_one_job(self, timeout=None):
         """Process a single job, if one is queued before `timeout`.
 
-        Only called by `work` and unit tests.
+        Jobs are submitted to our thread pool. Once the job has done processing
+        it will return the next link in the chain, which will be scheduled. Some
+        links are terminal, these links indicate the end of the chain for a
+        particalar package. If such a link is encountered the package is
+        deactivated and the next package is scheduled.
         """
-        # block until a job is waiting
         try:
             job = self.job_queue.get(timeout=timeout)
         except Queue.Empty:
@@ -207,7 +183,7 @@ class PackageQueue(object):
         result = self.executor.submit(job.run)
         result.add_done_callback(self._job_completed_callback)
 
-        if job.link.id in self.PACKAGE_COMPLETED_LINK_IDS:
+        if job.link.is_terminal():
             package_done_callback = functools.partial(
                 self._package_completed_callback, job.package
             )
@@ -221,34 +197,33 @@ class PackageQueue(object):
         self.shutdown_event.set()
 
     def _package_completed_callback(self, package, future):
-        """Called by an Executor on completion the last job in a package.
+        """Mark the package as inactive and schedule a new package.
 
-        The result of the future will always be None.
+        It is assumed that a package is only complete when a terminal link is
+        hit. When we hit a terminal link the future should not contain the next
+        job in the chain. If the future is a new job we raise an exception.
         """
         if future.result() is not None:
             raise RuntimeError(
                 "Unexpectedly received another job on package completion. "
-                "If the workflow has been modified, please update "
-                "PACKAGE_COMPLETED_LINK_IDS."
+                "Please verify the value of `end` in the workflow. "
             )
 
         self.deactivate_package(package)
         self.queue_next_job()
 
     def _job_completed_callback(self, future):
-        """Called by an Executor on completion of `job.run`.
+        """Schedule the next job in the chain.
 
-        The argument is a future, the result of which should be the next job
-        in the chain.
+        Retrieves the next job from the result from the previous job.
+        If there is no next_job return, otherwhise schedule the new job.
         """
         metrics.active_jobs_gauge.dec()
         next_job = future.result()
 
         if not next_job:
             return
-
-        # Special case for decision job here.
-        if isinstance(next_job, DecisionJob) and next_job.awaiting_decision:
+        elif isinstance(next_job, DecisionJob) and next_job.awaiting_decision:
             self.await_decision(next_job)
         else:
             self.schedule_job(next_job)
--- a/src/MCPServer/lib/server/queues.py
+++ b/src/MCPServer/lib/server/queues.py
@@ -183,7 +183,7 @@ class PackageQueue(object):
         result = self.executor.submit(job.run)
         result.add_done_callback(self._job_completed_callback)
 
-        if job.link.is_terminal():
+        if job.link.is_terminal:
             package_done_callback = functools.partial(
                 self._package_completed_callback, job.package
             )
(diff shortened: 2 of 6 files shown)


src/MCPServer/lib/server/workflow.py
```diff
@@ -194,6 +194,17 @@ def get_status_id(self, code):
             status_id = self._src["fallback_job_status"]
         return status_id
 
+    def is_terminal(self):
```
cole: Nit: I'd use a property here and return `bool(self._src["end"])`. But it doesn't matter much.

cole (approved):
Much better! 🎉

src/MCPServer/lib/server/workflow.py
jorikvankemenade: What do you mean exactly? If I use something like this:

```python
    @property
    def is_terminal(self):
        """Check if the link is indicated as a terminal link.

        Returns the value of the end property.
        If the value does not exist the link is assumed to be non-terminal.
        """
        return self._src["end"]
```

It won't work, because `end` is not a required field, so you will get key errors. I can annotate the entire function with property, but what would be the advantage of that? I am not very familiar with the property concept.

src/MCPServer/lib/server/queues.py
```diff
@@ -221,34 +197,33 @@ def stop(self):
...
         """
         metrics.active_jobs_gauge.dec()
         next_job = future.result()
 
         if not next_job:
             return
-
-        # Special case for decision job here.
-        if isinstance(next_job, DecisionJob) and next_job.awaiting_decision:
+        elif isinstance(next_job, DecisionJob) and next_job.awaiting_decision:
```
sevein, line 229: I think that the previous code style was calculated. The early returns gets that falsy job case out of your way to let the function body continue dealing with other cases. There is not right or wrong here though, but `elif` implies previous context that we don't need anymore?

src/MCPServer/lib/assets/workflow-schema-v1.1.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://www.archivematica.org/labs/workflow/schema/v1.json",
+  "$id": "https://www.archivematica.org/labs/workflow/schema/v1.1.json",
```
sevein: I would argue that we shouldn't remove the previous version. But let's just keep calling it v1 for now and avoid all the extra work around versioning until we promote this as a consumable language.

src/MCPServer/lib/server/workflow.py
sevein: I think it's just a common idiom for getters in Python? But you're right that you'd still need the exception handler, unless you want to perform the computation in the constructor.

src/MCPServer/lib/server/queues.py
```diff
@@ -221,34 +197,33 @@ def stop(self):
...
 
     def _package_completed_callback(self, package, future):
-        """Called by an Executor on completion the last job in a package.
+        """Mark the package as inactive and schedule a new package.
 
-        The result of the future will always be None.
+        It is assumed that a package is only complete when a terminal link is
+        hit. When we hit a terminal link the future should not contain the next
+        job in the chain. If the future is a new job we raise an exception.
         """
```
sevein, line 207: I like what you're doing here but I thought the bits about the execution context were also useful (*Called by an Executor...*). Didn't you find that useful when you read the method for the first time?

> If the future is a new job we raise an exception.

You've provided some useful context but I think we also have to draw a line where we stop commenting on *what* the code is doing because the code is self-describing. This idea I'm trying to communicate is better described [here](https://google.github.io/eng-practices/review/reviewer/looking-for.html#comments).

sevein (changes requested):
~`"bbfbecde-370c-4e26-8087-cfa751e72e6a"` is now a terminal link but `"d27fd07e-d3ed-4767-96a5-44a2251c6d0a"` and `"032cdc54-0b9b-4caf-86e8-10d63efbaec0"` aren't. Do you mind explaning why?~

Sorry, I realise you've already explained it, thanks!

src/MCPServer/lib/server/queues.py
jorikvankemenade: I see what you mean. The reason I changed it was because to me it all belongs to the same context, I see it as a switch on `next_job`. If there is no `next_job` we do nothing. In case we have a decision job we await the decision. In all other cases, we schedule the next job. 

The separation made me wonder why the early return was a special case. To me, it makes more sense now but if you prefer the old way I am more than happy to change it back.

src/MCPServer/lib/assets/workflow-schema-v1.1.json
jorikvankemenade: Yeah, I wasn't sure if versioning it was necessary or wanted. I'll revert the versioning changes.

src/MCPServer/lib/server/workflow.py
jorikvankemenade: > I think it's just a common idiom for getters in Python?

I believe so too, it should return a read-only version of the value so that is nice. I have changed it to a property and replaced the `KeyError` part with a `get` method on the dict.

src/MCPServer/lib/server/queues.py
jorikvankemenade: >  I thought the bits about the execution context were also useful

Originally I saw this as commenting about an implementation detail of the Python futures. However, looking at it again I see that this detail is actually coupled to how you can use the function. I.e. you cannot use the function to complete a package without using futures etc. So I agree that this adds some additional context that is relevant to the developer.

> This idea I'm trying to communicate is better described here.

I fully agree, missed the mark there. It is trivial to see in the code that we do this, and we already explained why - so that part was definitely obsolete.

jorikvankemenade:
@sevein I have added the requested changes in two new commits. If they are okay with you it is probably best to squash the whole PR into a single commit? Or I can squash the two new commits in the first three to keep some separation. Whatever you prefer.

src/MCPServer/lib/server/workflow.py
cole: I meant:

```python
    @property
    def is_terminal(self):
        try:
            return bool(self._src["end"])
        except KeyError:
            return False
```

It sounds like you have this, which is more concise:
```python
    @property
    def is_terminal(self):
        return bool(self._src.get("end"))
```

This is a nitpick and a purely stylistic change. The original approach is also fine! In fact I think there's an argument to be made that it's more consistent with the rest of the module.

Generally I like `@property` for methods without arguments that return a single value and don't have external side effects. Calling `bool` ensures the method always returns a boolean; that's more a habit from using [type hints](https://docs.python.org/3/library/typing.html) than anything else; there's not a huge amount of benefit here.

sevein (changes requested):
@jorikvankemenade, I think I've found a couple of extra links that should terminate. Could you please confirm?

- `16415d2f` (Create SIPs from TRIM transfer containers) which is similar to "Move transfer to backlog" or "Create SIP from transfer objects" but when dealing with TRIM transfers,
- `f8ee488b` (Handle unstored DIP) should also be added? Otherwise the processing of DIPs won't finalize if the user chooses not to store DIPs.

On the other hand, I don't understand why you chose `bbfbecde`. Wouldn't make more sense to add `89071669` which sits right behind `/activeTranfsers/standardTransfer`?

I've updated amflow so it highlights terminal links in orange. Check it out if that makes things easier for you.

src/MCPServer/lib/server/queues.py
```diff
@@ -221,34 +197,35 @@ def stop(self):
         self.shutdown_event.set()
 
     def _package_completed_callback(self, package, future):
-        """Called by an Executor on completion the last job in a package.
+        """Called by an executor on completion of a package. Marks the package
+        as inactive and schedules a new package.
```
sevein, line 201: One nit about your docstrings in this PR:

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings
> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).

jorikvankemenade: Check, this was a bit careless of me. Fixed it!

jorikvankemenade:
>  I think I've found a couple of extra links that should terminate

I would imagine there can be a couple of them. I just refactored the existing known links and added the one I was missing in my workflow: standard transfers. So I think odds are more will be found when testing Archivematica, especially with `concurrent_packages = 1`. The advantage of this approach is that rectifying these bugs should be simple.

> I've updated amflow so it highlights terminal links in orange. 

This is actually pretty helpful. One thing which baffles me is that I have 11 terminal links in the file in this PR. And only 7 show up in the amflow. Are certain error paths not drawn?

> `16415d2f` (Create SIPs from TRIM transfer containers) which is similar to "Move transfer to backlog" or "Create SIP from transfer objects" but when dealing with TRIM transfers,

I agree that this is a similar situation.

> `f8ee488b` (Handle unstored DIP) should also be added? Otherwise the processing of DIPs won't finalize if the user chooses not to store DIPs.

This is the end of the workflow for DIP, so yes it is terminal. Actually, `653b134f` should be terminal as well. Another thing that I notice when checking the DIP section is that there is no failure path for DIPS. Is this on purpose? And if so, we should still make sure that we can deactivate the package on failure.

> On the other hand, I don't understand why you chose `bbfbecde`. Wouldn't it make more sense to add `89071669` which sits right behind `/activeTranfsers/standardTransfer`?

I see why you would think this. I just used the `workflow.json` to select this transfer. Because of the watched directory structure, I missed that this was not the end of a workflow.  Looking at the amflow I think that your suggestion makes more sense. 

One general concern that dawned on me when looking at this is that for many links there is no error path. I think this means that there are some points left where the workflow of a package can stop, without the package being deactivated. I think we have three directions here:
1. Role with the punches. Set all test machines to `concurrent_packages = 1` and add terminal links each time a failure is encountered.
1. Manually verify the entire chain and set failure states for all points in the workflow. This is a tedious error-prone job, but should result in a more reliable workflow.
1. Think of a fallback mechanism that deactivates packages on failure. 

What do you think @sevein? What is your preference?

sevein:
> This is actually pretty helpful. One thing which baffles me is that I have 11 terminal links in the file in this PR. And only 7 show up in the amflow. Are certain error paths not drawn?

amflow has to ignore some very common links because otherwise graphviz takes forever to render the graph (30s+). I should add an option to force the rendering of all links for cases like this, will look into that. This is the [list](https://github.com/artefactual-labs/amflow/blob/0113346fc47ddd911e433696e9aeec40bd2c9f5d/internal/graph/dotviz.go#L17-L29).

> This is the end of the workflow for DIP, so yes it is terminal. Actually, 653b134f should be terminal as well. Another thing that I notice when checking the DIP section is that there is no failure path for DIPS. Is this on purpose? And if so, we should still make sure that we can deactivate the package on failure.

You're right. I'll take a look.

> What do you think @sevein? What is your preference?

It's relatively easy to introduce some sanity checks in the existing amflow validation function which uses an actual graph internally. That graph does not omit links, it's only the graphviz encoder that ignores them.

I think I'll merge this now since it's become a blocker for us! As long as I have more feedback on missing links I'll submit a PR and will CC you! Thank you so much for the help, Jorik.

sevein (approved):
👍

PR #1475, opened 2019-08-29
display dropdown modal is options greater then 10

ghost:
Changes to display the modal whit dropdown if the option list has over 10 items. 
This issue is already mentioned here: https://github.com/archivematica/Issues/issues/850

PR #850, opened 2017-12-15
Prevent recursion limit error during normative structmap creation

jrwdunham:
Prevent hitting the recursion limit using a for loop during METS normative structmap creation.

Replaces https://github.com/artefactual/archivematica/pull/825.

PR #1578, opened 2020-02-26
Fix enhanced decision point selector

replaceafill:
This PR fixes the modal dialog shown when a decision point has more than 10 options.

I started reviewing https://github.com/artefactual/archivematica/pull/1475 but found the same [issues Ross reported](https://github.com/artefactual/archivematica/pull/1475#pullrequestreview-316284704).

I consider this fix a bit more robust because:

1. It doesn't require changes to the style sheets: to make the `Select2` dropdown in the modal responsive I followed this suggestion in its documentation (see the `**Responsive Design - Percent Width** section in http://select2.github.io/select2/):

> The best way to ensure that Select2 is using a percent based width is to inline the style declaration into the tag.

2. It completes the HTML structure of the modal element as explained in the [Bootstrap documentation](https://getbootstrap.com/docs/3.3/javascript/#modals). I also set the `backdrop` option of the modal as [`static`](https://getbootstrap.com/docs/3.3/javascript/#modals-options) to have the user destroy the modal container either trhough the `x` icon or the `Cancel` button. 

The modal seems to work similarly in Firefox and Chrome and it looks like this:

Captura de pantalla de 2020-02-25 23-27-51

**Note**: I also found a limitation in the whole approach that we'll need to revise in the future: once Angular repaints the UI, triggering the `change` event of the original decision point dropdown doesn't work anymore. I commented the code accordingly.

Connected to https://github.com/archivematica/Issues/issues/850

### DIFF
--- a/src/dashboard/src/media/js/jobs.js
+++ b/src/dashboard/src/media/js/jobs.js
@@ -567,18 +567,27 @@ var BaseJobView = Backbone.View.extend({
       return UNKNOWN_COLOR;
     },
 
+  destroyEnhacedActionSelectContainer: function(statusObject)
+    {
+      $('.modal-backdrop.in').remove();
+      $('body').removeClass('modal-open');
+      $('#big-choice-select-modal').remove();
+      statusObject.bigSelectShowing = false;
+    },
+
   // augment dashboard action select by using a pop-up with an enhanced select widget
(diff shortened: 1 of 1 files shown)


sevein (approved):
This seems to be a good step forward addressing the problem described in #850. But I agree we still need the modal to survive to updates and embrace the fact that other users may be making the decision.

PR #1580, opened 2020-03-02
Enable zipped package transfer type and better management of types in future

ross-spencer:
This PR introduces unit testing for the majority of components relying on understanding a "transfer type" supported by Archivematica. 

After introducing unit tests, information about transfer types is centralized in Archivematica Common to enable sharing across the Dashboard and MCP Server, this will promote the ease of changing this information in future, e.g. adding information about a transfer type that can be easily cascaded throughout Archivematica.

Redundancy is reduced as much as possible across the code-base, for example, shared directory generation, is now done from the centralized "knowledge" of transfer types offered by Archivematica.

Finally, there is a proposed validation pathway introduced for the `api/v2beta/package` endpoint so that unit testing is better supported in future.

**Review notes:** _for those reviewing, while this is a pretty big PR, I have attempted to write a test for any change I have tried to make first, to avoid regressions, and then I have implemented a change, and so there should be a pattern of commits, "test" -> "implementation" -> "test" -> etc. etc. Hopefully this helps ease any reviewing burden._ 

Connected to archivematica/issues#1126
Related to archivematica/issues#682 

Acceptance tests supporting `zipped package` transfers have been added in this related PR:

Related to https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/158


src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+TRANSFER_STANDARD = "standard"
+TRANSFER_ZIPFILE = "zipped package"
+TRANSFER_UNZIPPED_BAG = "unzipped bag"
+TRANSFER_ZIPPED_BAG = "zipped bag"
+TRANSFER_DSPACE = "dspace"
+TRANSFER_MAILDIR = "maildir"
+TRANSFER_TRIM = "TRIM"
+TRANSFER_DATAVERSE = "dataverse"
+
+TRANSFER_DISk_IMAGE = "disk image"
```
replaceafill:
```suggestion
TRANSFER_DISK_IMAGE = "disk image"
```

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+selected by directory, or by file, e.g. standard transfer vs. zipped bags.
+
+This module should try and capture these characteristics where possible.
+"""
+from __future__ import absolute_import, unicode_literals
+
+import collections
+
+TRANSFER_STANDARD = "standard"
+TRANSFER_ZIPFILE = "zipped package"
```
replaceafill: Do you think `TRANSFER_ZIPPED_PACKAGE` would be more appropriate?

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+Transfers have different characteristics such as whether a transfer can be
+selected by directory, or by file, e.g. standard transfer vs. zipped bags.
+
+This module should try and capture these characteristics where possible.
+"""
+from __future__ import absolute_import, unicode_literals
+
+import collections
+
+TRANSFER_STANDARD = "standard"
```
replaceafill: I think this group should be prefixed as `TRANSFER_TYPE_...`. It'd make easier to understand them individually without having to resort to `archivematica_transfer_types.TRANSFER_STANDARD` or `amtypes.DSPACE`

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+WATCHED_ZIPFILE = "zippedPackage"
+WATCHED_UNZIPPED_BAG = "baggitDirectory"
+WATCHED_ZIPPED_BAG = "baggitZippedDirectory"
+WATCHED_DSPACE = "Dspace"
+WATCHED_MAILDIR = "maildir"
+WATCHED_TRIM = "TRIM"
+WATCHED_DATAVERSE = "dataverseTransfer"
+
+# TRANSFER_DISK_IMAGE is treated differently by src/MCPServer/lib/server/packages.py.
+TRANSFER_TYPES = (
```
replaceafill: Can't this be replaced with `ARCHIVEMATICA_TRANSFER_TYPES`? As far as I can tell, it's only used for `... in TRANSFER_TYPES`checks.

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+APPROVE_ZIPPED = "Approve zipped transfer"
+APPROVE_DSPACE = "Approve DSpace transfer"
+APPROVE_UNZIPPED_BAG = "Approve bagit transfer"
+APPROVE_ZIPPED_BAG = "Approve zipped bagit transfer"
+
+# Archivematica transfer profiles that describe a transfer's properties.
+TransferProfile = collections.namedtuple(
+    "TransferProfile", "transfer_type watched_directory"
+)
+ARCHIVEMATICA_TRANSFER_TYPES = {
```
replaceafill: Is the `ARCHIVEMATICA_` prefix necessary if we remove the `TRANSFER_TYPES` tuple above?

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+TRANSFER_ZIPPED_BAG = "zipped bag"
+TRANSFER_DSPACE = "dspace"
+TRANSFER_MAILDIR = "maildir"
+TRANSFER_TRIM = "TRIM"
+TRANSFER_DATAVERSE = "dataverse"
+
+TRANSFER_DISk_IMAGE = "disk image"
+
+WATCHED_STANDARD = "standardTransfer"
+WATCHED_ZIPFILE = "zippedPackage"
```
replaceafill: Do you think `WATCHED_ZIPPED_PACKAGE` would be more appropriate?

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+TRANSFER_UNZIPPED_BAG = "unzipped bag"
+TRANSFER_ZIPPED_BAG = "zipped bag"
+TRANSFER_DSPACE = "dspace"
+TRANSFER_MAILDIR = "maildir"
+TRANSFER_TRIM = "TRIM"
+TRANSFER_DATAVERSE = "dataverse"
+
+TRANSFER_DISk_IMAGE = "disk image"
+
+WATCHED_STANDARD = "standardTransfer"
```
replaceafill: As with the transfer type group, I think these would be better represented with a `WATCHED_DIRECTORY_NAME_...` prefix.

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+    TRANSFER_TRIM: TransferProfile(TRANSFER_TRIM, WATCHED_TRIM),
+    TRANSFER_DATAVERSE: TransferProfile(TRANSFER_DATAVERSE, WATCHED_DATAVERSE),
+}
+
+
+def retrieve_watched_dirs():
+    for type_ in ARCHIVEMATICA_TRANSFER_TYPES:
+        yield ARCHIVEMATICA_TRANSFER_TYPES[
+            type_
+        ].transfer_type, ARCHIVEMATICA_TRANSFER_TYPES[type_].watched_directory
```
replaceafill: I know I'm really short sighted but I'm having trouble seeing the value of the `TransferProfile` named tuple. If we convert `ARCHIVEMATICA_TRANSFER_TYPES` into a `{transfer_type: watched_directory_name, ...}` dictionary, this block of logic is equivalent to [`iteritems`](https://docs.python.org/2/library/stdtypes.html#dict.iteritems)

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def retrieve_watched_directory(transfer_type, return_key_error=False):
+    if transfer_type not in TRANSFER_TYPES:
+        if not return_key_error:
```
replaceafill: If `return_key_error` is `False` by default, what's the advantage of this approach instead of letting the caller to deal with the missing key?

src/MCPServer/lib/server/packages.py
```diff
@@ -55,59 +57,68 @@ def _get_setting(name):
...
 # There is also the hope that the watched directories can be deprecated in the
 # near future.
+ACTIVE_TRANSFERS = "activeTransfers"
 PACKAGE_TYPE_STARTING_POINTS = {
-    "standard": StartingPoint(
+    amtypes.TRANSFER_STANDARD: StartingPoint(
         watched_dir=os.path.join(
-            _get_setting("WATCH_DIRECTORY"), "activeTransfers/standardTransfer"
+            _get_setting("WATCH_DIRECTORY"),
+            os.path.join(ACTIVE_TRANSFERS, amtypes.WATCHED_STANDARD),
```
replaceafill: Can the inner `os.path.join` call be omited? I mean:

`os.path.join(_get_setting("WATCH_DIRECTORY"), ACTIVE_TRANSFERS, amtypes.WATCHED_STANDARD)`

src/MCPServer/lib/server/packages.py
```diff
@@ -323,7 +334,7 @@ def create_package(
     """
     if not name:
         raise ValueError("No transfer name provided.")
-    if type_ is None or type_ == "disk image":
+    if type_ is None or type_ == amtypes.TRANSFER_DISk_IMAGE:
```
replaceafill:
```suggestion
    if type_ is None or type_ == amtypes.TRANSFER_DISK_IMAGE:
```

src/MCPServer/lib/server/shared_dirs.py
```diff
@@ -243,6 +244,22 @@ def install_builtin_config(name):
...
+
+
+def generate_transfer_watched_dirs():
+    """Create a tuple of watched directories specifically related to transfers
+    based on the transfer profiles in amtypes.
+    """
+    transfer_dirs = ()
+    for transfer_type in amtypes.retrieve_watched_dirs():
+        transfer_dirs = transfer_dirs + (
+            os.path.join(ACTIVE_TRANSFERS_DIR, (transfer_type[1])),
```
replaceafill: I think iterating the result of `retrieve_watched_dirs` differently would avoid having to wonder what's this `[1]`.

src/MCPServer/tests/test_package.py
```diff
@@ -338,3 +340,26 @@ def test_move_file(self, tmp_path, processing_dir, transfer):
 
         transfer.refresh_from_db()
         assert Path(transfer.currentlocation) == dest_path
+
+
+ACTIVE_TRANSFERS_DIR = "activeTransfers"
```
replaceafill: Can't this be imported from `server.packages`?

src/archivematicaCommon/tests/test_archivematica_transfer_types.py
```diff
@@ -0,0 +1,39 @@
...
+def test_retrieve_watched_dirs():
+    dirs = amtypes.retrieve_watched_dirs()
+    assert len(list(dirs)) == len(EXPECTED_DIRS)
+    for dir_ in dirs:
+        if dir_ not in EXPECTED_DIRS:
+            assert (
+                False
+            ), "Unexpected watched directory in transfer watched directories: {}".format(
+                dir_
+            )
```
replaceafill:
```suggestion
        assert dir_ in EXPECTED_DIRS, "Unexpected watched directory in transfer watched directories: {}".format(
            dir_
        )
```

src/archivematicaCommon/tests/test_archivematica_transfer_types.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_watched_directory():
+    for expected in EXPECTED_DIRS:
+        assert amtypes.retrieve_watched_directory(expected[0]) == expected[1]
```
replaceafill:
```suggestion
    for transfer_type, watched_directory_name in EXPECTED_DIRS:
        assert amtypes.retrieve_watched_directory(transfer_type) == watched_directory_name
```

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,111 @@
...
+
+ARCHIVE_ZIP = ".zip"
+ARCHIVE_TGZ = ".tgz"
+ARCHIVE_TAR_GZ = ".tar.gz"
+
+ARCHIVE_TYPES = (ARCHIVE_ZIP, ARCHIVE_TGZ, ARCHIVE_TAR_GZ)
+ZIP_TYPE_TRANSFERS = (TRANSFER_ZIPFILE, TRANSFER_ZIPPED_BAG, TRANSFER_DSPACE)
+
+# Manual approval jobs associated with a transfer.
+APPROVE_STANDARD = "Approve standard transfer"
```
replaceafill, line 69: I think we could gain some clarity by grouping these into a `APPROVE_TRANSFER_JOB_NAMES/TYPES` or something and use `Q(jobtype_in=...)` in the `.filter(...)` call where these are used in the `api.views` module.

src/dashboard/src/components/api/views.py
```diff
@@ -819,16 +819,30 @@ def package(request):
...
     try:
         payload = json.loads(request.body)
         path = base64.b64decode(payload.get("path"))
     except (TypeError, ValueError):
         return helpers.json_response(
             {"error": True, "message": 'Parameter "path" cannot be decoded.'}, 400
         )
+    transfer_type = payload.get("type")
+    try:
+        if transfer_type not in amtypes.TRANSFER_TYPES and transfer_type is not None:
```
replaceafill, line 824: Is the `transfer_type is not None` check necessary?

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,150 @@ def test_copy_metadata_files_api(mocker):
...
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "transfer_type,expected_directory,alt_dir",
+    [
+        ("standard", "standardTransfer", None),
+        ("unzipped bag", "baggitDirectory", "standardTransfer"),
+        ("zipped bag", "baggitZippedDirectory", "standardTransfer"),
+        ("dspace", "Dspace", "standardTransfer"),
+        ("zipfile", "standardTransfer", "zippedPackage"),
```
replaceafill: If you're testing an invalid values here, could we make it more clear? `invalid type` maybe?

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,150 @@ def test_copy_metadata_files_api(mocker):
...
+
+def dashboard_login_and_setup(client, django_user_model, username, password):
+    django_user_model.objects.create_user(username=username, password=password)
+    client.login(username=username, password=password)
+    helpers.set_setting("dashboard_uuid", "test-uuid")
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "transfer_type,expected_directory,alt_dir",
```
replaceafill: What's the purpose of `alt_dir`?

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,150 @@ def test_copy_metadata_files_api(mocker):
...
+        "components.filesystem_ajax.views._copy_from_transfer_sources",
+        return_value="True",
+    )
+    mocker.patch(
+        "components.filesystem_ajax.helpers.check_filepath_exists", return_value=None
+    )
+    mocker.patch("shutil.move", return_value=None)
+    mocker.patch("tempfile.mkdtemp", return_value=tempfile.mkdtemp())
+    mocker.patch("shutil.rmtree", return_value=None)
+    # Invalid path
```
replaceafill: Don't get the comment?

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,150 @@ def test_copy_metadata_files_api(mocker):
...
+
+
+@pytest.mark.django_db
+def test_unapproved_transfers_endpoint(
+    client, django_user_model, username, password, mocker
+):
+    MESSAGE_SUCCESS = "Fetched unapproved transfers successfully."
+    UNAPPROVED_URL = "/api/transfer/unapproved/"
+    EXPECTED_DIRS = [
+        "SimpleTransferWithProcessingMCP.tgz",
```
replaceafill, line 659: A `.tgz` as expected *dir*?

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,150 @@ def test_copy_metadata_files_api(mocker):
...
+    dashboard_login_and_setup(client, django_user_model, username, password)
+    resp = client.get(UNAPPROVED_URL)
+    resp_parsed = json.loads(resp.content)
+    assert resp.status_code == 200
+    assert resp_parsed["message"] == MESSAGE_SUCCESS
+    assert len(resp_parsed["results"]) == len(EXPECTED_DIRS)
+    dirs = [x["directory"] for x in resp_parsed["results"]]
+    for directory in dirs:
+        if directory not in EXPECTED_DIRS:
+            assert False, "Unexpected transfer in unapproved transfers"
```
replaceafill:
```suggestion
        assert directory in EXPECTED_DIRS, "Unexpected transfer in unapproved transfers"
```

replaceafill (commented):
@ross-spencer Even though this is a bigger change than I expected, I think it's good overall. Most of my comments are difficulties I've had trying to understand the new code.

replaceafill:
Oh, I forgot to mention. We're getting rid of the `zipfile` string everywhere, correct? I've noticed it in a few places.

ross-spencer:
Thanks @replaceafill yes, it was bigger than I anticipated too, but I'm heartened by the response. I'll get to looking at these comments asap.

I wonder, can you help identify some of those uses of `zipfile` -- I noticed `zipfile` is unhelpfully used in some external assets and documentation, but didn't seem related to the feature, but that just made it harder to spot. 

In the new archivematica_transfer_type helper, I stuck with `ZIPFILE` as a `const` name for convenient lookup, but I can change that. There shouldn't be any "functional" uses of `zipfile` outside of any variable naming.

src/MCPServer/lib/server/packages.py
ross-spencer: Good spot. I've tried it slightly differently, this entire snippet could be changed, but I definitely missed this first time around.

src/archivematicaCommon/lib/archivematica_transfer_types.py
ross-spencer: Nice! I didn't know about this. V.Cool.

src/dashboard/src/components/api/views.py
ross-spencer: Oh yeah, in this case, `None` is technically valid and Archivematica will try to initiate a transfer as a standard transfer if it sees `None` here.

src/dashboard/tests/test_api.py
ross-spencer: Ah, not too much, I'll remove! :slightly_smiling_face:

src/dashboard/tests/test_api.py
ross-spencer: Where's the `flake8` for incomplete thoughts!! I don't remember what this was going to be either! :)

src/archivematicaCommon/lib/archivematica_transfer_types.py
ross-spencer: Yeah, I see, I've changed this now, so it's a bit shorter to use across modules.

src/archivematicaCommon/lib/archivematica_transfer_types.py
ross-spencer: My approach was initially to think about having a class to define the characteristics of a transfer type. Then I decided on the pattern of a named tuple. The point here was to have something that could be a little nicer to work with than a `dict` (I like dot notation). And then to promote extensibility where there is more to say about a transfer type, e.g. whether is a file-type transfer, e.g. a zip. As examples. Some of that information in the packages module around starting link etc. could also be ported here. Obviously I haven't quite realized this potentiality with this refactor. I was keeping a closer eye on the integration and making sure I didn't break anything, but I am hopeful for it in the future if we keep it. What do you think?

src/dashboard/tests/test_api.py
ross-spencer: This is just defined by the endpoint, you'll see in the model this is under `directory` and so I'm just following that convention here.

src/archivematicaCommon/lib/archivematica_transfer_types.py
replaceafill: I have trouble seeing potential designs. I usually think YAGNI if I notice I'm trying to come up with structures more complex of what I need and go with what's simpler. That gets you in other kinds of problems. Again short sighted :blush:

I bet @sevein can provide much better feedback on this.

src/archivematicaCommon/lib/archivematica_transfer_types.py
ross-spencer: I think this is just an artifact of how I was thinking about this refactor and trying to maintain all the previous functionality. Possibly a short-cut as I was writing and testing. I've put the responsibility back on the caller now which I think is cleaner too.

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -104,7 +92,7 @@ def retrieve_watched_dirs():
 
 
 def retrieve_watched_directory(transfer_type, return_key_error=False):
-    if transfer_type not in TRANSFER_TYPES:
+    if transfer_type not in ARCHIVEMATICA_TRANSFER_TYPES.keys():
```
replaceafill:
```suggestion
    if transfer_type not in ARCHIVEMATICA_TRANSFER_TYPES:
```

src/dashboard/src/components/api/views.py
```diff
@@ -829,7 +829,10 @@ def _package_create(request):
         )
     transfer_type = payload.get("type")
     try:
-        if transfer_type not in amtypes.TRANSFER_TYPES and transfer_type is not None:
+        if (
+            transfer_type not in amtypes.ARCHIVEMATICA_TRANSFER_TYPES.keys()
```
replaceafill:
```suggestion
            transfer_type not in amtypes.ARCHIVEMATICA_TRANSFER_TYPES
```

ross-spencer:
Okay, @replaceafill, I think that's 98.9% of requested changes made. We can change the structure for holding information about transfer types if needed, the tests make that easy now, but yeah, let me know.

src/MCPServer/lib/server/packages.py
```diff
@@ -323,7 +329,7 @@ def create_package(
     """
     if not name:
         raise ValueError("No transfer name provided.")
-    if type_ is None or type_ == "disk image":
+    if type_ is None or type_ == amtypes.TRANSFER_DISK_IMAGE:
```
replaceafill:
```suggestion
    if type_ is None or type_ == amtypes.TRANSFER_TYPE_DISK_IMAGE:
```

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -296,6 +296,8 @@ def _copy_to_start_transfer(
 
         # default to standard transfer
         type_subdir = amtypes.retrieve_watched_directory(transfer_type)
+        if not type_subdir:
+            type_subdir = amtypes.retrieve_watched_directory("standard")
```
replaceafill:
```suggestion
            type_subdir = amtypes.retrieve_watched_directory(amtypes.TRANSFER_TYPE_STANDARD)
```

replaceafill (approved):
Thanks @ross-spencer! Couple of small suggestions and found a typo. Hope @sevein can chime in regarding the data structure conversation. Otherwise this LGTM :+1:

src/dashboard/tests/test_api_v2beta.py
```diff
@@ -174,3 +175,98 @@ def test_avalon_err(self):
...
+
+@pytest.fixture
+def password():
+    return "test"
+
+
+def dashboard_login_and_setup(client, django_user_model, username, password):
+    django_user_model.objects.create_user(username=username, password=password)
+    client.login(username=username, password=password)
+    helpers.set_setting("dashboard_uuid", "test-uuid")
```
sevein, line 193: Why did you choose not to extend the existing xunit-style tests? There is some redundancy now with your changes that it's going to make the maintenance of this module harder.

src/dashboard/tests/test_api_v2beta.py
```diff
@@ -174,3 +175,98 @@ def test_avalon_err(self):
...
+
+
+def test_package_endpoint(client, django_user_model, username, password):
+    """Validate the behavior of the package endpoint."""
```
sevein, line 197: This function is testing both `GET /api/v2beta/package` and `POST /api/v2beta/package`, that's confusing!

src/dashboard/tests/test_api_v2beta.py
```diff
@@ -174,3 +175,98 @@ def test_avalon_err(self):
...
+    resp = client.post(
+        "/api/v2beta/package/",
+        json.dumps(test_transfer),
+        content_type="application/json",
+    )
+    assert resp.status_code == 500
+    assert json.loads(resp.content) == {
+        u"message": u"Package cannot be created",
+        u"error": True,
+    }
```
sevein, line 227: There are xunit-style tests in this module that testing similar concerns right? Wouldn't make more sense to improve them? They also mock the MCPClient.

src/dashboard/src/components/api/views.py
```diff
@@ -819,16 +811,33 @@ def package(request):
...
+        ):
+            raise (
+                ValueError(
+                    "Unexpected type of package provided '{}'".format(transfer_type)
+                )
+            )
+    except ValueError as err:
+        LOGGER.error("{}: {}".format(MESSAGE, err))
+        return helpers.json_response(
+            {"error": True, "message": "{}: {}".format(MESSAGE, err)}, 500
```
sevein, line 833: FYI we have a helper that could also be used here: `_error_response`. It could shorten this view significantly.

src/dashboard/src/components/api/views.py
```diff
@@ -819,16 +811,33 @@ def package(request):
...
+            transfer_type not in amtypes.TRANSFER_TYPES.keys()
+            and transfer_type is not None
+        ):
+            raise (
+                ValueError(
+                    "Unexpected type of package provided '{}'".format(transfer_type)
+                )
+            )
+    except ValueError as err:
+        LOGGER.error("{}: {}".format(MESSAGE, err))
```
sevein, line 831: It this the right level to use? Generally `ERROR` means that we're dealing with a serious problem that a system administrator needs to deal with.

src/dashboard/src/components/api/views.py
```diff
@@ -819,16 +811,33 @@ def package(request):
...
+        return helpers.json_response(
+            {"error": True, "message": "{}: {}".format(MESSAGE, err)}, 500
+        )
```
sevein: Why did you choose try/catch here? You may want to try a simpler form like:

```python
if __condition__:
    # do something
    return _error_response(...)
```

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,127 @@
...
+
+ - References to watched directories also exist in Workflow.json and MCP
+   server's shared_dirs.py
+     * src/MCPServer/lib/assets/workflow.json
+
+Transfers have different characteristics such as whether a transfer can be
+selected by directory, or by file, e.g. standard transfer vs. zipped bags.
+
+This module should try and capture these characteristics where possible.
+"""
```
sevein, line 32: We have unfortunate precedents such archivematicaFunctions, but I think that we have to stop calling things archivematica_XYZ, we can just refer to them as XYZ right?

src/MCPServer/tests/test_package.py
```diff
@@ -338,3 +341,23 @@ def test_move_file(self, tmp_path, processing_dir, transfer):
...
+        ("dataverse", os.path.join(ACTIVE_TRANSFERS_DIR, "dataverseTransfer")),
+        ("maildir", os.path.join(ACTIVE_TRANSFERS_DIR, "maildir")),
+        ("TRIM", os.path.join(ACTIVE_TRANSFERS_DIR, "TRIM")),
+    ],
+)
+def test_starting_point_watched_dirs(transfer_type, expected_directory):
+    assert PACKAGE_TYPE_STARTING_POINTS[transfer_type]
+    assert PACKAGE_TYPE_STARTING_POINTS[transfer_type].watched_dir.endswith(
+        expected_directory
+    )
```
sevein, line 363: What would be the benefit of this test? I don't think that we have to test our data.

src/MCPServer/tests/test_shared_dirs.py
```diff
@@ -0,0 +1,23 @@
...
+        "watchedDirectories/activeTransfers/standardTransfer",
+        "watchedDirectories/activeTransfers/TRIM",
+        "watchedDirectories/activeTransfers/zippedPackage",
+    ]
+    transfer_watched_dirs = shared_dirs.generate_transfer_watched_dirs()
+    assert len(transfer_watched_dirs) == 8
+    for dir_ in transfer_watched_dirs:
+        assert (
+            dir_ in EXPECTED_DIRS
+        ), "Unexpected directory in transfer watched directories: {}".format(dir_)
```
sevein, line 23: This test seems to put a lot of focus on the data with some undesirable coupling. I think that it'd be much interesting to test the `create` function - that would really give you the level of confidence that I think you're looking for.

src/MCPServer/lib/server/shared_dirs.py
```diff
@@ -300,6 +309,7 @@ def create():
         "www/AIPsStore/transferBacklog/originals",
         "www/DIPsStore",
     )
+    dirs = dirs + generate_transfer_watched_dirs()
```
sevein, line 312: At this point where we're manipulating the data I'd suggest to stop using tuples.

src/dashboard/tests/test_api.py
```diff
@@ -576,3 +579,145 @@ def test_copy_metadata_files_api(mocker):
...
+            "%sharedPath%watchedDirectories/activeTransfers/dataverseTransfer",
+        ),
+        ("maildir", "%sharedPath%watchedDirectories/activeTransfers/maildir"),
+        ("TRIM", "%sharedPath%watchedDirectories/activeTransfers/TRIM"),
+        ("unexpected transfer type", None),
+    ],
+)
+def test_get_modified_standard_transfer_path(transfer_type, expected_result):
+    ret = views.get_modified_standard_transfer_path(transfer_type)
+    assert ret == expected_result
```
sevein, line 723: Again I think it'd be easier if you just tested the implementation. You don't need all the data points for that right? It's going to become a maintenance burden.

sevein (changes requested):
Thanks Ross. I think generally there is a level of coupling with the data in your tests that seems undesirable. How many instances of our transfer types are we willing to keep around? Most of the times you could just mutate the original for the sake of testing the implementation right?

Your validation proposal for v2 isn't too off but I think it'd be ideal if MCPServer communicated the validation issue to the Dashboard in a meaningful way that could be later translated into the API space avoiding the seggregation of responsibilities. We haven't quite talked about it but the vision for v2 was to become part of MCPServer. Dashboard /api/v2 would just be a reverse proxy, /api/v1 could be managed by a translation layer that talked to MCPServer.

So I think that this PR has a lot of potential but it's way too late in the 1.11 release cycle to even consider it. My suggestion here is to set this work aside for now and submit a new PR with the actual change in the naming that we were looking for.

src/dashboard/src/components/api/views.py
```diff
@@ -819,16 +811,30 @@ def package(request):
...
```
sevein, line 834: What would be the value of using try/catch when the error that you're catching is the one that you're raising? It sounds like it'd be much simpler to write something like `if condition: do X; do Y`.

ross-spencer:
> So I think that this PR has a lot of potential but it's way too late in the 1.11 release cycle to even consider it. My suggestion here is to set this work aside for now and submit a new PR with the actual change in the naming that we were looking for.

@sevein we have been discussing this change in Slack,and the idea of doing nothing was not desirable. The idea of just changing the transfer type, was also undesirable as a standalone fix because every time we make changes around that requires a large cascade across components and we can't guarantee the efficacy of a fix without tests (which we didn't have). In short, it has never been easy to change or add a transfer type by Achivematica's standards. If we needed to change `Zipped package` again now, we could copy and replace, or we could use a new method where it is clear from module documentation what non-python components need to change, or the other changes can ripple down the code-base based on the new module.This change makes it easy, and testable. Any coupling you see that is undesirable right now is more reflective of Archivematica's platform, and because it is largely abstracted away in unit tests now, it's completely within our control how we change that coupling moving forward -- it does not say we can't change this in future. 

What I've done here is:

* Decoupled and removed redundancy across _all_ of Archivematica so changes can be centralized. The next time we add a transfer type, or if we had to change `Zipped packages` again, it is documented in a central module shared between components.
 
* Added integration level testing (at least something analogous to it, and exactly what you see) in the unit tests, because we haven't any other tests to guarantee the stability of our start transfer endpoints. You ask why am I testing the data we return? Because we haven't _any_ equivalent tests, and this is the standard we've set with our endpoints thus far. 

* Modernized some of the testing -- e.g. used `pytest` over `xunit` where parameterization was more easily used and aren't easily compatible with old-style tests. We have both style tests across AM, the only difference here is that it is in the same module. Additionally, unwrapping, decent, working tests, is not the same as making changes to untested code. At best you might change the inputs or outputs. At worst, you replace it with another test.

Your comments seem to flip-flop between why have I attempted something new to why am I following Archivematica's existing patterns... so I don't think that's particularly constructive. 

The criticism of the Dashboard validation would also make a lot more sense were it not framed as something "we might do" in the MCP server. What this offers is an easy to implement middle ground right now that means we have an integration test, and more user-feedback. And is not un-doable - but rather than the concept of un-doing it might be better to look at how this change promotes future improvements by making sure functionality isn't unwrapped without due care and caution. 

Two of us have looked at this. I think it clearly two or three steps forward, and if I thought it any step back I wouldn't submit the PR. I also have one approval, and we're about to hit regression testing, and I've backed up the changes with significant unit testing (focused on the specific end-points being used) sample-data, and AMAUAT testing. I didn't make a single change to code without making sure a test was written first to ensure nothing was going to break. I don't agree with the risks, but I think you'll see any potential risks a lot differently when you take that into account. In comparison to the workflow functionality just merged, this is very much surface level, vs. something far more significant to AM's heart. You note the stage in the release cycle, but there's little to support that we would have done any more significant testing on this during the release, other than a brief acceptance test, vs. regression testing which I think it always out concern across the platform. 

It looks like there are maybe one or two, surface level changes that you have suggested that aren't about stopping this PR that I might be able to make. There was a specific question about a data structure you were asked about though. 

I would need to analyze the surface changes a bit more but the premise of putting this aside significantly undermines an effort here, and I think would only be warranted if there was something significantly broken (which you're clearly not saying). I don't think it means too much to this PR that one day we might remove a transfer type (because clearly that day is not now) or change the shape of our data and improve our API offering. And also not that it doesn't quite fit a personal style. That should be resolved first, and then we need to look at what is sensible to be changed vs. the suggested nuclear option. Keep in mind, this is a set of tests to prove legacy functionality, and to create a platform for moving away from that one-day without breaking anything. It's not the end of this work by any means.

sevein:
> I would need to analyze the surface changes a bit more but the premise of putting this aside significantly undermines an effort here, and I think would only be warranted if there was something significantly broken (which you're clearly not saying).

Ok, thanks for the clarification and sorry for the flip-flopping. Let me try to suggest a way to move forward. We haven't had an opportunity to discuss these problems that you're trying to fix. Why don't you file a couple of new issues describing the need to factor out transfer types and the proposal around validation? It sounds like https://github.com/archivematica/Issues/issues/1126 could easily be addressed like you were planning to do in https://github.com/artefactual/archivematica/pull/1569 with a few extra changes right?

ross-spencer:
Ultimately this work won't make it into 1.11 with a decision made and a kind note given by the product owner.

------

From a developer to developer perspective, then what follows is a development note that hopes to demystify what was requested, vs. what was done. Because this impacts trust, and future agency. And this has a significant impact on those working cross-team. _The analysis was done entirely off the clock to promote discussion somewhere in the future..._


**NB.** the described outcome -- a solution that expanded on 1162, and described in Slack: 

> hi ... I think the feeling is there may be a fix to be done here, which at 
  least centers around getting the right name in the UI, and alignment with 
  the API. I think we can commit to a fix to be merge-able Monday, that could 
  also leave open some room to get that naming convention right, i.e. say its 
  coded with Zipped package for now following Evelyn's notes: 
  https://github.com/archivematica/Issues/issues/682#issuecomment-557633499 
  it could still be copy-and-replaced before merging if we have a better name. 
  From my initial PR work here, there's clearly a unit test missing that can be 
  added and that will benefit that idea. That unit test should be feasible, and 
  needs to say, whatever I call my API transfer type, if I submit code against 
  it I won't break the UI at the same time, which is something that 
  could/should be caught. That might also be an AMAUAT which I think is 
  possible now with Douglas' recent fixes to processing configs.

What this note addresses:

* So I think that this PR has a lot of potential but it's way too late in the 1.11 release cycle to even consider it.
* There is too much going on in this PR.

Because unfortunately, when a value judgement is made on a portion of code, i.e. it isn't quantified as something broken, or clearly incorrect, e.g. too broad vs. _"this changes the behaviour we expect of Archivematica..."_ if it isn't impossible to rebuke, it takes at least two or three times the effort.

So, breaking this down:

The only method we've had available to do this change in the past is to copy and replace a value. 

To add a new transfer type, just the scaffolding, is another chore, as the developer has to find all of the locations that each form part of a transfer "trigger".

To make the change the old-school way, we copy and replace against two search terms (`zipfile` and `zipped directory`):

_1. Process: Copy and replace, some terms use `zipfile`, some `zipped directory`._

**Commit:** https://github.com/artefactual/archivematica/pull/1586/commits/5104e3ec24b86719cd072b75f91a8dfd7fb81820

**Features:**

* Testing is entirely manual,
* Copy and replace is bad programming, redundancy introduces errors, even if we've tested. 
* Difficult to change, painful to code review more than once, e.g. if we change the value again, or if mistakes are made. 
* Redundancy across two modules, these are semantically, exactly the same properties. `Transfer X` in the Dashboard is the same as `Transfer X` in the server. In code, they're unconnected literals. 
* Because we're copying and pasting two different values across multiple locations, it takes 15-20 minutes to go through each. 
* We satisfy ourselves it will work because it worked with the previous values.
* To go through that process again, we repeat all of the above, hence the length of time taken and doubt of making the change the first time around. I.e. if we decide against `zipped package` when we come to try it out, 

**No. lines:** +15 −15

Now, the change that was submitted. 

_2. Process: Improve on 1, and isolate values into a shared module:_

**Commit:** https://github.com/artefactual/archivematica/pull/1581/commits/6fa249cbd1809cbb05c1e34cd379afc5c6f6d02d _**NB.** The commit is [annotated](https://github.com/artefactual/archivematica/pull/1581)._

**Features:**

* Link Dashboard and Server values using a shared module.
* Now we fulfil the promise of changing in one place, and cascading across the code.
* Opportunity to describe transfers and their properties, if that's something we want to do in future, e.g. if the model remains, and opportunities to not do this, if that's another direction to go. 
* New module becomes documentation helping developers in future to understand transfer types (while transfer types still exist). 

**Questions?** 

Would we put just `Zipped package` into a new module on its own and not change the other transfer types? _No._ That's neither good programming 101, or likely to pass CR. So changes/complexity, probably look something like (15 (lines affected in PR1) x No. transfer types.) where transfer types == 5 not 6 (Disk images are treated differently). The hypothesis being this would reflect about as minimal a change as possible to fulfil the spec and push things forward without repeating poor practices.

**No. lines:** +221 −75 (127 of which in the new lookup module)

The work is done at this point, but breaking the trend of not writing tests due to last-minute time constraints:

_3. Process: Write some tests,_ 

* Specifically testing entry points, specifically testing transfer types. 
* In a few instances, focuses on data relied on by the application to not change, so watched directories etc. This wouldn't matter so much, if, say, workflow.json was imported into a module so that its key values were discoverable, or the other way round, either way, we didn't have redundancy in either the workflow or the application see also: [here](https://gist.github.com/sevein/63e751acd37bde5590d680344ad8b3d5#watched-directories-dont-describe-whether-they-are-used-to-start-transfers). 
* Are all the tests valuable? Until we've unwired everything then yes. the tests both help maintain present state and make it maintainable; and in future, tell us what we've missed unwiring, e.g. once a single value from the transfer-type module is deleted, it also trickles down. 
* Tests are a lot easier to delete than application code. We're just testing legacy work, it can all be removed as we reach a future state where the tests might say something about a concrete set of interfaces. Don't need the test. Delete it. Application behaviour isn't impacted.
* Uses pytest, but through compatiblity issues: https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses separates pytest out. Ideally, if there were fewer, or there was time, would rewrite the remainder of the test case tests as in PR#1149. Ideally, there'd be a lot of tests added, or code refactored if this was an actually broadly scoped PR. The intent, _at least_, was a small incremental step. 
* There is clearly more room for generating trust in the application and I think we can achieve that with these small steps. 

**Finally,** 

4. Again, complete transparency, a final commit was tagged onto all of the original work, as a proposal. Designing it to be the final commit, and labelling it appropriately, I made sure it could be discussed and unwound if so decided.

**Commit:** https://github.com/artefactual/archivematica/pull/1580/commits/1cf2896770780539fde4c668d9c0b3fa650dc0b1

PR #1581, opened 2020-03-04
WIP: Demonstrate specific changes are to start transfer endpoints

ross-spencer:
This commit is exactly, the same as [dev/issue-682-zipped-package-type](https://github.com/artefactual/archivematica/pull/1580) and contains two commits:

* [Commit 1](https://github.com/artefactual/archivematica/pull/1581/commits/3d9176576a312ef64b7aadf4b4945ed2b7fad9d7): Changes required to simplify changing a transfer name in Archivematica/API.
* [Commit 2](https://github.com/artefactual/archivematica/pull/1581/commits/2c49263eae5a60ae47cd30389d4600e768a272a9): Proposed novel change to packages endpoint to perform early validation.

In comparison to [dev/issue-682-zipped-package-type](https://github.com/artefactual/archivematica/pull/1580) all tests and fixtures have been removed to demonstrate where the bulk of the code changes are in code.

### DIFF
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 """Package management."""
-from __future__ import absolute_import, division, print_function, unicode_literals
+from __future__ import absolute_import, division, unicode_literals
 
 import abc
 import ast
@@ -18,6 +18,8 @@ from django.utils import six
 import storageService as storage_service
 from archivematicaFunctions import strToUnicode
 from archivematicaFunctions import unicodeToStr
+import archivematica_transfer_types as amtypes
+
 from main import models
 
 from server.db import auto_close_old_connections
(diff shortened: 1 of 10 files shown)


src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,127 @@
...
+
+ - References to watched directories also exist in Workflow.json and MCP
+   server's shared_dirs.py
+     * src/MCPServer/lib/assets/workflow.json
+
+Transfers have different characteristics such as whether a transfer can be
+selected by directory, or by file, e.g. standard transfer vs. zipped bags.
+
+This module should try and capture these characteristics where possible.
+"""
```
ross-spencer, line 32: Adding the only documentation associated with transfer types as they are in Archivematica at present. For looking into the past, and into the future, e.g. to see where legacy wiring needs to be undone.

src/archivematicaCommon/lib/archivematica_transfer_types.py
```diff
@@ -0,0 +1,127 @@
...
+    for type_ in TRANSFER_TYPES:
+        yield TRANSFER_TYPES[type_].transfer_type, TRANSFER_TYPES[
+            type_
+        ].watched_directory
+
+
+def retrieve_watched_directory(transfer_type):
+    if transfer_type not in TRANSFER_TYPES:
+        return None
+    return TRANSFER_TYPES[transfer_type].watched_directory
```
ross-spencer, line 127: Written to be identical to any use of: 
```
TRANSFER_TYPE_DIRECTORIES = {	
    "standard": "standardTransfer",	
    "zipfile": "zippedDirectory",	
    "unzipped bag": "baggitDirectory",	
    "zipped bag": "baggitZippedDirectory",	
    "dspace": "Dspace",	
    "maildir": "maildir",	
    "TRIM": "TRIM",	
    "dataverse": "dataverseTransfer",	
}
```
in `filesystem_ajax/views.py` specifically with a view to drop-in the previous method if absolutely called for.

src/dashboard/frontend/app/browse/browse.controller.js
```diff
@@ -78,7 +78,7 @@ class BrowseController {
   // Determines whether a given file can be added to the transfer,
   // depending on the transfer type.
   file_can_be_added(file) {
-    if (this.transfer.type === 'zipped bag' || this.transfer.type === 'zipfile') {
+    if (this.transfer.type === 'zipped bag' || this.transfer.type === 'zipped package') {
```
ross-spencer, line 81: I would have liked to find a way to reference the other constants somehow, or at least reduce both JavaScript, and the HTML constants to one.

src/dashboard/src/components/api/views.py
```diff
@@ -442,14 +443,9 @@ def unapproved_transfers(request):
...
+    # TODO: Does Dataverse transfers need to be here?
     jobs = models.Job.objects.filter(
-        (
-            Q(jobtype="Approve standard transfer")
-            | Q(jobtype="Approve zipped transfer")
-            | Q(jobtype="Approve DSpace transfer")
-            | Q(jobtype="Approve bagit transfer")
-            | Q(jobtype="Approve zipped bagit transfer")
-        )
+        Q(jobtype__in=amtypes.APPROVE_TRANSFER_JOB_NAMES)
```
ross-spencer, line 448: An example optimisation driven by the new module.

src/MCPServer/lib/server/shared_dirs.py
```diff
@@ -261,16 +278,8 @@ def create():
...
-        "watchedDirectories/activeTransfers/baggitDirectory",
-        "watchedDirectories/activeTransfers/baggitZippedDirectory",
-        "watchedDirectories/activeTransfers/dataverseTransfer",
-        "watchedDirectories/activeTransfers/Dspace",
-        "watchedDirectories/activeTransfers/maildir",
-        "watchedDirectories/activeTransfers/standardTransfer",
-        "watchedDirectories/activeTransfers/TRIM",
-        "watchedDirectories/activeTransfers/zippedDirectory",
+        WATCHED_DIRS,
+        ACTIVE_TRANSFERS_DIR,
```
ross-spencer, line 282: Were there to be any doubt about this, (we had tests to prove the data which we've created to so rely on on AM of present), this code would also fail immediately on start-up, and then watched directories after that would simply not work for any transfer too. `black-box` tests cover AIP generation. We'd know pretty quick.

src/dashboard/src/components/api/views.py
```diff
@@ -515,9 +511,7 @@ def approve_transfer(request):
     transfer_file = watched_path.replace(
         SHARED_PATH_TEMPLATE_VAL, SHARED_DIRECTORY_ROOT
     )
-    if transfer_type in {"zipped bag", "zipfile", "dspace"} and os.path.isfile(
-        transfer_file
-    ):
+    if transfer_type in amtypes.ZIP_TYPE_TRANSFERS and os.path.isfile(transfer_file):
```
ross-spencer, line 514: As transfer types that share a property, we can pull them all together here by categorising them in our new module.

src/MCPServer/lib/server/packages.py
```diff
@@ -55,59 +57,63 @@ def _get_setting(name):
...
+ACTIVE_TRANSFERS_DIR = "activeTransfers"
+WATCHED_ACTIVE_TRANSFERS = os.path.join(
+    _get_setting("WATCH_DIRECTORY"), ACTIVE_TRANSFERS_DIR
+)
 PACKAGE_TYPE_STARTING_POINTS = {
-    "standard": StartingPoint(
+    amtypes.TRANSFER_TYPE_STANDARD: StartingPoint(
         watched_dir=os.path.join(
-            _get_setting("WATCH_DIRECTORY"), "activeTransfers/standardTransfer"
+            WATCHED_ACTIVE_TRANSFERS, amtypes.WATCHED_DIRECTORY_STANDARD
```
ross-spencer, line 67: More transfer-related properties. Could be associated with this data in a new module describing transfer types. (For the present).

ross-spencer: Why do we also change the watched directories in this PR though when we want to keep scope tight?

1. We can related these directly to the transfer type. The lookup here is based on them. That goes back to point 1 where a future state might describe these in a separate module, but we're not in the habit of moving swathes of code in one go.
2. The use of this on the left `"activeTransfers/standardTransfer"` is not system agnostic, it's not great practice, and has long been our practice to try and use `os.path.join(...)` wherever possible. It wasn't caught during PR previously, but there's an opportunity to make that change here. 
3. I would bring this up within reason in any CR, `"activeTransfers/standardTransfer"`: you're in this very line of code, why leave it in a half-state where it's improved in one aspect but this remains a pretty obvious flaw. It would also be reasonable to refuse to change it, but given the time to do it, an easy improvement.

PR #1584, opened 2020-03-06
MCPServer: relax logging on unexpected terminal links

sevein:
Our workflow has a number of terminal links that may actually continue doing
work, e.g. in case of errors. Under these circumstances, MCPServer logs an
ERROR but it allows processing to continue assuming that termination happens
somewhere down the same chain.

E.g. "Move transfer to backlog" is one of the possible ends of a transfer
package, unless the operation fails in which case we've manually confirmed that
the error path terminates properly - this last step is important otherwise our
queues are exhausted blocking system processing generally.

The ERROR logging level is misleading for sysadmins since we don't expect them
to correct the situation. This commit lowers the level to WARNING.

```
Unexpectedly received another job on package completion. Please verify the value of `end` in the workflow. Link abd6d60c-d50f-4660-a189-ac1b34fafe85.
```

amflow can help identify anomalies, e.g. these are "false positives" found with
`amflow check` at the time of writing this commit:

    [abd6d60c-d50f-4660-a189-ac1b34fafe85]
    Move transfer to backlog [children=1] [refsWD=false]

    [16415d2f-5642-496d-a46d-00028ef6eb0a]
    Create SIPs from TRIM transfer containers [children=2] [refsWD=true]

    [653b134f-4a37-4578-a286-7f2072e89f9e]
    Store DIP [children=1] [refsWD=false]

We've confirmed that the alternative paths take care of termination.

In the long term, it may be preferably to explore other ways to signal end of
processing.

Connects to https://github.com/archivematica/Issues/issues/1108.

### DIFF
--- a/src/MCPServer/lib/server/queues.py
+++ b/src/MCPServer/lib/server/queues.py
@@ -185,7 +185,7 @@ class PackageQueue(object):
 
         if job.link.is_terminal:
             package_done_callback = functools.partial(
-                self._package_completed_callback, job.package
+                self._package_completed_callback, job.package, job.link.id
             )
             result.add_done_callback(package_done_callback)
 
@@ -196,9 +196,8 @@ class PackageQueue(object):
         """
         self.shutdown_event.set()
 
-    def _package_completed_callback(self, package, future):
-        """Called by an executor on completion of a package. Marks the package
-        as inactive and schedules a new package.
+    def _package_completed_callback(self, package, link_id, future):
+        """Marks the package as inactive and schedules a new package.
 
         It is assumed that a package is only complete when a terminal link is
         hit. When we hit a terminal link the future should not contain the next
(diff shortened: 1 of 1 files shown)


sevein:
I've created a full graph generated with `amflow export --full` (not hiding frequent links), https://gist.github.com/sevein/7ce51848750a8654c33360d43ed389bd. This is useful to identify links that we've tagged as terminal, e.g. "Stored DIP" that also have a hidden exit path pointing to another terminal link.

image

CC @jorikvankemenade.

jraddaoui (approved):

src/MCPServer/lib/server/queues.py
```diff
@@ -196,20 +196,21 @@ def stop(self):
         """
         self.shutdown_event.set()
 
-    def _package_completed_callback(self, package, future):
-        """Called by an executor on completion of a package. Marks the package
-        as inactive and schedules a new package.
+    def _package_completed_callback(self, package, link_id, future):
+        """Called by an executor on completion of a job.
```
jorikvankemenade: Nit: maybe move this to the body of the commend and move the first line of the comment here.

jorikvankemenade (approved):
LGTM

PR #1587, opened 2020-03-10
MCPServer: fix update logic in package initializer

sevein:
It turns out that the update logic in the get_or_create_from_db_path method is
needed in some rare situations. We've seen low rate of errors in test
environments where the db location attribute is out of sync - while this commit
has shown zero error rates during testing hundreds of transfers.

Connects to https://github.com/archivematica/Issues/issues/1141.
Connects to https://github.com/archivematica/Issues/issues/1111.

### DIFF
--- a/src/MCPServer/lib/server/packages.py
+++ b/src/MCPServer/lib/server/packages.py
@@ -721,6 +721,11 @@ class SIPDIP(Package):
                     "diruuids": False,
                 },
             )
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Thanks @sevein!

Those two tests could parameterized into one but, considering the different model and property, it may not be worthy.

sevein:
Yeah, I agree. I didn't even think about it, thanks.

PR #1589, opened 2020-03-27
SIP Arrangement: stop using completedTransfers

sevein:
This commit updates our haphazardly structured SIP Arrangement functionality
so the transfer(s) extracted from Backlog are located in `sharedDir/tmp` instead
of using `completedTransfers`. The latter approach was problematic because it's
a watched directory triggering workflow chains inadvertently.

Connects to https://github.com/archivematica/Issues/issues/1157.

### DIFF
--- a/src/MCPClient/lib/clientScripts/copy_transfer_submission_documentation.py
+++ b/src/MCPClient/lib/clientScripts/copy_transfer_submission_documentation.py
@@ -32,6 +32,8 @@ django.setup()
 from bag import is_bag
 from main.models import File, SIP
 
+from archivematicaFunctions import find_transfer_path_from_ingest
+
 
 def call(jobs):
     for job in jobs:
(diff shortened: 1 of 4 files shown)


replaceafill (changes requested):
I think the `tmp` path changes in both jobs should only be applied in a backlog arrangement workflow.

They affect the other workflows (standard, unzipped bag, dataverse, etc). I [ran the `black-box` AMAUAT tag](https://github.com/artefactual/archivematica/files/4399412/amauat-failures.txt) on this branch and noticed the `Copy transfer submission documentation` job is failing consistently like this (unzipped bag example):

unzipped-bag-error

I think the `Copy transfers metadata and logs` job succeeds because of [this fallback logic it has](https://github.com/artefactual/archivematica/blob/148bed42897163a5c85603eb96cf3caeba838020/src/MCPClient/lib/clientScripts/copy_transfers_metadata_and_logs.py#L64-L65).

sevein:
Thanks for doing that, Douglas. Please take another look! Tests are passing now.

replaceafill (approved):
LGTM :+1:

PR #1596, opened 2020-04-13
Dashboard: Add processing config selector to GUI

tw4l:
This PR adds a dropdown component to the "Start transfer" button that enables users to select which processing configuration they would like to use with their transfer, as well as a new corresponding API endpoint that lists the available processing configurations. It was also necessary to change the base Archivematica template's doctype to HTML5 so that the Bootstrap "Split button dropdown" component would render correctly (Bootstrap [requires HTML5](https://getbootstrap.com/docs/3.4/css/#overview-doctype)).

Connected to https://github.com/archivematica/Issues/issues/1166


src/dashboard/src/components/api/views.py
```diff
@@ -806,6 +806,14 @@ def processing_configuration(request, name):
         return django.http.HttpResponse(content, content_type="text/xml")
 
 
+@_api_endpoint(expected_methods=["GET"])
+def processing_configurations(request):
+    """Return list of names of available processing configurations."""
+    config_dir = helpers.processing_config_path()
+    processing_configs = [filename[:-17] for filename in os.listdir(config_dir)]
+    return helpers.json_response({"processing_configurations": processing_configs})
```
ross-spencer: Nice pragmatic approach to this. Great to see it. When this is merged, we just need to make sure it is added to the API docs at least: https://wiki.archivematica.org/Archivematica_API#Processing_Configuration (they're not much, but it's what we've got for now!) Michelle is working on a modern rendering.

src/dashboard/tests/test_api.py
```diff
@@ -481,6 +481,16 @@ def setUp(self):
...
+        install_builtin_config("automated")
+
+    def test_list_processing_configs(self):
+        response = self.client.get(reverse(views.processing_configurations))
+        assert response.status_code == 200
+        payload = json.loads(response.content.decode("utf8"))
+        processing_configs = payload["processing_configurations"]
+        assert len(processing_configs) == 2
+        expected_names = ["default", "automated"]
+        assert all([a == b for a, b in zip(processing_configs, expected_names)])
```
ross-spencer: If possible we should avoid short names for variables, even if it seems trivial code can always be most descriptive. perhaps, `actual == expected for actual, expected` is a good fit here?

ross-spencer: It's reasonable to assume here that because the JSON you're returning is returning an array, and you're comparing it to an array, then you're comparing like for like, but `os.listdir` is a bit of a wildcard when the request is made to the processing configurations endpoint. For tests to pass consistently on environments then I think we need to order the result of the request here, or in the view itself. Alternatively, some fuzzier logic might need to be put in place here, like asking if each processing config is part of a set and then removing it until `len == 0`. There may be other approaches you can think of.

src/dashboard/src/components/api/views.py
```diff
@@ -806,6 +806,14 @@ def processing_configuration(request, name):
         return django.http.HttpResponse(content, content_type="text/xml")
 
 
+@_api_endpoint(expected_methods=["GET"])
+def processing_configurations(request):
+    """Return list of names of available processing configurations."""
+    config_dir = helpers.processing_config_path()
+    processing_configs = [filename[:-17] for filename in os.listdir(config_dir)]
```
ross-spencer: I wasn't sure what we were deleting here. It looks to be `ProcessingMCP.xml`? I feel like a const describing that, might help readability, and then `[filename[:len(MYCONST)]...` might help describe what we're doing better? What do you think?

ross-spencer (commented):
Nice work @tw4l. 

I've left comments on a few things that might be good to consider. There are a few small things around the reliability of tests that should probably be resolved. I've asked your opinion on a few small things.

Regarding the feature, then there are two things I might consider from an end-user perspective:

1. When I try to select a processing config, the text input cursor shows, rather than a pointer. Can this be overridden in the style sheets at all? Do you see the same?
2. Now that I have the power to select a processing configuration I really want to make sure that I select the right one. How do I know aside from the label `default` that this is always the one I am selecting if I don't choose another? I started a transfer earlier and it did make me wonder, what if I'm accidentally selecting the last option I chose (I had previously chosen `automated`? Automated is also the first in my list. Have you any methods you can think of to make it absolutely clear the default is in-fact default? Or let me know which config is being engaged when I've finally pressed "go" on a transfer? (NB. Given the "default" string is derived from the file-listing (if i read correctly), it won't be translated. Would it change any of the design decisions around highlighting the default, if for example we wanted folks to be able to select default ( `défaut` or `standaard`) in their own language?)

This might be nothing, or something for Peter and yourself and Sarah R to consider. And of course the client too. It's just something popping into my head as I think about how I'm working with the tool. 

In terms of final merge process. I recognize this is a feature branch for now. We have a bit of an unresolved approach to versioning, and I notice the API header states: `X-Archivematica-Version: 1.11.0`.

Would you be able to follow-up with the SAD team about what their expectations are with regards to the first features added as part of 1.12. It might be that you'll be asked to write the migration to update the version number which would (might) need to be merged first, e.g. [migration 0074](https://github.com/artefactual/archivematica/blob/3e52494735ebfeb0cabc477d95d692034f4b3142/src/dashboard/src/main/migrations/0074_version_number.py) but so far we haven't consistently done this at the beginning/middle/end of a set of release sprints. It seems reasonable to find best-practice and what it means going forward. It seems reasonable to me to perform this migration at the beginning but it might not be what folks want. There is also a [general ticket](https://github.com/archivematica/Issues/issues/383) in the issues repo that asks for better ways to do this, over pretty static migrations.

src/dashboard/tests/test_api.py
replaceafill: > There may be other approaches you can think of.

Maybe mocking `os.listdir` just for this test to ensure you always get the same file names?

src/dashboard/frontend/app/front_page/transfer_browser.html
```diff
@@ -29,17 +29,34 @@
...
+          data-toggle="dropdown"
+          aria-haspopup="true"
+          aria-expanded="false">
+      <span class="caret"></span>
+      <span class="sr-only">Toggle Dropdown</span>
+    </button>
+    <ul class="dropdown-menu">
+      <li class="processing-config-choice"
+          ng-repeat="processing_config in vm.transfer.processing_configs">
+        <a ng-click="vm.transfer.start(processing_config)">{{ "Start with" | translate }} "{{ processing_config }}" {{ "config" | translate }}</a>
```
replaceafill: I think an interpolated string would be better here:

```
<a ng-click="vm.transfer.start(processing_config)" translate>Start with {{ processing_config }} config</a>
```

You'd get a single translation string:

```
msgid "Start with {{ processing_config }} config"
msgstr ""
```

instead of multiple ones:
```
msgid "Start with"
msgstr ""
...
msgid "config"
msgstr ""
```

which provides better context for translators.

@tw4l Could you please also make sure once this is approved to [extract and update the translation files](https://github.com/artefactual/archivematica/tree/55af1eaa8cd6686903fc72c2b3790bb828f77930/src/dashboard/frontend#translations)? I can help if you need.

tw4l: @replaceafill I'd be happy to update the translation files. Thank you for offering to help! Would it make sense to do that before merging so that the compiled messages are part of the PR?

src/dashboard/src/components/api/views.py
tw4l: I've implemented as you suggested and it is a big readability improvement!

src/dashboard/tests/test_api.py
tw4l: Sorting the lists both in the view and the test (in commit 407798d) should solve this issue.

tw4l: Absolutely! Implemented in 426d28d.

src/dashboard/frontend/app/front_page/transfer_browser.html
replaceafill: @tw4l Yes, that makes sense. Thanks!

src/dashboard/tests/test_api.py
replaceafill: The reason I prefer mocking to guarantee iteration results is that the test won't break or need updating when/if we add a new standard processing configuration (default, automated, somethingnew).

tw4l: I'm happy to change it if you'd like, but this test shouldn't break if we add a new standard processing configuration in AM. It's kind of being mocked already. See the `setUp` method for the test class:

```
def setUp(self):
        self.client = Client()
        self.client.login(username="test", password="test")
        helpers.set_setting("dashboard_uuid", "test-uuid")
        settings.SHARED_DIRECTORY = tempfile.gettempdir()
        self.config_path = os.path.join(
            settings.SHARED_DIRECTORY,
            "sharedMicroServiceTasksConfigs/processingMCPConfigs/",
        )
        if not os.path.exists(self.config_path):
            os.makedirs(self.config_path)
        install_builtin_config("default")
        install_builtin_config("automated")
```
Since the test is using a tempdir as the sharedDirectory, and the "default" and "automated" configs are manually added using the `install_builtin_config` function from archivematicaCommon's `processing` module, even if we add new built-in processing configs moving forward, the test won't be affected unless we specifically modify that `setUp` method.

I added the `install_builtin_config("automated")` to accommodate this test but otherwise the `setup` method is unchanged from what was there before.

replaceafill: Ah! Thanks Tim. Sorry that I didn't catch the old behaviour from the diff. That should cover my concern then :+1:

tw4l:
Thank you for the helpful comments, @ross-spencer and @replaceafill!

I've tried to address all of the in-line comments. On the other issues you've raised:

* When hovering over items in the dropdown list, the user will now see a pointer cursor.
* UX issues related to knowing which processing configuration is being used:
	* Immediately after starting a transfer, a dismissable info alert will now appear just above the transfer browser indicating which processing configuration was used to start the transfer.
	* I've added translatable tooltips to the "Start transfer" and dropdown buttons that should help clarify their purpose for users.
	* I will also update the documentation to make clear that the "Start transfer" button always uses the "default" configuration, same as it has in previous versions of Archivematica.
* Message files have been updated to reflect the new UI strings in the Angular frontend app.

I've also separately bumped the versions of Archivematica and the Storage Service in their `qa/1.x` and `qa/0.x` branches, respectively.

I think there is an outstanding issue about the translatability of the processing configuration names themselves (and especially of our hard-coded configurations, "default" and "automated"). Their names are indeed derived from their filenames, not just in this work but everywhere in Archivematica. After looking into it a bit further and discussing with Peter, the work that would be necessary to address that seems too large to tackle at this time but I will file an issue in the Archivematica/Issues repository so that it's on our radar moving forward.

Please let me know if you'd like to see further changes, and thanks again!

tw4l:
It looks like the commit updating the frontend translation messages (https://github.com/artefactual/archivematica/commit/d21d39057d53c5db847da3400a7b90f0bb2258f8) on the branch isn't listed in the PR for some reason. Maybe related to the GH service hiccup earlier today?

tw4l:
Issue above is resolved! Pushing an extra commit brought the last one into the PR as well. Maybe a rare GH bug?

ross-spencer (approved):
Great PR, thanks Tim. Some really nice decisions made which clear up a number of my concerns, and I think will be a great feature! 

I noticed in comparison to the Backlog PR the resolved conversations were still interactive, so I don't know if there's a process thing, or, maybe more likely Github got confused? But yep. all worked well.

Thanks too for logging the new ticket against the processing configuration options. 

Nice work! LGTM :+1: 

For merging, we have a merge guideline [here](https://wiki.archivematica.org/Merging). You'll want to rebase this into a single feature commit, rebase against origin/qa/1.x and then merge. Being on a wiki, if you spot any improvements you want to make to the guideline please feel free.

PR #1597, opened 2020-04-14
Dashboard: Add new fields and column visibility options to Backlog

tw4l:
This PR adds information to the Backlog tab's tables for transfers and associated files and adds a column selector to the table which can be used to choose which columns are visible. More specifically, this PR introduces the following changes:

* Elasticsearch changes
	* Adds `size` and `accessionid` fields to the `transfers` ES index
	* Adds `pending_deletion` field to the `transferfiles` ES index
	* Adds logic to cascade `pending_deletion` status updates to to all related files in the `transferfiles` ES index when a user requests deletion of a transfer in the backlog
* DataTables changes
	* Adds "Size", "Accession number", and "Status" columns to the `transfers` table
	* Adds "Accession number" and "Status" columns to the `transferfiles` table
	* Upgrades DataTables to version 1.10.13
	* Adds the [DataTables Buttons extension](https://datatables.net/extensions/buttons/)
	* Uses the [colvis Button](https://datatables.net/reference/button/colvis) to allow users to control which columns are visible
	* Persists column visibility settings to the DashboardSettings model/database using the native DataTables [stateSaveCallback](https://datatables.net/reference/option/stateSaveCallback) and [stateLoadCallback](https://datatables.net/reference/option/stateLoadCallback) options
	* Sets default column visibility options for the first time the DataTable is initialized
* Other changes
	* Adds new URLs and views in the dashboard's `backlog` component (essentially thin IO layers that wrap around `helpers.get_setting` and `helpers.set_setting`) to handle the server-side logic of storing and retrieving DataTable state using the DashboardSettings model 
	* Adds tests for new views
	* Adds Font Awesome 4.7.0 to media files and base `layout.html` template. This is used to supply a Font Awesome icon for the "Select columns" button. Font Awesome 4 is already installed via npm in the `frontend` Angular app but was not accessible from Django templates and `src/dashboard/src/media/js/backlog/backlog-search.js` (where the DataTable is initialized). There may be a better way to handle this. I've used Font Awesome 4 for consistency until an Archivematica-wide migration to Font Awesome 5 could be done, as upgrading the `frontend` app seemed out of scope for this work and the docs for upgrading make a point that [4 and 5 should not be used together on the same project](https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4).
	* Updates `djangojs.po` message files in `src/dashboard/src/locale` to account for changes in JS, including new UI strings

Connected to https://github.com/archivematica/Issues/issues/1167


src/dashboard/src/media/css/backlog.css
```diff
@@ -11,6 +11,17 @@
   margin-top: 2px;
 }
 
+.dt-buttons {
```
ross-spencer, line 14: A tiny thing, but I notice the naming convention above is `dataTables.` I also prefer long names myself. It's one less question for the reader, "what's this initialism again?" Can we consider changing this to `dataTable.button`and below, as well?

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -191,11 +190,7 @@ def main(job, transfer_id, transfer_path, created_at):
     _index_transfer(job, transfer_id, transfer_path)
 
     logger.info("Calculating size...")
-    size = 0
-    for dirpath, _, filenames in scandir.walk(transfer_path):
-        for filename in filenames:
-            file_path = os.path.join(dirpath, filename)
-            size += os.path.getsize(file_path)
+    size = get_dir_size(transfer_path)
```
ross-spencer: If you inspect the object returned from ` _create_bag(transfer_id, transfer_path)` then I think you'll have [an object](https://github.com/LibraryOfCongress/bagit-python#create) that you can get the payload-oxum from directly, e.g.
```
mybag = _create_bag(transfer_id, transfer_path)
size = mybag.get("Payload-Oxum", 0) 
```
I'm not sure if there is any other info from the bag we need here? But worth considering.

src/MCPClient/lib/clientScripts/move_to_backlog.py
```diff
@@ -16,22 +16,21 @@
...
+from archivematicaFunctions import get_dir_size, get_setting
 from custom_handlers import get_script_logger
 from databaseFunctions import insertIntoEvents
 import elasticSearchFunctions
-from main.models import Agent, File, UnitVariable
 import storageService as storage_service
 
-from bagit import make_bag
-import metsrw
+from main.models import Agent, File, UnitVariable
```
ross-spencer: We should keep this ADR in mind -- or maybe while we're early on in 1.12 push for a decision on it as it will make all our lives easier: https://adr.archivematica.org/0004-isort-import-ordering.html before then, it's probably not worth moving the imports around too much. The reason it crossed my mind is here, I wonder if `import storageService ...` would actually appear below `from main.models` alphabetically?

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -25,6 +25,7 @@
 """
 from __future__ import absolute_import, print_function
 
+import bagit
```
ross-spencer: Depending on your `bagit` investigation in `move_to_bacllog.py` I might consider leaving these helpers in here for now. I'll add some more detail in the elasticsearch part of the code to consider too.

src/MCPClient/lib/clientScripts/move_to_backlog.py
ross-spencer: Reading through the logic a little more, I'd be tempted to amend the indexing function signature to either pass the bag object through (depending on if you need more information) or just the size, and then when we don't need to access the disk again to read the bag and extract the payload-oxum. It would be a small improvement in speed I imagine. But moreso, I feel like doing the same thing in multiple places doesn't quite feel right. But what do you think? We have the database as an option too, and I'll make that note below.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -29,22 +29,20 @@
...
 import sys
 import time
-from lxml import etree
 
 from django.db.models import Min, Q
-from main.models import File, Identifier, Transfer
+from elasticsearch import Elasticsearch, ImproperlyConfigured
+from elasticsearch.helpers import bulk
+from lxml import etree
+from six.moves import range
```
ross-spencer: Again, this is useful clean-up but we don't really have to worry about this too much going forward (I don't think?) Did you utilize isort to get you here?

src/dashboard/src/components/backlog/views.py
```diff
@@ -185,14 +195,27 @@ def search(request):
 
     results = [x["_source"] for x in hits["hits"]["hits"]]
 
+    def format_values(k, v):
+        """
+        Format search result values for display in Backlog tab.
+        """
+        if k == "size":
+            return "{0:.2f} MB".format(float(v))
+        return v
```
ross-spencer: There's a helper we've been using in a few views to display user friendly sizes, e.g. [here](https://github.com/artefactual/archivematica/blob/4f2abe13a40ae330462a01d38e1ab5b3ce46d950/src/dashboard/src/components/administration/views.py#L227-L230), it might be good to be consistent with those other places to use it here as well? Ref: [filesizeformat](https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#filesizeformat). I think if you were observe we're not consistent across the code-base, you'd be correct!! But we'll get there.

src/dashboard/src/components/backlog/views.py
```diff
@@ -17,23 +17,22 @@
...
 from django.utils.translation import ugettext as _
+import requests
 
 import elasticSearchFunctions
 import storageService as storage_service
 
-from components import advanced_search
-from components import decorators
-from components import helpers
+from components import advanced_search, decorators, helpers
```
ross-spencer, line 39: Nice.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -673,10 +676,20 @@ def index_transfer_and_files(client, uuid, path, printfn=print):
         if dt:
             ingest_date = str(dt.date())
 
+    transfer_size = get_dir_size(path)
```
ross-spencer: As mentioned, I'd probably consider extending the function signature to send the bag object, or size value along here. I think that would be the path of least resistance without repeating ourselves. 

I think there are clues in this function to alternatives, then extending the model seems sensible. I feel like lines `669-671` are pointing to that too and working into the function its own heuristic to get hold of information that should be persisted somewhere else. I feel like this would be my second favorite option to the bag constructor, but happy to chat more. And I think it would take a brief discussion in the AM channel with the team too. 

I don't think we try to store the accession number or ingest date in the baginfo do we? Could they also provide some of the information here? [Updating the bag info](https://github.com/LibraryOfCongress/bagit-python#update-bag-metadata). (The `External-Identifier: <uuid>` is a value Wellcome added in there for similar purposes, the structure is definitely something that could eventually be considered as a place to describe and retrieve information about our packages.)

src/dashboard/src/media/js/backlog/backlog-search.js
```diff
@@ -4,34 +4,106 @@ $(document).ready(function()
...
+      return $('#id_show_files').prop('checked') ? 'transferfiles/' : 'transfers/';
+    }
+
+    function get_default_hidden_columns() {
+      return $('#id_show_files').prop('checked') ? [2, 3] : [3, 4, 6];
+    }
+
+    function get_action_column_index() {
+      return $('#id_show_files').prop('checked') ? [4] : [7];
+    }
```
ross-spencer: This part of the code is new to me, so I wonder if you can help me understand it a bit more? Coming in cold, I can get a feel for what is being done here in the function definitions. The titles are descriptive. I am wondering though, is there an alternative to slice(?) notation in lines 25 and 29? That would ultimately make it easier to see what's going on. But I don't know if it is possible?

src/dashboard/src/components/backlog/views.py
```diff
@@ -185,14 +195,27 @@ def search(request):
...
+        """
+        Format search result values for display in Backlog tab.
+        """
+        if k == "size":
+            return "{0:.2f} MB".format(float(v))
+        return v
+
+    formatted_results = []
+    for result in results:
+        formatted_dict = {k: format_values(k, v) for k, v in result.items()}
```
ross-spencer: I wonder if the list comprehension is absolutely necessary? Would `result["size'] = format_values(result.get(size, 0))` have a similar affect before appending? Am I missing something about the data structure we're dealing with here that would prevent that? I feel like it could reduce complexity and improve readability a little, but what do you think?

src/dashboard/src/components/backlog/views.py
```diff
@@ -255,3 +278,34 @@ def download(request, uuid):
     return helpers.stream_file_from_storage_service(
         storage_service.download_file_url(uuid)
     )
+
+
+def save_state(request, table):
```
ross-spencer, line 313: Both these functions work really well. Nice use of these features.

src/dashboard/tests/test_backlog.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
```
ross-spencer, line 2: This is good, it helps with py3 compatibility. Was it why you did it? I think it reasonable you add it to the other scripts you're touching above, e.g. elasticsearchFunctions where it is missing, if so. 

I'd add the `utf-8` header too.

```
# -*- coding: utf-8 -*-
from __future__ import absolute_import
```

You can get more python 3 compatibility help from this command if you have `pylint` installed: `pylint --py3k lumberjack.py`.

Only if it's an angle you were coming from will it be useful :)

src/dashboard/tests/test_backlog.py
```diff
@@ -0,0 +1,49 @@
...
+        assert payload["columns"][0]["visible"] is True
+        assert payload["columns"][3]["visible"] is False
+
+    def test_load_datatable_state_404(self):
+        """Non-existent settings should return a 404"""
+        response = self.client.get(reverse(views.load_state, args=["nonexistent"]))
+        assert response.status_code == 404
+        payload = json.loads(response.content.decode("utf8"))
+        assert payload["error"] is True
+        assert payload["message"] == "Setting not found"
```
ross-spencer, line 49: These all seem pretty good and sensible! Nice work. Always great to increase our coverage.

ross-spencer (changes requested):
Thanks @tw4l this is all really good work and I think a satisfactory result from an end-user perspective already.

I've left a lot of comments, but I hope it doesn't contradict that view :) 

There are a few style comments. 

There is one main thing in the code that I'd like to consider how we can change which is the `get_size` functions you're introducing. The approach is great. But I have a few thoughts described about how we can remove the redundancy in trying to access those statistics twice the same way. I don't think it needs to be too much effort, but should just avoid adding a few lines and tests as part of this PR. There are a couple of approaches I've tried to describe. 

Second, the law of unintended consequences!! Now, I don't know if this needs to be fixed or not. It seems like it might reasonably be considered out of scope but certainly worth a conversation with folks.

The "status" field being indexed now, and being brought to the fore means it is easier to see where there is a bit of a flaw in AM's current wiring. 

If I have this correct (I've done a little testing). Then the deletion request from the backlog updates the elastic search index. Also, when the package is finally deleted from the storage service, it will no longer be returned, it is deleted, and so throughout, the package status remains up to date: `stored` -> `pending deletion` -> `no longer visible`.

If I do that the other way around, i.e. from the storage service, we haven't any functionality to ask about the package until deletion happens. So, up until the point of deletion the package is only ever stored, until is is deleted. I don't know if the index is updated at all that way around, so I can potentially download something and get a 500 error. 

It seems like we're missing something in AM to query the packages endpoint and seek the package status to update it when this table loads. I guess:

1. Can you recreate something similar your end? 
2. Is it something you can consider making sure works with the emphasis on the backlog tab? 

There are helpers in `amclient` to make this easier, but yep. It would obviously mean more work than anticipated.

src/MCPClient/lib/clientScripts/move_to_backlog.py
tw4l: I was actually working under the impression that this ADR was approved! As it was listed under "Currently accepted records." So I was trying to incrementally move toward the goal as I touched things. I only just noticed that the status on that ADR is actually "Status: proposed." Given that, I agree, we should wait until a decision is made before rearranging these imports, so I'll revert these to how they were. (That's also a good question about how to order "import" vs. "from ..."! I don't know the answer.)

src/MCPClient/lib/clientScripts/move_to_backlog.py
tw4l: These are great observations! I agree that we should avoid doing the same thing multiple places, and that we should be passing this information along rather than re-calculating it later. I also lean toward amending the indexing function signature. I'll look into amending the client script and corresponding indexing function to fit that model, with an eye toward making sure whatever changes are made also play nicely with our re-indexing scripts for existing transfers in the backlog.

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: I did! My code editor auto-applied isort due to a plugin - I've since fixed that in the plugin config, but figured might as well keep the changes since I was working under the assumption that the related ADR was approved but just not implemented yet.

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: Thanks, Ross! I'll keep this all in mind. On your last point, we are not currently storing the accession number or ingest date in the bag-info. Here's the contents of a bag-info file for a transfer I just saved into the backlog on my local dev instance of AM:
```
Bag-Software-Agent: bagit.py v1.7.0 <https://github.com/LibraryOfCongress/bagit-python>
Bagging-Date: 2020-04-22
External-Identifier: 9a033c72-8338-4863-8639-32af17d9f1db
Payload-Oxum: 1651855.18
```
The `External-Identifier` is the transfer UUID; the other info is generated by the bagit-python library.

tw4l: The `Bagging-Date` will likely correspond to the date that a transfer was started in many cases, but may not for especially large or long-running transfers...

src/dashboard/src/components/backlog/views.py
tw4l: Good to know! I'll bookmark that. I was trying to make the `size` property in the `transfers` index match how it's handled in the `aips` index, where it is [stored as a float that represents the size in MBs](https://github.com/artefactual/archivematica/blob/4f2abe13a40ae330462a01d38e1ab5b3ce46d950/src/archivematicaCommon/lib/elasticSearchFunctions.py#L450). So the formatting is about stripping off the trailing decimals rather than converting from bytes to another unit. My thinking was that since we don't want to force users with lots of data to re-index it's unlikely that we'll change how we're storing this in the `aips` index, and so we may as well make the two indices consistent. However I'm happy to make changes if you think it'd be better to handle differently.

src/dashboard/src/components/backlog/views.py
tw4l: You're quite right! Classic case of overengineering :sweat_smile:

src/dashboard/src/components/backlog/views.py
tw4l: Thank you!

src/dashboard/src/media/css/backlog.css
tw4l: Your point on readable variable names is well-taken! In this case, the CSS is targeting a DOM element created internally by the DataTable by the `.dt-buttons` class it's given by DataTables/Buttons. Not much I can do about the class name unfortunately. I could add a comment to the CSS to clarify what this is targeting if that'd be helpful.

src/dashboard/src/media/js/backlog/backlog-search.js
tw4l: I'd be happy to! These are using JavaScript's [ternary operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator), basically a short-hand for conditional logic.

`get_default_hidden_columns` could be rewritten without the ternary operator as:

```
function get_default_hidden_columns() {
  if ( $('#id_show_files').prop('checked') ) {
    return [2, 3];
  } else {
    return [3, 4, 6];
  }  
}
```
Personally I find the ternary form easier to read, but that might just be due to familiarity. I've noticed it used elsewhere in the Angular frontend app as well, but I'm not sure if we have standard practice/style for JS in Archivematica?

If you mean the logic rather than the syntax, these functions are returning column indices for the columns in the DataTable, for use in the DataTable initialization function. `get_default_hidden_columns` returns an array containing the indices for the columns which should be hidden by default if there's no saved state (we decided to hide all of the new columns by default). These indices change depending on whether the "Show files?" button is checked (and hence the files table, rather than transfers table is being displayed). `get_action_column_index` is returning an array containing the column index of the "Actions" column for the table that's currently displayed (transfers or files), so that I can disable sorting on that column.

Maybe `get_default_hidden_columns` should be renamed to `get_default_hidden_columns_index`? I'd also be happy to write some in-line comments explaining what's going on.

src/dashboard/tests/test_backlog.py
tw4l: I used one of the other test scripts in that directory as a template and kept the utf-8 header and `from __future__ import absolute_import` from the other file, figuring it couldn't hurt with 2/3 compatibility. 

I just checked and these are already in elasticsearchFunctions, so we should be good there. Thanks for the tip on pylint!

src/dashboard/src/media/css/backlog.css
ross-spencer: That's okay! And thanks Tim, that makes sense. A failure to context switch on my part.

src/MCPClient/lib/clientScripts/move_to_backlog.py
ross-spencer: It'll bring benefits once it's accepted for sure. Interesting this was from `isort` it wasn't quite what I expected, but it's automatic/less-subjective/etc. all good things. One less thing to worry about.

src/archivematicaCommon/lib/elasticSearchFunctions.py
ross-spencer: It took me a while to make the logic leap, but that's cool.

src/dashboard/src/components/backlog/views.py
ross-spencer: I feel from a long term perspective, And just in general if we have something that's reactive and future facing, e.g. as sizes of SIPs grow - if we can have another library that can deal with this formatting for us (rather than having a hard coded MB) then that would be great. We're also saying here "what's a human readable string?" So I feel you can do that with the Django helper. 

In either case there's a split between the model and the view, so we should be able to format whatever we're retrieving without re-indexing providing we know the data we're retrieving. So I think that might be a positive/okay/ 

That doesn't provide a way-forward though, and I hear you too, around consistency. 

This is what i see if I put into backlog a couple of small-ish files: 

image

I feel like this is an end-user question, so probably just worth asking the analyst on this project if they'd consider this desirable now or not. It might provide some ideas.

And just, a question given your familiarity with the code in this work. Given we're creating a sort of "display dependency" for the sake of consistency -- is it sensible to have a helper shared across Backlog and Archival Storage that at least codifies that there's a connection here?

src/dashboard/src/media/js/backlog/backlog-search.js
ross-spencer: Ah, yep, that's increasingly familiar to me. 

I'm not going to suggest changing from ternary to long-hand here especially if it's elsewhere in our code. 

Borrowing from the longhand though: `return [3, 4, 6];` -- my question was more specifically about the indices here. Do we only have numbers to refer to these columns? Have we any tools in our kit to make these self-describing? Do we only have integers to refer to columns or do we have access to a column name at all? e.g. key/value. 

No worries at all if not.

src/dashboard/tests/test_backlog.py
ross-spencer: A lesson for me in clarity, e.g. elasticsearchFunctions, or any others, e.g. `src/MCPClient/lib/clientScripts/move_to_backlog.py` etc. :wink: but since you're not specifically thinking about `py3` compatibility, that's a-okay. This would normally be out of scope of a CR review I feel (again unless there was a clear motivation). Nothing to do here. It'll get picked up when the Client scripts are migrated.

tw4l:
Thanks @ross-spencer for your insightful comments above! As you've seen, I started by addressing the inline comments yesterday and am planning to spend the morning today focusing on your larger question around state consistency when requesting deletions from the Storage Service rather than the backlog tab - that's a good catch!

I'll give an update here when the PR is ready for a re-review or if other questions come up in the meantime.

src/dashboard/src/media/css/backlog.css
tw4l: No worries!

src/MCPClient/lib/clientScripts/move_to_backlog.py
tw4l: It's definitely possible I may have rearranged isort's order to bring more in line with the example/recommended format in the ADR - they don't seem to *quite* always line up, especially when something like `django.setup()` is in the imports.

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: I think my comment was ambiguous! I *was* working under that assumption, but given that the ADR isn't approved, I'll return this to how it was for now. Thanks!

src/dashboard/tests/test_backlog.py
tw4l: You're kind but I'm going to go ahead and blame the reader (me) for this misunderstanding :smile:

src/MCPClient/lib/clientScripts/move_to_backlog.py
tw4l: Hi Ross! I made the changes you suggested in commit 3fff680. `get_dir_size` helper function has been refactored as `get_bag_size` and accepts a bag object and the path to the bag as arguments (the latter in case `Payload-Oxum` as an optional attribute is not present in bag-info.txt and we have to calculate size with a dirwalk), and `index_transfer_and_files` now expects the transfer's size as an argument rather than re-calculating.

src/archivematicaCommon/lib/archivematicaFunctions.py
tw4l: See above - I refactored as a general helper function to get a package size from a Bag object's payload-oxum, with a fallback for cases where there is no `Payload-Oxum`

src/dashboard/src/media/js/backlog/backlog-search.js
tw4l: Per the [DataTable docs](https://datatables.net/reference/option/columnDefs.targets) you should be able to target a column by a string matching the corresponding `<th>` text in the table header. In practice I found this doesn't work very reliably, especially when switching languages in the interface, whereas the integer values work consistently.

src/dashboard/src/components/backlog/views.py
tw4l: I discussed this briefly with the project analyst today and the decision was to take the question up with the product manager. I'll do that and update here when there's an answer.

Note to self: It should be possible to use [HTML5 data attributes](https://www.datatables.net/release-datatables/examples/advanced_init/html5-data-attributes.html) to keep sorting functional even if we auto-format the display using `filesizeformat`.

If we do create this sort of "display dependency" I will implement as a shared helper. But perhaps that's not necessary.

tw4l: After internal conversation, I've made changes to store size in the `transfers` index as bytes ("long" type in ES), and format the data for the DataTable using `filesizeformat`.

tw4l:
I've created an issue in the Archivematica Issues repository describing the problem of the Storage Service status and Elasticsearch `pending_deletion` fields getting out of sync when package deletions happen via the Storage Service: https://github.com/archivematica/Issues/issues/1189

This issue has been surfaced by our Backlog work, but is confirmed as being an issue for the Archival Storage table as well.

tw4l:
With commit 97ae098, the search view now checks each transfer's status with the Storage Service and updates ES appropriately when a package has been marked as pending deletion, regardless of whether the deletion request was made through the Archivematica GUI or directly with the Storage Service.

I've tested in my local environment and can confirm that the status is updated in the indices and the statuses between ES and the Storage Service are staying appropriately in sync. Because it takes time to update Elasticsearch, the status will typically update in the table after the next page refresh or new search.

The performance of this could be improved eventually by moving the ES updating logic closer to where the status changes happen in the Storage Service, but for now I think this solution solves our immediate problem and this is ready for re-review.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -286,6 +285,40 @@ def format_subdir_path(dir_path, path_prefix_to_repl):
...
+        for filename in filenames:
+            file_path = os.path.join(dirpath, filename)
+            size += os.path.getsize(file_path)
+    return size
+
+
+def get_bag_size(bag, path):
+    """Return size of BagIt Bag, using Payload-Oxum if present.
+
+    Payload-Oxum, like other Bag Metadata elements, is optional
```
ross-spencer: It looks like your comments are being truncated at about 64 characters? We can afford to go up to 79 characters here. `Black` in some cases in code will extend the line longer when it's code. Not sure if this is an editor setting you need to look at?

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -286,6 +285,40 @@ def format_subdir_path(dir_path, path_prefix_to_repl):
...
+
+    :param transfer_path: Bag object
+    :param path: path to Bag directory
+    :return: size in bytes (int)
+    """
+    oxum = bag.info.get("Payload-Oxum")
+    if oxum is not None:
+        return int(oxum.split(".")[0])
+    else:
+        return walk_dir(path)
```
ross-spencer: No need to change this, just a note on style. But you might find in something like `pylint` that this `else:` before the return is not needed, as you're always going to return once `if oxum is not None:` isn't satisfied.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -277,8 +276,10 @@ def _get_transfers_index_body():
                     },
                     "status": {"type": "text"},
                     "ingest_date": {"type": "date", "format": "dateOptionalTime"},
+                    "size": {"type": "long"},
```
ross-spencer, line 279: :+1:

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -685,10 +700,12 @@ def index_transfer_and_files(client, uuid, path, printfn=print):
     transfer_data = {
         "name": transfer_name,
         "status": status,
+        "accessionid": accession_id,
```
ross-spencer, line 703: For naming consistency we might have used an underscore here as well, `accession_id`.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -715,14 +745,6 @@ def _index_transfer_files(client, uuid, path, ingest_date, status="", printfn=pr
     # This should match the basename of the file.
     ignore_files = ["processingMCP.xml"]
 
-    # Get accessionId and name from Transfers table using UUID
-    try:
-        transfer = Transfer.objects.get(uuid=uuid)
-        accession_id = transfer.accessionid
-        transfer_name = transfer.currentlocation.split("/")[-2]
-    except Transfer.DoesNotExist:
-        accession_id = transfer_name = ""
```
ross-spencer, line 724: This is nice clean-up work. :+1:

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1341,6 +1364,16 @@ def mark_aip_stored(client, uuid):
 def mark_backlog_deletion_requested(client, uuid):
     _update_field(client, "transfers", uuid, "pending_deletion", True)
 
+    files = _document_ids_from_field_query(client, "transferfiles", "sipuuid", uuid)
+    if len(files) > 0:
```
ross-spencer: A nice thing about `_document_ids_from_field_query` if I read it correctly is that it will always return a list type, so if you jump straight into the loop, it won't iterate, but it won't give you a `TypeError` either, so you can avoid this length check. 

No need to change this either. Just a note that can help condense the code a little more in future.

src/archivematicaCommon/tests/test_archivematica_functions.py
```diff
@@ -0,0 +1,42 @@
...
+    bagit.make_bag(bag_dir.strpath)
+    # Replace Payload-Oxum with incorrect arbitrary value
+    bag_info = bag_dir.join("bag-info.txt")
+    size_oxum = 1056
+    bag_info.write("Payload-Oxum: {}.1".format(str(size_oxum)))
+    # Test returned value against expected
+    size_on_disk = walk_dir(bag_dir.strpath)
+    size = get_bag_size(bag_dir.strpath)
+    assert size == size_oxum
+    assert size != size_on_disk
```
ross-spencer: Am I reading this correctly, that we're testing the oxum size vs. the size on disk, and want to make sure they're not the same? I'm not quite sure what we want to assert here if so? 

I wonder if you're looking for the mock function assert_not_called to make sure `os.path.getsize(...)` isn't called? 

Ref: https://cpython-test-docs.readthedocs.io/en/latest/library/unittest.mock.html#unittest.mock.Mock.assert_not_called 

If you mock out the function call you don't have to worry about predicting what it will return either, just that the code pathway was never executed. 

Below, an assert_called_once: https://cpython-test-docs.readthedocs.io/en/latest/library/unittest.mock.html#unittest.mock.Mock.assert_called_once might complement that well.

src/dashboard/src/components/backlog/views.py
```diff
@@ -17,23 +17,26 @@
...
-from django.http import HttpResponse, Http404
-from django.shortcuts import render, redirect
+from django.http import Http404, HttpResponse
+from django.shortcuts import redirect, render
 from django.template import RequestContext
+from django.template.defaultfilters import filesizeformat
 from django.utils.translation import ugettext as _
+import requests
 
+from amclient import AMClient
```
ross-spencer, line 33: :+1:

src/dashboard/src/components/backlog/views.py
```diff
@@ -71,6 +74,31 @@ def check_and_remove_deleted_transfers(es_client):
...
+    """Check Storage Service to see if transfer has pending deletion and update ES if so.
+
+    :param uuid: Transfer UUID.
+    :param pending_deletion: Current pending_deletion value in transfers ES index.
+    :return: None
+    """
+    api_results = AMClient(
+        ss_api_key=get_setting("storage_service_apikey", None),
+        ss_user_name=get_setting("storage_service_user", "test"),
+        ss_url=get_setting("storage_service_url", None).rstrip("/"),
```
ross-spencer: I don't know if you need the explicit `None` as default here. I haven't a convincing default you can set. Though `86` probably shouldn't be `test`. And `87` will return an `AttributeError` if it really did return `None`. So you could have string type returns for all three, e.g. `""` but there might be a trickle down error to deal with. I think if blank strings as defaults hits the path on `90`:
```
    if not api_results:
        logger.warning("AIP not found in storage service: {}".format(uuid))
        return
```
Then that's probably the best approach. IDK what do you think?

ross-spencer (commented):
Thanks @tw4l there are two small things in the review comments in `backlog/views.py` and possibly `test_archivematica_funtions.py` that would be good to clear up. There's a note below about the download button too. The rest are small style comments that might be useful for future and you can choose to fix.

In testing I've noticed that the font has changed for the download button:

image

I didn't spot what might have caused that change in the code, but if you compare it to 1.11 (Sandbox):

image

Are you seeing the same in your branch? It would be great to get this back to the Serif font before merging.

Just a note around review process, but I feel the "Resolve conversation" button is more a tool for the reviewer. Practically it closes the conversation entirely between pushes (it seems that way as I couldn't reply once commits were against a block of code). But also, it's a check to help keep track of whether a review comment is or isn't resolved. So perhaps keep those open in future. I appreciate it can be a bit noisy. I tend to use a thumbs-up in the top-right corner of a comment when I'm noting something I think I've completed. 

Also, I like your commitment to the atomic commit, but perhaps there's a balance to seek, e.g. 

* https://github.com/artefactual/archivematica/pull/1597/commits/77f86ae763eae45ee39f2e387fcd963670024701
* https://github.com/artefactual/archivematica/pull/1597/commits/60ef92556ec5117cccaa08989d558b21408a94a0 
* https://github.com/artefactual/archivematica/pull/1597/commits/b011a3770e6a53e6f2b02edfdd7642465a89137f

Might be a little bit tiny when trying to determine the meaning with regards to the feature. Unfortunately this gets into subjective territory, so, perhaps see how things pan out for other reviewers in time, but you might consider:

* Grouping commits per module.
* Atomic commits for nit-picks not related to the feature, e.g. you've noticed a typo. 
* A feature/or functionality commit can be great too, but one could imagine setting that out as commits for say:
   * Scaffolding for the work you're about to do, e.g. interfaces/empty test units.
   * The feature code/specific functionalities across the code.
   * Tests. 

Something a little broader to give an overview about what was being done. Especially as one-liners might change between commits. 

This isn't a bad one as an example: https://github.com/artefactual/archivematica/pull/1597/commits/3fff680065756a804c21656daa6bfd3aa9dee13f 

But I don't want to take away from the good work here, and also, I appreciate commits are as much about helping your own process and progress through a feature as well. So whatever enables that best. I reviewed this as one feature block as I went through the code, so not being able to go commit by commit was fine. A little harder for re-review. (But Github wasn't able to render "changes since last review" which it normally can.)

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: "accessionid" (without an underscore) was already in some of the ES indices so I used that form of name in ES when adding it in new indices, while preferring `accession_id` in my Python code.

src/archivematicaCommon/tests/test_archivematica_functions.py
tw4l: Thanks for these tips! The refactored tests are much clearer and get more directly to the point of what we're trying to test.

src/dashboard/src/components/backlog/views.py
tw4l: I've replaced the default values with empty strings in all three cases (the values that had been before were copied from another usage of amclient elsewhere in the codebase, and you're right that they weren't helping).

On closer look at amclient, errors result in an int status code being returned, not None (as I was trying to catch here). I've added handling for the three error codes amclient can return for this particular method, including cases where the connection can't be made to the Storage Service, and made sure that the log message will include the status code for debugging.

tw4l:
Thank you, @ross-spencer! I'll take your comments about commits and not resolving comments to heart and try to do better moving forward :)

Oddly, the download button text font issue was due to the fontawesome icon within that button (which was already there in 1.11) actually being rendered now that the Dashboard has access to the fontawesome assets (which were there in 1.11 in the frontend app, but not in a way that they could be reached from this code). I've added a span for the button text with some CSS to revert the font-families back to what they should be.

I think I've addressed your other issues here as well, but please let me know if you see anything else!

src/dashboard/src/components/backlog/views.py
ross-spencer: > (the values that had been before were copied from another usage of amclient elsewhere in the codebase, and you're right that they weren't helping).

Oof! I bet I know who that was!! (the editor's eye is keener on other's work :slightly_smiling_face: )

src/archivematicaCommon/lib/elasticSearchFunctions.py
ross-spencer: That's unfortunate (we've used snake case in the indexes elsewhere). Thanks for following up.

src/archivematicaCommon/tests/test_archivematica_functions.py
ross-spencer: Nice work. Thanks Tim.

src/dashboard/src/components/backlog/views.py
ross-spencer: AMClient's responses could be cleaner. We've a ticket for that: https://github.com/archivematica/Issues/issues/488 we'll get there.

src/dashboard/src/components/backlog/views.py
```diff
@@ -71,6 +74,35 @@ def check_and_remove_deleted_transfers(es_client):
...
+    :param pending_deletion: Current pending_deletion value in transfers ES index.
+    :return: None
+    """
+    api_results = AMClient(
+        ss_api_key=get_setting("storage_service_apikey", ""),
+        ss_user_name=get_setting("storage_service_user", ""),
+        ss_url=get_setting("storage_service_url", "").rstrip("/"),
+        package_uuid=uuid,
+    ).get_package_details()
+    if api_results in (1, 2, 3):
```
ross-spencer, line 90: Pragmatic approach. Nice.

I think one thing we need to be doing across the code is providing as much context for something as possible. A pattern you might consider here, even though AMclient isn't the most sensible, is to describe what we're expecting and what we're testing againstt. 

So somewhere, this function maybe, might be a const, `AMCLIENT_ERRS = (1, 2, 3)` you might choose to comment that further. And then inline, the future reader sees `if api_results in AMCLIENT_ERRS: ...` which you might agree helps us to mot have to ask too much about the what and why here. It tells us this client returns errors, and we need to check against them. Might just be something to consider in future.

ross-spencer (approved):
LGTM! :+1: 

I noticed the encryption AMAUAT tests touch on the backlog, around downloading. I don't know if you tested them yet? I ran them successfully today, so:
```
1 feature passed, 0 failed, 26 skipped
7 scenarios passed, 0 failed, 70 skipped
55 steps passed, 0 failed, 658 skipped, 0 undefined
Took 8m25.510s
```
Is great! 

>  I'll take your comments about commits and not resolving comments to heart and try to do better moving forward

We touched base in Slack. As noted. It's a conversation (both a CR, and our practices) and your experience adds to how we can do better too, so thank you for your patience, and let's keep figuring out what good CR/PR principles look like! 

Thanks Tim. Looks like another great commit to core!

PR #1602, opened 2020-05-04
 Dashboard: Add new fields and column visibility options to Archival Storage

tw4l:
This PR rewrites the Archival Storage tab's tables using DataTables, adds new information to the tables, and adds a column selector that can be used to choose which columns are visible. More specifically, this PR makes the following changes:

* Elasticsearch changes
    * Adds file count and accession numbers (up to one for files, potentially many for AIPs) to ES indices
    * Adds raw field to`filePath` in files index to enable sorting on filepath
* Archival Storage changes
    * Adds JavaScript to initialize and render a DataTable used to browse/search AIPs and their files
    * Re-implements "Create AIC" feature to work with DataTable
    * Modifies HTML templates
    * Substantially changes views related to searching and listing AIPs and their files (e.g. `archival_storage.views.search` is now a JSON endpoint designed to work with DataTables)
    * Adds new URLs and views for saving and loading DataTable column visibility state in Dashboard Settings, following model from https://github.com/artefactual/archivematica/pull/1597

The "Show AICs" checkbox in the Archival Storage tab search box has been removed in this work. The checkbox did not include or exclude AICs from the search results (they are always included), but instead determined whether the "AIC" column in the search results table is shown. Given that column visibility is now the domain of the DataTable and new visibility selector, the checkbox has been removed so that there are not two potentially contradictory UI elements controlling the visibility of the same column on the screen. Due to this change and the visual changes to search results, we will need to update screenshots and perhaps some text as well in the Archivematica documentation.

Note that the changes to the Elasticsearch mapping (specifically, adding the `filePath.raw` field in the `aipfiles` index) will need to be applied to existing ES indices through reindexing or potentially [Put Mapping](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/indices-put-mapping.html). Based on the documentation alone, the latter technique should be able to handle adding new multi-fields to existing fields. Otherwise, sorting on some columns may be broken. Additionally, data for new fields such as accession numbers will not be available for already-indexed AIPs without reindexing -  empty strings will be displayed for this data in its stead when it is not available in the indices.
 
Connected to https://github.com/archivematica/Issues/issues/1168


src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1069,6 +1114,45 @@ def _list_files_in_dir(path, filepaths=None):
...
+        amdSec = _get_amdSec(admID, doc)
+        registration_detail_notes = ns.xml_xpath_premis(
+            amdSec,
+            ".//premis:event[premis:eventType='registration']/premis:eventOutcomeInformation/premis:eventOutcomeDetail/premis:eventOutcomeDetailNote",
+        )
+        if not registration_detail_notes:
+            return None
+        detail_text = registration_detail_notes[0].text
+        # Strip "accession#" from start of string
+        return detail_text[10:]
```
ross-spencer: I can see from the comment we're truncating `accession#` here with `[10:]` which is helpful, but it would be great if the code described that. We can remove the comment then. 

```python
accession_no_prefix = "accession#"
...
...
return detail_text[len(accession_no_prefix):]

File: src/archivematicaCommon/lib/elasticSearchFunctions.py [hunk 2]
```diff
@@ -698,13 +737,26 @@ def index_transfer_and_files(client, uuid, path, printfn=print):
...
 
     :param client: ElasticSearch client.
     :param uuid: UUID of the Transfer in the DB.
     :param path: path on disk, including the transfer directory and a
                  trailing / but not including objects/.
+    :param transfer_name: name of Transfer
+    :param accession_id: optional accession ID
+    :param ingest_date: date Transfer was indexed
     :param status: optional Transfer status.
     :param printfn: optional print funtion.
```
ross-spencer: It looks like `pending_deletion` is perhaps missing from this list of params?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1069,6 +1114,45 @@ def _list_files_in_dir(path, filepaths=None):
...
+
+    Look for a <premis:event> entry within file's amdSec that has a
+    <premis:eventType> of "registration". Return the text value (with leading
+    "accession#" stripped out) from the <premis:eventOutcomeDetailNote>.
+
+    If file file does not have an amdSec (i.e. is a metadata file)
+    or no matching <premis:event> entries is found, return None.
+
+    :param admID: admID for amdSec to parse
+    :param doc: METS document to parse
```
ross-spencer: I wonder if this docstring is right? (`METS document to parse`) If I read correctly, we've parsed the document by this point, so we're passing a copy of a `ElementTree object` from which to find the accession no?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -566,6 +581,14 @@ def _generator():
 
             indexData["FILEUUID"] = fileUUID
 
+            # Get accession number associated with file and add to list of
+            # unique accession ids for AIP if not already present
+            accessionid = _get_accession_number(admID, mets)
```
ross-spencer: We have an `amdSecInfo` in `570`, is it the same object at this point in the code? If so we can probably pass that section into this helper and avoid passing the larger object around.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -550,9 +567,7 @@ def _generator():
                 if len(set(uuids)) == 1:
                     fileUUID = uuids[0]
             else:
-                amdSecInfo = ns.xml_find_premis(
-                    mets, "mets:amdSec[@ID='{}']".format(admID)
-                )
+                amdSecInfo = _get_amdSec(admID, mets)
                 fileUUID = ns.xml_findtext_premis(
                     amdSecInfo,
                     "mets:techMD/mets:mdWrap/mets:xmlData/premis:object/premis:objectIdentifier/premis:objectIdentifierValue",
```
ross-spencer: With these new helpers being introduced, it would be really nice to bring that consistency to the other parts here, e.g. for `fileUUID`. Would you consider that in-scope?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1341,6 +1425,16 @@ def mark_aip_stored(client, uuid):
 def mark_backlog_deletion_requested(client, uuid):
     _update_field(client, "transfers", uuid, "pending_deletion", True)
 
+    files = _document_ids_from_field_query(client, "transferfiles", "sipuuid", uuid)
```
ross-spencer: `"transferfiles"` is repeated below on `1433`. Do we have a lookup table that describes our indexes anywhere? At the very least I think this function could benefit from removing the duplication. Perhaps consts for `transferfiles`, `sipuuid` and `pending_deletion`.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -635,8 +658,10 @@ def _generator():
     # It should be revisited once we make documents smaller.
     bulk(client, _generator(), chunk_size=50)
 
+    return (len(files), accession_ids)
 
-def index_transfer_and_files(client, uuid, path, printfn=print):
+
+def index_transfer_and_files(client, uuid, path, pending_deletion=False, printfn=print):
```
ross-spencer: I wonder if some of these changes are part of the backlog ones that were made and not part of this PR? 

I note too, that the Backlog and Archival Storage Tabs adopt two different strategies to access the accession_id. Is there a way to bring consistency between those? Or do you think the Archival Storage tab should specifically be looking at the XML and not what's in the database?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1399,6 +1399,16 @@ def _update_field(client, index, uuid, field, value):
 def mark_aip_deletion_requested(client, uuid):
     _update_field(client, "aips", uuid, "status", "DEL_REQ")
```
ross-spencer: This is fairly small, but per a previous comment, wherever possible, we should be avoiding repetition within modules as a whole, but definitely within individual functions, so if, for, example, I need to reference `"aipfiles"` even once in-line really, I want to encode that as a constant somewhere to be referenced. This will go to promote ease of change in future, they can be self-describin, and of course reduce the chance of introducing errors in the short-term.

In this function, I feel there may be more global ways to reference: `"aips"`, `"DEL_REQ"`, `"aipfiles"`. `"status"` and `AIPUUID` are they more local? might be defined within the local context? 

For the indexes `aipfiles`, `transferfiles`, we have a list: https://github.com/artefactual/archivematica/blob/4f2abe13a40ae330462a01d38e1ab5b3ce46d950/src/archivematicaCommon/lib/elasticSearchFunctions.py#L76. Which also be referring to constants.

ross-spencer: I notice too the function `mark_aip_deletion_requested` Is pretty much doing the same as `_mark_deletion_request`. 
```python
def mark_backlog_deletion_requested(client, uuid):
    _update_field(client, "transfers", uuid, "pending_deletion", True)

    files = _document_ids_from_field_query(client, "transferfiles", "sipuuid", uuid)
    for file_id in files:
        client.update(
            body={"doc": {"pending_deletion": True}},
            index="transferfiles",
            doc_type=DOC_TYPE,
            id=file_id,
        )
```
Do you think the two can be brought together, perhaps something like this? (untested);
```python
def mark_aip_deletion_requested(client, uuid):
	_mark_deletion_request("aipfiles", "aips", uuid, "status", "AIPUUID", "DEL_REQ")

def mark_backlog_deletion_requested(client, uuid):
	_mark_deletion_request("transferfiles", "transfers", uuid, "pending_deletion", "sipuuid", True)

def _mark_deletion_request(index, field, uuid, status_field, package_uuid, value):
    _update_field(client, field, uuid, status, value)

    files = _document_ids_from_field_query(client, index, package_uuid, uuid)
    for file_id in files:
        client.update(
            body={"doc": {status: value}},
            index=index,
            doc_type=DOC_TYPE,
            id=file_id,
        )        
```
I'm just using the hard-coded literals here as a convenience, but perhaps it can be built out?

src/dashboard/src/media/css/archival_storage.css
```diff
@@ -104,3 +104,14 @@ div.search-summary p {
...
+.dt-buttons {
+  margin-left: 45px;
+  margin-top: 9px;
+}
+
+.dt-buttons button.buttons-collection {
+  padding: 0;
+  border: none;
+  background: none;
+}
```
ross-spencer: Github doesn't like this! :wink: 

You should be able to ask your IDE to create a newline if it doesn't exist. I don't know if that's why there's no newline here? But it's one less thing to thing about if configured to do so.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -19,8 +19,6 @@ along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 
 $(document).ready(function() {
 
-  localizeTimestampElements();
```
ross-spencer: I'm not sure why this was removed, but I'm also not sure what it was doing :slightly_smiling_face: have you some more context around this?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -359,57 +362,61 @@ def search_augment_file_results(es_client, raw_results):
...
-            messages.error(request, "Error creating AIC")
-            logger.exception(
-                "Error creating AIC: Error creating directory {}".format(destination)
-            )
-            return redirect("archival_storage_index")
-
-        # Create SIP in DB
-        mcp_destination = destination.replace(shared_dir, "%sharedPath%") + "/"
-        databaseFunctions.createSIP(mcp_destination, UUID=temp_uuid, sip_type="AIC")
+    # Use AIP UUIDs to fetch their names, which is used to produce files below
```
ross-spencer: It feels like I've been picking up on a lot of style points. Sorry about that! Here, and in a few places in the PR, try and make sure that comments are [complete sentences](https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html) terminated with a full-stop. These are close! And I notice you've been doing that with `docstrings` which is great.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -359,57 +362,61 @@ def search_augment_file_results(es_client, raw_results):
...
 
-        return redirect("components.ingest.views.aic_metadata_add", temp_uuid)
-    else:
+    # Create SIP (AIC) directory in staging directory
+    temp_uuid = str(uuid.uuid4())
+    destination = os.path.join(staging_dir, temp_uuid)
+    try:
+        os.mkdir(destination)
+        os.chmod(destination, 0o770)
+    except os.error:
```
ross-spencer: Do we care about what the `os.error` is enough to consider adding it to the error logging?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -359,57 +362,61 @@ def search_augment_file_results(es_client, raw_results):
...
         messages.error(request, "Error creating AIC")
-        logger.error("Error creating AIC: Form not valid: {}".format(aic_form))
+        logger.exception(
+            "Error creating AIC: Error creating directory {}".format(destination)
+        )
         return redirect("archival_storage_index")
 
+    # Create SIP in DB
+    mcp_destination = destination.replace(shared_dir, "%sharedPath%") + "/"
```
ross-spencer: You've been doing a great job of using `os.join()`, consider making use of that here, and if it's just the slash you need, `os.paht.sep`. os.path.join("some-value", "") should also give you a slash-terminated string.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -359,57 +362,61 @@ def search_augment_file_results(es_client, raw_results):
...
+    databaseFunctions.createSIP(mcp_destination, UUID=temp_uuid, sip_type="AIC")
+
+    # Create files with filename = AIP UUID, and contents = AIP name
+    for aip in results["hits"]["hits"]:
+        filepath = os.path.join(destination, aip["_source"]["uuid"])
+        with open(filepath, "w") as f:
+            os.chmod(filepath, 0o660)
```
ross-spencer: Perhaps consider placing `0x660` and `0x770` into meaningful variables somewhere:

```python
DIRECTORY_PERMISSIONS = 0o770
FILE_PERMISSIONS = 0o660
```
And then referencing them later on where they become self-describing in-line.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -17,6 +17,121 @@ You should have received a copy of the GNU General Public License
...
+  } else {
+    target.hide('fade', {}, 250);
+  }
+}
+
+function renderArchivalStorageSearchForm(search_uri, on_success, on_error) {
+  // create new form instance, providing a single row of default data
+  var search = new advancedSearch.AdvancedSearchView({
+    el: $('#search_form_container'),
+//    allowAdd: false,
```
ross-spencer: This should be removed if it's non-functional.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -42,7 +168,10 @@ $(document).ready(function() {
   function render_file_actions(file_uuid) {
     var download_href = '/archival-storage/download/aip/file/' + file_uuid + '/';
     var text_span = '<span class="button-text-span">' + gettext('Download') + '</span>';
-    return '<a class="btn btn-default fa-download fa" target="_blank" href="' + download_href + '"> ' + text_span + '</a>';
+    return '<a class="btn btn-default fa-download fa" target="_blank" href="'
+      + download_href + '"> '
+      + text_span
+      + '</a>';
```
ross-spencer: I really like that you're splitting these string manipulation functions over multiple lines.

src/dashboard/tests/test_archival_storage.py
```diff
@@ -207,7 +207,7 @@ def setUp(self):
         self.client = Client()
         self.client.login(username="test", password="test")
         helpers.set_setting("dashboard_uuid", "test-uuid")
-        self.data = '{"time":1586880124134,"columns":[{"visible":true},{"visible":true},{"visible":true},{"visible":false},{"visible":false},{"visible":true},{"visible":false},{"visible":true}]}'
+        self.data = '{"time":1588609847900,"columns":[{"visible":true},{"visible":true},{"visible":false},{"visible":true},{"visible":false},{"visible":false},{"visible":true},{"visible":true},{"visible":false},{"visible":true}]}'
```
ross-spencer: Have you more context about moving the archival storage tests out of its own component). I think that this might be going backwards on a decision that we implemented in the PR here: https://github.com/artefactual/archivematica/pull/1448#issuecomment-510092888 where components are separated into their own responsibilities. Depending on what drove this, I feel it might need reverting.

src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
```diff
@@ -5,9 +5,7 @@
     <div id="search_form_container" class="clearfix"></div>
     <div id="search_form_submit_container" class="clearfix">
       <span style="margin-left: 10px">{% trans "Show files?" %}</span>
-      <span style="margin-left: 5px"><input id="id_show_files" type="checkbox" {{ checked_if_in_file_mode }} /></span>
-      <span style="margin-left: 10px">{% trans "Show AICs?" %}</span>
```
ross-spencer, line 9: I am wondering if show AICs was intended to be removed? 

I noticed it in testing:

image

But I think it needs to be here? 

image

src/dashboard/src/templates/archival_storage/archival_storage.html
```diff
@@ -0,0 +1,70 @@
+{% extends "layout_fluid.html" %}
```
ross-spencer, line 1: This page is really nice and clean. :+1:

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -49,56 +49,170 @@
...
+            aip_status = api_results[0]["status"]
+        except IndexError:
+            logger.info("AIP not found in storage service: {}".format(aip_uuid))
+            continue
+
+        if aip_status == "DELETED":
+            elasticSearchFunctions.delete_aip(es_client, aip_uuid)
+            elasticSearchFunctions.delete_aip_files(es_client, aip_uuid)
+        elif aip_status != "DEL_REQ":
+            elasticSearchFunctions.mark_aip_stored(es_client, aip_uuid)
```
ross-spencer: I was wrestling with where and how to describe this. I spotted in earlier commits the repetition in the elastic search module, but hadn't realized there was a little here too.

What I am seeing is instances of strings such as `DELETED`, or `DEL_REQ` appearing throughout the code, in some cases, they're just field values, in others they become index lookup values, and here, flags which we want to act upon.

```
DEL_REQ:
    9 matches across 3 files
DELETED:
    2 matches across 2 files    
UPLOADED:
    6 matches across 2 files    
aipfiles
    11 matches across 2 files
```

I can see largely, this is really a legacy issue which we need to do better on as a team to improve upon. Given the extent of your changes though and how much code we're touching, I feel we should take the opportunity to do a little of that clean up in your PR. 

A sensible place for describing these status values might be the elasticsearch functions for now, so here, I'd expect to see the module imported and referenced, `elasticSearchFunctions.STATUS_DEL_REQ` as an example. (IDK, the module import is a bit long too, so, import as `es`? `es.STATUS_DEL_REQ`. And then similarly, these values would be referenced in the ES module. I think it's only this file, Backlog, and elasticsearch they're all used? (Backlog seems out of scope now). 

In another review comment I've noted things like the index names too. You might see other values you want to reuse you might do this one, or perhaps find a nicer balance for yourself, but does that seem okay to work through as a result of this first CR?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -49,56 +49,170 @@
...
+    :return: None
+    """
+    api_results = AMClient(
+        ss_api_key=helpers.get_setting("storage_service_apikey", ""),
+        ss_user_name=helpers.get_setting("storage_service_user", ""),
+        ss_url=helpers.get_setting("storage_service_url", "").rstrip("/"),
+        package_uuid=uuid,
+    ).get_package_details()
+
+    AMCLIENT_ERROR_CODES = (1, 2, 3)
```
ross-spencer: Nice! From the backlog review! 

One thing about this pattern, is that this is now the third? time we're calling AMClient in long-form. Perhaps this intrepid coder should have considered it when using it too :wink: but do you think we should now consider testing it out into a helper module (I'm not sure off the top of my head where it would sit): But...
```
=== My helper module.py ===
def setup_amclient();
   client = AMClient(
       ss_api_key=helpers.get_setting("storage_service_apikey", ""),
       ss_user_name=helpers.get_setting("storage_service_user", ""),
       ss_url=helpers.get_setting("storage_service_url", "").rstrip("/"),  
   return client
)
```

```
=== archival_storage/views.py ===
amclient = my_helper_module.setup_amclient()
amclient.package_uuid = "123"
api_results = amclient.get_package_details()
...etc...
```
I don't know how you'd feel about doing it within the scope of this feature? I'd be comfortable that if the problem makes sense to you, that perhaps you log a new issue against it and describe it, if that were okay? Although :hand: I can take the issue logging on if you like. Let me know! 

(One approach might be to create the helper now, and log an issue about AMClient's other uses as well, that would have the benefit of likely not having to come back to this function to test it following a re-write).

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -121,101 +235,94 @@ def search(request):
...
-            "created": fields["created"],
-            "isPartOf": _get_es_field(fields, "isPartOf"),
-            "AICID": _get_es_field(fields, "AICID"),
-            "countAIPsinAIC": _get_es_field(fields, "countAIPsinAIC", "(unknown)"),
-            "type": "AIC" if "AIC#" in _get_es_field(fields, "AICID", "") else "AIP",
-            "status": AIP_STATUS_DESCRIPTIONS[
-                _get_es_field(fields, "status", "UPLOADED")
-            ],
-            "encrypted": _get_es_field(fields, "encrypted", False),
+            "name": fields.get("name"),
```
ross-spencer: Nice use of `.get()` in all of these places. Does the index handle `None` okay? Should default be `""`?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -121,101 +235,94 @@ def search(request):
...
+            "uuid": fields.get("uuid"),
+            "file_count": counts[fields["uuid"]],
```
ross-spencer:
```suggestion
            "file_count": counts.get("fields", []).get("uuid"))
```
Too obtuse?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -240,10 +347,11 @@ def search_augment_file_results(es_client, raw_results):
                 AIPSTOREPATH + "/", "AIPsStore/"
             )
 
-        except:
+        except ElasticsearchException:
```
ross-spencer, line 352: Nice, we have a few plain `Except:` in our code. Always good to see this refined.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -253,57 +361,61 @@ def search_augment_file_results(es_client, raw_results):
...
-        mcp_destination = destination.replace(shared_dir, "%sharedPath%") + "/"
-        databaseFunctions.createSIP(mcp_destination, UUID=temp_uuid, sip_type="AIC")
+    # Use AIP UUIDs to fetch their names, which is used to produce files below
+    query = {"query": {"terms": {"uuid": aip_uuids}}}
+    es_client = elasticSearchFunctions.get_client()
+    results = es_client.search(
+        body=query,
+        index="aips",
+        _source="uuid,name",
+        size=elasticSearchFunctions.MAX_QUERY_SIZE,  # return all records
```
ross-spencer:
```suggestion
        size=elasticSearchFunctions.MAX_QUERY_SIZE, 
```
Really, only `MAX_QUERY_SIZE` so `10000`.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -686,3 +673,34 @@ def view_aip(request, uuid):
...
+    """
+    setting_name = "{}_datatable_state".format(table)
+    state = helpers.get_setting(setting_name)
+    if state:
+        return HttpResponse(
+            json.loads(state), content_type="application/json", status=200
+        )
+    return helpers.json_response(
+        {"error": True, "message": "Setting not found"}, status_code=404
+    )
```
ross-spencer, line 697: What are the possibilities here for not-duplicating `backlog/views.py`? Different components, different view, but similar function with nearly identical input and output. I don't have an instinct.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -101,33 +100,314 @@ $(document).ready(function() {
...
+      return false;
+    });
+  }
+
+  return search;
+}
+
+$(document).ready(function() {
+
+  set_create_aic_visibility();
```
ross-spencer, line 136: One small change in behavior I notice is that the create AIC button will show now without performing a search. I personally don't think that's a bad thing. Having to set its display if a search is performed is extra logic to deal with. That being said, it's not my decision! :p And it does point to the logic for the set visibility function not working as anticipated. So it probably does need looking at if you can also confirm that behavior for yourself.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -101,33 +100,314 @@ $(document).ready(function() {
...
+              'extend': 'colvis',
+              'text': '<i class="fa fa-cog" aria-hidden="true"></i> ' + gettext('Select columns')
+          }
+        ],
+      'language': {
+        'sEmptyTable':            pgettext('DataTable - sEmptyTable',         'No data available in table'),
+        'sInfo':                  pgettext('DataTable - sInfo',               'Showing _START_ to _END_ of _TOTAL_ entries'),
+        'sInfoEmpty':             pgettext('DataTable - sInfoEmpty',          'Showing 0 to 0 of 0 entries'),
+        'sInfoFiltered':          pgettext('DataTable - sInfoFiltered',       '(filtered from _MAX_ total entries)'),
+        'sInfoPostFix':           '', // pgettext('DataTable - sInfoPostFix',        ''),
```
ross-spencer: Do you have more context for this line? Should the part following `//` be removed?

ross-spencer (changes requested):
Thanks @tw4l this is really good overall. A lot of work, and tenacity too to make that archival storage enhancement!

Also, thank you for the AMAAUAT changes as you've done this. I'll finalize the review on those as part of acceptance of this. Or at least, that's something I anticipate doing. 

To the CR comments inline then I hope they make sense, seem fair, etc. Let me know if you have any questions. 

I've done a little playing around with the feature, it looks great, and yeah, I had a lot of confidence in the code and quality of the feature from your backlog changes. 

On the structure of this PR then I think as a bit of a follow up to the backlog PR, would just try and think about some of the advice in [this article](https://mentalized.net/journal/2018/12/12/anatomy-of-a-good-commit/) around the focus and completeness of commits. What I'm seeing in the commits is your process, which is great. You're making changes based on an intellectual separation, and making sure they're available to folk, it is a distributed source control after all. 

What I think might need to happen is a step between that, and then submitting the PR, where I think we want to see it wrapped into much smaller complete chunks. As an example, as a reviewer, it would help a lot to not have commits such as https://github.com/artefactual/archivematica/pull/1602/commits/a637456488f9a5fda211abcd5c7887c1464ee05c and https://github.com/artefactual/archivematica/pull/1602/commits/14d4dc17b6a856a9385e43933a5b5bcba6d04779 included. They both represent initial work in progress, but working through the code-review, I was evaluating them on merit and making style/logic suggestions to find them change further on. (One other example I didn't have enough context for was https://github.com/artefactual/archivematica/pull/1602/commits/2b3460c97c6585ebbb450e3611ac43ea31bb53d8 which might have benefited from a longer commit message body, but yeah, had this branch been based previously on the backlog feature?) 

So in the article, that's the part about focus and completeness. You might still have a lot of commits, but each one ideally is the final form of _something_. It takes a bit of time and git-fu, but yeah, think about how `git rebase -i`, `git commit --amend`, and `cherry-pick` can help you when it comes to submitting the PR. Perhaps in some cases a new branch. Seeing your working style, then perhaps this time around, you might have reworked this into a new branch and submitted a PR against that. The extent of the changes here means that might only have been 3-4 commits. Extending Elasticsearch. Adding the Datatable end-points in preparation for adding the Datatables. Introducing the Datatables; adding tests; as a potential layout. Really just examples.

Thank you for your patience on the code-review. I was really hoping to have had this done earlier this week but had to react to a few other priorities.

tw4l:
Thank you @ross! I'm glad that overall you think this feature is on the right track.

I also really appreciate the time and care you've taken here in explaining how this and future work could be improved in terms of process and commit scoping. There is a bit of a mental adjustment that needs to be made going from working almost exclusively as a solo developer to developing on a team, and I'm still working on making that adjustment (we're always growing, yeah?). Your advice here is helpful and well-taken, and I appreciate the care with which you have pointed to specific commits as examples and even suggested a working practice that might balance my own personal creative/development process with delivering work where each commit has appropriate focus and completeness. Coincidentally, I spent some time yesterday with Google's [How to do a code review](https://google.github.io/eng-practices/review/reviewer/) and Author's Guide documentation yesterday (which I believe Jesús and maybe also you suggested to me when I started at Artefactual), which similarly helped me to see the issue from the perspective of the reviewer and future readers/developers interacting with the codebase.

With all this in mind, I will go through the inline comments and address them starting today and likely picking up/getting to the point where it's ready for re-review next week.

A question for you: do you think it would make sense to re-structure this work on this or a new branch? It strikes me that the best way to improve at anything is typically consideration followed by plenty of practice. That said, if doing that would interfere with your review or be too much from a resourcing perspective, I'm also happy to wait until the next opportunity.

ross-spencer:
@tw4l that's great to hear. It's tough to figure out how to talk about these things and be constructive at the same time. From my-perspective, I would say you could do two things, and it would be find at this point in the CR:

1. Rebase what you've done into a single commit as if it were the feature. 
2. Create CR-response commits on top of that. 

Does that sound okay? I think then we'd just be reviewing CR commits and that is a model I've used in the past :)

tw4l:
@ross-spencer Yes, sounds like a plan! For my part, I'm doing my best to keep a [growth mindset](https://youtu.be/krGwiTCPo4g) about these things, and I appreciate the opportunity to learn.

Expect a request for re-review next week!

src/dashboard/src/components/archival_storage/views.py
tw4l: I was wondering this too but didn't land on an answer that feels satisfactory. I considered making an API endpoint but that doesn't feel right, as external users would have no use for it. There's not much here to abstract into a shared helper either.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: The intention was to render the AIP created dates in readable form using the local timezone, but it didn't actually work! I replaced it with:
```
function render_aip_created_date(created) {
    return timestampToLocal(created);
  }
```
(`timestampToLocal` is an already existing helper function in `misc.js`)

src/dashboard/src/components/archival_storage/views.py
tw4l: This (like most of the `create_aic` function, including the subjects of your next two comments) I left as-is, focusing my changes in parsing the AIP UUIDs from the request. I'm not quite sure why the diff is showing this as new code.

That said, you're right that we could do better! While I'm here I'll consider making some touch-ups where it seems like I can make improvements without spending too much time.

src/dashboard/tests/test_archival_storage.py
tw4l: Thanks for this context! I moved the archival storage tests to `src/dashboard/tests` because with the exceptions of `administration` and `archival_storage` (until this change), the tests for all of the dashboard components live at `src/dashboard/tests` and I found it confusing to have the dashboard tests and fixtures scattered across several locations.

The way I'm reading the linked PR, I think this change is in-line with that discussion recommendation, as it moves toward consolidation of the dashboard app tests. This comment for example references `src/dashboard/tests` as an expected location for dashboard tests: https://github.com/artefactual/archivematica/pull/1448#issuecomment-510458475.

There is definitely some ambiguity around the word "component", which I think may be referring in some cases in this conversation to the Django apps (`dashboard`, `MCPServer`, `MCPClient`) rather than the dashboard components. However it's also very possible that I'm misunderstanding! What do you think?

src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
tw4l: Good catch! I meant to include a note about this in the PR description and forgot.

Contrary to what I assumed would happen, the "Show AICs" checkbox doesn't include or exclude AICs from the search results (they are always included), but instead determines whether the "AIC" column in the search results table is shown. Given that column visibility is now the domain of the DataTable and new visibility selector, I removed this checkbox intentionally so that we don't have two potentially contradictory UI elements controlling the visibility of the same column.

I'll run this by the product owner and other internal stakeholders on Slack for validation, as you're right, this shouldn't just be a surprise for folks.

sromkey: That makes sense to me!! "Show AICs" has always been unnecessarily confusing anyway :confounded:

src/dashboard/tests/test_archival_storage.py
ross-spencer: I feel the conversation could be clearer for sure. My understanding is different. I think the ultimate conclusion is to have tests per component.

My reference point given there is doubt would be the history of the modules themselves, and for each:

* [Administration](https://github.com/artefactual/archivematica/commits/stable/1.11.x/src/dashboard/src/components/administration/tests).
* [Archival storage](https://github.com/artefactual/archivematica/commits/stable/1.11.x/src/dashboard/src/components/archival_storage/tests).

The tests only landed there this last tranche of development for `1.11`. e.g. [Storage service enhancements](https://github.com/artefactual/archivematica/commit/0410e5d0639222c4e46244a9c46291267895f881#diff-a6384982158121f309654f1f7166fd44) which was connected to the previous discussion.

I would stick with that consistency for now and move the archival storage test back. That's not to say it can't be resolved in this PR, but perhaps separating concerns, it could be logged as a new issue to be discussed. And we can tackle it in a future issue. It seems like maybe the strategy does need encoding somewhere? From a change/commit/PR perspective, then doing this in a future issue would give us each something to reference that says, doing (x) because of (y). Does that sound okay for now?

src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
ross-spencer: Sounds good. It's way more intuitive now too. I think that rationale and impact should just be placed in the PR comment, and likely testing notes associated with the change (i,e, users can expect a slight difference in screens!). If screenshots in the docs are impacted by the change, which it seems like they will be anyway with the extent of the data tables improvements, then the removal of the check-box is a good detail to highlight to make sure that it's included. We actually have a `docs needed` label which we can use for the issue associated with this one.

src/dashboard/tests/test_archival_storage.py
tw4l: Yes, sounds good for now. Thanks Ross!

src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
tw4l: Will do, thanks!

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: > And it does point to the logic for the set visibility function not working as anticipated. So it probably does need looking at if you can also confirm that behavior for yourself.

Leaving aside for a moment the question of whether the Create AIC button should show before the user runs a search, is inconsistent visibility something that you noticed in your testing? In my testing, the set visibility function worked fine once I remembered to call the function when the document is ready in addition to when search results are refreshed.

tw4l: On the larger question of whether the button should show on page refresh, I had a conversation with @sromkey @ablwr and the decision was to show the button regardless of whether a user has run a search. Rationale: "In terms of UX/design, we should go with whatever causes the least amount of surprise for the user, and I think in general an initially-hidden button that you have to uncover through searching is a 'surprise'".

Based on that conversation, I will also add a tooltip to the "Create an AIC" button that clarifies that clicking on it will create an AIC from the currently displayed search results.

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: There was some of the backlog work here, as I based my work on that since the backlog PR hadn't been merged anywhere when I started and there were some shared dependencies. My apologies if some of it remained! I'll include auditing for remaining bits from the backlog PR as part of my revisions.

You raise a good point about the inconsistency! My thinking was that AIP indexing should look up accession numbers in the METS XML for two reasons:

1. The AIP indexing function is already doing quite a lot of METS processing, including getting other information from the file `amdSec`s, so the additional lookup time for getting the accession number is pretty minimal (database lookups would almost certainly be quicker, but rewriting all of the AIP indexing seems out of scope for this project).

2. Since we support [indexing from the AIPs directly](https://github.com/artefactual/archivematica/blob/0722067836c3430f2a371f15583be633a6db7fe5/src/dashboard/src/main/management/commands/rebuild_elasticsearch_aip_index_from_files.py) even when there's no record of the AIP in the database, it seemed like the better approach.

I notice now that the reindexing script for transfers contains [some logic](https://github.com/artefactual/archivematica/blob/0722067836c3430f2a371f15583be633a6db7fe5/src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py#L306) which handles this issue/edge case in a different way, by writing the information such as the accession number from the METS into the database prior to indexing if there's not already a database entry for the transfer. That could potentially be an interesting tack to consider taking with the AIPs as well if we were to rewrite the indexing functions?

ross-spencer: > There was some of the backlog work here, as I based my work on that since the backlog PR hadn't been merged anywhere when I started and there were some shared dependencies. My apologies if some of it remained! I'll include auditing for remaining bits from the backlog PR as part of my revisions.

I see. This is really a Git hygiene thig. I think it's definitely complicated here by tackling these two features, independently, but also sharing dependencies. Ideally you want to be working off a clean git reference, so I think the backlog's last commit ref is https://github.com/artefactual/archivematica/commit/ef6931b7a789b6009867bb607c58495cde246e98 so at some point you need to rebase against that if you're working against it and features of it are part of the PR to review.  

There becomes a clean delineation between concerns. 

The biggest impact is really the potential to introduce changes to the previous work (re-write, add new lines/ formatting, etc.). But I think we should find that eventual consistency when rebasing here. So rebase might be another next step. I don't think you'll see merge conflicts, but I think that can be the other risk of manually bringing consistency. 

It's a bit of a dance, and another point we'll need to consider if our feature merge strategies are modified. 

This is what I see in my head about how we've asked this work be approached:

image

And how I look at our current approach:

image

(Though of course there are always exceptions.)

All that to say, I recognize its complexity.

## Accession number from XML

That all seems reasonable. 

|  That could potentially be an interesting tack to consider taking with the AIPs as well if we were to rewrite the indexing functions? 

If the suggestion is at one point, these two sets of scripts would ideally be one that: 1. Consume the METS, 2. Index the METS. Then there's definitely a draw to that in terms of maintenance. 

An aside. I don't think I had  formed an opinion when reviewing about using [`metsrw`](https://adr.archivematica.org/0001-unification-of-mets-creation.html). Other than it maybe seemed out of scope. But the reindex script is already using it, and seems to make accession id easily available. Is it something to consider in scope of this work to use METS Reader-writer for any XML functions you're adopting? 

It looks like a lot of the XML work is already pre-existing, so perhaps it really is just out of scope for now.

tw4l: _Re: git hygeine_: I see your point here and the the diagrams make the difference between what's expected and what I submitted for this PR very clear. I've rebased on https://github.com/artefactual/archivematica/commit/ef6931b7a789b6009867bb607c58495cde246e98 and will be more mindful moving forward of my git practice. Genuinely, thank you for helping me with some of these growing pains.

_Re: XML_: 

> If the suggestion is at one point, these two sets of scripts would ideally be one that: 1. Consume the METS, 2. Index the METS. Then there's definitely a draw to that in terms of maintenance.

I wasn't thinking that so much as musing that the re-indexing script for AIPs not associated with a pipeline could borrow the approach from re-indexing transfers not associated with a pipeline (namely, reading information from the XML into the database, which enables the general indexing functions to be more efficient by retrieving information from the database rather an XML). Really just musing here, as 1) I don't think this is a priority and 2) that's definitely well out of scope for this work.

> An aside. I don't think I had formed an opinion when reviewing about using metsrw. Other than it maybe seemed out of scope. But the reindex script is already using it, and seems to make accession id easily available. Is it something to consider in scope of this work to use METS Reader-writer for any XML functions you're adopting?
> It looks like a lot of the XML work is already pre-existing, so perhaps it really is just out of scope for now.

My inclination is that using efficiently metsrw in the the parts of the AIP indexing functions I've modified for this work would more or less require rewriting the functions, and that seemed out of scope for now to me as well. It would be a great improvement though, I agree. Do you think logging an issue for it would be an okay compromise for the moment?

src/dashboard/src/components/archival_storage/views.py
tw4l: This function is taking values from the ES index (which does handle `None` values - it casts them to null, which Elasticsearch is fine with for keyword fields - though if we have a missing name, I think that would be an indication of a larger issue!) and formatting them before returning them to the frontend. The frontend does expect a string, so that would make a better default!

src/dashboard/src/components/archival_storage/views.py
tw4l: Removed the comment since the line is self-describing

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: See thread under comment https://github.com/artefactual/archivematica/pull/1602#discussion_r425065493

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: I've implemented a solution based on your comment here and tested it. Nice suggestion!

src/dashboard/src/components/archival_storage/views.py
tw4l: Agreed! Great suggestion. I've implemented the change for everything in scope for this PR. I would certainly consider doing the same for the Backlog/transfer indexes as a quick separate PR to the SFU dev branch prior to merging into `qa/1.x`. Wouldn't take long and it makes a nice improvement. Would you be willing to CR that down the line?

src/dashboard/src/components/archival_storage/views.py
tw4l: I think what we want is `"file_count": counts.get(fields.get("uuid"), 0),`, since the resulting value should be an integer.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: Yes, this should be (and now has been) removed. It allows you to add a message to the table such as "All records shown are derived from real information." ([from legacy DataTables docs](https://legacy.datatables.net/ref)), but we don't use this feature.

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: Resolved

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: Resolved in rebase

src/archivematicaCommon/lib/elasticSearchFunctions.py
tw4l: Very sensible suggestion! Since the `fileUUID` bit is in the same block I'm touching for the accession number, it seemed reasonable to touch this up.

src/dashboard/src/components/archival_storage/views.py
tw4l: Change made

src/dashboard/src/components/archival_storage/views.py
tw4l: This makes sense to me, and I'm happy to log the issue!

> (One approach might be to create the helper now, and log an issue about AMClient's other uses as well, that would have the benefit of likely not having to come back to this function to test it following a re-write).

Agreed! One fewer thing to come back to later :)

src/archivematicaCommon/lib/elasticSearchFunctions.py
ross-spencer: NP @tw4l. If it's helpful at all that's good. These are small things in the context of much more significant work you're doing here :slightly_smiling_face: 

> It would be a great improvement though, I agree. Do you think logging an issue for it would be an okay compromise for the moment? 

That's okay. I don't think it needs a follow up issue. The general principle of replacing `metsrw` is a question we should ask in most PRs touching METS as a by-product of the ADR which I think is largely how we'll see it addressed in future. It would be great as its own maintenance task at some point. I think it's absolutely valid to be out of scope. And these are good reasons. It's not an insignificant undertaking.

src/dashboard/src/components/archival_storage/views.py
ross-spencer: Absolutely!

src/dashboard/src/components/archival_storage/views.py
tw4l: I added this setup function and an `AMCLIENT_ERROR_CODES` constant to `archivematicaFunctions.py`. Seemed like the most reasonable place to put them, that would make them available across all the different AM modules. Let me know if you think they belong somewhere else!

tw4l:
Hi @ross-spencer, this is ready for re-review. I think I've addressed all of your comments above except the one about not repeating the `save_state` and `load_state` views, as I haven't come up with a better solution than simply repeating them across the two components (as much as that does bother me a bit from a DRY perspective). Thanks!

src/dashboard/src/templates/archival_storage/archival_storage.html
```diff
@@ -55,7 +55,12 @@ <h1>
         </small>
       </h1>
 
-    <button id="create-aic-btn" class="btn btn-primary">{% trans "Create an AIC" %}</button>
+    {% blocktrans with "Create an AIC from current search results" as create_aic_title %}
+    <button
+      id="create-aic-btn"
+      class="btn btn-primary"
+      title="{{ create_aic_title }}">Create an AIC</button>
```
ross-spencer: I haven't worked a lot on the UI, out of interest, does `Create an AIC` need a translation placeholder as well?

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -69,6 +74,16 @@ def get_dashboard_uuid():
...
 
 
+def setup_amclient():
+    """Initialize and return an AMClient instance."""
+    client = AMClient(
+        ss_api_key=get_setting("storage_service_apikey", ""),
+        ss_user_name=get_setting("storage_service_user", ""),
+        ss_url=get_setting("storage_service_url", "").rstrip("/"),
+    )
+    return client
```
ross-spencer, line 84: Perfect, thank you for doing this!

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
ross-spencer: >  I think in general an initially-hidden button that you have to uncover through searching is a 'surprise'".

_Archivematica (TM) for PS4 (TM)_ :wink: 

I like it. Good decision, and one less thing to think about. Nice work all.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -101,33 +99,314 @@ $(document).ready(function() {
...
+
+  }
+
+  function set_create_aic_visibility() {
+    if ($('#id_show_files').prop('checked')) {
+      $("#create-aic-btn").hide();
+    } else {
+      $("#create-aic-btn").show();
+    }
+  }
```
ross-spencer, line 379: On the subject of AICs ^^^ can we remove this now?

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -101,33 +99,314 @@ $(document).ready(function() {
...
+    var table_uuids = $('#id_show_files').prop('checked')
+      ? []
+      : $('#archival-storage-entries').DataTable().column(1).data();
+    // Add each UUID in the returned object to our clean aip_uuids array.
+    for (i = 0; i < table_uuids.length; i++) {
+      aip_uuids.push(table_uuids[i]);
+    }
+    // Redirect window to create_aic, passing AIPs UUIDs in the POST request.
+    $.redirectPost('/archival-storage/search/create_aic/', {'uuids': aip_uuids});
+  });
```
ross-spencer, line 411: And this?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -637,6 +667,11 @@ def _generator():
     # It should be revisited once we make documents smaller.
     bulk(client, _generator(), chunk_size=50)
 
+    file_count = len(files)
+    accession_ids_list = list(accession_ids)
```
ross-spencer, line 671: Just a small thing, but I really appreciate this extra verbosity here. It's one less thing to compute as a reader.

ross-spencer (approved):
@tw4l great work. 

I am approving this, with one thing that might need to be removed, and one question, that may also need to be resolved.

1. A question inline about AIC detritus in the JavaScript now we're defaulting to displaying the button. (If this does need removing, and everything looks good functionally, then :+1: )
2. The tiniest thing in the Create AIC button name itself and translations.

_These are small things that shouldn't hold you up by needing another CR and shouldn't hinder your progress if they don't need fixing._

As for the rest, then it's great! The clean-up work across the view and elastic search functions really makes a difference as a reader. That work has been a long-time coming so I'm glad you've been able to tackle it. 

If there is an opportunity to follow up with the Backlog cleanup there at some-point then it's appreciated. I also appreciate that our hopes don't always work out with so much on, so I wonder if you wouldn't mind adding a GH issue in the short-term? Something like:

      1. We've cleaned up variable usage for archival storage indices in this PR.
      2. We should do the same for Backlog. 

Does that sound okay? 

Once again, nice work. Now onto the merging!! (I suspect you'll need a branch approval between your feature branch and then merging into QA per our discussions earlier. Feel free to holler if so!)

src/dashboard/src/templates/archival_storage/archival_storage.html
tw4l: That string is being translated already because it's within a [{% blocktrans %} block](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#blocktrans-template-tag), which is handy and was new to me!

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: We're still hiding the button when the "show files?" checkbox is ticked, as you can't create an AIC from individual files, so this needs to stay.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: Likewise, we need to keep this so that the button functionality works to create an AIC from AIPs.

tw4l:
Thanks @ross-spencer ! I will merge this into the feature branch now, and then do the work to rebase the feature branch and submit the PR for merging into QA today. Thanks for offering to do a quick CR for that - I'll definitely take you up on the offer! Since I'm planning to work on this today and it won't take much additional time, I'll address the backlog readability improvements and [implement the amclient helper across the dashboard app](https://github.com/archivematica/Issues/issues/1215) as commits in the PR for merging into QA. No time like the present :)

src/dashboard/src/templates/archival_storage/archival_storage.html
tw4l: As @sevein rightly pointed out, the `{% blocktrans %}` approach passed not only the strings but also the HTML into the message files. I've updated it so that only the strings are being translated.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
ross-spencer: Ah, perfect!

PR #1604, opened 2020-05-09
Fix dashboard header overlaps

jraddaoui:
Header overlap with content and navigation overlap with logo and status.

- Remove static property from top-bar.
- Remove body padding.
- Add content padding.
- Add display and padding to logo link.
- Remove float from logo link in small widths.
- Move status before navigation bar.
- Remove absolute position from status.

Refs https://github.com/archivematica/Issues/issues/1034.


jraddaoui:
**Before:**
before

**After:**
after

sevein (approved):

PR #1606, opened 2020-05-15
Add CSRF middleware

replaceafill:
This adds Django's `CsrfViewMiddleware` to the dashboard settings and
the `{% csrf_token %}` tag to the form templates.

The wrapper for the API endpoints has been marked as CSRF exempt to
allow API calls to reach authentication, but the CSRF checks are still
performed by the Tastypie's authentication backends when needed.

Connected to https://github.com/archivematica/Issues/issues/1235

### DIFF
--- a/src/dashboard/src/components/api/views.py
+++ b/src/dashboard/src/components/api/views.py
@@ -29,6 +29,7 @@ import re
 from django.db.models import Q
 import django.http
 from django.conf import settings as django_settings
+from django.views.decorators.csrf import csrf_exempt
 import six
 
 from tastypie.authentication import (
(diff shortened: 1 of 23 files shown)


src/dashboard/src/components/api/views.py
```diff
@@ -74,6 +75,7 @@ def _api_endpoint(expected_methods):
     def decorator(func):
         """The decorator applied to the endpoint."""
 
+        @csrf_exempt
```
sevein, line 78: So much work behind this one-line change!

sevein (approved):
It works great, thanks.

PR #1610, opened 2020-06-01
Use uncompressed StaticFilesStorage for local dev

tw4l:
Fixes https://github.com/archivematica/Issues/issues/1218 in local Docker development environment

Results of running `make test-dashboard` before change: `24 failed, 111 passed, 6 skipped, 32 warnings in 28.35 seconds`

After change: `8 failed, 134 passed, 6 skipped, 32 warnings in 16.59 seconds`

Some of the tests that are still failing seem to be known issues (e.g. with running `test_rights.py` tests in Docker), but there may still be some other tests that need to be modified following the Django 1.11 upgrade (e.g. `tests/test_transfer.py::TestTransferViews::test_metadata_edit`).

### DIFF
--- a/src/dashboard/src/settings/local.py
+++ b/src/dashboard/src/settings/local.py
@@ -30,6 +30,12 @@ from .base import *
 DEBUG = True
 TEMPLATES[0]["OPTIONS"]["debug"] = True
 
+# Use Django uncompressed StaticFilesStorage rather than compressed Whitenoise
+# storage in local development so that tests run correctly.
+# See: https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/
+# #django.contrib.staticfiles.storage.ManifestStaticFilesStorage.manifest_strict
+STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
+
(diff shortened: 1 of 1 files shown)


src/dashboard/src/settings/local.py
```diff
@@ -30,6 +30,12 @@
 DEBUG = True
 TEMPLATES[0]["OPTIONS"]["debug"] = True
 
+# Use Django uncompressed StaticFilesStorage rather than compressed Whitenoise
+# storage in local development so that tests run correctly.
```
replaceafill, line 34: >so that tests run correctly

Since the intent is to make tests work, I wouldn't switch the `local` settings but the `tests` ones. That brings [a different problem](https://github.com/archivematica/Issues/issues/1218#issuecomment-637034567) unfortunately.

AFAIU `Whitenoise` works fine for development, but let me know if I'm missing something.

Edit: FWIW I've been taking the (very questionable) approach of relying on `tox` lately instead of `make test-...` commands in the docker environment. I guess I'm just trying to make Travis happy until we figure out https://github.com/archivematica/Issues/issues/1140

PR #1612, opened 2020-06-03
Set CSRF header through httpProvider config

replaceafill:
~~This [interceptor][0] adds the `X-CSRFToken` header to all the `POST`,
`PUT` and `DELETE` requests made with Restangular.~~

@jraddaoui recommended trying [angular configuration to set the CSRF header][1]. It works nicely and it's a much cleaner solution :+1:

Connected to https://github.com/archivematica/Issues/issues/1221

[0]: https://code.angularjs.org/1.5.11/docs/api/ng/service/$http#interceptors
[1]: https://code.angularjs.org/1.5.11/docs/api/ng/service/$http#cross-site-request-forgery-xsrf-protection


jraddaoui (approved):
:clap: Thanks @replaceafill!

PR #1617, opened 2020-06-04
Add management command to update ES mappings for Archivematica 1.12

tw4l:
Fixes https://github.com/archivematica/Issues/issues/1228

This management command uses Put Mapping to update the mappings for the `aips` and `aipfiles` indices, to ensure proper sorting in the Archival Storage tab DataTable in Archivematica 1.12.

This command must be run immediately following the upgrade to Archivematica 1.12, prior to new AIPs being ingested, as Put Mapping is unable to change the data type of existing ES fields. New string fields will be automatically set to "text" rather than "keyword" if not explicitly defined in the mapping, which will prevent the Put Mapping command from working.

Note that reindexing is still necessary in cases where users want to fully populate the new fields for AIPs ingested prior to Archivematica 1.12.

### DIFF
--- a/src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
+++ b/src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+"""Update Elasticsearch mappings for Archivematica 1.12
+
+This command updates the Elasticsearch mappings for the aips and
+aipfiles indices for Archivematica 1.12 to enable sorting on all fields
+displayed in the new Archival Storage DataTable and populates the new
+filePath.raw subfield in the aipfiles index using an Update By Query.
+
+Execution example:
+
+./manage.py update_elasticsearch_mappings
+"""
+from __future__ import absolute_import, print_function
+
+import sys
+
+from django.conf import settings
+from elasticsearch import ElasticsearchException
+
+from main.management.commands import DashboardCommand
+import elasticSearchFunctions as es
+
+
+class Command(DashboardCommand):
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        # Check that the AIPs index is enabled before proceeding.
+        if es.AIPS_INDEX not in settings.SEARCH_ENABLED:
+            self.error(
+                "The AIPs indexes are not enabled. Please, make sure to "
+                "set the *_SEARCH_ENABLED environment variables to `true` "
+                "to enable the AIPs and Transfers indexes, or to `aips` "
+                "to only enable the AIPs indexes."
+            )
+            sys.exit(1)
+
+        try:
+            es.setup_reading_from_conf(settings)
+            es_client = es.get_client()
+        except ElasticsearchException:
+            self.error("Error: Elasticsearch may not be running.")
+            sys.exit(1)
+
+        # Update the AIPs index mappings.
+        es_client.indices.put_mapping(
+            index=es.AIPS_INDEX,
+            doc_type=es.DOC_TYPE,
+            body={
+                "properties": {
+                    "accessionids": {"type": "keyword"},
+                    "status": {"type": "keyword"},
+                    "file_count": {"type": "integer"},
+                }
+            },
+        )
+
+        # Update the AIP files index mapping.
+        es_client.indices.put_mapping(
+            index=es.AIP_FILES_INDEX,
+            doc_type=es.DOC_TYPE,
+            body={
+                "properties": {
+                    "accessionid": {"type": "keyword"},
+                    "status": {"type": "keyword"},
+                    "filePath": {
+                        "type": "text",
+                        "analyzer": "file_path_and_name",
+                        "fields": {"raw": {"type": "keyword"}},
+                    },
+                }
+            },
+        )
+
+        # Perform an update by query on the aipfiles index to populate
+        # the filePath.raw subfield from existing text values. We do
+        # not specify a query to ensure that all documents are updated.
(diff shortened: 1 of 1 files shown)


src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
```diff
@@ -0,0 +1,75 @@
...
+                "to only enable the AIPs indexes."
+            )
+            sys.exit(1)
+
+        es.setup_reading_from_conf(settings)
+        es_client = es.get_client()
+
+        try:
+            es_client.info()
+        except ElasticsearchException:
```
jraddaoui: Unlike you would expect, we check the indexes existence while setting up the client in ... https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematicaCommon/lib/elasticSearchFunctions.py#L202.

So you may want to include the setup and client retrieval in the try block instead of the info call.

jraddaoui: Ups, clicked single comment but I'm on it @tw4l ;)

src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Update AIPs index mappings
+        es_client.indices.put_mapping(
+            index=es.AIPS_INDEX,
+            doc_type=es.DOC_TYPE,
+            body={
+                "properties": {
+                    "accessionids": {"type": "keyword"},
+                    "status": {"type": "keyword"},
+                    "file_count": {"type": "integer"},
```
jraddaoui, line 54: We really need to normalize field names in our indexes, but it may be to late to do it in here. I added a comment about this in ... https://github.com/archivematica/Issues/issues/404.

src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
```diff
@@ -0,0 +1,75 @@
...
+            index=es.AIP_FILES_INDEX,
+            doc_type=es.DOC_TYPE,
+            body={
+                "properties": {
+                    "accessionid": {"type": "keyword"},
+                    "status": {"type": "keyword"},
+                    "filePath": {
+                        "type": "text",
+                        "analyzer": "file_path_and_name",
+                        "fields": {"raw": {"type": "keyword"}},
```
jraddaoui, line 70: Is this new raw field populated in existing documents when the mapping is modified? If that's not the case, I think it's the only one that we could populate in this script based on the current value of the text field. You could use the [update by query API](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/docs-update-by-query.html) and a [Painless script](https://www.elastic.co/guide/en/elasticsearch/painless/6.8/painless-examples.html) to do so if it's needed.

Considering that we can't do the same for the other new fields, feel free to disregard if this is something outside the scope of this feature. Nevertheless, it would be nice to note this limitations in the Archivematica documentation.

jraddaoui (approved):
This is really nice @tw4l and it's looking great!

It raises the question (as the reindex_from_remote_cluster script that I added did) about how we handle this kind of upgrades and the trade offs of having multiple one go commands needed for specific versions. I had a quick chat with @sevein yesterday about this and we considered that it's a bigger issue that we need to analyze separately, so I think this is okay to merge.

Please, make sure the upgrade process gets documented properly for the usage of this script. Thanks!

src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
tw4l: Yes, agreed @jraddaoui! I tried to balance consistency in the indices with keeping a reasonable scope for the project (especially given that the scope of work involved for the Archival Storage tab was larger than estimated) and not requiring too many changes to existing installations, but in retrospect I would have focused more on normalizing field names. And now that I'm more familiar with Elasticsearch, I see that `accessionids` for example could have just been `accessionid`, since ES will natively handle turning a field into an array if there are multiple values of the same type (and of course, ideally would be `accession_id` across the different indices).

Thanks for adding the comment to https://github.com/archivematica/Issues/issues/404 so this stays on our radar.

src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
tw4l: It looks like the new raw field is not updated, so good catch! I will look into populating that field in this script as I have time this week.

I'll also make sure that the limitations of the script are clear in the Archivematica documentation and on our radar for 1.12 client upgrades, so that everyone's aware of what this script does and doesn't do, and that full reindexing is necessary if one wants to fully populate these new fields for AIPs ingested prior to the upgrade.

jraddaoui: Thinking it through ... If we still have to document that limitation for all the other fields, it probably doesn't deserve the effort to fix it for this one. Totally up to you ;)

tw4l:
Relates to https://github.com/archivematica/Issues/issues/1257

src/dashboard/src/main/management/commands/update_elasticsearch_mappings.py
tw4l: Hi @jraddaoui - coming back to this after a little while, but I went ahead and added the query by update to populate the `filePath.raw` subfield because some Archivematica users may not want or be able to fully reindex and this seems like a rather crucial field to be able to sort on. Thanks for the suggestion!

PR #1618, opened 2020-06-04
Fix Archival Storage Chrome and sorting issues

tw4l:
Fixes a Chrome issue and sorting issues discovered in internal QA for https://github.com/archivematica/Issues/issues/1168.

**Chrome issue**
Clicking the browser's back button in Chrome to return to the Archival Storage table directly after clicking a link to the AIP detail page from the Show files page was causing a cryptic DataTables error, due to mismatches between the cached DOM and the JSON returned from the AJAX call to the Archival Storage search endpoint. Due to differences in browser functionality, there was no such error in Firefox.

This PR fixes the issue by using the browser's Navigation Timing API to check if a user arrived to the current page via their browser's back or forward buttons when the page first loads, and triggering a page refresh immediately after the document is ready if so.

With this approach, in Chrome if a user goes from the "files" view in the Archival Storage tab to an AIP detail page and then hits back, they will be brought back to the Archival Storage tab's default display (listing all of the AIPs in storage) rather than exactly where they were before clicking on the link to the AIP page.

This does introduce a difference in behavior between browsers when a user takes the particular series of steps as outlined above, but consensus on the project team was that this is an acceptable trade-off, especially given that another solution was not readily apparent after a few hours of research and investigation.

**Sorting issues**
This PR also updates the ES mappings for the `aip` and `aipfiles` indices to explicitly define the mappings for new fields in the indices and ensure that all string fields used for sorting are type "keyword", not "text" (in the case of the `aipfiles` index's `filePath` field, a new raw "keyword" subfield was added so as not to break fulltext searching on the filepath).

In new Archivematica 1.12 installations where there is no extant Elasticsearch mapping, this will entirely solve the problem.

In upgraded instances with existing mappings, the mappings will need to be updated via Put Mapping immediately following the upgrade to 1.12. See https://github.com/archivematica/Issues/issues/1228 for more details, and the associated PR for a new Django dashboard management command to perform the index mapping updates via Put Mapping.

Connected to archivematica/Issues#1168.

### DIFF
--- a/src/dashboard/src/media/js/archival_storage/archival_storage_search.js
+++ b/src/dashboard/src/media/js/archival_storage/archival_storage_search.js
@@ -132,7 +132,6 @@ function renderArchivalStorageSearchForm(search_uri, on_success, on_error) {
 }
 
 $(document).ready(function() {
-
   set_create_aic_visibility();
 
   var search = renderArchivalStorageSearchForm(null, null, null);
@@ -231,7 +230,9 @@ $(document).ready(function() {
   }
 
   function get_datatable() {
-    if ($('#id_show_files').prop('checked')) {
+    var showFiles = $('#id_show_files').prop('checked');
+
+    if (showFiles) {
       var cols = [
         {sTitle: gettext('Thumbnail'), mData: 'FILEUUID', mRender: render_thumbnail },
         {sTitle: gettext('File'), mData: 'filePath', mRender: render_filepath },
@@ -341,7 +342,7 @@ $(document).ready(function() {
       'sAjaxSource': '/archival-storage/search?' + search.toUrlParams(),
       'aoColumns': cols,
       'fnServerData': function(sSource, aoData, fnCallback) {
-        aoData.push({ 'name': 'file_mode', 'value': $('#id_show_files').prop('checked') });
+        aoData.push({ 'name': 'file_mode', 'value': showFiles });
         $.getJSON(sSource, aoData, function(json) {
           fnCallback(json);
         });
--- a/src/dashboard/src/media/js/misc.js
+++ b/src/dashboard/src/media/js/misc.js
@@ -139,3 +139,23 @@ $(document).ready(
                 $(this).closest('.preview-help-text').children('.preview').show();
               });
 });
+
+// Return true if user navigated to page via browser back/forward button.
+// Return false if user navigated directly or if browser doesn't support Navigation Timing API.
+function navigatedFromBackOrForwardButton() {
+  if (window.performance) {
+    // Navigation Timing v2 spec
+    var navEntries = window.performance.getEntriesByType('navigation');
+    if (navEntries.length > 0 && navEntries[0].type === 'back_forward') {
+      return true;
+    // Navigation Timing v1 spec (deprecated but widely supported)
+    } else if (window.performance.navigation
+        && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD) {
+      return true;
+    } else {
+      return false;
+    }
+  } else {
+    return false;
+  }
+}
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -233,6 +233,9 @@ def _get_aips_index_body():
                     },
                     "size": {"type": "double"},
                     "uuid": {"type": "keyword"},
+                    "accessionids": {"type": "keyword"},
+                    "status": {"type": "keyword"},
+                    "file_count": {"type": "integer"},
                 },
             }
         },
@@ -271,6 +274,7 @@ def _get_aipfiles_index_body():
                     "origin": {"type": "text"},
                     "identifiers": {"type": "keyword"},
                     "accessionid": {"type": "keyword"},
+                    "status": {"type": "keyword"},
                 },
             }
         },
--- a/src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
+++ b/src/dashboard/src/templates/archival_storage/_archival_storage_search_form.html
@@ -5,7 +5,7 @@
     <div id="search_form_container" class="clearfix"></div>
     <div id="search_form_submit_container" class="clearfix">
       <span style="margin-left: 10px">{% trans "Show files?" %}</span>
-      <span style="margin-left: 5px"><input id="id_show_files" type="checkbox" /></span>
+      <span style="margin-left: 5px"><input id="id_show_files" type="checkbox" autocomplete="off"/></span>
       <input class="btn btn-primary" type="button" id="search_submit" value="{% trans 'Search archival storage' %}" style="margin-left: 10px">
     </div>
   </div>
--- a/src/dashboard/src/templates/archival_storage/archival_storage.html
+++ b/src/dashboard/src/templates/archival_storage/archival_storage.html
@@ -12,7 +12,6 @@
   <script type="text/javascript" src="{% static 'vendor/datatables/js/dataTables.buttons.min.js' %}"></script>
   <script type="text/javascript" src="{% static 'vendor/datatables/js/buttons.bootstrap.min.js' %}"></script>
   <script type="text/javascript" src="{% static 'vendor/datatables/js/buttons.colVis.min.js' %}"></script>
-  <script type="text/javascript" src="{% static 'js/misc.js' %}"></script>
   <script type="text/javascript" src="{% static 'js/advanced-search-query-creator.js' %}"></script>
   <script type="text/javascript" src="{% static 'js/archival_storage/archival_storage_search.js' %}"></script>
 {% endblock %}


src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -132,6 +132,13 @@ function renderArchivalStorageSearchForm(search_uri, on_success, on_error) {
 }
 
 $(document).ready(function() {
+  // If user arrived via browser back or forward button, reload page to prevent
+  // a DataTables error in Chrome that results from inconsistencies between the
+  // cached DOM and the data returned by the AJAX call.
+  var backOrForwardNavigation = navigatedFromBackOrForwardButton();
+  if (backOrForwardNavigation === true) {
+    window.location.reload();
+  }
```
ross-spencer: It does look like checking this back status is what we need here, but I have a feeling that reloading the entire page is a bit of a "smell" so to speak. It feels a bit like a hammer and we haven't quite got a nail. The clue for me is that we shouldn't really have different behavior between Chrome and Firefox these days when we have the tools to bring them in line. I also agree right now it doesn't quite seem possible to get exactly the desired outcome you'd like, but perhaps we can get closer to something more consistent? 

I'd like to propose the following:

Create a global: `var backOrForwardNavigation = false;` 

And then in `get_datatable()`:

```Javascript
  function get_datatable() {
    // If user arrived via browser back or forward button, reload page to prevent
    // a DataTables error in Chrome that results from inconsistencies between the
    // cached DOM and the data returned by the AJAX call.
    var showFiles = $('#id_show_files').prop('checked');
    if (backOrForwardNavigation === false && showFiles === true) {
       backOrForwardNavigation = navigatedFromBackOrForwardButton();
       if (backOrForwardNavigation === true) {
         showFiles = false;
         $('#id_show_files').prop('checked', false);
       }
    }
```
The complete function I managed to arrive at:
```Javascript
  function get_datatable() {
    // If user arrived via browser back or forward button, reload page to prevent
    // a DataTables error in Chrome that results from inconsistencies between the
    // cached DOM and the data returned by the AJAX call.
    var showFiles = $('#id_show_files').prop('checked');
    if (backOrForwardNavigation === false && showFiles === true) {
       backOrForwardNavigation = navigatedFromBackOrForwardButton();
       if (backOrForwardNavigation === true) {
         showFiles = false;
         $('#id_show_files').prop('checked', false);
       }
    }
    if (showFiles) {
      var cols = [
        {sTitle: gettext('Thumbnail'), mData: 'FILEUUID', mRender: render_thumbnail },
        {sTitle: gettext('File'), mData: 'filePath', mRender: render_filepath },
        {sTitle: gettext('File UUID'), mData: 'FILEUUID'},
        {sTitle: gettext('AIP'), mData: 'sipname', mRender: render_file_aip_info },
        {sTitle: gettext('Accession number'), mData: 'accessionid', defaultContent: ''},
        {sTitle: gettext('Status'), mData: 'status', mRender: render_status },
        {sTitle: gettext('Actions'), mData: 'FILEUUID', mRender: render_file_actions }
      ];
    }
    else {
      var cols = [
        {sTitle: gettext('Name'), mData: 'name', mRender: render_aip_name },
        {sTitle: gettext('UUID'), mData: 'uuid'},
        {sTitle: gettext('AIC'), mData: 'AICID', mRender: render_aip_aic },
        {sTitle: gettext('Size'), mData: 'size'},
        {sTitle: gettext('File count'), mData: 'file_count', defaultContent: ''},
        {sTitle: gettext('Accession numbers'), mData: 'accessionids', mRender: render_aip_accession_ids},
        {sTitle: gettext('Created'), mData: 'created', mRender: render_aip_created_date },
        {sTitle: gettext('Status'), mData: 'status', mRender: render_status },
        {sTitle: gettext('Encrypted'), mData: 'encrypted', mRender: render_aip_encrypted },
        {sTitle: gettext('Actions'), mData: 'uuid', mRender: render_aip_actions }
      ];
    }

    return $('#archival-storage-entries').dataTable({
        'dom': 'rtiBp',
        'stateSave': true,
        'stateDuration': 60 * 60 * 24 * 365 * 10, // Set state duration to 10 years.
        // Don't save information other than column visibility.
        'stateSaveParams': function(settings, data) {
            delete data.search;
            delete data.start;
            delete data.order;
            delete data.length;
            for ( var i=0 ; i< data.columns.length ; i++ ) {
              delete data.columns[i].search;
            }
        },
        'stateSaveCallback': function(settings, data) {
            $.ajax({
                url: '/archival-storage/save_state/' + get_state_url_params(showFiles),
                data: JSON.stringify(data),
                dataType: 'json',
                type: 'POST',
                headers: {'X-CSRFToken': getCookie('csrftoken')},
                success: function() {},
                // Handle case where there is no saved state.
                error: function () {
                  callback(null);
              }
            }
            );
        },
        'stateLoadCallback': function (settings, callback) {
            $.ajax({
                url: '/archival-storage/load_state/' + get_state_url_params(showFiles),
                dataType: 'json',
                success: function (json) {
                  callback(json);
                },
                // Handle case where there is no saved state.
                error: function () {
                  callback(null);
                }
            });
        },
        'columnDefs': [
          {
            'targets': get_default_hidden_column_indices(),
            'visible': false
          },
          {
            'targets': get_unorderable_column_indices(),
            'orderable': false
          }
        ],
        'buttons': [
          {
              'extend': 'colvis',
              'text': '<i class="fa fa-cog" aria-hidden="true"></i> ' + gettext('Select columns')
          }
        ],
      'language': {
        'sEmptyTable':            pgettext('DataTable - sEmptyTable',         'No data available in table'),
        'sInfo':                  pgettext('DataTable - sInfo',               'Showing _START_ to _END_ of _TOTAL_ entries'),
        'sInfoEmpty':             pgettext('DataTable - sInfoEmpty',          'Showing 0 to 0 of 0 entries'),
        'sInfoFiltered':          pgettext('DataTable - sInfoFiltered',       '(filtered from _MAX_ total entries)'),
        'sInfoPostFix':           '',
        'sInfoThousands':         pgettext('DataTable - sInfoThousands',      ','),
        'sLengthMenu':            pgettext('DataTable - sLengthMenu',         'Show _MENU_ entries'),
        'sLoadingRecords':        pgettext('DataTable - sLoadingRecords',     'Loading...'),
        'sProcessing':            pgettext('DataTable - sProcessing',         'Processing...'),
        'sSearch':                pgettext('DataTable - sSearch',             'Search:'),
        'sZeroRecords':           pgettext('DataTable - sZeroRecords',        'No matching records found'),
        'oPaginate': {
            'sFirst':             pgettext('DataTable - sFirst',              'First'),
            'sLast':              pgettext('DataTable - sLast',               'Last'),
            'sNext':              pgettext('DataTable - sNext',               'Next'),
            'sPrevious':          pgettext('DataTable - sPrevious',           'Previous')
        },
        'oAria': {
            'sSortAscending':     pgettext('DataTable - sSortAscending',      ': activate to sort column ascending'),
            'sSortDescending':    pgettext('DataTable - sSortDescending',     ': activate to sort column descending')
        }
      },
      'bLengthChange': false,
      'bFilter': false,
      'bAutoWidth': false,
      'sPaginationType': 'full_numbers',
      'bServerSide': true,
      'sAjaxSource': '/archival-storage/search?' + search.toUrlParams(),
      'aoColumns': cols,
      'fnServerData': function(sSource, aoData, fnCallback) {
        aoData.push({ 'name': 'file_mode', 'value': showFiles });
        $.getJSON(sSource, aoData, function(json) {
          fnCallback(json);
        });
      },
    });
  }
```
Notice  `showFiles` should replace anywhere we use `$('#id_show_files').prop('checked');`.

I recognize instead of using variables in a lot of these functions we're just going straight to the JQuery values. I suspect that's an age-old pattern we've used, but it makes it difficult to control individual state changes like this. 

So, the final result here is we end up with the same behavior on Firefox and Chrome, and we don't refresh the whole page. 

Does that sound like an approach? Or does it push you along at all? 

(I should note, this might need a little more work to test and perform properly, for example I haven't got the show files check box to uncheck yet... but yes, avoiding different behavior and a full reload would be great if we can).

src/dashboard/src/media/js/misc.js
```diff
@@ -139,3 +139,23 @@ $(document).ready(
                 $(this).closest('.preview-help-text').children('.preview').show();
               });
 });
+
+// Return true if user navigated to page via browser back/forward button.
+// Return false if user navigated directly or if browser doesn't support Navigation Timing API.
+function navigatedFromBackOrForwardButton() {
+  if (window.performance) {
+    // Navigation Timing v2 spec
```
ross-spencer, line 147: Here and below we should continue to try and use complete sentences terminated with a full-stop if possible.

src/dashboard/src/media/js/misc.js
```diff
@@ -139,3 +139,23 @@ $(document).ready(
...
+    var navEntries = window.performance.getEntriesByType('navigation');
+    if (navEntries.length > 0 && navEntries[0].type === 'back_forward') {
```
ross-spencer, line 149: Interesting we have `back_forward` here and `window.performance.navigation.TYPE_BACK_FORWARD` below. Do we have a similar constant in the V2 spec?

src/dashboard/src/media/js/misc.js
```diff
@@ -139,3 +139,23 @@ $(document).ready(
                 $(this).closest('.preview-help-text').children('.preview').show();
               });
 });
+
+// Return true if user navigated to page via browser back/forward button.
+// Return false if user navigated directly or if browser doesn't support Navigation Timing API.
+function navigatedFromBackOrForwardButton() {
+  if (window.performance) {
```
ross-spencer, line 146: This is a small thing, but if we align these checks to the happy path, and flip the IF to return earlier we can bring the indents in by one: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 

So:
```javascript
  if !(window.performance) {
    return false;
  }
  // Navigation Timing v2 spec
  var navEntries = window.performance.getEntriesByType('navigation');
  if (navEntries.length > 0 && navEntries[0].type === 'back_forward') {
    return true;
  // Navigation Timing v1 spec (deprecated but widely supported)
  } else if (window.performance.navigation
      && window.performance.navigation.type == window.performance.navigation.TYPE_BACK_FORWARD) {
    return true;
  } else {
    return false;
  }
```
It reduces visual complexity a little which I feel is important here where we end the function with:
```javascript
      return true;
    } else {
      return false;
    }
  } else {
    return false;
  }
}
```
It gets harder to switch back between the lines. 

I think the other if/else if/else might be fine I didn't think too hard about it! :slightly_smiling_face:

ross-spencer (changes requested):
HI @tw4l I feel this is nearly there, but have requested some additional thinking around aligning Chrome and Firefox if possible. I've suggested some code that might help. Let me know. I'm going to be around a little longer if you'd like to chat.

src/dashboard/src/media/js/misc.js
tw4l: Oddly I couldn't find one! Here's the section in the spec: https://www.w3.org/TR/navigation-timing-2/#sec-performance-navigation-types. All of the MDN docs and other places I looked used the "back_forward" string rather than a constant as well.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: Thanks @ross-spencer! I share your gut feeling that the page reload is a bit of a blunt force approach, and appreciate your proposal. While trying to implement something like what you suggest, I realized that every time the user hits back in Chrome from the AIP details page, `showFiles` is always evaluated as `false` on line `var showFiles = $('#id_show_files').prop('checked');`, regardless of the status of the checkbox. Hence why your code to uncheck the "show files" checkbox isn't working - the conditions to satisfy the if statement around it are never met. Whereas in Firefox, `showFiles` appears to be evaluated correctly based on whether or not the "show files" checkbox is checked.

I think I might have figured out a way around this. Will keep investigating and update when I have something more definitive.

tw4l: Interestingly, just setting `var showFiles = $('#id_show_files').prop('checked');` and then using the `showFiles` constant through the rest of the `get_datatable` function is enough to resolve the DataTables error, though without some extra work that results in new UI inconsistencies.

tw4l: Note for self: once `navigatedFromBackOrForwardButton` returns true, it continues to return true until the page is reloaded or the user clicks through to a new page. In practice, this means adding something like this to the start of `get_datatable()`:

```
backOrForwardNavigation = navigatedFromBackOrForwardButton();
if (backOrForwardNavigation === true) {
  showFiles = false;
  $('#id_show_files').prop('checked', false);
}
```
prevents the "show files" checkbox from working as expected until the page is reloaded or the user goes to another page and then returns to the Archival Storage tab by a means other than their browser's back or forward buttons.

tw4l: Interesting blog post related to this issue: https://madhatted.com/2013/6/16/you-do-not-understand-browser-history

The following stood out to me as a potentially relevant passage for understanding what's going on in this issue:

"Safari and Firefox implement something we can call the 'bfcache' (back-forward cache). This cache stores the current state of the DOM in memory. When a page is re-visited via the back button, no requests are made. The DOM itself is served straight up from cache."

So we're not getting the issue in Firefox because Firefox is simply returning the DOM from its cache, whereas Chrome is attempting to build the page from scratch (though seemingly in a way that is persisting the "show files" checkbox state, which triggers the DataTables error).

This won't solve our issue now, but it looks like Google added an experimental back-forward cache in Chrome 79, just a few months ago: https://www.zdnet.com/article/chrome-79-released-with-tab-freezing-back-forward-caching-and-loads-of-security-features/

tw4l: It turns out that the checkbox state was being cached by `autocomplete` in the Archival Storage search form! Pushing a much simpler fix that brings consistency between browsers in just a minute.

tw4l:
Hi @ross-spencer, this should be ready for re-review. Your comments really helped me dig in and understand exactly what was causing this issue - thanks for the encouragement to dig deeper and the guidance!

In the end, it turned out that the errant checkbox behavior was due to [form autocompletion](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion), and that setting `autocomplete=off` for the "show files" checkbox in the Archival Storage search form solves the issue and makes the user experience identical between Firefox and Chrome for all variations of Archival Storage user behaviour that I could think to test.

In my estimation, this is preferable to the Navigation History approach I took earlier, as this newer approach gets more directly to the source of the problem of the original bug. I also realized along way that that the previous approach of checking whether the user had arrived at the page via the browser back/forward buttons introduced its own problems, such as a new bug where the "show files" checkbox wasn't working properly until a page reload.

The MDN link on form autocompletion above points out that turning autocomplete off on a form breaks a WCAG 2.1 rule on [identifying input purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html). From my reading of that rule, however, I don't know that the "show files" checkbox is the type of input they have in mind, as it's not something that a user's browser would have information to autofill - the provided examples are text inputs for things like names and addresses that are likely to be repetitive across many different sites/apps.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -231,7 +230,9 @@ $(document).ready(function() {
   }
 
   function get_datatable() {
-    if ($('#id_show_files').prop('checked')) {
+    var showFiles = $('#id_show_files').prop('checked');
```
ross-spencer, line 233: Thank you for introducing this variable :+1:

ross-spencer (approved):
Fascinating stuff @tw4l. 

> I also realized along way that that the previous approach of checking whether the user had arrived at the page via the browser back/forward buttons introduced its own problems, such as a new bug where the "show files" checkbox wasn't working properly until a page reload.

Yeah, it was becoming clear that it was a little temperamental to work with. That being said, I really appreciate that you've gone through a solution and exhausted it and then come back to this new approach which has a certain amount of elegance. I learned from it, and I think these are all good things to keep in mind in the future.  

On WCAG. The MDN article seems typically inflexible describing the impact of turning off `autocomplete` as "breaking" WCAG. Though I can appreciate where it comes from. And I also hink we acknowledge at Artefactual we do need to constantly be working at this. 

Delving into the [guidelines on the spec](https://www.w3.org/TR/WCAG21/#input-purposes) a little today to see what concerns there might be there I feel we can say with reasonable confidence that the field we're turning this off is not one of the "required" types. I think we'd struggle to find a compelling argument to put it on our list to improve accessibility too, so I suspect we're good :wink: 

What I can see as well is that you've focused on turning this off for the specific field and not the entire form which is perfect :ok_hand: I would have been looking for that during review too.

PR #1620, opened 2020-06-08
Remove navigatedFromBackOrForwardButton helper

tw4l:
Remove misc.js helper function that was accidentally included following squash and merge of https://github.com/artefactual/archivematica/pull/1618, which included a fix for Chrome DataTable bug that happened when a user navigated to Archival Storage from an AIP detail page via their browser's back button.


ross-spencer (approved):
LGTM! :+1: 

goldblum recycles

PR #1624, opened 2020-06-29
Enable reindexing of indices from packages in Storage Service

tw4l:
This PR introduces a new management task `rebuild_aip_index_from_storage_service.py` for reindexing AIPs from packages in the Storage Service, and adds a `--reindex-from-storage-service` flag to `rebuild_transfer_backlog.py` which enables the same for Transfers. Collectively, these changes allow for reindexing from any non-deleted Transfers and AIPs/AICs in the Storage Service, including those stored in encrypted storage locations.

Both tasks poll the Storage Service API to determine which packages should be indexed, filtering out deleted packages and those not associated with the specified origin pipeline (defaulting to the current pipeline if a value is not provided).

`rebuild_aip_index_from_storage_service.py` downloads a copy of the AIP METS file (and the AIC METS file, for AICs) into a temporary directory and uses that locally downloaded METS file for reindexing. I've made this its own task as there were substantial enough changes with the existing task that it seemed to warrant being separated, but I would be happy to merge them together if that is preferred. 

`rebuild_transfer_backlog.py`, when called with the  `--reindex-from-storage-service` flag, downloads a copy of each transfer package from the Storage Service into a temporary directory, uses that locally downloaded copy to index the transfer, and then immediately deletes the local copy when it is no longer needed.

These tasks are slower than the existing indexing tasks and could be made more efficient. The reasons for the performance hit seem to be that:
* Although `rebuild_aip_index_from_storage_service.py` downloads the METS file and not the whole AIP, the Storage Service decrypts the entire package by moving a copy of the package to an internal Storage Service location in order to support downloading this one file.
* It would likely be more efficient (assuming more performant decryption for downloading individual files, if the issue described immediately above is fixed) to download the transfer METS file and use that as the source of information for indexing rather than downloading the entire transfer, like how AIPs are handled. This would require modifying the indexing functions for transfers in `elasticSearchFunctions.py` to support reading from a transfer METS file for information about files with a transfer rather than walking the transfer directory.

This PR also introduces a few helper functions in `archivematicaFunctions.py`, `fileOperations.py`, and `storageService.py` to support the new changes,

Fixes https://github.com/archivematica/issues/issues/734 (however, some problems also noted in that issue, such as not being able to arrange and create a SIP from encrypted transfers in the Arrangement tab, are still outstanding as they are not directly related to reindexing and should be addressed in separate issues. I will create those issues as needed as a follow-up to creating this PR and link them below).

Related to https://github.com/archivematica/Issues/issues/1228

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -34,14 +34,18 @@ import pprint
 import re
 from uuid import uuid4
 
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
+
 import scandir
 import six
 from lxml import etree
 from main.models import DashboardSetting
-from namespaces import NSMAP
+from namespaces import NSMAP, xml_find_premis
 
 from amclient import AMClient
-
 
 REQUIRED_DIRECTORIES = (
     "logs",
@@ -59,6 +63,23 @@ MANUAL_NORMALIZATION_DIRECTORIES = [
 ]
 
 AMCLIENT_ERROR_CODES = (1, 2, 3, 4, -1)
+
+# Package UUID suffix is a single dash followed by a UUID v4 with hyphens.
+PACKAGE_UUID_SUFFIX_LENGTH = 37
+
+# Package extension constants here are copied from Storage Service's
+# storage_service.common.utils module.
+COMPRESS_EXTENSION_7Z = ".7z"
+COMPRESS_EXTENSION_BZIP2 = ".bz2"
+COMPRESS_EXTENSION_GZIP = ".gz"
+
+COMPRESS_EXTENSIONS = (
+    COMPRESS_EXTENSION_7Z,
+    COMPRESS_EXTENSION_BZIP2,
+    COMPRESS_EXTENSION_GZIP,
+)
+
+PACKAGE_EXTENSIONS = (".tar",) + COMPRESS_EXTENSIONS
 
 
 def get_setting(setting, default=""):
@@ -434,3 +455,105 @@ def find_transfer_path_from_ingest(transfer_path, shared_path):
         return path
 
     raise Exception("Transfer directory not physically found")
+
+
+def find_aic_mets_filename(mets_root):
+    """Find name of AIC METS file within AIP METS document.
+
+    :param mets_root: AIP METS document root.
+
+    :returns: AIC METS filename or None.
+    """
+    return xml_find_premis(
+        mets_root, "mets:fileSec/mets:fileGrp[@USE='metadata']/mets:file/mets:FLocat"
+    ).get("{" + NSMAP["xlink"] + "}href")
+
+
+def find_aip_dirname(mets_root):
+    """Find name of AIP directory within AIP METS document.
+
+    :param mets_root: AIP METS document root.
+
+    :returns: AIP dirname or None.
+    """
+    return xml_find_premis(mets_root, "mets:structMap/mets:div").get("LABEL")
+
+
+def find_aips_in_aic(aic_root):
+    """Find extent of AIPs in AIC within AIC METS document.
+
+    :param aic_root" AIC METS document root.
+
+    :returns: Count of AIPs in AIC or None.
+    """
+    extent = xml_find_premis(
+        aic_root,
+        "mets:dmdSec/mets:mdWrap/mets:xmlData/dcterms:dublincore/dcterms:extent",
+    )
+    try:
+        return re.search("\d+", extent.text).group()
+    except AttributeError:
+        return None
+
+
+def package_name_from_path(current_path, remove_uuid_suffix=False):
+    """Return name of package without file extensions from current path.
+
+    This helper works for all package types (e.g. transfer, AIP, AIC).
+
+    :param current_path: Current path to package.
+    :param remove_uuid_suffix: Optional boolean to additionally remove
+    UUID suffix.
+
+    :returns: Package name minus any file extensions.
+    """
+    path = Path(current_path)
+    name, chars_to_remove = path.name, 0
+    if remove_uuid_suffix is True:
+        chars_to_remove = PACKAGE_UUID_SUFFIX_LENGTH
+    for suffix in reversed(path.suffixes):
+        if suffix not in PACKAGE_EXTENSIONS:
+            break
+        chars_to_remove += len(suffix)
+    # Check if we have characters to remove to avoid accidentally
+    # returning an empty string with name[:-0].
+    if not chars_to_remove:
+        return name
+    return name[:-chars_to_remove]
+
+
+def relative_path_to_aip_mets_file(uuid, current_path):
+    """Return relative path to AIP METS file.
+
+    :param uuid: AIP UUID.
+    :param current_path: Current path to AIP.
+
+    :returns: Relative path to AIP METS file.
+    """
+    package_name_without_extensions = package_name_from_path(current_path)
+    mets_name = "METS.{}.xml".format(uuid)
+    mets_path = "{}/data/{}".format(package_name_without_extensions, mets_name)
+    return mets_path
+
+
+def filter_packages_by_status_and_pipeline(
+    package_list, valid_statuses=("UPLOADED", "DEL_REQ"), pipeline_uuid=None
+):
+    """Filter packages by status and origin pipeline.
+
+    :param package_list: List of package info returned by Storage
+    Service (list).
+    :param valid_statuses: Acceptable statuses for filter (tuple).
+    :param pipeline_uuid: Acceptable pipeline UUID for filter (str).
+
+    :returns: Filtered package list.
+    """
+    if pipeline_uuid is None:
+        pipeline_uuid = get_dashboard_uuid()
+    origin_pipeline = "/api/v2/pipeline/{}/".format(pipeline_uuid)
+    return [
+        package
+        for package in package_list
+        if package["status"] in valid_statuses
+        and package["origin_pipeline"] == origin_pipeline
+    ]
--- a/src/archivematicaCommon/tests/test_archivematica_functions.py
+++ b/src/archivematicaCommon/tests/test_archivematica_functions.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 import bagit
 from mock import patch
 
-from archivematicaFunctions import get_bag_size, walk_dir
+import archivematicaFunctions as am
 
 
 def test_get_bag_size(tmpdir):
@@ -22,7 +22,7 @@ def test_get_bag_size(tmpdir):
     # Assertions
     with patch("archivematicaFunctions.os") as mock_os:
         # Test returned value against expected
-        size = get_bag_size(bag, bag_dir.strpath)
+        size = am.get_bag_size(bag, bag_dir.strpath)
         assert size == int(size_oxum.split(".")[0])
         # Verify directory was not walked
         mock_os.path.getsize.assert_not_called()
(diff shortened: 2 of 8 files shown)


tw4l:
Related recently-created issues:
* https://github.com/archivematica/Issues/issues/1244
* https://github.com/archivematica/Issues/issues/1245
* https://github.com/archivematica/Issues/issues/1246
* https://github.com/archivematica/Issues/issues/1247

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -434,3 +455,61 @@ def find_transfer_path_from_ingest(transfer_path, shared_path):
...
+    name, chars_to_remove = path.name, 0
+    if remove_uuid_suffix is True:
+        chars_to_remove = PACKAGE_UUID_SUFFIX_LENGTH
+    for suffix in reversed(path.suffixes):
+        if suffix not in PACKAGE_EXTENSIONS:
+            break
+        chars_to_remove += len(suffix)
+    # Check if we have characters to remove to avoid accidentally returning an
+    # empty string with name[:-0].
+    if not chars_to_remove > 0:
```
replaceafill:
```suggestion
    if not chars_to_remove:
```

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -434,3 +455,61 @@ def find_transfer_path_from_ingest(transfer_path, shared_path):
...
+
+    :returns: Relative path to AIP METS file.
+    """
+    package_name_without_extensions = package_name_from_path(current_path)
+    mets_name = "METS.{}.xml".format(uuid)
+    mets_path = "{}/data/{}".format(package_name_without_extensions, mets_name)
+    return mets_path
+
+
+def filter_packages(package_list, pipeline_uuid=None):
```
replaceafill: The call to this function from `rebuild_transfer_backlog.py` which doesn't have a comment explaining that this will filter out packages with `DELETED` status is a bit unclear. Maybe it could be renamed to `filter_deleted_packages` or pass an iterable explicitly with the status you want to filter out?

src/archivematicaCommon/tests/test_archivematica_functions.py
```diff
@@ -40,10 +40,42 @@ def test_get_bag_size_bag_missing_oxum(tmpdir):
...
+        },
+    ]
+    for test_package in test_packages:
+        current_path = test_package["current_path"]
+
+        package_name = am.package_name_from_path(current_path)
+        assert package_name == test_package["package_name"]
+
+        package_name_without_uuid = am.package_name_from_path(
+            current_path, remove_uuid_suffix=True
```
replaceafill, line 84: Is it worth adding a test case with `remove_uuid_suffix=False`?

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -166,22 +188,70 @@ def populate_data(self, es_client, transfer_backlog_dir):
...
+                )
+            except storageService.Error:
+                self.error(
+                    "Transfer {} not indexed. Unable to download from Storage Service.".format(
+                        transfer_uuid
+                    )
+                )
+                continue
+            # Transfers are downloaded as .tar files, so we extract files before indexing.
+            extract_package(local_package, temp_backlog_dir)
```
replaceafill: Should potential errors in the extraction be handled here?

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -166,22 +188,70 @@ def populate_data(self, es_client, transfer_backlog_dir):
...
+            local_package_sans_extension = am.package_name_from_path(local_package)
```
replaceafill: Do you think `_without_` might be clearer than `_sans_`?

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -0,0 +1,326 @@
...
+        # Ignore elasticsearch-py logging events unless they're errors.
+        logging.getLogger("elasticsearch").setLevel(logging.ERROR)
+        logging.getLogger("archivematica.common").setLevel(logging.ERROR)
+
+        # Verify ES is accessible
+        try:
+            es.setup_reading_from_conf(django_settings)
+            es_client = es.get_client()
+        except ElasticsearchException:
+            self.error("Error: Elasticsearch may not be running")
```
replaceafill: Is it worth logging the exception like `rebuild_transfer_backlog.py` does?

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -0,0 +1,326 @@
...
+        aips_in_aic = None
+        if is_aic is True:
+            aips_in_aic = get_aips_in_aic(package_info, mets_download_path, temp_dir)
+
+        package_name = am.package_name_from_path(
+            package_info["current_path"], remove_uuid_suffix=True
+        )
+
+        if delete_before_reindexing is True:
+            es.delete_aip(es_client, uuid)
```
replaceafill, line 269: Is it worth logging which package is about to be deleted like `rebuild_elasticsearch_aip_index_from_files.py` does?

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -70,11 +77,17 @@ def add_arguments(self, parser):
             "--transfer-backlog-dir", default=self.DEFAULT_TRANSFER_BACKLOG_DIR
         )
         parser.add_argument("--no-prompt", action="store_true")
+        parser.add_argument("--reindex-from-storage-service", action="store_true")
```
replaceafill: Not sure if it's already planned in the scope of this project, but maybe it's worth filing a documentation issue about updating the `Rebuild the indexes` section of the docs to include these new arguments.

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -166,22 +188,70 @@ def populate_data(self, es_client, transfer_backlog_dir):
...
-                    size = walk_dir(str(transfer_dir))
+                    size = am.walk_dir(str(transfer_dir))
                 _import_pipeline_dependant_transfer(
                     self, es_client, self.stdout, transfer_dir, transfer_uuid, size
                 )
             processed += 1
         self.success("{} transfers indexed!".format(processed))
 
+    def populate_data_from_storage_service(self, es_client, pipeline_uuid):
+        """Populate data in search indices and/or database from files temporarily downloaded from Storage Service.
```
replaceafill: This might require further discussion but I think the introduction of `black` has changed what we consider long lines. Docstrings and comments are also a bit special. and [pep8 has](https://www.python.org/dev/peps/pep-0008/#id19):

> ... it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

replaceafill (commented):
@tw4l I think this is on the right track and I really like the METS approach. I think most of my comments don't require significant changes.

However, although I agree with you that the new functionality is substantial enough to have its own command, I think the duplication of code between the new task and `rebuild_elasticsearch_aip_index_from_files` will make maintenance harder.

The `get_aips_in_aic` functions for example have some common functionality. Or the parsing of common arguments (`--delete` and `--delete-all`), and the initial part of the `handle` method (where Elasticsearch is tested and the `--delete-all` argument is handled). Do you think it's worth extracting some of these? Or maybe file an issue and do it later?

src/archivematicaCommon/lib/archivematicaFunctions.py
tw4l: I refactored the function a bit and renamed it to `filter_packages_by_status_and_pipeline`, which I think makes it clearer.

src/archivematicaCommon/tests/test_archivematica_functions.py
tw4l: Agreed and done.

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
tw4l: Good catch -  I didn't realize that the 72 character limit was strict for docstrings and comments. I've reworked this docstring and the others so that they wrap at 72 characters.

tw4l:
Hi @replaceafill - thank you for taking a look at this, and for the helpful suggestions. I think I've addressed your comments in the follow-up commits, including pulling out some of the task setup and shared `get_aips_in_aic` logic into shared helpers as requested to help with maintenance. I also added code in `rebuild_aip_index_from_storage_service.py` to delete METS files immediately after indexing to avoid any potential disk space issues, and renamed the `--reindex-from-storage-service` flag to just `--from-storage-service` to make it a little less unwieldy. Please let me know if you'd like to see any further changes.

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
tw4l: Yes, good point!

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -496,11 +536,15 @@ def relative_path_to_aip_mets_file(uuid, current_path):
     return mets_path
 
 
-def filter_packages(package_list, pipeline_uuid=None):
-    """Filter out deleted packages and packages from other pipelines.
+def filter_packages_by_status_and_pipeline(
+    package_list, valid_statuses=["UPLOADED", "DEL_REQ"], pipeline_uuid=None
```
replaceafill: Even though the function doesn't mutate `valid_statuses` I think it's worth to use a `tuple` for default arguments. Alternatively you could use `None` and then set its initial value. See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
replaceafill: This stills assumes extraction worked, correct?

I was wondering what would happen if the `subprocess.call` fails with a return code other than `0`. Should we at least log that?

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
```diff
@@ -248,8 +257,16 @@ def populate_data_from_storage_service(self, es_client, pipeline_uuid):
                         transfer_uuid,
                         transfer["size"],
                     )
+                    transfer_indexed = True
             shutil.rmtree(temp_backlog_dir)
-            processed += 1
+            if transfer_indexed is True:
```
replaceafill:
```suggestion
            if transfer_indexed:
```

replaceafill (approved):
Thank you @tw4l! LGTM :+1:

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
replaceafill: I noticed you filed https://github.com/archivematica/Issues/issues/1254 Thanks!

src/dashboard/src/main/management/commands/rebuild_transfer_backlog.py
tw4l: Yes, good point! I had taken a more indirect route of throwing an error when the directory expected to be extracted wasn't present, but you're right that it would be better to explicitly check the return code. I will add that before merging.

src/archivematicaCommon/lib/archivematicaFunctions.py
tw4l: TIL!

PR #1626, opened 2020-07-07
Enable download of complete Archival Storage Elasticsearch results as CSV and enables display of storage location

ross-spencer:
Enable download of complete Archival Storage Elasticsearch results as CSV and enables display of storage location

This series of commits adds two new features as well as tests to:

* Download the AIP Elasticsearch index as CSV.
* Enable display of storage location in the Archival Storage tab. 

Connected to archivematica/issues#1213
Connected to archivematica/issues#1214
Relies on https://github.com/artefactual/archivematica/pull/1625

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -57,6 +57,18 @@ AIP_FILES_INDEX = "aipfiles"
 TRANSFERS_INDEX = "transfers"
 TRANSFER_FILES_INDEX = "transferfiles"
 
+ES_FIELD_AICID = "AICID"
+ES_FIELD_ACCESSION_IDS = "accessionids"
+ES_FIELD_AICCOUNT = "countAIPsinAIC"
+ES_FIELD_CREATED = "created"
+ES_FIELD_ENCRYPTED = "encrypted"
+ES_FIELD_FILECOUNT = "file_count"
+ES_FIELD_LOCATION = "location"
+ES_FIELD_NAME = "name"
+ES_FIELD_SIZE = "size"
+ES_FIELD_STATUS = "status"
+ES_FIELD_UUID = "uuid"
+
 
 class ElasticsearchError(Exception):
     """ Not operational errors. """
@@ -226,16 +238,17 @@ def _get_aips_index_body():
             DOC_TYPE: {
                 "date_detection": False,
                 "properties": {
-                    "name": {
+                    ES_FIELD_NAME: {
                         "type": "text",
                         "fields": {"raw": {"type": "keyword"}},
                         "analyzer": "file_path_and_name",
                     },
-                    "size": {"type": "double"},
-                    "uuid": {"type": "keyword"},
-                    "accessionids": {"type": "keyword"},
-                    "status": {"type": "keyword"},
-                    "file_count": {"type": "integer"},
+                    ES_FIELD_SIZE: {"type": "double"},
+                    ES_FIELD_UUID: {"type": "keyword"},
+                    ES_FIELD_ACCESSION_IDS: {"type": "keyword"},
+                    ES_FIELD_STATUS: {"type": "keyword"},
+                    ES_FIELD_FILECOUNT: {"type": "integer"},
+                    ES_FIELD_LOCATION: {"type": "keyword"},
                 },
             }
         },
@@ -263,7 +276,7 @@ def _get_aipfiles_index_body():
                     "AIPUUID": {"type": "keyword"},
                     "FILEUUID": {"type": "keyword"},
                     "isPartOf": {"type": "keyword"},
-                    "AICID": {"type": "keyword"},
+                    ES_FIELD_AICID: {"type": "keyword"},
                     "indexedAt": {"type": "double"},
                     "filePath": {
                         "type": "text",
@@ -274,7 +287,7 @@ def _get_aipfiles_index_body():
                     "origin": {"type": "text"},
                     "identifiers": {"type": "keyword"},
                     "accessionid": {"type": "keyword"},
-                    "status": {"type": "keyword"},
+                    ES_FIELD_STATUS: {"type": "keyword"},
                 },
             }
         },
@@ -287,16 +300,16 @@ def _get_transfers_index_body():
         "mappings": {
             DOC_TYPE: {
                 "properties": {
-                    "name": {
+                    ES_FIELD_NAME: {
                         "type": "text",
                         "fields": {"raw": {"type": "keyword"}},
                         "analyzer": "file_path_and_name",
                     },
-                    "status": {"type": "text"},
+                    ES_FIELD_STATUS: {"type": "text"},
                     "ingest_date": {"type": "date", "format": "dateOptionalTime"},
-                    "size": {"type": "long"},
-                    "file_count": {"type": "integer"},
-                    "uuid": {"type": "keyword"},
+                    ES_FIELD_SIZE: {"type": "long"},
+                    ES_FIELD_FILECOUNT: {"type": "integer"},
+                    ES_FIELD_UUID: {"type": "keyword"},
                     "accessionid": {"type": "keyword"},
                     "pending_deletion": {"type": "boolean"},
                 }
@@ -320,7 +333,7 @@ def _get_transferfiles_index_body():
                     "fileuuid": {"type": "keyword"},
                     "sipuuid": {"type": "keyword"},
                     "accessionid": {"type": "keyword"},
-                    "status": {"type": "keyword"},
+                    ES_FIELD_STATUS: {"type": "keyword"},
                     "origin": {"type": "keyword"},
                     "ingestdate": {"type": "date", "format": "dateOptionalTime"},
                     # METS.xml files in transfers sent to backlog will have ''
@@ -334,8 +347,8 @@ def _get_transferfiles_index_body():
                         "format": "dateOptionalTime",
                         "ignore_malformed": True,
                     },
-                    "created": {"type": "double"},
-                    "size": {"type": "double"},
+                    ES_FIELD_CREATED: {"type": "double"},
+                    ES_FIELD_SIZE: {"type": "double"},
                     "tags": {"type": "keyword"},
                     "file_extension": {"type": "keyword"},
                     "bulk_extractor_reports": {"type": "keyword"},
@@ -398,6 +411,7 @@ def index_aip_and_files(
     aips_in_aic=None,
     identifiers=None,
     encrypted=False,
+    location="",
     printfn=print,
 ):
     """Index AIP and AIP files with UUID `uuid` at path `path`.
@@ -478,21 +492,22 @@ def index_aip_and_files(
     printfn("Indexing AIP ...")
 
     aip_data = {
-        "uuid": uuid,
-        "name": name,
+        ES_FIELD_UUID: uuid,
+        ES_FIELD_NAME: name,
         "filePath": aip_stored_path,
-        "size": int(aip_size) / (1024 * 1024),
-        "file_count": files_indexed,
+        ES_FIELD_SIZE: int(aip_size) / (1024 * 1024),
+        ES_FIELD_FILECOUNT: files_indexed,
         "origin": get_dashboard_uuid(),
-        "created": created,
-        "AICID": aic_identifier,
+        ES_FIELD_CREATED: created,
+        ES_FIELD_AICID: aic_identifier,
         "isPartOf": is_part_of,
-        "countAIPsinAIC": aips_in_aic,
+        ES_FIELD_AICCOUNT: aips_in_aic,
         "identifiers": identifiers,
         "transferMetadata": aip_metadata,
-        "encrypted": encrypted,
-        "accessionids": accession_ids,
-        "status": STATUS_UPLOADED,
+        ES_FIELD_ENCRYPTED: encrypted,
+        ES_FIELD_ACCESSION_IDS: accession_ids,
+        ES_FIELD_STATUS: STATUS_UPLOADED,
+        ES_FIELD_LOCATION: location,
     }
 
     _wait_for_cluster_yellow_status(client)
@@ -552,11 +567,11 @@ def _index_aip_files(client, uuid, mets, name, identifiers=None, aip_metadata=No
         "filePath": "",
         "fileExtension": "",
         "isPartOf": is_part_of,
-        "AICID": aic_identifier,
+        ES_FIELD_AICID: aic_identifier,
         "METS": {"dmdSec": {}, "amdSec": {}},
         "origin": get_dashboard_uuid(),
         "accessionid": "",
-        "status": STATUS_UPLOADED,
+        ES_FIELD_STATUS: STATUS_UPLOADED,
     }
 
     # Index all files in a fileGrup with USE='original' or USE='metadata'.
@@ -739,13 +754,13 @@ def index_transfer_and_files(
     printfn("Indexing Transfer ...")
 
     transfer_data = {
-        "name": transfer_name,
-        "status": status,
+        ES_FIELD_NAME: transfer_name,
+        ES_FIELD_STATUS: status,
         "accessionid": accession_id,
         "ingest_date": ingest_date,
-        "file_count": files_indexed,
-        "size": int(size),
-        "uuid": uuid,
+        ES_FIELD_FILECOUNT: files_indexed,
+        ES_FIELD_SIZE: int(size),
+        ES_FIELD_UUID: uuid,
         "pending_deletion": pending_deletion,
     }
 
@@ -829,12 +844,12 @@ def _index_transfer_files(
                     "fileuuid": file_uuid,
                     "sipuuid": uuid,
                     "accessionid": accession_id,
-                    "status": status,
+                    ES_FIELD_STATUS: status,
                     "origin": dashboard_uuid,
                     "ingestdate": ingest_date,
-                    "created": create_time,
+                    ES_FIELD_CREATED: create_time,
                     "modification_date": modification_date,
-                    "size": size,
+                    ES_FIELD_SIZE: size,
                     "tags": [],
                     "file_extension": file_extension,
                     "bulk_extractor_reports": bulk_extractor_reports,
@@ -1205,7 +1220,10 @@ def search_all_results(client, body, index):
 
 
 def get_aip_data(client, uuid, fields=None):
-    search_params = {"body": {"query": {"term": {"uuid": uuid}}}, "index": AIPS_INDEX}
+    search_params = {
+        "body": {"query": {"term": {ES_FIELD_UUID: uuid}}},
+        "index": AIPS_INDEX,
+    }
 
     if fields:
         search_params["_source"] = fields
@@ -1364,7 +1382,7 @@ def get_transfer_file_info(client, field, value):
 
 
 def remove_backlog_transfer(client, uuid):
-    _delete_matching_documents(client, TRANSFERS_INDEX, "uuid", uuid)
+    _delete_matching_documents(client, TRANSFERS_INDEX, ES_FIELD_UUID, uuid)
 
 
 def remove_backlog_transfer_files(client, uuid):
@@ -1401,7 +1419,7 @@ def _remove_transfer_files(client, uuid, unit_type=None):
 
 
 def delete_aip(client, uuid):
-    _delete_matching_documents(client, AIPS_INDEX, "uuid", uuid)
+    _delete_matching_documents(client, AIPS_INDEX, ES_FIELD_UUID, uuid)
 
 
 def delete_aip_files(client, uuid):
@@ -1428,7 +1446,7 @@ def _delete_matching_documents(client, index, field, value):
 
 
 def _update_field(client, index, uuid, field, value):
-    document_id = _document_id_from_field_query(client, index, "uuid", uuid)
+    document_id = _document_id_from_field_query(client, index, ES_FIELD_UUID, uuid)
 
     if document_id is None:
         logger.error(
@@ -1442,7 +1460,7 @@ def _update_field(client, index, uuid, field, value):
 
 
 def mark_aip_stored(client, uuid):
-    _update_field(client, AIPS_INDEX, uuid, "status", STATUS_UPLOADED)
+    _update_field(client, AIPS_INDEX, uuid, ES_FIELD_STATUS, STATUS_UPLOADED)
 
 
 def mark_aip_deletion_requested(client, uuid):
@@ -1453,7 +1471,7 @@ def mark_aip_deletion_requested(client, uuid):
     :return: None.
     """
     _mark_deletion_request(
-        client, AIPS_INDEX, AIP_FILES_INDEX, "AIPUUID", uuid, "status", "DEL_REQ"
+        client, AIPS_INDEX, AIP_FILES_INDEX, "AIPUUID", uuid, ES_FIELD_STATUS, "DEL_REQ"
     )
 
 
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -200,6 +200,46 @@ def get_location(path=None, purpose=None, space=None):
     return return_locations
 
 
+def location_description_from_slug(aip_location_slug):
+    """Retrieve the location resource description
+
+    The location slug can be retrieved by microservices using the
+    %AIPsStore%% argument. This helper enables easy access to the
+    resource description provided at the URL.
+
+    Example slugs:
+
+       * /api/v2/location/3e796bef-0d56-4471-8700-eeb256859811/
+       * /api/v2/location/default/AS/"
+
+    :param string aip_location: storage location URI slug
+    :return: storage service location description
+    :rtype: dict
+    """
+    API_SLUG = "/api/v2/"
+    JSON_MIME = "application/json"
+    CONTENT_TYPE_HDR = "content-type"
+    service_uri = _storage_service_url()
+    service_uri = service_uri.replace(API_SLUG, aip_location_slug)
+    response = {}
+    with ss_api_timer(function="get_location"):
+        response = _storage_api_session().get(service_uri)
+    if not response or response.status_code != 200:
+        LOGGER.warning(
+            "Cannot retrieve storage location description from storage service: %s",
+            response.status,
+        )
+        return {}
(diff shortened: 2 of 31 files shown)


src/dashboard/src/requirements/base.txt
```diff
@@ -39,16 +39,18 @@ lxml==3.5.0               # via -r base.in, metsrw
 metsrw==0.3.15            # via -r base.in
 mysqlclient==1.4.6        # via -r base.in, agentarchives
 ndg-httpsclient==0.5.1    # via -r base.in
+numpy==1.16.6             # via pandas
+pandas==0.24.2            # via -r base.in
```
sevein, line 43: Please don't see it as an objection but as a curiosity - the size of numpy + pandas together is near the size of the rest of our dependencies in dashboard, see https://gist.github.com/sevein/43bb237f7139e22c9f16e5bfddb69295.

src/archivematicaCommon/lib/storageService.py
```diff
@@ -200,6 +200,46 @@ def get_location(path=None, purpose=None, space=None):
...
+    """
+    API_SLUG = "/api/v2/"
+    JSON_MIME = "application/json"
+    CONTENT_TYPE_HDR = "content-type"
+    service_uri = _storage_service_url()
+    service_uri = service_uri.replace(API_SLUG, aip_location_slug)
+    response = {}
+    with ss_api_timer(function="get_location"):
+        response = _storage_api_session().get(service_uri)
+    if not response and response.status_code != 200:
```
tw4l: Should this `and` be an `or`? Or possibly `if not (response and response.status_code == 200):`? The parentheticals might not be the most idiomatic but I find sometimes they help to clarify intent.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -57,6 +57,18 @@
...
+ES_FIELD_ACCID = "accessionids"
+ES_FIELD_AICCOUNT = "countAIPsinAIC"
+ES_FIELD_CREATED = "created"
+ES_FIELD_ENCRYPTED = "encrypted"
+ES_FIELD_FILECOUNT = "file_count"
+ES_FIELD_LOCATION = "location"
+ES_FIELD_NAME = "name"
+ES_FIELD_SIZE = "size"
+ES_FIELD_STATUS = "status"
+ES_FIELD_UUID = "uuid"
```
tw4l: Nice!

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -57,6 +57,18 @@
 TRANSFERS_INDEX = "transfers"
 TRANSFER_FILES_INDEX = "transferfiles"
 
+ES_FIELD_AICID = "AICID"
+ES_FIELD_ACCID = "accessionids"
```
tw4l: I might consider renaming this to something like `ES_FIELD_ACCESSION` or `ES_FIELD_ACCESSIONID` for the sake of readability -  `ES_FIELD_AICID` and `ES_FIELD_ACCID` are a bit easy to mistake for each other.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
     return table_columns[file_mode][index]
 
 
+def _normalize_and_capitalize_csv_header():
+    """Capitalize the column headers of a Pandas DataFrame
+
+    Because the Archivematica code uses a mix of snake case, camel case
+    and not well formed variants of either plus all-caps we a dict
+    provides the lo-fi way to do this without changing significant parts
```
tw4l: This could use a bit of editing.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
+    column_lookup[es.ES_FIELD_STATUS] = _("Status")
+    column_lookup["type"] = _("Type")
+    column_lookup[es.ES_FIELD_ENCRYPTED] = _("Encrypted")
+    column_lookup[es.ES_FIELD_LOCATION] = _("Location")
+    return column_lookup
+
+
+def search_as_csv(data_frame, file_name=CSV_FILE_NAME, mime_type=CSV_MIMETYPE):
+    if mime_type != CSV_MIMETYPE:
+        mime_type = CSV_MIMETYPE
```
tw4l: If mime_type is always `CSV_MIMETYPE`, I might consider removing mime_type from the function signature and deleting these lines, and then later in the function using the constant directly, e.g.:

```
    response = StreamingHttpResponse(
        data_frame.to_csv(
            encoding=ENCODING, quoting=csv.QUOTE_ALL, index=False, chunksize=1024
        ),
        content_type=CSV_MIMETYPE,
    )
    response[CONTENT_DISPOSITION_HDR] = CONTENT_DISPOSITION
    response[MIME_HDR] = "{}; charset={}".format(CSV_MIMETYPE, ENCODING)
```

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
+    for i, cell in enumerate(data_frame[ACCESSION_IDS_FIELD]):
+        normalized_value = FIELD_SEPARATOR.join(cell)
+        data_frame.at[i, ACCESSION_IDS_FIELD] = normalized_value
+    CREATED_FIELD = es.ES_FIELD_CREATED
+    data_frame[CREATED_FIELD] = data_frame[CREATED_FIELD].apply(str)
+    for i, cell in enumerate(data_frame[CREATED_FIELD]):
+        normalized_value = make_aware(
+            datetime.fromtimestamp(float(cell)), timezone=get_current_timezone()
+        )
+        data_frame.at[i, CREATED_FIELD] = normalized_value
```
tw4l: Given the aim of enabling future extensibility re: other download types besides CSV (which I think you've done a nice job of balancing here vs. the immediate needs of the project), I might consider moving the bits of this formatting code that would be identical for any file type - so likely date normalization but possibly not the accession number join, which seems tailored for CSV specifically - to `search_as_file`.

Another option might be to move each of these normalization steps to their own function that takes a DataFrame as input and returns a DataFrame with a particular column formatted, and then call them from `search_as_file` or `search_as_csv` as appropriate.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
+    CONTENT_DISPOSITION = 'attachment; filename="{}"'.format(file_name)
+    for i, cell in enumerate(data_frame[ACCESSION_IDS_FIELD]):
+        normalized_value = FIELD_SEPARATOR.join(cell)
+        data_frame.at[i, ACCESSION_IDS_FIELD] = normalized_value
+    CREATED_FIELD = es.ES_FIELD_CREATED
+    data_frame[CREATED_FIELD] = data_frame[CREATED_FIELD].apply(str)
+    for i, cell in enumerate(data_frame[CREATED_FIELD]):
+        normalized_value = make_aware(
+            datetime.fromtimestamp(float(cell)), timezone=get_current_timezone()
+        )
```
tw4l, line 252: Nice!

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
+        data_frame.at[i, CREATED_FIELD] = normalized_value
+    data_frame.columns = [
+        _normalize_and_capitalize_csv_header().get(column_name, column_name)
+        for column_name in data_frame.columns
+    ]
```
tw4l: A comment clarifying purpose might be helpful here. Is it to translate the header row?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
     :param request: Django request object.
     :return: A JSON object including required metadata for the datatable and
     the search results.
     """
+    REQUEST_FILE = "requestFile"
+    MIMETYPE = "mimeType"
+    RETURN_ALL = "returnAll"
+    FILE_NAME = "fileName"
+
+    # active = request.POST.get('active') == 'true'
```
tw4l: Can this be removed?

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -186,32 +195,117 @@ def get_es_property_from_column_index(index, file_mode):
...
+    request_file = request.GET.get(REQUEST_FILE, "").lower() == "true"
+    file_mime = request.GET.get(MIMETYPE)
+    file_name = request.GET.get(FILE_NAME)
+
+    # Configure page-size requirements for the search.
+    PAGE_SIZE = 10
```
tw4l: Perhaps `DEFAULT_PAGE_SIZE` would be a better name?

src/dashboard/src/media/css/archival_storage.css
```diff
@@ -124,3 +124,12 @@ div.search-summary p {
...
 }
+
+#download-csv-btn {
+  float: right;
+}
+
+.button-text-span {
+  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-size: 14px;
+}
```
tw4l: This can be deleted - identical styles for `.button-text-span` are already present on lines 119-122 (I know because I added them based on your feedback on my Archival Storage PR! :smile: )

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -132,7 +132,12 @@ function renderArchivalStorageSearchForm(search_uri, on_success, on_error) {
 }
 
 $(document).ready(function() {
-  set_create_aic_visibility();
+  // Show files is a page state. We collect that state information early for
+  // use in each of the functions that then need it.
+  var showFiles = $('#id_show_files').prop('checked');
```
tw4l, line 137: Very nice! I'm glad you got around to doing this properly :smile:

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -254,6 +257,7 @@ $(document).ready(function() {
         {sTitle: gettext('Created'), mData: 'created', mRender: render_aip_created_date },
         {sTitle: gettext('Status'), mData: 'status', mRender: render_status },
         {sTitle: gettext('Encrypted'), mData: 'encrypted', mRender: render_aip_encrypted },
+        {sTitle: gettext('Location'), mData: 'location' },
```
tw4l, line 260: Given the display problems of "Store AIP in standard Archivematica Directory" mentioned in https://github.com/archivematica/Issues/issues/1271, a render function here could be useful to strip out the "Store AIP in " substring if present. Of course, we'd have to the do the same in the DataFrame normalization for this to make its way to the CSV and there's almost certainly a better solution with a bit more time and effort, but something to consider in the interim.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
```diff
@@ -413,4 +423,45 @@ $(document).ready(function() {
...
+    btn = $("#download-csv-btn");
+    if (showFiles) {
+      btn.hide();
+    } else {
+      btn.show();
+    }
+  }
+
+  // Return the contents of the Elasticsearch index as a CSV file.
+  document.getElementById(downloadCSVButton).onclick = function() { exportFile(csvMimeType, csvFileName) };
```
tw4l, line 443: Good job on the vanilla JS!

src/dashboard/src/templates/archival_storage/archival_storage.html
```diff
@@ -61,6 +61,11 @@ <h1>
       {% trans "Create an AIC" %}
     </button>
 
+    <a id="download-csv-btn" class="btn btn-default fa-download fa"
+      title="{% trans 'Create an AIC from current search results' %}">
```
tw4l: This title should be updated. Currently "Create an AIC from current search results" is displaying as the hovertext/tooltip for the download CSV button.

src/dashboard/src/templates/archival_storage/archival_storage.html
```diff
@@ -61,6 +61,11 @@ <h1>
       {% trans "Create an AIC" %}
     </button>
 
+    <a id="download-csv-btn" class="btn btn-default fa-download fa"
```
tw4l, line 64: Feel free to ignore if I'm off-base, but I seem to remember this being a border-less button in the wireframe. If so it might be good to add some CSS like the following for this element:
```
border: none;
background: none;
```

src/MCPClient/tests/test_index_aip.py
```diff
@@ -0,0 +1,55 @@
...
+
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+sys.path.append(os.path.abspath(os.path.join(THIS_DIR, "../lib/clientScripts")))
+
+from index_aip import retrieve_storage_location_description
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "slug,return_value,expected_result",
```
tw4l: Great usage of pytest :smile:  I'm learning a thing or two here...

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -105,7 +109,7 @@ def dashboard_login_and_setup(client, django_user_model, username, password):
 def get_streaming_response(streaming_content):
     response_text = ""
     for response_char in streaming_content:
-        response_text = "{}{}".format(response_text, response_char.decode("utf8"))
+        response_text = "{}{}".format(response_text, response_char)
```
replaceafill: Is this breaking functionality somehow at the moment? If I remember correctly I added it because it fails in Django 2.0 with Python 3 because the responses are binary.

src/MCPClient/tests/test_index_aip.py
```diff
@@ -0,0 +1,55 @@
...
+    [
+        (
+            "/api/v2/location/3e796bef-0d56-4471-8700-eeb256859811/",
+            {"description": "a description"},
+            "a description",
+        ),
```
tw4l: Might be worth adding a test case with both description and path, to test that it defaults to using the description if both are present.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
ross-spencer: Suspect this was accidentally removed @replaceafill while i was wrestling with streaming content and missed in my comb-through. Will be sure to change it back.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -200,6 +204,154 @@ def test_get_pointer_known_pointer(
...
+            "isPartOf": None,
+            "size": "200.1\xa0KB",
+            "type": "AIC",
+        },
+        {
+            "status": "Stored",
+            "encrypted": True,
+            "AICID": None,
+            "countAIPsinAIC": None,
+            "accessionids": ["Àà", "Éé", "Îî", "Ôô", "Ùù"],
```
tw4l, line 255: I really appreciate the Unicode testing here. Good looking out! :100:

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -200,6 +204,154 @@ def test_get_pointer_known_pointer(
...
+
+    streamed_content = b"".join([content for content in response.streaming_content])
+    csv_file = StringIO(streamed_content)
+    data_frame = pd.read_csv(csv_file, header=0, encoding="utf8")
+
+    # Make sure that our headers come out as expected and in the right order.
+    assert list(data_frame.columns.values) == ordered_headers
+
+    # Make some assertions about the quality of the rest of our data,
+    # particularly those we're encoding as
```
tw4l: This comment could use a bit of editing.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -200,6 +204,154 @@ def test_get_pointer_known_pointer(
...
+    accession_ids = data_frame["Accession IDs"]
+    assert pd.isnull(accession_ids[0])
+    assert accession_ids[1] == "; ".join(mock_augmented_result[1].get("accessionids"))
+
+    encyypted = data_frame["Encrypted"]
+    assert encyypted[0] == mock_augmented_result[0].get("encrypted")
+    assert encyypted[1] == mock_augmented_result[1].get("encrypted")
```
tw4l: I think `encyypted` here should be `encrypted`.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -200,6 +204,154 @@ def test_get_pointer_known_pointer(
...
+    Given the ability to download the Elasticsearch AIP index table as a
+    CSV, make sure that the new pathway through the code does not
+    impact the regular Archival Search component by performing a
+    rudimentary test to just ensure that we can successfully NOT
+    download the results as a CSV.
+    """
+    dashboard_login_and_setup(client, django_user_model, username, password)
+    MOCK_TOTAL = 10
+    AUG_RESULTS = {"mock": "response"}
+    CONTENT_TYPE_HDR = "content-type"
```
tw4l: Feel free to ignore, but `CONTENT_TYPE_HDR` and `JSON_MIME` could be a constants at file scope, since they are used in a few different tests.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -200,6 +204,154 @@ def test_get_pointer_known_pointer(
...
+    size = data_frame["Size"]
+    assert size[0] == mock_augmented_result[0].get("size")
+    assert size[1] == mock_augmented_result[1].get("size")
+
+    location = data_frame["Location"]
+    assert location[0] == mock_augmented_result[0].get("location")
+    assert location[1] == mock_augmented_result[1].get("location")
+
+
+def test_search_as_csv_invalid_route(
```
tw4l, line 312: Nice coverage.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -534,3 +535,32 @@ def relative_path_to_aip_mets_file(uuid, current_path):
     mets_name = "METS.{}.xml".format(uuid)
     mets_path = "{}/data/{}".format(package_name_without_extensions, mets_name)
     return mets_path
+
+
+def retrieve_storage_location_description(aip_location_slug, logger=None):
```
tw4l: I'm curious about why this is in `archivematicaFunctions.py`. I think I would have expected it to be in `storageService.py`.

tw4l (changes requested):
Thanks @ross-spencer, this is really good work!

I've suggested some minor changes and made some comments where I had questions, but overall the feature is working quite well and I like the approach you've taken. The quality and coverage of the tests (including ensuring that we have proper Unicode support) is especially great, and you've done a really nice job of balancing keeping this extensible (e.g. to set us up for other download file types in the future) without over-engineering it.

In terms of the reliance on #1625, personally I think it would be fine to merge that first or to close that PR and consider the issue solved in this commit (I don't find the new pip-compile noisiness too much an impediment to code review, though I can certainly see and think it's good you raised the issue), but perhaps that's something to clarify with @sevein.

src/dashboard/src/templates/archival_storage/archival_storage.html
ross-spencer: Good spot. Thanks for the CSS tip too. I appreciate that.

src/MCPClient/tests/test_index_aip.py
ross-spencer: Good call.

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
ross-spencer: True!

src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
ross-spencer: Good call. And indeed, `CONTENT_TYPE` had already been used at file scope so I was just repeating myself adding it again here and without looking there previously.

src/archivematicaCommon/lib/archivematicaFunctions.py
ross-spencer: I'll have time to change this if the logic doesn't track, so just let me know. My rationale was simply that while something from the storage service is needed here the function isn't responsible for communicating with the storage service which I think the storage service functions should do. Here we've got something more specific to the Archivematica application itself. It's not a perfect abstraction mind.

src/dashboard/src/templates/archival_storage/archival_storage.html
ross-spencer: Here lieth the evidence of copy-and-paste!

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
ross-spencer: I appreciate your thinking here. I was also wondering the same but decided against it for a handful of reasons, primarily, I think this is first and foremost a very temporary issue - would we create a render function if that string in the storage service was more suitable today? But second, we really have very little control over the data downstream. If that string was translated (it turns out it's not available for translation yet) would we strip out similar non-English variants? How would we couple what strings we try and format here with those that might actually be constructed between the storage service vs. Archivematica? 

For me anyway, it's a case of you get out what you put in, and we can avoid the maintenance burden of an additional function here that might need to tackle some of those challenges (or would soon be redundant) by making sure the storage service description is more fit for display by default, and then helping users be more aware of that so they're also constructing useful descriptors.

src/dashboard/src/components/archival_storage/views.py
ross-spencer: It sure can! And who even knows what I was trying to do with this one! :slightly_smiling_face:

src/dashboard/src/components/archival_storage/views.py
ross-spencer: The second option is quite attractive and complements the first too so I've combined both for the version I'll submit and I think it may help the code be more descriptive too. Nice work.

src/archivematicaCommon/lib/storageService.py
ross-spencer: Correct! Good spot. I went with the `or` this time around. I also verified this worked with the tests, but they don't need changing as the function still returned as expected even if via a different pathway. But good to have the fix.

src/dashboard/src/media/js/archival_storage/archival_storage_search.js
tw4l: Very sensible. And a good point too on the issue of not having much control over the data downstream, including translations.

src/archivematicaCommon/lib/archivematicaFunctions.py
tw4l: Right, I see where you're coming from. My thinking was that this seems very similar to the [filter_packages](https://github.com/artefactual/archivematica/blob/5d46475461c49997464338adc2be538f4e66a70d/src/archivematicaCommon/lib/storageService.py#L567) function from the recent reindexing work, which similarly does not directly call the Storage Service itself but handles the response of such a call. I do see that the heavy lifting (and Storage Service communication) is being done by `storage_service.location_description_from_slug`.

I don't feel strongly that this needs to change, but I am trying to determine (largely for myself) how to think about separation of functions between these modules moving forward.

ross-spencer:
Thanks @tw4l I've submitted some changes, and left comments on most things here. Let me know what you think of course. 

WRT: 

> In terms of the reliance on #1625, personally I think it would be fine to merge that first or to close that PR and consider the issue solved in this commit (I don't find the new pip-compile noisiness too much an impediment to code review, though I can certainly see and think it's good you raised the issue),...

Then that's good to hear that it doesn't impede you. I also don't mind too much. That being said, within reason, the less noisy a commit or a PR is to review the better. You may see in future CRs noisy requirements files are clues as to something having not being done right - usually an update having being missed in a previous PR. So it's helpful to at least have those as separate commits. This PR just happens to be the first adding to the requirements since logging #1625. It's really incidental that there's  any connection at all. #1625 was as much about researching why the change, than it was about the change. Otherwise all you'd see here is https://github.com/artefactual/archivematica/pull/1626/commits/2cccdf446aac9b4df937396626cdeb724c983c9a around reqs.

src/archivematicaCommon/lib/archivematicaFunctions.py
ross-spencer: @tw4l that's a really good point, and I did make that suggestion so let me be consistent and move it :slightly_smiling_face: and perhaps it's something we can figure out more in future.

ross-spencer: Done! 

> but I am trying to determine (largely for myself) how to think about separation of functions between these modules moving forward.

It's a good point and honestly I don't know. I flipped-flopped before submitting the PR. And I wasn't thinking too much about that other function but it's useful to think about in hindsight. I feel the consistency is better now at least.

src/dashboard/src/media/css/archival_storage.css
```diff
@@ -124,3 +124,17 @@ div.search-summary p {
...
+  background: none;
+  margin-top: 10px;
+}
+
+#unused {
+  border: none;
+  background: none;
+  margin-bottom: 0px;
+  padding-bottom: 0px;
+}
```
tw4l: Is this rule set leftover from the development process?

tw4l (approved):
Thanks @ross-spencer. This is great work and a feature that I'm sure many Archivematica users will be happy about.

I left one comment about a new CSS rule set introduced in commit 7a5db7706abe2c6e542f76c1b079568a40c1bfc6 that looks like it should maybe be deleted before squashing and merging. Otherwise LGTM!

src/dashboard/src/media/css/archival_storage.css
ross-spencer: :facepalm: doh! yes! Good catch. Will remove @tw4l.

tw4l: Sounds good, thanks! :smile:

PR #1628, opened 2020-07-13
Fix bugs in new task for reindexing AIPs from Storage Service

tw4l:
Fixes two bugs in the new AIP reindexing command added in commit 443fc2e66a101a0160951f06fa2f3a01717c9ce0:

* The task was not previously filtering out replicated AIPs, which was leading to a `lxml.etree.XMLSyntaxError` error for replicated AIPs during indexing (specifically, in the `index_aips_and_files` function in `elasticSearchFunctions.py`)
* The `--uuid` flag to reindex a single AIP was throwing an `AttributeError` because it was trying to get a property from a list, rather than from the first member of that list, and was incorrectly expecting `process_package` to return a tuple rather than a boolean.

Bugs addressed in this PR:
* https://github.com/archivematica/Issues/issues/1262
* https://github.com/archivematica/Issues/issues/1263

Connected to https://github.com/archivematica/issues/issues/734

### DIFF
--- a/src/archivematicaCommon/lib/archivematicaFunctions.py
+++ b/src/archivematicaCommon/lib/archivematicaFunctions.py
@@ -534,26 +534,3 @@ def relative_path_to_aip_mets_file(uuid, current_path):
     mets_name = "METS.{}.xml".format(uuid)
     mets_path = "{}/data/{}".format(package_name_without_extensions, mets_name)
     return mets_path
-
-
-def filter_packages_by_status_and_pipeline(
-    package_list, valid_statuses=("UPLOADED", "DEL_REQ"), pipeline_uuid=None
-):
-    """Filter packages by status and origin pipeline.
-
-    :param package_list: List of package info returned by Storage
-    Service (list).
-    :param valid_statuses: Acceptable statuses for filter (tuple).
-    :param pipeline_uuid: Acceptable pipeline UUID for filter (str).
-
-    :returns: Filtered package list.
-    """
-    if pipeline_uuid is None:
-        pipeline_uuid = get_dashboard_uuid()
-    origin_pipeline = "/api/v2/pipeline/{}/".format(pipeline_uuid)
-    return [
-        package
-        for package in package_list
-        if package["status"] in valid_statuses
-        and package["origin_pipeline"] == origin_pipeline
-    ]
--- a/src/archivematicaCommon/lib/storageService.py
+++ b/src/archivematicaCommon/lib/storageService.py
@@ -562,3 +562,44 @@ def download_package(package_uuid, download_directory):
     if local_filename is None:
         raise Error("Unable to download package {}".format(local_filename))
     return local_filename
+
+
+def filter_packages(
+    package_list,
+    statuses=("UPLOADED", "DEL_REQ"),
+    package_types=("AIP", "AIC", "transfer", "DIP"),
+    pipeline_uuid=None,
+    filter_replicas=False,
+):
+    """Filter packages by status and origin pipeline.
+
+    :param package_list: List of package info returned by Storage
+    Service (list).
+    :param statuses: Acceptable statuses for filter (tuple). Defaults
+    to filtering out deleted packages.
+    :param package_types: Acceptable package types for filter (tuple).
+    :param pipeline_uuid: Acceptable pipeline UUID for filter (str).
+    :param filter_replicas: Option to filter out replicas (bool).
+
+    :returns: Filtered package list.
+    """
+    if pipeline_uuid is None:
+        pipeline_uuid = am.get_dashboard_uuid()
+    origin_pipeline = "/api/v2/pipeline/{}/".format(pipeline_uuid)
+
+    if filter_replicas:
+        return [
+            package
+            for package in package_list
+            if package["status"] in statuses
+            and package["package_type"] in package_types
+            and package["origin_pipeline"] == origin_pipeline
+            and package["replicated_package"] is None
+        ]
+    return [
+        package
+        for package in package_list
+        if package["status"] in statuses
+        and package["package_type"] in package_types
+        and package["origin_pipeline"] == origin_pipeline
+    ]
--- a/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
+++ b/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
@@ -39,6 +39,7 @@ import logging
 from lxml import etree
 import os
 import shutil
+import sys
 import tempfile
 
 from elasticsearch import ElasticsearchException
@@ -47,6 +48,8 @@ import archivematicaFunctions as am
 import elasticSearchFunctions as es
 from main.management.commands import DashboardCommand, setup_es_for_aip_reindexing
 import storageService
+
+PACKAGE_TYPES_TO_INDEX = ("AIP", "AIC")
 
 
 def get_aips_in_aic(mets_root, temp_dir, uuid):
@@ -114,109 +117,75 @@ class Command(DashboardCommand):
         logging.getLogger("elasticsearch").setLevel(logging.ERROR)
         logging.getLogger("archivematica.common").setLevel(logging.ERROR)
 
-        # Setup es_client and delete indices if required.
-        es_client = setup_es_for_aip_reindexing(self, options["delete_all"])
-
         # Create temporary directory for downloaded METS files.
         temp_dir = tempfile.mkdtemp()
 
-        # Set pipeline UUID to use for filtering packages to reindex.
         pipeline_uuid = options["pipeline"]
+        delete_all = options["delete_all"]
 
         delete_before_reindexing = False
-        if options["delete"] is True:
+        if options["delete"]:
             delete_before_reindexing = True
 
-        # Index a single package and return if uuid option is enabled.
         if options["uuid"]:
-            package_uuid = options["uuid"]
-            self.info("Reindexing package {}".format(package_uuid))
-            package_info = storageService.get_file_info(uuid=package_uuid)
-
-            is_aic = False
-            package_type = package_info.get("package_type")
-            if package_type and package_type == "AIC":
-                is_aic = True
-
-            (index_success, error_message) = self.process_package(
-                es_client,
-                package_info,
-                temp_dir,
-                delete_before_reindexing,
-                is_aic=is_aic,
+            aips_to_index = storageService.get_file_info(uuid=options["uuid"])
+            # If we're indexing only one AIP, don't delete the indices.
+            delete_all = False
+        else:
+            # For bulk operations, index all AIPs and AICs associated
+            # with the pipeline that are not deleted or replicas.
+            packages = storageService.get_file_info()
+            aips_to_index = storageService.filter_packages(
+                packages,
+                package_types=PACKAGE_TYPES_TO_INDEX,
+                pipeline_uuid=pipeline_uuid,
+                filter_replicas=True,
             )
-            if index_success is True:
-                self.info("Successfully reindexed package {}".format(package_uuid))
-            else:
-                self.error(
-                    "Error indexing package {0}. Details: {1}".format(
-                        package_uuid, error_message
-                    )
-                )
+        aips_to_index_count = len(aips_to_index)
 
-            self.info("Cleaning up")
-            shutil.rmtree(temp_dir)
-            self.info("Indexing complete.")
-            return
+        # If there's nothing to index, log error and quit.
+        if not aips_to_index_count:
+            self.error("No AIPs found to index. Quitting.")
+            sys.exit(1)
 
+        # Setup es_client and delete indices if required.
+        es_client = setup_es_for_aip_reindexing(self, delete_all)
         self.info("Rebuilding 'aips' and 'aipfiles' indices")
 
+        # Index packages.
         packages_not_indexed = []
-
-        # Index non-deleted AIPs associated with this pipeline.
-        aips = storageService.get_file_info(package_type="AIP")
-        filtered_aips = am.filter_packages_by_status_and_pipeline(
-            aips, pipeline_uuid=pipeline_uuid
-        )
-        filtered_aip_count = len(filtered_aips)
-
         aip_indexed_count = 0
-        for aip in filtered_aips:
+        for aip in aips_to_index:
+            is_aic = False
+            if aip["package_type"] == "AIC":
+                is_aic = True
             index_success = self.process_package(
-                es_client, aip, temp_dir, delete_before_reindexing
+                es_client, aip, temp_dir, delete_before_reindexing, is_aic=is_aic
             )
-            if index_success is True:
+            if index_success:
                 aip_indexed_count += 1
             else:
                 packages_not_indexed.append(aip["uuid"])
-
-        # Index non-deleted AICs associated with this pipeline.
-        aics = storageService.get_file_info(package_type="AIC")
-        filtered_aics = am.filter_packages_by_status_and_pipeline(
-            aics, pipeline_uuid=pipeline_uuid
-        )
-        filtered_aic_count = len(filtered_aics)
-
-        aic_indexed_count = 0
-        for aic in filtered_aics:
-            index_success = self.process_package(
-                es_client, aic, temp_dir, delete_before_reindexing, is_aic=True
-            )
-            if index_success is True:
-                aic_indexed_count += 1
-            else:
-                packages_not_indexed.append(aic["uuid"])
 
         # Clean up and report on packages indexed.
         self.info("Cleaning up")
         shutil.rmtree(temp_dir)
 
         if packages_not_indexed:
-            indexing_stats = "{0} of {1} AIPs and {2} of {3} AICs".format(
-                aip_indexed_count,
-                filtered_aip_count,
-                aic_indexed_count,
-                filtered_aic_count,
-            )
-            self.info(
-                "Indexing complete. Indexed {0}. Packages not indexed: {1}.".format(
-                    indexing_stats, ", ".join(packages_not_indexed)
+            self.error(
+                "Indexing complete. Indexed {count} of {total} AIPs/AICs. Packages not indexed: {uuids}.".format(
+                    count=aip_indexed_count,
+                    total=aips_to_index_count,
+                    uuids=", ".join(packages_not_indexed),
                 )
             )
         else:
-            self.info(
-                "Indexing complete. Indexed {0} AIPs and {1} AICs".format(
-                    aip_indexed_count, aic_indexed_count
+            pluralized_aips_aics_term = (
+                "AIP/AIC" if aip_indexed_count == 1 else "AIPs/AICs"
+            )
+            self.success(
+                "Indexing complete. Successfully indexed {count} {term}.".format(
+                    count=aip_indexed_count, term=pluralized_aips_aics_term
                 )
             )
 
@@ -254,7 +223,7 @@ class Command(DashboardCommand):
             return False
 
         aips_in_aic = None
-        if is_aic is True:
+        if is_aic:
             mets_root = etree.parse(mets_download_path)
             aips_in_aic = get_aips_in_aic(mets_root, temp_dir, uuid)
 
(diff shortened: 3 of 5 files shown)


src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -536,6 +536,17 @@ def relative_path_to_aip_mets_file(uuid, current_path):
...
 
+def filter_replicated_aips(aips):
+    """Filter out replicated AIPs.
+
+    :param aips: List of package info for AIPs returned by Storage
+    Service.
+
+    :returns: Filtered AIP list.
+    """
+    return [aip for aip in aips if aip["replicated_package"] is None]
```
ross-spencer: This might be a useful helper in `AMClient` where we already have a helper for compressed packages as an example. I am not convinced it sits particularly well here as its own helper in Archivematica given it is just a single list comprehension. 

As this is dealing with storage service functions, a nice approach might be to modify the storage service helper. 

* `storageService.get_file_info(package_type="AIP")`

If you were interacting with an API, you might see a filter. Something like:

* `storageService.get_file_info(package_type="AIP", filter_replicas=True)`
* `storageService.get_file_info(package_type="AIP", filter=REPLICAS)` (where replicas is an `enum` like value.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -131,32 +131,36 @@ def handle(self, *args, **options):
...
-            package_info = storageService.get_file_info(uuid=package_uuid)
+            try:
+                package_info = storageService.get_file_info(uuid=package_uuid)[0]
+            except IndexError:
+                self.error(
+                    "Error indexing package {0}. "
+                    "Package with supplied UUID not found in Storage Service".format(
+                        package_uuid
+                    )
+                )
```
ross-spencer: Is this `black` formatted?

ross-spencer: An `AttributeError` is mentioned in the PR comment, is this the error that we're covering here? Or is this a tangential issue?

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -131,32 +131,36 @@ def handle(self, *args, **options):
...
+                return
```
ross-spencer: As the return from this function will be `None`. We should consider removing the explicit return here.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -131,32 +131,36 @@ def handle(self, *args, **options):
...
-                )
 
             self.info("Cleaning up")
             shutil.rmtree(temp_dir)
-            self.info("Indexing complete.")
+            if index_success:
+                self.success("Success! Indexing complete.")
+            else:
+                self.error("Failure. Indexing complete.")
             return
```
ross-spencer: We can remove this `return` too.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -131,32 +131,36 @@ def handle(self, *args, **options):
...
-                    )
-                )
 
             self.info("Cleaning up")
             shutil.rmtree(temp_dir)
-            self.info("Indexing complete.")
+            if index_success:
+                self.success("Success! Indexing complete.")
+            else:
+                self.error("Failure. Indexing complete.")
```
ross-spencer: Shall we continue to test for `True` here like in the original line at `148`? If it isn't true, can we return immediately after 147? Or do we still need to commit to clean up? Is there any more information that the user (or us) need here, will we get enough information to track back from: `Failure. Indexing complete.` do you think?

ross-spencer (commented):
Hi @tw4l I've left a few comments which I wouldn't mind getting your feedback on. 

In terms of process, I'm surprised this PR hasn't got an issue associated with it. Or, to be clear, I think archivematica/Issues#734 might otherwise have been moved into `review` and then one new issues associated with the new changes to this work so that there was good visibility prior to this PR, and simply because they are new issues. 

I think we've done it this way previously (without an issue) for your previous PR, but it can lead to rushing out the work we're doing where we should measure twice and cut once as much as possible. 

I'll drop you a line in Slack too.

src/archivematicaCommon/lib/archivematicaFunctions.py
ross-spencer: It dawns on me too, is there a way to only filter once? What information are we getting from the `storage_service.get_file_info()`? 

Is it possible to:

1. Retrieve all AIPs and AICs.
2. Filter replicated packages.
3. Direct the workflow in these functions by testing here whether something is an AIC or an AIP? 

It should simplify the code in rebuild AIP index, and then mean only filtering once. Basically we can remove some complexity (possibly).

tw4l: Hi @ross-spencer - thanks, this is good insight. I've taken the spirit of that you've said here and tried to address it in commit 7cca2c9624865044c95e160104a729fe01777361, which makes a single call to the storage service for packages, does all of the filtering in one shot (and moves that helper to `storageService.py`, where I think it's a better fit), and manages to remove a fair bit of duplicative code in the process. Let me know what you think.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: The `AttributeError` was being raised during `package_type = package_info.get("package_type")`, because `package_info` was the list returned by the Storage Service.

In this PR, I changed it so that `package_info` is the first item of the returned list: `package_info = storageService.get_file_info(uuid=uuid)[0]`. If there are no items in the list (i.e. a package with the specified UUID was not found in the Storage Service), it will raise an `IndexError`, so that's what I'm catching here.

I added some more information on the original bugs in this issue: https://github.com/archivematica/Issues/issues/1263

tw4l: The multi-line comment was run through `black`, but I had split the lines to try to keep line size reasonable. The changes made since have made that moot.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: The return will be `None`, but we need this return so that the task only indexes the one package we requested with the `--uuid` tag. Otherwise, the task will index this one package, and then move on to indexing all AIPs.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: See above.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: I think this is cleaned up a bit now. The messages here were a bit vague because `process_package` outputs more detailed information for each package to the terminal as well.

tw4l:
Thank you @ross-spencer! This is ready for re-review.

I've created issues for each of the bugs being addressed in this PR and linked them in the PR description up top. Moving forward I will stick to the process of filing an issue first, then the related PR. I take your point about measuring twice and cutting once.

Some details on the changes made here are in the in-line comments above, but in general I have refactored to:
* Handle truthiness consistently (with `if x` instead of `if x is True`, especially where we absolutely know values will be booleans)
* Move the logic for indexing a single package to its own function
* Retrieve package info from the Storage Service and filter it once (with a modified version of the previous filtering helper, which I've now moved from `archivematicaFunctions.py` to `storageService.py`, which seems like a better home for it. This also required a slight tweak to the transfer reindexing task). Ideally this filtering could all happen as part of the call to the Storage Service but that would require more major changes, and I hope you will agree that this is a good compromise. It will be a bit faster and, as you suggested, simplifies the `handle` function, which should pay off in terms of readability and maintainability.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: Ordering went off with the refactor, so that comment might not be so helpful. I believe this return is necessary to prevent the task from continuing to index other packages when the --uuid flag is used.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
ross-spencer: Cool. It's just a little jarring as we have `Flake8` [ignoring line-lengths](https://github.com/artefactual/archivematica/blob/396015a44723e6b2dd2f4d7bed1443514a429c12/tox.ini#L38) (E501) up to a certain point to be [compatible with `black`](https://stackoverflow.com/a/59184533) so it's a bit more obvious when the line hasn't been auto-formatted.

src/archivematicaCommon/lib/storageService.py
```diff
@@ -562,3 +562,50 @@ def download_package(package_uuid, download_directory):
...
+    """
+    if pipeline_uuid is None:
+        pipeline_uuid = am.get_dashboard_uuid()
+    origin_pipeline = "/api/v2/pipeline/{}/".format(pipeline_uuid)
+
+    if statuses is None:
+        statuses = ("UPLOADED", "DEL_REQ")
+
+    if package_types is None:
+        package_types = ("AIP", "AIC", "transfer", "DIP")
```
ross-spencer: What do you think about removing these five lines and making them the default parameters in the function signature?
```python
def filter_packages(
    package_list,
    statuses=("UPLOADED", "DEL_REQ"),
    package_types=("AIP", "AIC", "transfer", "DIP"),
    pipeline_uuid=None,
    filter_replicas=False,
): 
```
It seems like the logic would be about the same? It would really only have an impact if someone supplied `None` to the function, which might be a bit odd, but IDK?

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
ross-spencer: That makes sense. I feel like when there's a return like this in the function it says there might be something not quite right with the flow-control. So, in the code above we're directing the flow if `uuid` is set. Given that, I'd almost be tempted to say:
```python
if options["uuid"]:
   ...do thing
else:
   ...do the other thing
```
But that doesn't quite feel right either. I suppose a `return` does keep everything to the left quite nicely. 

But then I thought to myself what else might we be missing.

I had one observation, and that was the setup for Elasticsearch came too early. What happens if we retrieve zero packages from the storage service? e.g. there's a connection error? I think we will have deleted our index and didn't need to.

The second observation I had with the new function for indexing a single package is that we really wanted to do the same thing across both main sections of the code. Given the storage service helper always returns a list, we can treat a list of 1 (or zero) the same we treat a list of 100, with a few small exceptions written along the way.

I have modified your functions here. So I have removed `index_single_package` and moved the bulk of the unique work inline so that most of the logic is the same for bulk or single operations. Also, we now only setup Elasticsearch just in time, so that we only do anything with the index if there is something to do. See what you think:

```diff
diff --git a/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py b/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
index f766ee77..d48c05fa 100644
--- a/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
+++ b/src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
@@ -116,9 +116,6 @@ class Command(DashboardCommand):
         logging.getLogger("elasticsearch").setLevel(logging.ERROR)
         logging.getLogger("archivematica.common").setLevel(logging.ERROR)
 
-        # Setup es_client and delete indices if required.
-        es_client = setup_es_for_aip_reindexing(self, options["delete_all"])
-
         # Create temporary directory for downloaded METS files.
         temp_dir = tempfile.mkdtemp()
 
@@ -129,29 +126,27 @@ class Command(DashboardCommand):
         if options["delete"]:
             delete_before_reindexing = True
 
-        # Index a single package and return if uuid option is enabled.
-        if options["uuid"]:
-            index_success = self.index_single_package(
-                options["uuid"], es_client, temp_dir, delete_before_reindexing
-            )
-            if index_success:
-                self.success("Success! Indexing complete.")
-            return
-
         self.info("Rebuilding 'aips' and 'aipfiles' indices")
-
         packages_not_indexed = []
-
-        # Index AIPs and AICs that are associated with this pipeline
-        # that are not deleted or replicas.
-        packages = storageService.get_file_info()
-        aips_to_index = storageService.filter_packages(
-            packages,
-            package_types=PACKAGE_TYPES_TO_INDEX,
-            pipeline_uuid=pipeline_uuid,
-            filter_replicas=True,
-        )
+        if not options["uuid"]:
+            # Index AIPs and AICs that are associated with this pipeline
+            # that are not deleted or replicas.
+            packages = storageService.get_file_info()
+            aips_to_index = storageService.filter_packages(
+                packages,
+                package_types=PACKAGE_TYPES_TO_INDEX,
+                pipeline_uuid=pipeline_uuid,
+                filter_replicas=True,
+            )
+        else:
+            aips_to_index = storageService.get_file_info(uuid=options["uuid"])
         aips_to_index_count = len(aips_to_index)
+        if aips_to_index_count > 1:
+            # For bulk operations setup es_client and delete indices if required.
+            es_client = setup_es_for_aip_reindexing(self, options["delete_all"])
+        elif aips_to_index_count:
+            # For single item operations, setup es_client and don't delete indices.
+            es_client = setup_es_for_aip_reindexing(self, False)
         aip_indexed_count = 0
         for aip in aips_to_index:
             is_aic = False
@@ -184,37 +179,6 @@ class Command(DashboardCommand):
                 )
             )
 
-    def index_single_package(self, uuid, es_client, temp_dir, delete_before_reindexing):
-        """Index single package by UUID.
-
-        :param uuid: Package UUID.
-        :param es_client: Elasticsearch client.
-        :param temp_dir: Path to tempdir for downloaded METS files.
-        :param delete_before_reindexing: Boolean of whether to delete
-        package from indices prior to reindexing.
-
-        :returns: Boolean indicating success.
-        """
-        self.info("Reindexing package {}".format(uuid))
-        try:
-            package_info = storageService.get_file_info(uuid=uuid)[0]
-        except IndexError:
-            self.error("Error indexing package {0}. UUID not found.".format(uuid))
-            return False
-
-        is_aic = False
-        package_type = package_info.get("package_type")
-        if package_type and package_type == "AIC":
-            is_aic = True
-
-        index_success = self.process_package(
-            es_client, package_info, temp_dir, delete_before_reindexing, is_aic=is_aic
-        )
-
-        self.info("Cleaning up")
-        shutil.rmtree(temp_dir)
-        return index_success
-
     def process_package(
         self, es_client, package_info, temp_dir, delete_before_reindexing, is_aic=False
     ):
```
You might see other optimizations, but does that help at all? (I wasn't quite sure how to write this down in a better narrative, so hoped maybe the code-diff would work as well).

ross-spencer (commented):
Thanks @tw4l. It's all looking a lot cleaner. Nice work bringing those two functions together for the storage service helpers. 

There's a chunky comment/demonstration inline, let me know if that helps. And I hope it's not overreaching too much. I was wrestling with how to jot the point down more effectively.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: Yeah, fair!

src/archivematicaCommon/lib/storageService.py
tw4l: Yeah, agreed that making these the default parameters makes it cleaner and easier to read, and from the docstring one should know not to supply `None` so I think that's an okay tradeoff.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: It did help! I refactored based on your comments here (FWIW, the diff really did help me come around to your point) and the resulting code is shorter, more readable, and handles the case of no packages being returned by the Storage Service much more gracefully.

tw4l:
Hi @ross-spencer - thanks, the inline diff really did help and your thinking brought some really nice changes to the code (I also added you as a co-author on the last commit, and will for the merge commit as well if that's okay with you).

I've flagged this for re-review - let me know if you'd like to see other changes.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -114,109 +117,72 @@ def handle(self, *args, **options):
...
-        # Set pipeline UUID to use for filtering packages to reindex.
         pipeline_uuid = options["pipeline"]
 
         delete_before_reindexing = False
-        if options["delete"] is True:
+        if options["delete"]:
             delete_before_reindexing = True
 
-        # Index a single package and return if uuid option is enabled.
         if options["uuid"]:
```
ross-spencer, line 130: My instinct is still to protect against deletion of the entire index, if the user is only re-indexing a single UUID. The impact of that mistake goes up quite high the bigger the index is. Right now, if `--delete-all` is selected, also `-u` that's definitely possible in this script. What do you think about avoiding that?

There are various ways to do this, but maybe setting `options["'delete_all"]` to `False` here would be enough?

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
```diff
@@ -114,109 +117,72 @@ def handle(self, *args, **options):
...
-                aip_indexed_count,
-                filtered_aip_count,
-                aic_indexed_count,
-                filtered_aic_count,
-            )
-            self.info(
-                "Indexing complete. Indexed {0}. Packages not indexed: {1}.".format(
-                    indexing_stats, ", ".join(packages_not_indexed)
+            self.error(
+                "Indexing complete. Indexed {0} of {1} AIPs/AICs. Packages not indexed: {1}.".format(
```
ross-spencer:
```suggestion
                "Indexing complete. Indexed {} of {} AIPs/AICs. Packages not indexed: {}.".format(
```
I think you want `{2}` here! :wink: that being said it might be a good (inadvertent) demonstration of where we can let the positional arguments do the work, and avoid explicit indexes where possible.

I'd chose `{}` in this case. For other instances if you're reusing variables during string concatenation, then the named approach works quite nicely as a self-describing and very explicit option:
```python
            self.error(
                "Indexing complete. Indexed {demo_1} of {demo_2} AIPs/AICs. Packages not indexed: {demo_3}.".format(
                    demo_1=aip_indexed_count,
                    demo_2=aips_to_index_count,
                    demo_3=", ".join(packages_not_indexed),
                )
            )
```
There's a whole bunch of string concatenation methods but I tend to stick with this one for AM and personally.

ross-spencer (approved):
Nearly there @tw4l with one definite change needed for a string concat further below, and a request to change the logic around deletion a little bit. Let me know if you have any queries. 

Also, I had forgotten to do this, maybe you were already doing it? But `pylint --py3k`
```
$ pylint --py3k src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
Using config file /home/ross-spencer/.pylintrc

------------------------------------
Your code has been rated at 10.00/10
``` 
perfect score: :100: /:100: !!

Approving so you can fix those up and get onwards and upwards once those changes are in.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: Honestly I went back and forth on this because I didn't want to unnecessarily limit users, but ultimately I think you're right that the impact of that mistake outweighs the remote possibility that someone would want to delete their entire index and reindex only one package (especially since I can't think of a use case for the latter). I'll make the change before merging.

src/dashboard/src/main/management/commands/rebuild_aip_index_from_storage_service.py
tw4l: Good catch! This named option is nice. It reminds me of f-strings, which I tend to prefer when possible.

PR #1548, opened 2020-01-29
Change name of sanitize names microservices/jobs

sallain:
See the [issue](https://github.com/archivematica/Issues/issues/230) for rationale for this change.

**Transfer tab**

In 1.10 and earlier, the microservice and job names were:

```
Microservice: Clean up names
   Job: Sanitize Transfer name
   Job: Sanitize object's file and directory names
```

They have been changed to:

```
Microservice: Change transfer filenames
   Job: Change transfer name
   Job: Change object and directory filenames
```

**Ingest tab**

In 1.10 and earlier, the microservice and job names were:

```
Microservice: Clean up names
   Job: Load Dublin Core metadata from disk
   Job: Sanitize SIP name
```

They have been changed to:

```
Microservice: Change SIP filenames
   Job: Load Dublin Core metadata from disk
   Job: Change SIP name
```

I have also changed the names where they appear in `tests/fixtures`.

Connected to archivematica/Issues#230

PR #1629, opened 2020-07-16
Update last few job names to use more inclusive language

sallain:
Just catching the final few job names that used "sanitize". This work
was started in archivematica/Issues#230 but somehow I missed these ones!

Connected to archivematica/Issues#1163

### DIFF
--- a/src/MCPServer/lib/assets/workflow.json
+++ b/src/MCPServer/lib/assets/workflow.json
@@ -820,7 +820,7 @@
                 "stdout_file": "%SIPLogsDirectory%filenameCleanup.log"
             },
             "description": {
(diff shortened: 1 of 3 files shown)


replaceafill (approved):
I just noticed in https://github.com/artefactual/archivematica/pull/1548 you also changed some of the fixture files to match. Do you plan to do the same here? I found a few remaining in:

* `src/dashboard/tests/fixtures/jobs-sip-complete-clean-up-last.json`
* `src/dashboard/tests/fixtures/jobs-sip-complete.json`

sallain:
@replaceafill ahh good catch! I'd forgotten I'd done that. I will fix them up too.

PR #1630, opened 2020-07-16
Add CAS authentication to Archivematica dashboard

tw4l:
This PR bring support for [CAS (Central Authentication Service)](https://www.apereo.org/projects/cas) Single Sign-On to the Archivematica dashboard. The PR uses the [django-cas-ng](https://github.com/django-cas-ng/django-cas-ng) client library for CAS support, which in turn uses the [python-cas](https://github.com/python-cas/python-cas) library.

When the `ARCHIVEMATICA_DASHBOARD_DASHBOARD_CAS_AUTHENTICATION` environment variable is true, unauthenticated users are redirected to the configured CAS server for authentication. If they authenticate successfully, they are returned to Archivematica, where a local user account is created with their username if one does not already exist, and an API key is auto-generated.

This PR also adds some optional configuration choices:
- If `AUTH_CAS_AUTOCONFIGURE_EMAIL` is true, an email address will be added for the local user account, following the rule `USER@DOMAIN`, where the domain name is specified in `AUTH_CAS_EMAIL_DOMAIN`.
- If `AUTH_CAS_CHECK_ADMIN_ATTRIBUTES` is true, Archivematica will read through user attributes returned by the CAS server in the p3/serviceValidate step (available in CAS versions 2 and higher), which are exposed through the `cas_user_authenticated` signal sent by the `django-cas-ng` client library immediately after the user is authenticated. If the key-value pair configured in `AUTH_CAS_ADMIN_ATTRIBUTE` and `AUTH_CAS_ADMIN_ATTRIBUTE_VALUE` is found, the user's `is_superuser` attribute is set to True; otherwise, it is set to False. This check happens for both new and existing users, so that changes in group membership on the CAS server (e.g. a user being added or removed from an administrator group) are applied in Archivematica on the next login.

The CAS server address can be configured with the `AUTH_CAS_SERVER_URL` environment variable. If it's not set, the default is set to a public CAS demo server (https://django-cas-ng-demo-server.herokuapp.com/cas/) for testing purposes. The username/pw combo `admin / django-cas-ng` can be used to authenticate on the demo server.

Because the `AUTH_CAS_CHECK_ADMIN_ATTRIBUTES` attribute parsing can be difficult to test manually (i.e. is reliant on external infrastructure), I've made sure that the tests focus heavily on this aspect of the integration, and tried not to duplicate what's already well-covered in the `django-cas-ng` tests. We have an internal development/QA VM which is configured with the SFU CAS server for an additional layer of testing/QA. Feel free to ping me on Slack for the details.

Issue https://github.com/archivematica/Issues/issues/1270 was discovered during development, which prevents a user in Archivematica `qa/1.x` from viewing their profile if `ALLOW_USER_EDITS` is set to false (as it is for CAS, Shibboleth, and LDAP auth). Commit 31c773ca3d2b64ade0588c82c9ae7aa8aeb14153 fixes this issue by updating URLs in the template so that they are appropriately namespaced to match values expected following the upgrade to Django 1.11.

Connected to https://github.com/archivematica/issues/issues/1211

Related Storage Service PR: https://github.com/artefactual/archivematica-storage-service/pull/538


src/dashboard/src/components/accounts/signals.py
```diff
@@ -1,11 +1,84 @@
...
+        return
+
+    username = kwargs.get("user")
+    attributes = kwargs.get("attributes")
+
+    if not attributes:
+        return
+
+    is_administrator = _user_is_administrator(attributes)
+    user = User.objects.get(username=username)
```
sevein: A future-proof way to reference the user model is [get_user_model()](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#referencing-the-user-model). In practice I think it's not going to matter.

src/dashboard/src/components/accounts/signals.py
```diff
@@ -1,11 +1,82 @@
...
+    :param cas_attributes: Attributes dict returned by CAS server.
+
+    :returns: True if expected value is found, otherwise False.
+    """
+    ADMIN_ATTRIBUTE = settings.CAS_ADMIN_ATTRIBUTE
+    ADMIN_ATTRIBUTE_VALUE = settings.CAS_ADMIN_ATTRIBUTE_VALUE
+    if (ADMIN_ATTRIBUTE is None) or (ADMIN_ATTRIBUTE_VALUE is None):
+        logger.error(
+            "Error determining if new user is an administrator. Please "
+            "be sure that env variables AUTH_CAS_ADMIN_ATTRIBUTE and "
```
sevein: I'd call them "CAS settings" or "application settings" instead of "env variables".

src/dashboard/src/components/accounts/signals.py
```diff
@@ -1,11 +1,82 @@
...
+from django.dispatch import receiver
 from django_auth_ldap.backend import populate_user
+from django_cas_ng.signals import cas_user_authenticated
 
 from components.helpers import generate_api_key
 
+logger = logging.getLogger("archivematica.dashboard")
+
+
+def _user_is_administrator(cas_attributes):
```
sevein: You may want to name it something like `_cas_user_is_administrator` to avoid confusions.

src/dashboard/src/components/accounts/signals.py
```diff
@@ -1,11 +1,82 @@
...
+    attributes = kwargs.get("attributes")
+
+    if not attributes:
+        return
+
+    user = User.objects.get(username=username)
+    is_administrator = _user_is_administrator(attributes)
+    if user.is_superuser != is_administrator:
+        user.is_superuser = is_administrator
+        user.save()
```
sevein: We could try to put a lock on the user row to avoid potential concurrency issues. I think this could work:

```python
User.objects.select_for_update().get(username=username)
```

src/dashboard/src/settings/components/cas_auth.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+
+from os import environ
+
+from django.core.exceptions import ImproperlyConfigured
+
+CAS_DEMO_SERVER_URL = "https://django-cas-ng-demo-server.herokuapp.com/cas/"
```
sevein, line 12: A comment here mentioning that the default uses a [live demo](https://djangocas.dev/blog/django-cas-ng-example-project/#live-demo) would be nice.

src/dashboard/tests/test_cas.py
```diff
@@ -0,0 +1,249 @@
...
+        "service": service,
+        TEST_CAS_ADMIN_ATTRIBUTE: TEST_CAS_ADMIN_ATTRIBUTE_VALUE_POSITIVE,
+    }
+    pgtiou = None
+    return user, attributes, pgtiou
+
+
+@pytest.mark.skipif(
+    not settings.CAS_AUTHENTICATION, reason="tests will only pass if CAS is enabled"
+)
```
sevein, line 71: Could we set the toggle from setUp to ensure that the test runs regardless what's in the settings module?

src/dashboard/src/settings/base.py
```diff
@@ -501,6 +507,20 @@
...
+    # Insert CAS after the authentication middleware
+    MIDDLEWARE.insert(
+        MIDDLEWARE.index("django.contrib.auth.middleware.AuthenticationMiddleware") + 1,
+        "django_cas_ng.middleware.CASMiddleware",
+    )
+
+    from .components.cas_auth import *  # noqa
 
 LDAP_AUTHENTICATION = config.get("ldap_authentication")
 if LDAP_AUTHENTICATION:
```
sevein: It occurs to me that we're allowing users to set up multiple authentication backends. Do we want them to be combined or is that even possible at all?

sevein (changes requested):
Nicely done! I've been testing locally and it works! I've suggested some minor changes.

src/dashboard/src/components/accounts/signals.py
tw4l: Yup, works well!

src/dashboard/tests/test_cas.py
tw4l: I agree it would be great to make the tests run regardless of settings.

I spent an hour or two trying different variations of overriding settings with the decorators described at https://docs.djangoproject.com/en/1.11/topics/testing/tools/#overriding-settings but haven't been able to get it to work. There is a warning in the Django docs that the changes may happen too late to take effect for some settings, and I see a similar comment [in our test settings](https://github.com/artefactual/archivematica/blob/5d46475461c49997464338adc2be538f4e66a70d/src/dashboard/src/settings/test.py#L67-L68). Did you have something in mind that I might not be thinking of?

src/dashboard/src/settings/base.py
tw4l: It's definitely [possible and supported to have multiple authentication backends](https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#specifying-authentication-backends). The custom CAS backend is being added to `AUTHENTICATION_BACKENDS` after `django.contrib.auth.backends.ModelBackend` rather than replacing it, which works fine because Django tries each authentication backend until one works.

I think what would happen if someone enabled more than one authentication scheme in our current settings depends on the combination.

If it was CAS and OIDC (there is an [unmerged PR](https://github.com/artefactual/archivematica/pull/1536) which I used as a reference for the CAS work), I suspect everything would work but the user would only be able to authenticate via CAS because the CAS middleware redirects to the CAS server instead of the Archivematica login screen, where OIDC is presented as an option for login next to local authentication. So I don't think there would be too much value in combining them. If we added something like OAuth as an option in the Archivematica login screen, my guess is we'd see a similar behaviour. But it's also very possible there would be errors - I'm not sure.

But presumably OIDC + OAuth + local login could work just fine, with all three presented as options via the Archivematica login screen.

I ran an experiment where I enabled both CAS and Shibboleth in my `docker-compose.yml`. I was redirected to the CAS server for login and was authenticated but then a `NoReverseMatch` exception was raised because the `shibboleth` namespace wasn't be found. It seems like we could work out these sorts of issues with different combinations of auth schemes, but for schemes like CAS and Shibboleth I think one (likely the one whose middleware is earlier in the list) will take precedence, which begs the question of why one would want to enable them both.

What do you think? Should we build in more safeguards here?

tw4l:
Thank you @sevein ! I addressed most of your comments and followed up on your last two questions with some thoughts/questions of my own and a few requests for guidance :) 

I'll plan to make changes to the Storage Storage PR next week after we worth through some of the issues here a little bit more.

src/dashboard/src/components/accounts/signals.py
```diff
@@ -1,11 +1,83 @@
...
+
+    if not attributes:
+        return
+
+    User = get_user_model()
+    user = User.objects.select_for_update().get(username=username)
+    is_administrator = _cas_user_is_administrator(attributes)
+    if user.is_superuser != is_administrator:
+        user.is_superuser = is_administrator
+        user.save()
```
sevein: Thanks for doing this! Looking at it again I'm realizing that we're introducing a couple of new issues,
* The lock is retained longer that we need since _cas_user_is_administrator is executed within the transaction window, but you could easily avoid that by making that call before the transaction starts, and
* I think that we're not freeing the lock - we'd need to wrap the database-related code using `with transaction.atomic: ...`.

And we don't have to do this at all but I was wondering if this would also be a good case for [conditional expressions](https://docs.djangoproject.com/en/1.8/ref/models/conditional-expressions/#conditional-update) and if that would be a better way to solve it.

src/dashboard/src/settings/base.py
sevein: Thank you, that's useful. I think so, yeah. If we don't have capacity to test all permutations, we can't guarantee it's safe nor its long-term support then I'd rather make them mutually exclusive. If the user really wanted to combine them they can always write their own settings module, but at that point I think it'd be cleat that it's outside our recommendations.

src/dashboard/tests/test_cas.py
sevein: I see! We have all these tests that will only run if the corresponding toggle is enabled, but in practice we never do. It sounds like we'll need multiple runs, e.g. `pytest --ds=settings.test.cas tests/test_cas.py` or similar.

src/dashboard/src/settings/base.py
tw4l: That makes sense to me. After some further thought it's CAS in particular that doesn't play nicely with the other SSO methods (because its middleware circumvents the Archivematica login screen, where other SSO options would be presented to a user or handled within the existing form), so I've added a check in the CAS settings in `base.py` that raises an `ImproperlyConfigured` exception if Shibboleth or LDAP is enabled in addition to CAS, as well as an explanatory comment. Please let me know if you think that resolves the issue.

src/dashboard/src/components/accounts/signals.py
tw4l: Yes, good point! In the past I've relied on Django managing transactions for me. A reading of the Django [database transactions](https://docs.djangoproject.com/en/1.11/topics/db/transactions/) docs was very instructive.

I've moved `_cas_user_is_administrator` and added the `transaction.atomic` context manager, which should resolve the new issues.

Conditional updates are new to me - thanks for pointing it out! From my reading, they seem most appropriate for cases where an update is required and the trick is to determine which value to set. I think this case is somewhat different, where most logins after the first will not require updates to the user instance, since updates will only be necessary if the administrator attribute is added or removed for a user on the CAS server (which I assume happens rarely). So we want to check to see if our application state is out of date and update only if necessary.

tw4l:
Hi @sevein ! I've responded in-line - good catch on needing to wrap the update logic in a transaction block. I think it should be good now but let me know if you'd like to see further changes.

sevein (approved):

PR #1632, opened 2020-07-28
Update package and file status in dashboard after deletion request is rejected

tw4l:
This PR fixes a bug whereby the statuses for transfers and their files, as well as for AIP files, were not updated after a package deletion request was rejected in the Storage Service.

It also includes a commit which modifies the Backlog rendering function to display "Deletion requested" rather than "Delete requested" for the Status column, to fix an inconsistency in language between the Backlog and Archival Storage tabs.

Connected to https://github.com/archivematica/Issues/issues/1273 and https://github.com/archivematica/Issues/issues/1189

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -1470,7 +1470,7 @@ def mark_aip_deletion_requested(client, uuid):
     :param uuid: AIP UUID.
     :return: None.
     """
-    _mark_deletion_request(
+    _update_field_for_package_and_files(
         client, AIPS_INDEX, AIP_FILES_INDEX, "AIPUUID", uuid, ES_FIELD_STATUS, "DEL_REQ"
     )
 
@@ -1482,7 +1482,7 @@ def mark_backlog_deletion_requested(client, uuid):
     :param uuid: Transfer UUID.
     :return: None.
     """
-    _mark_deletion_request(
+    _update_field_for_package_and_files(
         client,
         TRANSFERS_INDEX,
         TRANSFER_FILES_INDEX,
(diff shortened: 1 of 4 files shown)


src/dashboard/src/components/backlog/views.py
```diff
@@ -67,6 +75,8 @@ def check_and_remove_deleted_transfers(es_client):
         if status == es.STATUS_DELETED:
             es.remove_backlog_transfer_files(es_client, transfer_uuid)
             es.remove_backlog_transfer(es_client, transfer_uuid)
+        elif status == es.STATUS_UPLOADED:
+            es.revert_backlog_deletion_request(es_client, transfer_uuid)
```
ross-spencer, line 79: Cool. This makes sense to me. Nice and simple.

src/dashboard/src/components/archival_storage/views.py
```diff
@@ -92,8 +98,8 @@ def check_and_remove_deleted_aips(es_client):
         if aip_status == es.STATUS_DELETED:
             es.delete_aip(es_client, aip_uuid)
             es.delete_aip_files(es_client, aip_uuid)
-        elif aip_status != es.STATUS_DELETE_REQUESTED:
-            es.mark_aip_stored(es_client, aip_uuid)
+        elif aip_status == es.STATUS_UPLOADED:
```
ross-spencer, line 101: :+1:

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1493,10 +1493,46 @@ def mark_backlog_deletion_requested(client, uuid):
...
+        TRANSFERS_INDEX,
+        TRANSFER_FILES_INDEX,
+        "sipuuid",
+        uuid,
+        "pending_deletion",
+        False,
+    )
+
+
+def _update_field_for_package_and_files(
```
ross-spencer, line 1532: This is really nice to see.

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -1493,10 +1493,46 @@ def mark_backlog_deletion_requested(client, uuid):
     )
 
 
-def _mark_deletion_request(
+def revert_aip_deletion_request(client, uuid):
```
ross-spencer, line 1496: This and `revert_backlog_deletion_request` are both sensible wrappers here. Nice.

ross-spencer (approved):
LGTM @tw4l :+1: 

I was looking at `test_elasticsearch_functions.py` tonight with a view to pointing you in the direction of adding some tests but I didn't manage to mock anything out that I think would be useful. It's a good habit to add tests for fixes like this to avoid regressions later on, but it might not be worth it here. Did you happen to look at any at all? I didn't get at all close in the end. So I don't think I'd try it here. I ran through a couple of scenarios with my AM and storage service and it's all looking good.

tw4l:
Thanks @ross-spencer ! I really appreciate the quick CR. I did look at testing, though admittedly briefly, and was similarly stumped as to how to mock up that was useful relative to the amount of time it would take to get there. Today I tried doing something of an integration test by indexing data in ES and then modifying it using the elasticSearchFunctions helpers and testing the results but was a lot of added overhead. Will aim for better coverage moving forward but for this fix I'm going to merge as-is, with the two of us having tested in local AM/Storage Service environments.

PR #1635, opened 2020-08-07
Fix arrangement of transfers with BagIt logs/metadata

sevein:
`copy_from_arrange_to_completed_common` commands Storage Service to copy the
arranged contents into a new SIP. Additionally, this function copies the
corresponding logs and metadata contents from the original transfers. The latter
requires to know whether the transfer(s) are in BagIt format since the accessor
would be different, i.e. BagIt packages include a payload directory (data) not
used in old-style transfers.

`data/objects/forkleaf-sundew.jpg` or `data/objects/roundleaf-sundew.jpg` are
two examples of files that this function successfully identifies as members of
BagIt transfers because they're found inside the "data" payload directory.
However, this function fails to identify BagIt transfers when arranging files
such `data/logs/BagIt/bagit.txt` or `data/metadata/manifest-md5.txt`, which can
be found in transfers originally submitted to Archivematica in BagIt format.

This commit fixes the identification routine so the format of the transfer is
identified consistently.

Fixes https://github.com/archivematica/Issues/issues/1267.

### DIFF
--- a/src/dashboard/src/components/filesystem_ajax/views.py
+++ b/src/dashboard/src/components/filesystem_ajax/views.py
@@ -474,7 +474,9 @@ def copy_from_arrange_to_completed(
     """
     if filepath is None:
         filepath = b64decode_string(request.POST.get("filepath", ""))
-    logger.info("copy_from_arrange_to_completed: filepath: %s", filepath)
+    logger.info(
+        "copy_from_arrange_to_completed: filepath: %s, sip_uuid: %s", filepath, sip_uuid
+    )
     # can optionally pass in the UUID to an unstarted SIP entity
     if sip_uuid is None:
         sip_uuid = request.POST.get("uuid")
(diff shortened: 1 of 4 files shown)


src/dashboard/tests/test_filesystem_ajax.py
```diff
@@ -10,6 +10,11 @@
 from django.test.client import Client
 import pytest
 
+try:
+    import mock
```
replaceafill: Why is the `try/except` necessary here? I noticed `mock` is included in the `test` and `dev` requirements and other test modules in the `dashboard` package don't have it.

src/dashboard/src/components/filesystem_ajax/views.py
```diff
@@ -549,14 +551,14 @@ def copy_from_arrange_to_completed_common(filepath, sip_uuid, sip_name):
             ).split("/", 1)
             transfer_name = transfer_parts[0]
             # Determine if the transfer is a BagIt package
-            is_bagit = True if transfer_parts[1].startswith(u"data/objects") else False
+            is_bagit = transfer_parts[1].startswith(u"data/")
             # Copy metadata & logs to tmp/, where later scripts expect
             for directory in ("logs", "metadata"):
-                source = [DEFAULT_BACKLOG_PATH, transfer_name, directory]
+                source = [DEFAULT_BACKLOG_PATH, transfer_name, directory, "."]
```
replaceafill, line 557: Not suggesting to change it now, but are these trailing dots necessary to turn the paths into directories? I think I've seen (and used) spaces instead.

replaceafill (approved):
LGTM :+1:

src/dashboard/src/components/filesystem_ajax/views.py
sevein: I was assuming this is what SS was expecting, isn't it?

src/dashboard/tests/test_filesystem_ajax.py
sevein: Good point, will fix.

PR #1640, opened 2020-08-13
Report size even if du returns non-zero exit code

tw4l:
When `du` encounters directories for which it doesn't have permissions it returns an exit code of 1, causing `subprocess.check_output` to throw a `CalledProcessError`. This results in `usage_get_directory_used_bytes` returning 0 when run against a directory containing any files or sudirectories for which the `archivematica` user doesn't have permissions, and ultimately in the Dashboard's "Processing storage usage" page reporting that 0 bytes are used on disk:

image

This commit modifies how a non-zero `du` exit code is handled. Now, a warning is logged but the bytes reported by `du` are read from `CalledProcessError.output` so that the total number of bytes in directories the archivematica user can access is still reported in the "Processing storage usage" page. We default to 0 only if `du` did not print a byte count to stdout.

Note that that directories and files for which `archivematica` doesn't have permissions won't be included in the total, meaning that the size reported won't be 100% accurate. Changes to deployments (namely, mounting storage at a directory such as `/var/archivematica` rather than at root) may also be required to circumvent this problem fully.

Connected to https://github.com/archivematica/Issues/issues/1281

### DIFF
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -392,12 +392,16 @@ def _usage_get_directory_used_bytes(path):
         # the archivematica user doesn't have permissions. In this case
         # du still prints an output to stdout so we try to catch it
         # from CalledProcessError.output.
-        logger.warning(
-            "Non-zero exit code while determining usage of %s. Some directories may be missing from total.",
-            path,
-        )
-        byte_count = err.output.split("\t")[0]
-        return byte_count if byte_count else 0
+        byte_count = 0
+        try:
+            byte_count = int(err.output.split("\t")[0])
+            logger.warning(
+                "Non-zero exit code while determining usage of %s. Some directories may be missing from total.",
+                path,
+            )
+        except (AttributeError, ValueError):
+            logger.exception("Unable to determine usage of %s.", path)
+        return byte_count
 
 
 def _usage_clear_context(request, dir_id):
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -396,7 +396,8 @@ def _usage_get_directory_used_bytes(path):
         try:
             byte_count = int(err.output.split("\t")[0])
             logger.warning(
-                "Non-zero exit code while determining usage of %s. Some directories may be missing from total.",
+                "Non-zero exit code while determining usage of %s. "
+                "Some directories may be missing from total.",
                 path,
             )
         except (AttributeError, ValueError):
--- a/src/dashboard/src/components/administration/views.py
+++ b/src/dashboard/src/components/administration/views.py
@@ -386,9 +386,18 @@ def _usage_get_directory_used_bytes(path):
     except OSError:
         logger.exception("No such directory: %s", path)
         return 0
-    except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/administration/views.py
```diff
@@ -386,9 +386,18 @@ def _usage_get_directory_used_bytes(path):
...
+        # 1 because there were some directories in the path for which
+        # the archivematica user doesn't have permissions. In this case
+        # du still prints an output to stdout so we try to catch it
+        # from CalledProcessError.output.
+        logger.warning(
+            "Non-zero exit code while determining usage of %s. Some directories may be missing from total.",
+            path,
+        )
+        byte_count = err.output.split("\t")[0]
+        return byte_count if byte_count else 0
```
sevein: Two things that we should probably consider:
* `err.output` can be `None`
* Using `int(byte_count)` (raises `ValueError`) would ensure that we're returning an integer

sevein (changes requested):

src/dashboard/src/components/administration/views.py
tw4l: Thanks @sevein ! https://github.com/artefactual/archivematica/pull/1640/commits/860508a9616edd1ba08d9dd13feb3abaa382d5db refactors this to return 0 for an `AttributeError` (if `err.output` is `None`) or `ValueError` (if `err.output` can't be cast to an `int`), and otherwise return the number of bytes as an `int`.

src/dashboard/src/components/administration/views.py
```diff
@@ -386,9 +386,22 @@ def _usage_get_directory_used_bytes(path):
...
+        # the archivematica user doesn't have permissions. In this case
+        # du still prints an output to stdout so we try to catch it
+        # from CalledProcessError.output.
+        byte_count = 0
+        try:
+            byte_count = int(err.output.split("\t")[0])
+            logger.warning(
+                "Non-zero exit code while determining usage of %s. Some directories may be missing from total.",
+                path,
+            )
```
sevein: Function parameters can take multiline strings without extra symbols, e.g.:

```suggestion
            logger.warning(
                "Non-zero exit code while determining usage of %s. "
                "Some directories may be missing from total.",
                path,
            )
```

sevein (approved):

amayita:
It is still not accurate 100%, see for example am112bionic:
In `/var/log/archivematica/dashboard/dashboard.log` I still see:
```
WARNING   2020-08-28 15:38:29  archivematica.dashboard:views:_usage_get_directory_used_bytes:401:  
Non-zero exit code while determining usage of /. Some directories may be missing from total.
```
(Moot, I know) ;)

I have tested this on just 2 servers and I am consistently see a ~1GB difference (note these servers are bot under any real use, could it be more in a production server?):

Reported usage is 8.4GB vs real usage being 9.6GB.
In another just deployed test server I saw: Reported usage is 6.5GB vs real usage being 7.4GB.

Screenshot at 2020-08-28 21-39-24

Screenshot_2020-08-28 Archivematica Dashboard - Usage(1)

amayita:
So after discussing it with my people^w support team mates (LOL), we do believe that this is a big improvement, and would be happy to see this go into the next release.

PR #1641, opened 2020-08-20
Ensure that transfer name does not display as blank in Archivematica

ross-spencer:
Ensures that we cannot return `None` when we try and retrieve a directory/transfer name from the Job class. We now ensure the path is terminated with a slash if the string isn't empty, and the default should the regex group still return `None` will be to return the SIP UUID which looked like the original intention of the author. 

Connected to https://github.com/archivematica/Issues/issues/1250

### DIFF
--- a/src/dashboard/tests/test_models.py
+++ b/src/dashboard/tests/test_models.py
@@ -5,6 +5,8 @@ from django.db import IntegrityError
 from django.test import TestCase
 import mock
 
+import pytest
+
 from main import models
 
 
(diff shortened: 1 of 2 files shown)


src/dashboard/tests/test_models.py
```diff
@@ -99,3 +101,75 @@ def test_sip_arrange_create_many_with_integrity_error(mocker):
     # If bulk creation fails each SIPArrange is saved individually
     assert arrange1_mock.save.called_once()
     assert arrange2_mock.save.called_once()
+
+
+class TestJobModel(object):
+    """This had better work..."""
```
tw4l: Maybe this docstring could be removed? xD

tw4l (approved):
Thanks for this @ross-spencer. LGTM - I left only one very minor comment on a docstring leftover from the development process that should probably be modified before merging. The logic here is clear and very easy to read, and I am reliably seeing the transfer name appear in my local development environment and in am112centos, which I repurposed to test this out.

Nice work - I'm glad to see that your persistence with this issue paid off!

src/dashboard/tests/test_models.py
ross-spencer: Ha! That's a good one from me. Good spot too. Thanks Tessa.

PR #1643, opened 2020-08-26
Increase polling interval and make sure that the configuration setting is actually consumed to be used by jobs.js

ross-spencer:
The default interval was previously not instantiated from the configuration settings and so Archivematica would always find itself using the default 5s regardless. This fix ensures that a configured value will work, and also increases that default for a better user-experience.

Connected to archivematica/issues#1017

### DIFF
--- a/src/dashboard/src/media/js/jobs.js
+++ b/src/dashboard/src/media/js/jobs.js
@@ -586,8 +586,30 @@ var BaseJobView = Backbone.View.extend({
           // clone action selector
           var $proxySelect = $select.clone().css({'width': '100%'});
 
+          modalDlg = `
+            <div class="modal" id="big-choice-select-modal">
+              <div class="modal-dialog">
+                <div class="modal-content">
+                  <div class="modal-header">
+                    <button type="button" class="close" id="big-choice-select-close" data-dismiss="modal">×</button>
+                    <h3>
+          ` + gettext('Select an action...') + `
+                    </h3>
+                  </div>
+                  <div class="modal-body" id="big-choice-select-body">
+                  </div>
+                  <div class="modal-footer">
+                    <a href="#" class="btn btn-default" data-dismiss="modal" id="big-choice-select-cancel">
+          ` + gettext('Cancel') + `
+                    </a>
+                  </div>
+                </div>
+              </div>
+            </div>
+          `;
+
           // display action selector in modal window
-          $('<div class="modal" id="big-choice-select-modal"><div class="modal-dialog"><div class="modal-content"><div class="modal-header"><button type="button" class="close" id="big-choice-select-close" data-dismiss="modal">×</button><h3>' + gettext('Select an action...') + '</h3></div><div class="modal-body" id="big-choice-select-body"></div><div class="modal-footer"><a href="#" class="btn btn-default" data-dismiss="modal" id="big-choice-select-cancel">' + gettext('Cancel') + '</a></div></div></div></div>').modal({show: true, backdrop: 'static'});
+          $(modalDlg).modal({show: true, backdrop: 'static'});
           $('#big-choice-select-body').append($proxySelect);
 
           // style clone as Select2
@@ -621,8 +643,6 @@ var BaseJobView = Backbone.View.extend({
 });
 
 BaseAppView = Backbone.View.extend({
-
-  interval: window.pollingInterval ? window.pollingInterval * 1000: 5000,
 
   idle: false,
 
@@ -693,6 +713,12 @@ BaseAppView = Backbone.View.extend({
 
   initialize: function(options)
     {
+      const milliseconds = 1000;
+      const defaultPollingInterval = 10000;
+
+      this.interval = window.pollingInterval ? window.pollingInterval * milliseconds : defaultPollingInterval;
+      console.log("".concat("Polling interval configured as: ", this.interval));
+
       this.statusUrl = options.statusUrl;
 
       _.bindAll(this, 'add', 'remove');
--- a/src/dashboard/src/settings/base.py
+++ b/src/dashboard/src/settings/base.py
@@ -98,6 +98,11 @@ CONFIG_MAPPING = {
         "option": "agentarchives_client_timeout",
         "type": "float",
     },
+    "polling_interval": {
+        "section": "Dashboard",
+        "option": "polling_interval",
+        "type": "int",
+    },
     "prometheus_enabled": {
         "section": "Dashboard",
         "option": "prometheus_enabled",
@@ -143,6 +148,7 @@ storage_service_client_timeout = 86400
 storage_service_client_quick_timeout = 5
 agentarchives_client_timeout = 300
 prometheus_enabled = False
+polling_interval = 10
 site_url =
 time_zone = UTC
 
@@ -437,8 +443,8 @@ else:
 
 # Dashboard internal settings
 GEARMAN_SERVER = config.get("gearman_server")
-POLLING_INTERVAL = 5  # Seconds
-STATUS_POLLING_INTERVAL = 5  # Seconds
+POLLING_INTERVAL = config.get("polling_interval")
+
 TASKS_PER_PAGE = 10  # for paging in tasks dialog
 UUID_REGEX = "[\w]{8}(-[\w]{4}){3}-[\w]{12}"
 
--- a/src/dashboard/install/README.md
+++ b/src/dashboard/install/README.md
@@ -204,6 +204,14 @@ variables or in the gunicorn configuration file.
     - **Type:** `float`
     - **Default:** `300`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_POLLING_INTERVAL`**:
+    - **Description:** describes the interval (in seconds) at which the
+    dashboard client will request an update from the server, e.g. to refresh
+    the microservice jobs display.
+    - **Config file example:** `Dashboard.polling_interval`
+    - **Type:** `integer`
+    - **Default:** `10`
+
 - **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PROMETHEUS_ENABLED`**:
     - **Description:** Determines if Prometheus metrics should be collected.
     - **Config file example:** `Dashboard.prometheus_enabled`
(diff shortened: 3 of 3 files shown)


src/dashboard/src/settings/base.py
```diff
@@ -437,8 +437,7 @@
 
 # Dashboard internal settings
 GEARMAN_SERVER = config.get("gearman_server")
-POLLING_INTERVAL = 5  # Seconds
-STATUS_POLLING_INTERVAL = 5  # Seconds
+POLLING_INTERVAL = 10  # Seconds
```
scollazo: This was indeed needed!

Besides it might be out of scope here, could you change that "10" into a configurable variable, as the gearman_server (or others). I think the config system also allows for setting them as environment variables...

But it would be good if we could change this value without messing with the content of base.py...

scollazo (changes requested):

src/dashboard/src/settings/base.py
ross-spencer: Thanks Santi. Will take a look and try and get this in - should be able too. Nice suggestion.

src/dashboard/src/media/js/jobs.js
```diff
@@ -713,7 +713,7 @@ BaseAppView = Backbone.View.extend({
 
   initialize: function(options)
     {
-      this.interval = window.pollingInterval ? window.pollingInterval * 1000: 5000
+      this.interval = window.pollingInterval ? window.pollingInterval * 1000: 10000
```
tw4l: I would add a space between `1000` and `10000` here so that the ternary expression's a bit easier to read - i.e. `? window.pollingInterval * 1000 : 10000`

tw4l: It seems like there's a missing semi-colon at the end of this line as well? (they're optional in JS but we seem to use them elsewhere `jobs.js` so might be good to be consistent)

src/dashboard/src/settings/base.py
tw4l: Agreed - something like `os.environ.get("POLLING_INTERVAL", 10)` would be nice here. The settings for SSO options in `src/dashboard/src/settings/components` might provide some good examples.

src/dashboard/src/media/js/jobs.js
```diff
@@ -586,8 +586,30 @@ var BaseJobView = Backbone.View.extend({
           // clone action selector
           var $proxySelect = $select.clone().css({'width': '100%'});
 
+          modalDlg = `
```
tw4l, line 589: This is nice cleanup, Ross! Much much more readable :1st_place_medal:

src/dashboard/src/settings/base.py
```diff
@@ -438,7 +438,6 @@
 # Dashboard internal settings
 GEARMAN_SERVER = config.get("gearman_server")
 POLLING_INTERVAL = 5  # Seconds
-STATUS_POLLING_INTERVAL = 5  # Seconds
```
tw4l, line 441: I do notice in the network tab of my browser's dev tools that there seem to be two polling jobs:

image

`/transfer/status/?(timestamp)` is controlled by `POLLING_INTERVAL` and I've confirmed that changing that value is reflected in the Dashboard (the screenshot shows a polling interval of 16s, which is working nicely)

`/status/?(timestamp)` seems to happen every 5 seconds and responds with what I'm assuming is the number of transfers, SIPs, and DIPs awaiting decisions. Is it possible this `STATUS_POLLING_INTERVAL` is or was related to that?

tw4l (commented):
Hi @ross-spencer, nice work! This is looking great. I've verified that `POLLING_INTERVAL` values work as expected, that it's defaulting to 10 seconds, and that tests are passing.

I've left a few comments/questions and a +1 on Santi's suggestion to be able to set `POLLING_INTERVAL` from the environment. It might be a good idea as well with that to do a bit of type checking so that we use the default value if what's set in the environment can't be cast to an int.

One last thing is that I like the user experience of a 10 second default polling time in relation to the dropdowns but the change does have a noticeable impact on how how the quickly the microservices and jobs are updated in the UI. To me the trade-off seems worth it, but it might be worth checking in with @sromkey if you haven't already to see if she agrees.

src/dashboard/src/settings/base.py
ross-spencer: It's an interesting observation. The status that you noticed is controlled by some (unfortunately) hard-coded values, but this one specifically should be this one [here](https://github.com/artefactual/archivematica/blob/c5e7a0fb4aaaf5d30d20e46707b66f788d7a0fba/src/dashboard/src/media/js/status.js#L88). (Verified by changing on the fly). I don't think we know the original intent of `STATUS_POLLING_INTERVAL` but your hypothesis is a pretty good one (it does seem that way). Douglas describes its non-use in the issue attached to the PR and I just confirmed that and removed it here.

ross-spencer:
Okay, I think this might now address CR comments and makes the setting configurable from the environment. Tested in Docker-compose by adding the following setting and modifying the integer value: 

* `ARCHIVEMATICA_DASHBOARD_DASHBOARD_POLLING_INTERVAL: 5`. 

We're looking at finishing this CR and then cherry-picking into `stable/1.12.x` _ideally overnight or early tomorrow._ 

For @sromkey I have recorded two video showing the difference in the microservices requested by Tessa. 

**5 second interval**

https://drive.google.com/file/d/1b2UcGvib2Ua7qs4Cyd_cn6O6dczugDUj/view?usp=sharing

**10 second interval**

https://drive.google.com/file/d/1WE3tJOTBgG8-dI_2tIq8ypEcgjVVF_Ay/view?usp=sharing

Note, as the refresh is cyclical, it might take longer or shorter for the first microservice group to appear depending on whether a count has just reset or not whatever that value is.

sromkey:
> One last thing is that I like the user experience of a 10 second default polling time in relation to the dropdowns but the change does have a noticeable impact on how how the quickly the microservices and jobs are updated in the UI. To me the trade-off seems worth it, but it might be worth checking in with @sromkey if you haven't already to see if she agrees.

I agree, I think the tradeoff is worth it! Thanks for the demonstration @ross-spencer

ross-spencer:
@tw4l in regard to type checking, I have added: https://github.com/artefactual/archivematica/pull/1643/commits/e6e15cc52631c87505d66659e1b00d6583f91f73. Have you an example of what was done for CAS? - it feels, perhaps, a little too defensive for me in that I believe the server just wouldn't start if the environment wasn't configured properly, and there wouldn't be an impact if the server is already running. Let me know what you think though. I can keep this additional try/except in or modify it as required.

src/dashboard/src/settings/base.py
```diff
@@ -437,8 +443,13 @@
 
 # Dashboard internal settings
 GEARMAN_SERVER = config.get("gearman_server")
-POLLING_INTERVAL = 5  # Seconds
-STATUS_POLLING_INTERVAL = 5  # Seconds
+
+DEFAULT_POLLING_INTERVAL = 10
```
sevein: Hey Ross, config defaults are defined in `CONFIG_DEFAULTS` above.

ross-spencer: sorry @sevein have you an example how to use that here?

sevein: I.e. `config.get("polling_interval")` is going to return `10` when the user doesn't provide a value since you indicated it in `CONFIG_DEFAULTS`.

ross-spencer: Right, it does - I was testing that. But Tessa raised the issue of casting in the PR commentary below, and if config.get("polling_interval") can't cast either of those to an integer then the default won't work like that. So I suspect I am missing something - but also, have a look at my comment around the try/except here. I have only added this "default" for that particular last commit.

sevein: `appconfig.Config` is using [getint](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getint) for coercion under the covers for `int` attributes like your new `polling_interval` so we don't have to worry.

ross-spencer: I don't think it's working how you think it is so that's why I'd ask you to try it. 

If I set `ARCHIVEMATICA_DASHBOARD_DASHBOARD_POLLING_INTERVAL: "fff"`: I get this error:

```
archivematica-dashboard_1        |   File "/src/archivematicaCommon/lib/env_configparser.py", line 14, in wrapper
archivematica-dashboard_1        |     return fn(*args, **kwargs)
archivematica-dashboard_1        |   File "/src/archivematicaCommon/lib/env_configparser.py", line 70, in getint
archivematica-dashboard_1        |     return ConfigParser.SafeConfigParser.getint(self, *args, **kwargs)
archivematica-dashboard_1        |   File "/usr/local/lib/python2.7/ConfigParser.py", line 359, in getint
archivematica-dashboard_1        |     return self._get(section, int, option)
archivematica-dashboard_1        |   File "/usr/local/lib/python2.7/ConfigParser.py", line 356, in _get
archivematica-dashboard_1        |     return conv(self.get(section, option))
archivematica-dashboard_1        | ValueError: invalid literal for int() with base 10: 'fff'
```
Really just trying to honor the code-review here and I've left a note about what that mean, or not mean here. Have a think about the intention and let me know.

tw4l:
Hey @ross-spencer - re: the type casting, the example I had remembered seeing was https://github.com/artefactual/archivematica/blob/stable/1.12.x/src/dashboard/src/settings/components/ldap_auth.py#L71-L76 in the LDAP settings. But I didn't consider the relation to `CONFIG_MAPPING` which sets the type explicitly, so I'm afraid I may have led you down a bit of a dead end. My apologies for that!

> it feels, perhaps, a little too defensive for me in that I believe the server just wouldn't start if the environment wasn't configured properly, and there wouldn't be an impact if the server is already running.

That's very fair, and we have a pattern set out with int values set in the default config like `elasticsearch_timeout` so please feel free to disregard my original comment and go with the established pattern here.

Edit: Replace `CONFIG_DEFAULTS` with `CONFIG_MAPPING`

tw4l (approved):
To follow up on the last comment, I would drop https://github.com/artefactual/archivematica/commit/e6e15cc52631c87505d66659e1b00d6583f91f73 but otherwise this LGTM.

ross-spencer:
Ah, cool. Thanks for sharing that Tessa. Glad I wasn't far off. I thought I might be missing something. Will drop that commit, and then sqush those changes and prepare a cherry-pick branch. This can be merged a bit later on, but I appreciate the approval.

src/dashboard/src/settings/base.py
sevein: Ok, I see. I appreciate the effort. I read Tessa's suggestion but misssed [this comment](https://github.com/artefactual/archivematica/pull/1643#discussion_r520666589). I think that the level of strictness is beneficial, this is the only window we have to report errors that is not disruptive. Fail-fast as opposed to masking the issue, if that makes sense. It'd be nice though if the [`env_configparser`](https://github.com/artefactual/archivematica/blob/stable/1.12.x/src/archivematicaCommon/lib/env_configparser.py) reported a friendlier error that includes the section and option values from the config parser.

sevein (approved):

scollazo (approved):
lgtm!

PR #1645, opened 2020-08-28
fpr: stop using get_models and model.rel

sevein:
This commit, similar to ea5f700b7f1a613e3225a997c31f621c099c47bd, changes the fpr app so it stops using deprecated APIs in recent releases of Django.

Fixes https://github.com/archivematica/Issues/issues/1295.


tw4l (approved):
LGTM except for one minor issue: flake8 is failing with `src/dashboard/src/fpr/views.py:4:1: F401 'django.apps.apps' imported but unused` when running the tests via tox. I think the `from django.apps import apps` import just needs to be removed from `views`.

sevein:
Ops thanks, looks like Travis is not publishing the status of the build anymore, but here are the [results](https://travis-ci.org/github/artefactual/archivematica/builds/722976891).

PR #1646, opened 2020-09-08
Add username check migration

sevein:
This commit adds a new migration that interrupts the process when usernames
are longer than they should.

Fixes https://github.com/archivematica/Issues/issues/1278.

### DIFF
--- a/src/dashboard/src/main/migrations/0078_username_check.py
+++ b/src/dashboard/src/main/migrations/0078_username_check.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""Ensures that the username attribute does not exceed maximum expected length.
+
+Column ``auth_user.username`` has a max_length of 150 characters since Django
+brought migration  ``auth.0012_alter_user_first_name_max_length``.
+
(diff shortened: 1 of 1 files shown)


src/dashboard/src/main/migrations/0078_username_check.py
```diff
@@ -0,0 +1,63 @@
...
+_DJANGO_USERNAME_MAX_LENGTH = 150
+
+
+def data_migration_up(apps, schema_editor):
+    User = apps.get_model("auth", "User")
+    result = User.objects.annotate(username_len=Length("username")).aggregate(
+        Max("username_len")
+    )
+    maxlen = result["username_len__max"]
+    interactive = "--no-input" not in sys.argv
```
jraddaoui, line 39: Nice! Not the best flag name for this case but I see you're just "reusing" an existing one.

jraddaoui (approved):
Looks great to me @sevein, thanks!

sevein:
It looks like the problem with Travis persists. Build status is here: https://travis-ci.org/github/artefactual/archivematica/builds/725364787.

PR #1647, opened 2020-09-09
MCPServer: enqueue next job after decisions are processed

sevein:
This commit updates the job completion callback so work can be continued when another package is awaiting.

It needs to be tested before merging.

Connects to https://github.com/archivematica/Issues/issues/1306.

### DIFF
--- a/src/MCPServer/lib/server/queues.py
+++ b/src/MCPServer/lib/server/queues.py
@@ -229,6 +229,7 @@ class PackageQueue(object):
             return
         elif isinstance(next_job, DecisionJob) and next_job.awaiting_decision:
             self.await_decision(next_job)
+            self.queue_next_job()
         else:
             self.schedule_job(next_job)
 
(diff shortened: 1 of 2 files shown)


replaceafill (approved):
I tested the steps to reproduce this issue in `qa/1.x` and then in this branch and can see that the dashboard keeps responding as expected after restarting the services, so this looks good to me :+1:

Nitpick: more inline comments in the new tests would be useful to understand the jobs interactions in the queue. They were easier to understand after I (accidentally) read `src/MCPServer/tests/test_integration.py::test_workflow_integration`

Speaking of `test_workflow_integration`, it also seems flaky. I got the usual:

```console
>       assert package_queue.job_queue.qsize() == 1
E       assert 0 == 1
E        +  where 0 = <bound method Queue.qsize of <Queue.Queue instance at 0x7fd8a4f3b690>>()
E        +    where <bound method Queue.qsize of <Queue.Queue instance at 0x7fd8a4f3b690>> = <Queue.Queue instance at 0x7fd8a4f3b690>.qsize
E        +      where <Queue.Queue instance at 0x7fd8a4f3b690> = <server.queues.PackageQueue object at 0x7fd8a3d155d0>.job_queue
src/MCPServer/tests/test_integration.py:230: AssertionError
```
in one of my first `tox` runs. But after that it never failed again :shrug:

PR #1660, opened 2020-12-09
Remove font-awesome symlink

sevein:
Symlinks are being problematic in our package builds. This commit
vendors font-awesome instead of linking to the instance made availble in
node_modules. Long term, we should probably adopt a tool like
django-compressor or similar.

We're also including this in qa/1.x.

Connects to https://github.com/archivematica/Issues/issues/1300.


mamedin (approved):

sevein:
289f32f5 in qa/1.x

PR #1661, opened 2020-12-14
Add configurable password validation to Archivematica

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1332

This PR adds password validation to the Archivematica dashboard. The validators employed include:
- Django's MinimumLengthValidator (default length: 8 characters)
- Django's CommonPasswordValidator
- Django's UserAttributeSimilarityValidator
- A custom PasswordComplexityValidator, which checks that passwords contain at least 3 of: lower-case characters, upper-case characters, numbers, special characters.

The validators are enabled by default but can be disabled individually via the environment. Since the client requirement for password complexity is not universally seen as a positive and some clients may want to disable it (see for instance the newest [NIST password guidelines](https://en.wikipedia.org/wiki/Password_policy#NIST_guidelines), which discourage composition rules), we decided to make each validator individually configurable rather than providing one option to enable/disable all of them at once.

Password validation is enforced by Django when creating or editing users in the dashboard GUI as well as through the `createsuperuser` and `changepassword` management commands.

Commit a4be60c disables password validation for the local environment. To test that password validation is working in a Docker development environment, either drop that commit locally or comment out `AUTH_PASSWORD_VALIDATORS = []` in ` src/dashboard/src/settings/local.py`.

See also the accompanying PR for the Archivematica Storage Service: https://github.com/artefactual/archivematica-storage-service/pull/553.

### DIFF
--- a/src/dashboard/install/README.md
+++ b/src/dashboard/install/README.md
@@ -162,6 +162,30 @@ variables or in the gunicorn configuration file.
     - **Type:** `string`
     - **Default:** `127.0.0.1:4730`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_MINIMUM_LENGTH`**:
+    - **Description:** sets minimum length for user passwords.
+    - **Config file example:** `Dashboard.password_minimum_length`
+    - **Type:** `integer`
+    - **Default:** `8`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_COMMON_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that occur in a [list of common passwords](https://github.com/django/django/blob/stable/1.11.x/django/contrib/auth/common-passwords.txt.gz).
+    - **Config file example:** `Dashboard.password_disable_common_validation`
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_USER_ATTRIBUTE_SIMILARITY_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that are too similar to their username and other user attributes.
+    - **Config file example:** `Dashboard.password_disable_user_attribute_similarity_validation`
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_COMPLEXITY_VALIDATION`**:
+    - **Description:** disables password validation that checks that passwords contain at least three of: lower-case characters, upper-case characters, numbers, special characters.
+    - **Config file example:** `Dashboard.password_disable_complexity_validation`
+    - **Type:** `boolean`
+    - **Default:** `false`
+
 - **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_SHIBBOLETH_AUTHENTICATION`**:
     - **Description:** enables the Shibboleth authentication system.
     - **Config file example:** `Dashboard.shibboleth_authentication`
--- a/src/dashboard/src/settings/base.py
+++ b/src/dashboard/src/settings/base.py
@@ -63,6 +63,26 @@ CONFIG_MAPPING = {
         "option": "gearman_server",
         "type": "string",
     },
+    "password_minimum_length": {
+        "section": "Dashboard",
+        "option": "password_minimum_length",
+        "type": "int",
+    },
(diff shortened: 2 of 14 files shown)


tw4l:
Hi @ross-spencer - I've updated the composition validator and tests and moved the basic auth settings into a separate component, to mirror the changes we made in the accompanying Storage Service PR. I've also rebased the branch on `qa/1.x`, since there had been a few commits since I opened the PR. It should be ready for review now but don't hesitate to ping if I've missed something!

I'll update the message files prior to merging but haven't done so yet so that it's a bit easier to wade through the PR for code review.

src/dashboard/install/README.md
```diff
@@ -162,6 +162,30 @@ variables or in the gunicorn configuration file.
...
     - **Default:** `127.0.0.1:4730`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_MINIMUM_LENGTH`**:
+    - **Description:** sets minimum length for user passwords.
+    - **Config file example:** `Dashboard.password_minimum_length`
+    - **Type:** `integer`
+    - **Default:** `8`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_COMMON_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that occur in a list of common passwords.
```
ross-spencer: Oh, just a thought, when we come to write the docs for this it might be a good idea to link to the Django module/listing of common passwords - I feel it has educational value. (Also, I am sure it will come up in future RFP/client queries) _what does this list look like?!_

src/dashboard/install/README.md
```diff
@@ -162,6 +162,30 @@ variables or in the gunicorn configuration file.
...
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_USER_ATTRIBUTE_SIMILARITY_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that are too similar to their username and other user attributes.
+    - **Config file example:** `Dashboard.password_disable_user_attribute_similarity_validation`
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_COMPLEXITY_VALIDATION`**:
```
ross-spencer: Should it be? (It looks like this might just be the documentation and then when it's read in the config it's okay. (If I read it correctly))
```suggestion
- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_PASSWORD_DISABLE_COMPLEXITY_VALIDATION`**:
```

src/dashboard/tests/test_password_validation.py
```diff
@@ -0,0 +1,46 @@
...
+        ("abcdeFGHIJK", True),
+        ("ABÇDEF1234", True),
+        ("àbcdef!$@", True),
+        # Passwords with three types pass validation.
+        ("abcdef123!@#", False),
+        ("ABCdef12345", False),
+        ("#@!abcdefGHIJK", False),
+        # Passwords with all four types pass validation.
+        ("abcDEF123!@#", False),
+        ("#l33t#PASSWORD!", False),
```
ross-spencer, line 33: lol!

src/dashboard/src/settings/components/auth.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+
+# flake8: noqa
```
ross-spencer: I'm not getting a Flake8 error on this file, I wonder if this can be removed? But what prompted the `noqa` flag?

src/dashboard/src/settings/base.py
```diff
@@ -140,6 +160,10 @@
 elasticsearch_timeout = 10
 search_enabled = true
 gearman_server = 127.0.0.1:4730
+password_minimum_length = 8
+password_disable_common_validation = False
+password_disable_user_attribute_similarity_validation = False
+password_disable_complexity_validation = False
```
ross-spencer, line 166: I was wondering about why `password` was the prefix, vs. the action, but this groups them nicely. Nice work.

ross-spencer (approved):
Solid work. One question about the documentation in the `README.md` and a curiosity around `flake8: noqa` - either way, you've got this! Looks like a merge when ready! :+1:

src/dashboard/src/settings/components/auth.py
tw4l: In the fresh light of morning, I'm not sure! I think just a copy-paste from the header of `base.py` or `local.py`? But I think you're right that it's not needed here

PR #1665, opened 2021-01-09
Pass extra parameters for AtoM DO creation

replaceafill:
This sends the AIP name and relative path of the file that are needed
by AtoM's extract file action for metadata-only DIP uploads from
Archivematica.

It also bumps the `agentarchives` version for a [related change](https://github.com/artefactual-labs/agentarchives/pull/57).

Connected to https://github.com/archivematica/Issues/issues/1334


sbreker (approved):
Looks great @replaceafill!

PR #1667, opened 2021-01-14
Improve password validation UX

tw4l:
Connected to https://github.com/archivematica/issues/issues/1332

This pull request consists of a commit with UX improvements identified during internal QA for the password validation feature. The fixes consist of:

- Style form error messages bright red to make them more visible for users.
- Ensure that password mismatch message only displays when accurate for `UserCreationForm` and `UserChangeForm` (i.e. not if error is just that password doesn't match validation criteria).
- Make form behaviour more consistent between forms with respect to where and how error messages appear for invalid passwords and non-matching passwords.

The code in this commit is heavily based on [the `UserCreationForm` and `UserChangeForm` code in the Django source code](https://github.com/django/django/blob/master/django/contrib/auth/forms.py).

Here's what it looks like in practice:

### Invalid password

Screenshot - Add new user - invalid password

### Passwords don't match

Screenshot - Add new user - passwords don't match

### Invalid password AND passwords don't match

Screenshot - Add new user - invalid password and passwords don't match

### Invalid password

Screenshot - Edit user - invalid password

### Passwords don't match

Screenshot - Edit user - passwords don't match

### Invalid password AND passwords don't match

Screenshot - Edit user - invalid password and password's don't match

### DIFF
--- a/src/dashboard/src/components/accounts/forms.py
+++ b/src/dashboard/src/components/accounts/forms.py
@@ -26,6 +26,7 @@ from django.contrib.auth.password_validation import (
     validate_password,
 )
 from django.core.exceptions import ValidationError
+from django.utils.translation import gettext as _
 
 from main.models import UserProfile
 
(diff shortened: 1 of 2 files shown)


ross-spencer (approved):
The change looks nice Tessa. I can see the pattern in the code-base gain of repeating key literals in-line throughout, e.g. `"Administrator"`. I feel like this stuff becomes super apparent during review, but here the additions of `"password", "password_mismatch", "password1" and maybe "password2"` could all be encoded once somewhere and repeated as a const/var inline.    I do believe the benefit to be here in review, but then in future, changes are just made in one place and so easier to see where changes trickle down e.g. during refactor. I don't see any need to change it now. Sorry this took a while to get to today. LGTM :+1:

tw4l:
Thanks @ross-spencer, much appreciated!

PR #1676, opened 2021-01-26
Fix docs for pulling submodules

tw4l:
Setting up my new Archivematica development environment today, I noticed that the included instructions for pulling submodules don't work (at least when cloning over ssh, I didn't try https). My submodule dirs were all uninitialized/empty when the clone completed. This commit reverts the instructions to their previous form, which fixed it for me locally.

Connects to https://github.com/archivematica/Issues/issues/392.

### DIFF
--- a/hack/README.md
+++ b/hack/README.md
@@ -133,8 +133,7 @@ For more information, please consult the Elasticsearch `6.x`
 If you haven't already, create a directory to store this repository using git
 clone, and pull all the submodules:
 
-    $ git clone https://github.com/artefactual/archivematica.git
-    $ git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


sevein (commented):
Thank @tw4l. That must be happening because the default branch doesn't have submodules yet (`stable/1.12.x`). In addition to the change you've made, do you think that adding `--branch qa/1.x` to the clone command would make things clearer?

tw4l:
Thanks @sevein, and good insight! The ssh equivalent of `git clone https://github.com/artefactual/archivematica.git --branch qa/1.x --recurse-submodules` does the trick for me and is nice and compact so I've updated accordingly.

sevein (approved):
Thanks!

PR #1678, opened 2021-01-26
Add audit log middleware to dashboard

tw4l:
This commit adds middleware which adds an `X-Username` header with the name of the authenticated user to each response sent from the dashboard. This provides a mechanism for audit logging of user behaviour via the nginx access logs.

Connected to: https://github.com/archivematica/Issues/issues/1341

### DIFF
--- a/src/dashboard/install/README.md
+++ b/src/dashboard/install/README.md
@@ -242,6 +242,12 @@ variables or in the gunicorn configuration file.
     - **Type:** `boolean`
     - **Default:** `False`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_AUDIT_LOG_MIDDLEWARE`**:
+    - **Description:** enables X-Username header with authenticated HTTP responses.
+    - **Config file example:** `Dashboard.audit_log_middleware`
+    - **Type:** `boolean`
+    - **Default:** `False`
+
 - **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_SITE_URL`**:
     - **Description:** the public address of this service.
     - **Config file example:** `Dashboard.site_url`
--- a/src/dashboard/src/middleware/common.py
+++ b/src/dashboard/src/middleware/common.py
@@ -76,6 +76,19 @@ class ElasticsearchMiddleware(MiddlewareMixin):
             )
 
 
+class AuditLogMiddleware(object):
+    """Add X-Username header with authenticated user to responses."""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if request.user.is_authenticated:
+            response["X-Username"] = request.user.get_username()
+        return response
+
+
 SHIBBOLETH_REMOTE_USER_HEADER = getattr(
     settings, "SHIBBOLETH_REMOTE_USER_HEADER", "REMOTE_USER"
 )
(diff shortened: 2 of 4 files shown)


src/dashboard/install/README.md
```diff
@@ -242,6 +242,12 @@ variables or in the gunicorn configuration file.
     - **Type:** `boolean`
     - **Default:** `False`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_AUDIT_LOG_MIDDLEWARE`**:
+    - **Description:** enables logging of authenticated users with responses.
```
ross-spencer:
```suggestion
    - **Description:** enable X-Username header with authenticated HTTP responses.
```

src/dashboard/src/middleware/common.py
```diff
@@ -76,6 +76,36 @@ def process_exception(self, request, exception):
...
+    logs for auditing purposes.
+
+    To make use of this header in the nginx conf, add the following to
+    `log_format`:
+
+    user=$upstream_http_x_username
+
+    For more information on this middleware's use as part of an
+    Archivematica auditing solution, see the README for auditmatica:
+    https://github.com/artefactual-labs/auditmatica.
```
ross-spencer: Same here as for the storage service. The context is fascinating but not particularly needed in this docstring. We can use the docs proper or the `auditmatica` repo.

src/dashboard/src/middleware/common.py
```diff
@@ -76,6 +76,36 @@ def process_exception(self, request, exception):
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        username = None
+        if request.user.is_authenticated:
+            username = request.user.get_username()
+        response["X-Username"] = username
```
ross-spencer: Similar note here to storage service re: header set to `None`. Do we really need to add the header at all if a user isn't authenticated?

src/dashboard/src/settings/base.py
```diff
@@ -172,6 +177,7 @@
 storage_service_client_quick_timeout = 5
 agentarchives_client_timeout = 300
 prometheus_enabled = False
+audit_log_middleware = False
```
ross-spencer, line 180: Works nicely. I modified it in the Docker environment to give it a whirl.

src/dashboard/tests/test_middleware.py
```diff
@@ -50,3 +51,28 @@ def test_authenticated_user_passes(self):
...
+
+    def test_audit_log_middleware_adds_username(self):
+        with self.modify_settings(
+            MIDDLEWARE={"append": "middleware.common.AuditLogMiddleware"}
+        ):
+            response = self.client.get("/transfer/", follow=True)
+            self.assertTrue(response.has_header("X-Username"))
+            self.assertEqual(response["X-Username"], self.user.username)
+
+    def test_audit_log_middleware_unauthenticated(self):
```
ross-spencer, line 72: Same comment as for the storage service about just setting up expectations for both of these test functions. E.g. around the assertion `None` is what we'd like for logging out. I guess logging can speak for itself but perhaps a docstring doesn't hurt. 

Tests look nice though :+1:

ross-spencer (commented):
Thanks @tw4l largely the same comments as the storage service. Looking good and nice additions.

tw4l:
@ross-spencer also flagging this one as ready for re-review - thanks!

src/dashboard/install/README.md
ross-spencer: Thanks for the change here!

src/dashboard/tests/test_middleware.py
ross-spencer: This helps. Thanks Tessa.

ross-spencer (approved):
LGTM! :+1:

PR #1683, opened 2021-02-03
Specifies a dedicated tab as output in which all tasks are displayed

Slange-Mhath:
This will fix issue archivematica/issues#85 
specifying a dedicated output window as target to display tasks
in one single tab rather than opening one after every click on the
tasks gear.


ross-spencer (approved):
Looks great! Thanks Sebastian. :+1: 

Just waiting on Sarah to hear about next steps and a mile-stone to merge,

PR #1686, opened 2021-02-18
Stream transfer files from local disk if available

tw4l:
This PR changes the `filesystem_ajax` component's `download_by_uuid` view to stream transfer files from the local disk whenever possible, and only request them from the Storage Service if they're not locally available. This resolves an issue where transfer files were not able to be previewed or downloaded when the transfer backlog was on a pipeline local filesystem and the Storage Service was on a remote machine, and lowers the overhead for some file requests by skipping the need to rsync a transfer to the Storage Service in order to stream a single file to the client.

Connected to https://github.com/archivematica/Issues/issues/1333


tw4l:
Jenkins, please build this

sevein (approved):

PR #1695, opened 2021-03-03
Fix SIP arrangement when pkgs share the same name

sevein:
This commit fixes SIP arrangement when two or more SIPs are using the same
name. It ensures that the SIP object that we store in the database is using the
new path that we have assigned after the name clash was identified, i.e.
"../Foobar_1" as opposed to "../Foobar". Not doing this was causing the
MCPServer to create a whole new SIP, cuasing the `MultipleObjectsReturned`
error seen in our logs.

It also introduces a change that shows a warning to the user when an
empty SIP is being arranged since that is not a workflow that we
support.

Connects to https://github.com/archivematica/Issues/issues/1423.

### DIFF
--- a/src/dashboard/frontend/app/arrangement/arrangement.controller.js
+++ b/src/dashboard/frontend/app/arrangement/arrangement.controller.js
@@ -271,9 +271,15 @@ controller('ArrangementController', ['$scope', 'gettextCatalog', '$uibModal', '$
     };
 
     var on_failure = error => {
+      var message = gettextCatalog.getString('SIP could not be started! Check dashboard logs.');
+
+      if (error.hasOwnProperty("data") && error.data.hasOwnProperty("code") && "ERR_NO_FILES" == error.data.code) {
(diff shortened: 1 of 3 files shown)


src/dashboard/frontend/app/arrangement/arrangement.controller.js
```diff
@@ -271,9 +271,17 @@ controller('ArrangementController', ['$scope', 'gettextCatalog', '$uibModal', '$
     };
 
     var on_failure = error => {
+      var message = gettextCatalog.getString('SIP could not be started! Check dashboard logs.');
+
+      if (error.hasOwnProperty("data") && error.data.hasOwnProperty("code")) {
+        if ("ERR_NO_FILES" == error.data.code) {
```
replaceafill: Should these two `if`s be combined?

replaceafill (approved):
LGTM :+1: but a test that exposes the issue would be nice if you have a chance.

replaceafill (approved):
I really appreciate our test coverage being extended :blush:

PR #1700, opened 2021-03-12
Delete set_agent_code management command

sevein:
set_agent_code is an undocumented management command that we have used
in the past when running forks of Archivematica where we wanted to use
a different preservation system identifier.

In Archivematica 1.13, the preservation system identifier is not going
to be maintained in the database but generated by the `version` module.
Forks can tweak that module as needed.

Connects to https://github.com/archivematica/Issues/issues/383.


src/dashboard/src/installer/steps.py
```diff
@@ -64,10 +58,12 @@ def setup_pipeline(org_name, org_identifier, site_url):
...
 
     # Update Archivematica version in DB
-    set_agent_code("Archivematica-" + get_version(), pk=1)
+    Agent.objects.default_system_agent().update(
+        identifiervalue=get_preservation_system_identifier()
+    )
 
     if org_name != "" or org_identifier != "":
-        agent = get_agent()
+        agent = Agent.objects.default_organization_agent()
```
replaceafill, line 66: Could we also use `default_organization_agent` here? https://github.com/artefactual/archivematica/blob/c37d5af3edcf757670c4bd03f3c9a29b8e2cc73c/src/dashboard/src/components/administration/views.py#L479

replaceafill (approved):
LGTM, and based on the name of your branch I think you want https://github.com/archivematica/Issues/issues/383 as the connected issue, right?

PR #1706, opened 2021-03-16
Split AMAUAT ipc scenarios

replaceafill:
Currently, the `ipc` AMAUAT tag takes around 50 minutes to finish. This kind of abuses `behave`'s tag expressions to split the run into `access` and `preservation` scenarios which shortens the runs to about 30 minutes.

A more appropriate solution could be to modify the `test-at-behave` rule in the `Makefile` to allow such expressions instead of its `--tags=$(TAGS)` variable.

https://github.com/archivematica/Issues/issues/1115


replaceafill:
Closing this for now as I think we need to address those Firefox issues first. Also this feels a bit like deviating from the development workflow where you usually run the whole tag at once.

PR #1708, opened 2021-03-18
Change wording of filename changes

sevein:
This commit renames functions, modules, variable names, etc., according
to our convention where terms like sanitization or clean-up are considered
non-inclusive and inaccurate.

Changes in the UI happened in a previous release, see https://github.com/artefactual/archivematica/pull/1629.

These are changes in the artifacts we are producing that will impact users:
- `logs/filenameCleanup.log` is now `logs/filenameChanges.log`, and
- Event type now recorded as "filename change" instead of "name cleanup", and
- Event detail note now uses the term "new name" instead of "cleaned up name".

Connects to https://github.com/archivematica/Issues/issues/1079.


replaceafill (approved):
LGTM!

PR #1709, opened 2021-03-18
Add caller UID and GID to Django manage rules

replaceafill:
This allows these `Makefile` rules to write to the host filesystem,
for example when migration files or static assets need to be created.

Connected to https://github.com/archivematica/Issues/issues/31


sevein (approved):
Much needed, thanks!

PR #1701, opened 2021-03-12
Isolate test_sanitize tests

sevein:
TestSanitize class, part of the test_sanitize module, was leaving database
records behind after running. Inadvertently, we wrote other tests depending on
these records. This commit updates the way that fixtures are loaded to
ensure proper isolation.

Connects to https://github.com/archivematica/Issues/issues/383.

PR #1704, opened 2021-03-16
Delete preservation system agent from database

sevein:
This pull request introduces a new `Agent.objects.extend_queryset_with_preservation_system` method used to extend agent look-up iterables when we want to include the preservation system agent. This allows us to stop maintaining the preservation system agent as an actual database object. Most of the work was in updating our tests which had their own logic to inject the old database-based agent.

Connects to https://github.com/archivematica/Issues/issues/383.

PR #1710, opened 2021-03-21
Add pytest-randomly plugin

replaceafill:
I noticed [this issue](https://github.com/artefactual/archivematica/pull/1701#discussion_r598306160) with a MCPClient test being affected by other tests. I've also seen [tests "leaking" data in the Storage Service](https://github.com/archivematica/Issues/issues/1048#issuecomment-572785442) before.

So I think we should probably randomize the test runs to catch this kind of problem using the [`pytest-randomly`](https://github.com/pytest-dev/pytest-randomly) plugin. It needs to be pinned to version `1.2.3` which is the last one with Python 2 support.

It randomizes by default with no configuration or command line changes, but you could disable it in the development environment with the `PYTEST_ADDOPTS` environment variable like this:

```
env PYTEST_ADDOPTS="-p no:randomly" make test-mcp-client
```

In case we catch a test that fails because of the order in which it's executed, the `pytest` session output shows the `randomly-seed` that was used to generated such order:

```
$ make test-mcp-client 
...
mcpclient run-test: commands[0] | py.test
=============================================================================== test session starts ================================================================================
platform linux2 -- Python 2.7.17, pytest-4.6.11, py-1.10.0, pluggy-0.13.1
Using --randomly-seed=1616348307
django: settings: settings.testmysql (from env)
rootdir: /src, inifile: pytest.ini
plugins: randomly-1.2.3, pythonpath-0.7.3, mock-1.13.0, cov-2.4.0, django-3.10.0
collected 273 items
```

so we can reproduce the same order by passing that seed with the `PYTEST_ADDOPTS` environment variable:

```
$ env PYTEST_ADDOPTS='--randomly-seed=1616348307' make test-mcp-client
```

That session output is shown in the GitHub action output under the `Run make rule` step.

I've also temporarily marked the `test_sanitize_transfer_with_multiple_files` test until we investigate and fix it.

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -8,5 +8,6 @@ pytest-cov==2.4.0
 pytest-django>=3,<4
 pytest-mock>=1,<2
 pytest-pythonpath
+pytest-randomly==1.2.3
 tox
 vcrpy>=1,<2
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -83,7 +83,8 @@ pytest-cov==2.4.0         # via -r requirements-dev.in
 pytest-django==3.10.0     # via -r requirements-dev.in
 pytest-mock==1.13.0       # via -r requirements-dev.in
 pytest-pythonpath==0.7.3  # via -r requirements-dev.in
-pytest==4.6.11            # via -r requirements-dev.in, pytest-cov, pytest-django, pytest-mock, pytest-pythonpath
+pytest-randomly==1.2.3    # via -r requirements-dev.in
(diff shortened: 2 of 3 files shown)


sevein (approved):
Thank you!

FYI, I think https://github.com/artefactual/archivematica/pull/1704 fixes the underlying issue.

replaceafill:
Confirmed! The `test_sanitize_transfer_with_multiple_files` works again on multiple runs reusing the database. And a new order problem I found yesterday with `--randomly-seed=1616432139` when `test_verify_checksum.py` runs first has also been fixed :pray:

PR #1713, opened 2021-03-23
Override processing config labels

sevein:
In c37d5af, we started pulling processing config field labels
straight from workflow labels. In some cases, these labels need to be
customized as the original ones don't bring enough context.

This pull request also updates test_pid_components to stop using loaddata. This
is something we've been doing lately in MCPClient tests to ensure that tets are
not leaving data behind.

Connects to https://github.com/archivematica/Issues/issues/869.


replaceafill (approved):
Thanks! The processing config labels look like the old ones and `--randomly-seed=1616511316` passes now :+1:

PR #1716, opened 2021-03-24
Add make rules for running the tests in Python 3.6

replaceafill:
This updates the `tox` configuration and adds four new `Makefile` rules to the development environment for running the component tests in Python 3.6:

* `make test-mcp-server-py36`
* `make test-mcp-client-py36`
* `make test-dashboard-py36`
* `make test-archivematica-common-py36`

Neither of these pass yet in Python 3.6 as you can see in [this CI run](https://github.com/artefactual/archivematica/actions/runs/683667061) so they're not automatically run yet in the development environment or in CI and they shouldn't break all the previous Python 2.7 rules or the existing Docker services.

I also updated a relative path that caused trouble in this new setup in the "no mutable globals" test.

Connected to https://github.com/archivematica/Issues/issues/878

### DIFF
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -219,17 +219,26 @@ test-build:  ## Build archivematica-tests image.
 		--build-arg TARGET=archivematica-tests \
 			../
 
-__TOXENVS_MCPSERVER := mcpserver
+__TOXENVS_MCPSERVER := py27-mcpserver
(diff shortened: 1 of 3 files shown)


tw4l (approved):
Working great and I like the approach. Nice work! :smile:

PR #1721, opened 2021-04-01
Add logger for AJAX middleware #1353

Slange-Mhath:
This commit will introduce a logger which writes errors
into the log file and thus fixes the issue [1353](https://github.com/archivematica/Issues/issues/1353)

### DIFF
--- a/src/dashboard/src/middleware/common.py
+++ b/src/dashboard/src/middleware/common.py
@@ -19,6 +19,7 @@ from __future__ import absolute_import
 
 import sys
 import traceback
+import logging
 
 from django.conf import settings
 from django.http import HttpResponseServerError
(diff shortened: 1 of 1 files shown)


src/dashboard/src/middleware/common.py
```diff
@@ -27,18 +28,29 @@
 from django.utils.deprecation import MiddlewareMixin
 from shibboleth.middleware import ShibbolethRemoteUserMiddleware
 
+
```
sevein: We don't need this extra line.

src/dashboard/src/middleware/common.py
```diff
@@ -27,18 +28,29 @@
...
 import elasticsearch
 
 
+logger = logging.getLogger(__name__)
```
sevein: Please use this form:

    logger = logging.getLogger("archivematica.dashboard")

Otherwise it'll be given a name we don't want to use.

sevein (changes requested):
Thank you, just a couple of minor requests.

sevein (approved):
Thank you!

sevein (approved):

PR #1724, opened 2021-04-07
Workflow: Include policy validation events in transfer METS

tw4l:
This PR moves the "Generate METS.xml document" job later in the transfer workflow, which allows `validation` events from "Policy checks for originals" to be written into the transfer METS file.

Connected to https://github.com/archivematica/issues/issues/781

 For comparison:

**current qa/1.x**

qa-1
qa-2

**new workflow**

new-1
new-2

I have done manual testing with a number of transfer types and it appears to be working correctly, with the exception of a few known bugs:
* https://github.com/archivematica/Issues/issues/1438 (Dataverse transfers skip "Policy checks for originals" and "Examine contents")
* https://github.com/archivematica/Issues/issues/1007#issuecomment-567553434 (DSpace transfers hang on "Policy checks for originals")

Otherwise when "Policy checks for originals" is set to Yes, there is a matching FPR rule, and a transfer is stored in the backlog, the transfer METS file downloaded from the backlog contains validation events from "Policy checks for originals". As far as I can tell, DSpace transfers are continuing to work correctly following the changes as long as "Policy checks for originals" is not attempted.

### DIFF
--- a/src/MCPServer/lib/assets/workflow.json
+++ b/src/MCPServer/lib/assets/workflow.json
@@ -2496,7 +2496,7 @@
       "config": {
         "@manager": "linkTaskManagerUnitVariableLinkPull",
         "@model": "TaskConfigUnitVariableLinkPull",
-        "chain_id": "675acd22-828d-4949-adc7-1888240f5e3d",
+        "chain_id": "307edcde-ad10-401c-92c4-652917c993ed",
         "variable": "postExtractSpecializedProcessing"
       },
       "description": {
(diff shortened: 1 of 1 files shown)


tw4l:
The "Test ipc on Firefox" AMAUAT is failing but I think it's unrelated - I'm getting the same failure on `qa/1.x`.

tw4l:
Confirmed that the AMAUAT issue is not specific to this branch. See CI results in this test PR, which is based on qa/1.x: https://github.com/artefactual/archivematica/pull/1725

This seems to be related to the following issue under review: https://github.com/archivematica/Issues/issues/1115

sevein (approved):
Excellent!

PR #1726, opened 2021-04-14
Fix processing configuration reset for virus scanning

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1439

In addition to fixing the immediate issue and adding tests, this PR also reduces some duplication of processing configuration-related constants and functions between `archivematicaCommon` and` MCPServer`, and makes some minor UX tweaks related to virus scanning (namely, changing the label of the virus scanning option to be more consistent with other boolean choices and increasing its visibility by raising it to the top of the edit processing configuration screen). Following these commits, the edit processing configuration screen looks like this:

image


src/dashboard/src/components/administration/tests/test_processing_config.py
```diff
@@ -93,3 +95,21 @@ def test_name_field_is_validated(self, mock_conf_fields):
...
+            assert actual_file.read() == DEFAULT_PROCESSING_CONFIG
+
+    def test_reset_automated_processing_config(self):
+        response = self.client.get("/administration/processing/reset/automated/")
+        self.assertEqual(response.status_code, 302)
+        processing_config = os.path.join(
+            helpers.processing_config_path(), "automatedProcessingMCP.xml"
+        )
+        with open(processing_config) as actual_file:
+            assert actual_file.read() == AUTOMATED_PROCESSING_CONFIG
```
sevein, line 115: I'm realizing these tests will likely fail unless the environment has created the shared directory beforehand. Maybe this is true in many other places, I haven't checked. Not entirely sure what would be a good way to get around that, maybe patching django_settings.SHARED_DIRECTORY with the path of a temporary directory created by this test?

sevein (approved):
Nicely done! :+1:

src/dashboard/src/components/administration/tests/test_processing_config.py
sevein: FYI I'm fixing this from a different pull request. It passed with `--randomly-seed=1618426567` because test_api is running first.

PR #1727, opened 2021-04-14
Allow tag combinations in test-at-behave rule

replaceafill:
`behave` allows to [control the test runs with tags](https://behave.readthedocs.io/en/stable/gherkin.html#controlling-your-test-run-with-tags) using `and` combinations like:

```
--tags=wip --tags=slow
    This will select all the cases tagged both “wip” and “slow”.
```

This modifies our `test-at-behave` rule to split the `TAGS` variable by space providing similar functionality:

```
make test-at-behave TAGS="ipc preservation"
```

which becomes:

```
--tags=ipc --tags=preservation
```

The comma separated form of the `TAGS` variable (e.g. `make test-at-behave TAGS=aip-encrypt,aip-encrypt-mirror`) is still supported.

I've also modified the `acceptance-tests` CI workflow to:

* use this new rule functionality to split the run time of the `ipc` tag into shorter `preservation` and `access` scenario based runs. This is much cleaner than [my previous attempt](https://github.com/artefactual/archivematica/pull/1706)
* run the `ipc` based tags only in Chrome until we are able to determine what makes the tag fail in GitHub
* run the `black-box` tag only once since it doesn't rely on UI specific functionality

Connected to https://github.com/archivematica/Issues/issues/1115

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -15,7 +15,8 @@ jobs:
           - "aip-encrypt-mirror"
           - "black-box"
           - "icc"
-          - "ipc"
+          - "ipc preservation"
+          - "ipc access"
           - "mo-aip-reingest"
           - "tcc"
           - "tpc"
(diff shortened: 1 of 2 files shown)


sevein (approved):
:+1:

PR #1729, opened 2021-04-22
Fix archivematicaCommon tests in Python 3

replaceafill:
This also adds the `archivematicaCommon` Python 3 tests to the `test` CI workflow and updates the `Upload coverage report` step to include the Python version as a flag for codecov.io.

Connected to https://github.com/archivematica/Issues/issues/809


sevein (approved):
:heart_eyes:

PR #1730, opened 2021-04-22
Fix dashboard and MCPServer tests in Python 3

replaceafill:
Connected to https://github.com/archivematica/Issues/issues/807

### DIFF
--- a/src/MCPServer/tests/test_translation.py
+++ b/src/MCPServer/tests/test_translation.py
@@ -2,13 +2,19 @@
 
 from __future__ import unicode_literals
 
+import six
+
 from server.translation import UNKNOWN_TRANSLATION_LABEL, TranslationLabel
 
 
 def test_translation_label(mocker):
     mocker.patch("server.translation.FALLBACK_LANG", "en")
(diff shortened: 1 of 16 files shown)


src/MCPServer/tests/test_translation.py
```diff
@@ -2,13 +2,19 @@
...
 
 def test_translation_label(mocker):
     mocker.patch("server.translation.FALLBACK_LANG", "en")
     tr = TranslationLabel({"en": "cat", "es": "gato"})
-    assert repr(tr) == "TranslationLabel({u'en': u'cat', u'es': u'gato'})"
+    # TODO: revise this
+    if six.PY2:
+        assert repr(tr) == "TranslationLabel({u'en': u'cat', u'es': u'gato'})"
+    else:
+        assert repr(tr) == "TranslationLabel({'en': 'cat', 'es': 'gato'})"
```
sevein: I'd just get rid of the assertion, we don't care what's the value returned. I was just probably trying to be as much explicit as possible.

sevein (approved):
:+1:

PR #1731, opened 2021-04-27
Fix MCPClient tests in Python 3

replaceafill:
Connected to https://github.com/archivematica/Issues/issues/808

### DIFF
--- a/src/MCPClient/lib/archivematicaClient.py
+++ b/src/MCPClient/lib/archivematicaClient.py
@@ -53,8 +53,8 @@ task to run next).
 # @subpackage archivematicaClient
 # @author Joseph Perry <joseph@artefactual.com>
 
-import ConfigParser
-import cPickle
+import six.moves.configparser
+import six.moves.cPickle
 from functools import partial
 import logging
 import os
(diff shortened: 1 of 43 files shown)


sevein (approved):
:partying_face:

src/MCPClient/lib/archivematicaClient.py
```diff
@@ -107,19 +108,16 @@ def get_supported_modules(file_):
 @auto_close_db
 def handle_batch_task(gearman_job, supported_modules):
     module_name = supported_modules.get(gearman_job.task)
-    gearman_data = cPickle.loads(gearman_job.data)
+    gearman_data = six.moves.cPickle.loads(gearman_job.data)
```
sevein, line 111: Didn't we have to deal with a pickle version protocol change between Py2 and 3?

replaceafill: I set `protocol=0` in all the pickle `dump` calls in https://github.com/artefactual/archivematica/pull/1731/commits/ac1152745c460da10617a9de996218ad9b6c45fd

PR #1732, opened 2021-04-29
create_mets_v2: populate DMDIDs in normative structMap

sevein:
In the context of https://github.com/archivematica/Issues/issues/768, I've written some more tests focusing on create_mets_v2. The new tests could eventually replace those in test_create_aip_mets.py. I'm also introducing a main function in the client module to make testing more accessible. In the same commit, I've started using `html.escape` in Python 3 replacing `cgi.escape` which is removed from Python 3.8 and stopped accepting bytes.

The main change is to populate existing DMDID attributes into the normative structMap.

Connects to https://github.com/archivematica/Issues/issues/768.
Connects to https://github.com/archivematica/Issues/issues/808.


src/MCPClient/tests/test_create_mets_v2.py
```diff
@@ -35,3 +48,268 @@ def test_createDMDIDsFromCSVMetadata_finds_non_ascii_paths(mocker):
...
+    empty_dir_path.mkdir()
+
+    return empty_dir_path
+
+
+@pytest.fixture()
+def metadata_csv(sip, sip_path, objects_path):
+    (objects_path / "metadata").mkdir()
+    metadata_csv = objects_path / "metadata" / "metadata.csv"
+    metadata_csv.write_text("Filename,dc.title\n" "objects/file1,File 1")
```
replaceafill:
```suggestion
    metadata_csv.write_text("Filename,dc.title\nobjects/file1,File 1")
```

src/MCPClient/tests/test_create_mets_v2.py
```diff
@@ -35,3 +48,260 @@ def test_createDMDIDsFromCSVMetadata_finds_non_ascii_paths(mocker):
...
+    assert len(objects_div) == 2
+    assert len(dublincore) == 3
+    assert dublincore[0].tag == "{http://purl.org/dc/elements/1.1/}title"
+    assert dublincore[0].text == "Hello World Contents"
+    assert dublincore[1].tag == "{http://purl.org/dc/elements/1.1/}identifier"
+    assert dublincore[1].text == "12345"
+    assert dublincore[2].tag == "{http://purl.org/dc/terms/}isPartOf"
+    assert dublincore[2].text == "23456"
+
+    print(etree.tostring(mets_xml))
```
replaceafill: Is this `print` call needed?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -979,9 +980,7 @@ def createFileSec(
     :param directoryPath: Path to recursively traverse and create METS entries for
     :param parentDiv: structMap div to attach created children to
     :param baseDirectoryPath: SIP path
-    :param baseDirectoryName: Name of the %var% for the SIP path
     :param fileGroupIdentifier: SIP UUID
```
replaceafill: Do you think `fileGroupIdentifier` could be renamed to `sip_uuid`?

replaceafill (approved):
LGTM and thanks for taking care of the `cgi` module deprecation :+1:

replaceafill (approved):
Did a bit of testing and it looks good! It's nice that it works with multiple DMDIDs too like `<mets:div LABEL="artwork" TYPE="Directory" DMDID="dmdSec_5 dmdSec_6">`

PR #1733, opened 2021-05-03
Add rights CSV validation, make act not required in CSV workflow

bonniegee:
Does the following:
* Adds a rights csv endpoint to the validation API (connected to archivematica/issues#563)
* Modifies the rights csv parser so that an act is not required (fixes archivematica/issues#1445)
* Adds or updates associated tests

For the rights csv validation, there were some decisions I made that I wanted to note:
* There were some differences in the language used (e.g., in error messages and variable names) between the Avalon validator and the rights csv parser; for the most part, I used the existing rights csv language
* Though the rights csv parser does not require "basis," I required it in the validator
* Though the rights csv parser does not check that the csv adheres to the PREMIS data dictionary (e.g,. that a copyright rights statement has a copyright status), I checked for this in the validator

### DIFF
--- a/src/dashboard/src/components/api/validators.py
+++ b/src/dashboard/src/components/api/validators.py
@@ -293,104 +293,100 @@ class RightsValidator(BaseValidator):
             )
         return True
 
-        @staticmethod
-        def _check_restriction(row, columns):
-            """If any grant fields exist, validate grant information.
+    @staticmethod
+    def _check_restriction(row, columns):
+        """If any grant fields exist, validate grant information.
 
-            :param row: list: metadata fields
-            :param columns: dict: column names and indexes
-            """
-            grant_fields = [
-                "grant_act",
-                "grant_restriction",
-                "grant_start_date",
-                "grant_end_date",
-                "grant_note",
-            ]
-            if [g for g in grant_fields if row[columns.get(g)]]:
-                if not row[columns.get("grant_act")]:
-                    raise ValidationError("No act specified.")
-                elif row[columns.get("grant_act")].lower() not in [
-                    "disallow",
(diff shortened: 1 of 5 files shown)


sevein (commented):
:exploding_head: this is looking great, @bonniegee. Tests are passing too. I'll take a closer look later today.

Do you have time to update our API docs? There is a [validation section](https://github.com/artefactual/archivematica-docs/blob/1.13/dev-manual/api/api-reference-archivematica.rst#validate) that could use some rewording.

bonniegee:
Thanks so much, @sevein! I can definitely update the API docs this week.

src/dashboard/src/components/api/validators.py
```diff
@@ -244,7 +244,181 @@ def validate(self, string):
...
+        csvr = csv.reader(csvfile)
+        for i, row in enumerate(csvr):
+            if i == 0:
+                self._check_column_data(row)
+                columns = {k: v for v, k in enumerate(row)}
+            if i >= 1:
+                basis = self._check_basis(row, columns)
+                if basis.lower() in ["copyright", "statute"]:
+                    getattr(self, "_check_{}".format(basis.lower()))(row, columns)
+                self._check_documentation
```
sevein: I've just realized this is missing the actual function call.

sevein (changes requested):
Thank you! Just one minor issue before merging. :+1:

src/dashboard/src/components/api/validators.py
bonniegee: Thanks for the catch--just fixed!

sevein (approved):

PR #1734, opened 2021-05-10
Add new config attribute "workflow_file"

Slange-Mhath:
This PR will introduce the new
attribute workflow_file, which will allow users
to specify their own customized workflow file as
ARCHIVEMATICA_MCPSERVER_WORKFLOW_FILE in the
docker-compose.yml.
Default behavior stays the same if given an
empty String.
This will fix the issue [1441](https://github.com/archivematica/Issues/issues/1441).

### DIFF
--- a/src/MCPServer/install/README.md
+++ b/src/MCPServer/install/README.md
@@ -191,6 +191,12 @@ This is the full list of variables supported by MCPServer:
     - **Type:** `int`
     - **Default:** 4
 
+- **`ARCHIVEMATICA_MCPSERVER_WORKFLOW_FILE`**:
+    - **Description:** the path to the user-customised workflow file. If this is empty, Archivematica will load the default workflow. This configuration attribute is not recommended for general use, but only for users with development skills, an understanding for the Archivematica workflow engine, and are willing to maintain their own workflow file.
+    - **Config file example:** `MCPServer.workflow_file`
+    - **Type:** `string`
+    - **Default:** `""`
+
 - **`ARCHIVEMATICA_MCPSERVER_MCPSERVER_WORKER_THREADS`**:
     - **Description:** the number of threads used to handle MCPServer jobs.
     - **Config file example:** `MCPServer.worker_threads`
--- a/src/MCPServer/lib/server/workflow.py
+++ b/src/MCPServer/lib/server/workflow.py
@@ -27,6 +27,7 @@ from jsonschema.exceptions import ValidationError
 
 from server.jobs import Job
 from server.translation import FALLBACK_LANG, TranslationLabel
+from django.conf import settings as django_settings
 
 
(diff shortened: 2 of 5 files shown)


hack/docker-compose.yml
```diff
@@ -108,6 +108,7 @@ services:
       ARCHIVEMATICA_MCPSERVER_SEARCH_ENABLED: "${AM_SEARCH_ENABLED:-true}"
       ARCHIVEMATICA_MCPSERVER_MCPSERVER_PROMETHEUS_BIND_PORT: "7999"
       ARCHIVEMATICA_MCPSERVER_MCPSERVER_PROMETHEUS_BIND_ADDRESS: "0.0.0.0"
+      ARCHIVEMATICA_MCPSERVER_WORKFLOW_FILE: ""
```
sevein: I don't think we need this line. We don't list environment strings here unless necessary.

src/MCPServer/install/README.md
```diff
@@ -190,6 +190,12 @@ This is the full list of variables supported by MCPServer:
     - **Config file example:** `MCPServer.rpc_threads`
     - **Type:** `int`
     - **Default:** 4
+  
+- **`ARCHIVEMATICA_MCPSERVER_WORKFLOW_FILE`**:
+    - **Description:** the path to the user-customised workflow file.
```
sevein: Can you clarify what's the expected behaviour when empty? E.g. `If empty, Archivematica will load the default workflow.`

src/MCPServer/lib/server/workflow.py
```diff
@@ -35,6 +36,7 @@
 )
 
 DEFAULT_WORKFLOW = os.path.join(ASSETS_DIR, "workflow.json")
+WORKFLOW_FILE = os.path.join(ASSETS_DIR, "")
```
sevein: I don't think you need this constant.

sevein (changes requested):
Thanks, this is great. Just a few minor suggestions before merging.

src/MCPServer/install/README.md
```diff
@@ -190,6 +190,12 @@ This is the full list of variables supported by MCPServer:
     - **Config file example:** `MCPServer.rpc_threads`
     - **Type:** `int`
     - **Default:** 4
+  
+- **`ARCHIVEMATICA_MCPSERVER_WORKFLOW_FILE`**:
+    - **Description:** the path to the user-customised workflow file. If this is empty, Archivematica will load the default workflow.
```
sevein: I almost forget! We should make very clear that this configuration attribute is not recommended for general use, but only for users with development skills that understand the Archivematica workflow engine and are willing to maintain their own workflow document.

sevein (changes requested):
Sorry, we need one more change! :pray:

sevein (approved):
Thank you! :balloon:

PR #1735, opened 2021-05-17
Create package status field

sevein:
This pull request adds a status attribute to the Transfer and SIP models that MCPServer updates in order to reflect processing status in the database at the package level. The goal is to simplify the identification of packages that completed processing so they can be purged - but this is being addressed in the next pull request.

MCPServer can't recover processing state after crashes or restarts so the same strategy remains: packages in unknown or in-progress will be marked as failed. Eventually, some packages may resume processing if their contents are identified within one of the existing watched directories, forcing MCPServer to update the status attribute back to in progress.

Connects to https://github.com/archivematica/Issues/issues/1239.


src/MCPServer/tests/test_package.py
```diff
@@ -376,3 +377,42 @@ def test_move_file(self, tmp_path, processing_dir, transfer):
...
+            DIP,
+            models.SIP,
+        ),
+    ],
+)
+@pytest.mark.django_db(transaction=True)
+def test_package_statuses(package_class, model_class):
+    package_id = uuid.uuid4()
+    model_class.objects.create(pk=package_id)
+    package = package_class("/tmp", package_id)
```
replaceafill: I don't think the package classes do anything on their own with the `path` currently, but maybe use the `tmp_path` fixture just in case?

replaceafill (approved):
This looks great to me. I understand that:

> MCPServer can't recover processing state after crashes or restarts so the same strategy remains: packages in unknown or in-progress will be marked as failed. Eventually, some packages may resume processing if their contents are identified within one of the existing watched directories, forcing MCPServer to update the status attribute back to in progress.

means we don't need to add data migrations for existing transfers and sips, correct?

src/MCPServer/tests/test_package.py
sevein: Fixed! Thanks.

sevein:
> [...] means we don't need to add data migrations for existing transfers and sips, correct?

Exactly!

PR #1737, opened 2021-05-20
Add bytes column to Archival Storage CSV

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1450

This is one potential solution to the linked issue. Another might be to use a consistent unit of measure in the Archival Storage web UI and CSV, e.g. to always use MB.


src/dashboard/src/components/archival_storage/tests/test_archival_storage.py
```diff
@@ -248,6 +249,7 @@ def test_search_as_csv(client, mocker, django_user_model, username, password, tm
             "file_count": 2,
             "location": "/var/archivematica/AIPStore",
             "isPartOf": None,
+            "bytes": "200100",
```
sevein: Shouldn't this be a `number`?

sevein (approved):
👍

PR #1738, opened 2021-05-21
Optimize download speeds

sevein:
This commit uses iter_content (requests.Response) to make Storage
Service downloads faster and more efficient.

Suggested by Marcus Elizondo (@elizondom).

Connects to https://github.com/archivematica/Issues/issues/1448.

### DIFF
--- a/src/dashboard/src/components/helpers.py
+++ b/src/dashboard/src/components/helpers.py
@@ -310,10 +310,21 @@ def processing_config_path():
 def _prepare_stream_response(
     payload, content_type, content_disposition, preview_file=False
 ):
-    """Prepare the streaming response to return to the caller."""
+    """Prepare the streaming response to return to the caller.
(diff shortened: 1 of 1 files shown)


src/dashboard/src/components/helpers.py
```diff
@@ -310,10 +310,21 @@ def processing_config_path():
 def _prepare_stream_response(
     payload, content_type, content_disposition, preview_file=False
 ):
-    """Prepare the streaming response to return to the caller."""
+    """Prepare the streaming response to return to the caller.
+
+    It iterates over the response data (payload) to avoid reading the conten at
```
replaceafill:
```suggestion
    It iterates over the response data (payload) to avoid reading the content at
```

replaceafill (approved):
Thanks!

PR #1739, opened 2021-05-21
Include pending DIPs in Ingest red bubble count

sevein:
Connects to https://github.com/archivematica/Issues/issues/1366.


replaceafill (approved):
Nice!

PR #1740, opened 2021-05-25
Make the code Python 3 compatible

replaceafill:
This updates the MCPServer, MCPClient and dashboard to run in Python
2.7 and Python 3.6:

* All the shebangs of the Python scripts have been updated to
  `#!/usr/bin/env python` so they can run in both versions

* All the gearman tasks are now registered with a binary name and
  several encoding issues raised by the AMAUATs have been fixed

* The `clientScripts` directory has been added to the MCPClient
  `PYTHONPATH` and all the `sys.path` modifications in its tests have
  been removed

* A Django migration script has been added to make all the FPR
  `pythonScript`s compatible with both versions

* A `PYTHON_VERSION` argument has been added to the docker development
  environment which allows the components to be run in both versions
  with Python 3.6 being the default one

* The `ensure-no-mutable-globals.py` and `checkformigrations.sh`
  scripts are now also run in Python 3.6 in the GitHub test CI
  workflow

* The `archivematica-acceptance-tests` submodule has been updated to
  include a fix for the `@transfer-backlog` tag

* The GitHub AMAUAT CI workflow runs in both versions at the same time

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   test:
     if: github.event.label.name == 'AMAUAT'
-    name: "Test ${{ matrix.tag }} on ${{ matrix.browser }}"
+    name: "Test ${{ matrix.tag }} ${{ matrix.python_version }} on ${{ matrix.browser }}"
     runs-on: "ubuntu-18.04"
     strategy:
       matrix:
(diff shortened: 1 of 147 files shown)


sevein (approved):
Big day! Thank you so much. Lots of smart decisions here. Running it locally, it seems to work great too. It feels kind of magical to recreate my containers and boom! I'll update Ansible and packages with the new path consequently.

PR #1741, opened 2021-05-25
Disallow hiding active packages

sevein:
This commit makes the "active" attribute of a package available to the
front-end so it is posible to reject user requests attempting to hide
them.

A window.alert dialog is used to reject the user action, but this can
be later improved, e.g. hiding the button which needs some refactoring
in the rendering layer.

A similar check is added to the back-end endpoint. The `mark_hidden`
view now rejects hiding active packages returning a 409 conflict code.

It fixes https://github.com/archivematica/Issues/issues/1446.


replaceafill (approved):
It works great and thank you for extending the test coverage there!

PR #1742, opened 2021-05-25
Convert escape_string to text

sevein:
MySQLdb.escape_string always returns a sequence of bytes. This commit
converts it to text so we can concatenate in Python 3.

Connects to https://github.com/archivematica/Issues/issues/809.


replaceafill (approved):
So it begins! Thank you! :pray:

PR #1748, opened 2021-05-27
Update assign_file_uuids.py and update_size_and_checksum.py client scripts to parse METS once

tw4l:
Connected to https://github.com/archivematica/issues/issues/1429

Previously, the `assign_file_uuids.py` and `update_size_and_checksum.py` client scripts ran once per file. This meant that during full reingest, the METS file was parsed and read into memory again for each file. This commit modifies the scripts to run on a per-directory basis, parsing the METS file only once. This results in a significant speed-up for these jobs during full reingest.

I've put together some quick rough benchmarks from conducting full reingests of a scientific dataset consisting of 278 files ([source](https://spectrum.library.concordia.ca/986807/)) using the "automated" processing configuration in my local Docker dev environment. The benchmarks measure the time elapsed in the "Assign file UUIDs to objects" and "Assign checksums and file sizes to objects" Jobs in the Transfer microservice "Microservice: Assign file UUIDs and checksums":

This branch:
- Assign file UUIDs to objects: 3 seconds
- Assign checksums and file sizes to objects: 3 seconds

`qa/1.x`:
- Assign file UUIDs to objects: 2 minutes, 27 seconds
- Assign checksums and file sizes to objects: 2 minutes, 11 seconds

That computes to a speed-up of ~43-49x the previous performance for these scripts in the tested sample data set. The difference is even larger as the number of files in a SIP grows. With a sample dataset consisting of 10 copies of the dataset above, resulting in a transfer of 2780 small files (the total compressed AIP is still only ~11 MB), "Assign file UUIDs to objects" took 1 minute, 5 seconds on this branch, compared to 14 hours, 48 minutes in `qa/1.x`, which is a speed-up of ~819x.

This is with 1 MCPClient instance - it's possible that the `qa/1.x` baselines would be faster with more instances deployed but that also increases the chances for out-of-memory errors with large METS files.

There may be other `MCPClient` client scripts that could similarly benefit from switching to a per-directory script and parsing the METS file once rather than on a per-file basis - I haven't done a complete analysis of the other client scripts.


src/archivematicaCommon/tests/test_archivematica_functions.py
```diff
@@ -3,10 +3,52 @@
...
+    except EEXIST:
+        if path.dirpath().isdir():
+            pass
+        else:
+            raise
+
+    path.write_text(u"Test file created by %s" % __file__, encoding="utf-8")
+
+
+def test_find_mets_file_match(tmpdir):
```
sevein: If you have some time, it'd be nice to use `tmp_path` instead. It's the pytest fixture that returns a `pathlib.Path` and has a `touch` method.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -266,6 +271,26 @@ def find_metadata_files(sip_path, filename, only_transfers=False):
...
+    mets_paths = glob.glob(os.path.join(src, "METS.*.xml"))
+
+    if len(mets_paths) == 1:
+        return mets_paths[0]
+    elif len(mets_paths) == 0:
+        raise METSDiscoveryError("No METS file found in %s" % src)
+    else:
+        raise METSDiscoveryError(
+            "Multiple METS files found in %s: %r" % (src, mets_paths)
+        )
```
sevein, line 289: Do you think we could leverage built-in requests like IOError?
I'm not sure it's a good idea but worth considering. E.g.

```python
raise OSError(errno.EEXIST, "Multiple METS files found ...")
```

I think that'd be pretty close to the built-in IOError subclasses FileExistsError (EEXIST) and FileNotFoundError (ENOENT) that we see in Python 3.

src/MCPClient/lib/clientScripts/update_size_and_checksum.py
```diff
@@ -111,39 +93,105 @@ def get_file_info_from_mets(job, shared_path, file_):
...
+    else:
+        files = get_sip_file_queryset(sip_uuid, filter_subdir)
+
+    mets_file = None
+    if files:
+        try:
+            mets_file = find_mets_file(sip_directory)
+        except METSDiscoveryError as err:
+            job.print_output("METS file not found: {}".format(err))
+            pass
```
sevein: I think we don't need `pass` here.

src/MCPClient/lib/clientScripts/update_size_and_checksum.py
```diff
@@ -111,39 +93,105 @@ def get_file_info_from_mets(job, shared_path, file_):
...
+            file_path = file_.currentlocation.replace(
+                SIP_REPLACEMENT_PATH_STRING, sip_directory
+            )
+        # See if it's a Transfer and in particular a Archivematica AIP transfer.
+        # If so, try to extract the size, checksum and checksum function from the
+        # original METS document.
+        if (
+            file_.transfer
+            and (not file_.sip)
+            and file_.transfer.type == "Archivematica AIP"
```
sevein: Do you think this could be abstracted as a File model method?

src/MCPClient/lib/clientScripts/assign_file_uuids.py
```diff
@@ -123,88 +107,117 @@ def get_file_info_from_mets(job, sip_directory, file_path_relative_to_sip):
...
+    try:
+        mets_file = find_mets_file(sip_directory)
+    except METSDiscoveryError as err:
+        job.print_output("METS file not found: {}".format(err))
+        pass
+    if mets_file:
+        job.print_output(
+            "Reading METS file", mets_file, "for reingested file information."
+        )
+        logger.info("Archivematica AIP: reading METS file %s.", mets_file)
```
sevein: We should probably take this chance to either stop printing or logging. The latter uses stderr (JobContext adds the logger handler automatically), so we end up with two events. See screenshot: https://i.imgur.com/FTnek03.png.

sevein (changes requested):
Nice job, Tessa. I've added comments on some minor issues but it's working great. If you don't have the time then merge right away.

FYI, in this context, I think https://github.com/archivematica/Issues/issues/313 is relevant. At some point we have to measure the impact of these long transactions and if it's a problem then find better ways to write.

tw4l:
Thanks @sevein! This should be ready for re-review now.

On the question of database transactions, I ran some tests using per-file transactions and then batches of 10 and 100 files. Here were the results, with the same 2780 file dataset as above:

```
# Baseline - One big transaction (prior to revisions)

- Job Assign file UUIDs to objects: 20 seconds
- Job Assign checksums and file sizes to objects: 19 seconds
- Job Assign file UUIDs to objects (reingest): 62 seconds
- Job Assign checksums and file sizes to objects (reingest): 54 seconds

# Per-file transactions

- Job Assign file UUIDs to objects: 41 seconds
- Job Assign checksums and file sizes to objects: 40 seconds
- Job Assign file UUIDs to objects (reingest): 94 seconds
- Job Assign checksums and file sizes to objects (reingest): 73 seconds

# Batch of 10 files/transaction

- Job Assign file UUIDs to objects: 24 seconds
- Job Assign checksums and file sizes to objects: 24 seconds
- Job Assign file UUIDs to objects (reingest): 85 seconds
- Job Assign checksums and file sizes to objects (reingest): 59 seconds

# Batch of 100 files/transaction

- Job Assign file UUIDs to objects: 20 seconds
- Job Assign checksums and file sizes to objects: 19 seconds
- Job Assign file UUIDs to objects (reingest): 81 seconds
- Job Assign checksums and file sizes to objects (reingest): 59 seconds
```

I've gone with a batch size of 10 files per transaction in this PR, since that seemed like a nice balance to me between performance and keeping the transactions relatively short. But I'd be happy to change that or do further research if you'd prefer.

src/dashboard/src/main/models.py
```diff
@@ -750,6 +752,17 @@ def add_custom_identifier(self, scheme, value):
         """Allow callers to add custom identifiers to the model's instance."""
         self.identifiers.create(type=scheme, value=value)
 
+    @property
+    def in_reingested_aip(self):
+        """Return boolean indicating if file is in reingested AIP"""
+        if not self.transfer:
+            return False
+        return (
+            self.transfer
```
sevein: We can remove this line since we know it's already true.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -535,3 +557,9 @@ def relative_path_to_aip_mets_file(uuid, current_path):
     mets_name = "METS.{}.xml".format(uuid)
     mets_path = "{}/data/{}".format(package_name_without_extensions, mets_name)
     return mets_path
+
+
+def chunk_list(input_list, chunk_size=10):
+    """Yield successive chunks of size chunk_size from input list."""
+    for index in range(0, len(input_list), chunk_size):
+        yield input_list[index : index + chunk_size]
```
sevein: I'm wondering if we should write something that looks more like [`more_itertools.grouper`](https://github.com/more-itertools/more-itertools/blob/e899af25e95a2300d63b832f8a7861d55b979371/more_itertools/recipes.py#L284-L297) since `zip_longest` is written in C and it looks like it could work with any sequence.

src/dashboard/src/main/tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+        completed_at=timezone.now(),
+    )
+
+    models.Job.objects.create(
+        sipuuid=sip.pk, unittype="unitSIP", createdtime=timezone.now()
+    )
+    models.File.objects.create(uuid=uuid.uuid4(), sip=sip)
+    models.Access.objects.create(sipuuid=sip.pk)
+
+    return sip
```
sevein, line 40: It's great that you did this. I've been resisting a little because I wasn't sure what could be good reusable and combinable fixtures for every module, but you just came up with very good ones. I'm hoping we can have just a single conftest for all Archivematica components once we standardize the application layout and we stop messing with PYTHONPATH, lots of room there for cleaning things up. Thanks!

sevein (approved):
Veeeery nice, thanks Tessa!

src/dashboard/src/main/tests/conftest.py
tw4l: My pleasure! And that sounds like a great goal - it'd be really nice to have a standard set of fixtures across the project.

src/archivematicaCommon/lib/archivematicaFunctions.py
tw4l: Oh nice! Yeah I can do that for sure.

src/archivematicaCommon/lib/archivematicaFunctions.py
```diff
@@ -30,6 +30,7 @@
 import errno
 import glob
 import hashlib
+from itertools import zip_longest
```
sevein: Sorry, I didn't know neither - will need `from six.moves import zip_longest`.

tw4l: Ah nice, thanks!

PR #1754, opened 2021-06-04
Upgrade dependencies

sevein:
Doing this before we publish v1.13.0-rc.2.


replaceafill (approved):
~~Should we upgrade the Python 2 requirements too?~~

:facepalm: Sorry, didn't notice the requirements.txt name :sweat_smile:

PR #1759, opened 2021-06-09
Make javascript catalog view login exempt

replaceafill:
The `installer.middleware.ConfigurationCheckMiddleware` class requires
authentication for accessing the javascript catalog view at `/jsi18n`
which produces a JS `Uncaught SyntaxError` in the login form when its
`<script>` tag receives HTML instead of JS.

This fixes it by adding the javascript catalog path to the list of
login exempt URLs.


sevein (approved):
👏

PR #1760, opened 2021-06-09
Fix preview file pane in the appraisal tab

replaceafill:


sevein (approved):
👏

PR #1762, opened 2021-06-10
Update filegrpuse if file already has UUID

tw4l:
Connected to https://github.com/archivematica/issues/issues/1429

This commit fixes a defect in https://github.com/artefactual/archivematica/commit/b6183e3a7be3c893916a4bd20a8ded1e68fa561c whereby a routine in the original `assign_file_uuids` client script that checks whether a file already has a UUID and updates its `filegrpuse` if so was missed when converting the script to run per-directory rather than per-file. This resolves an issue with manual normalization and makes all of the AMAUATs pass again.


replaceafill (approved):
LGTM :+1:

PR #1767, opened 2021-06-22
Add rebuild_transfer_backlog and related fixes

hakamine:
- Add the following options to rebuild_transfer_backlog:
    --delete-all: to delete entire index (do not delete it by default)
    --uuid: to only index a specific transfer
    --skip-to: start indexing from a specific transfer
    ---delete: before re-reindexing a transfer, will delete any data found in Elasticsearch with a matching UUID

-   Add missing  modification_date assignment in elasticSearchFunctions._index_transfer_files() to prevent "local variable referenced before assignment" exception

Connected to https://github.com/archivematica/Issues/issues/1457 and https://github.com/archivematica/Issues/issues/1458


sevein (approved):
I've tested the various optional arguments you're introducing. They seem very useful and work as expected. Thank you!

PR #1770, opened 2021-06-29
Fix access target decoding on AtoM DIP upload

replaceafill:
Connected to https://github.com/archivematica/Issues/issues/1470


sevein (approved):
LGTM

PR #1773, opened 2021-07-08
Fix default duration in purge command

replaceafill:
This fixes the default duration of the purge command and set it to 6 hours.

Connected to https://github.com/archivematica/Issues/issues/1239


sevein (approved):
Ops, thank you! :pray:

PR #1777, opened 2021-09-02
Fix access target encoding issue for DIP upload

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1496


tw4l:
This is now ready for review. Thanks!

sevein (approved):
This is working, thank you! The test we had in test_upload_qubit didn't trigger this error because it was creating the Access object beforehand. Adding a new test where the Access object was left uncreated triggered the error, and using six.ensure_binary fixed it.

sevein:
efea8adc in stable/1.13.x

PR #1779, opened 2021-09-14
Update MCPClient forking model

sevein:
This pull request is a rebase of Cole's https://github.com/artefactual/archivematica/pull/1499 with some small adjustements and fixes. It introduces a new prefork execution model in MCPClient based on a process pool. The size of the pool is configurable and dynamically adjusted by the parent process. Workers are automatically restarted to free resources and will reuse database connections when possible. Logging is centralized with a queue that is shared with pool processes. Metrics are captured using mmap and collected by the parent process. For more details, see https://gist.github.com/sevein/f4ff0e024c1b394eedb196e63bdeb91c.

[Two new settings](https://github.com/artefactual/archivematica/pull/1779/files#diff-83fe977352d9deeb2b488bf52de01cf869f7c680283b4eca7f3b6f42362c200c) were added:
* **`ARCHIVEMATICA_MCPCLIENT_WORKERS`**
  Number of workers. If undefined, it defaults to the number of CPUs available on the machine. Only client modules that define `concurrent_instances` will perform concurrent execution of tasks (e.g. [identify_file_format](https://github.com/artefactual/archivematica/blob/2a83f0c47243881e363cb3db031e51b089c4094a/src/MCPClient/lib/clientScripts/identify_file_format.py#L20-L21)).
* **`ARCHIVEMATICA_MCPCLIENT_MAX_TASKS_PER_CHILD`**
  Maximum number of tasks a worker can execute before it's replaced by a new process in order to free resources.

While qa/1.x is still buliding on Python 2 and Python 3 (we're dropping the former soon), this pull request targets Python 3 only.

Connects to https://github.com/archivematica/Issues/issues/943 and https://github.com/archivematica/Issues/issues/1482.


sevein:
AMAUAT tests pass if SS is in stable/0.19.x. I was originally trying against qa/0.x but something there broke, but that can be addressed separately.

src/MCPClient/lib/client/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+    """The multi-process directory must be wiped between runs."""
+    shutil.rmtree(prometheus_tmp_dir)
+
+
+# Set up the temporary directory that the Prometheus client library will use to
+# store metrics of MCPClient and its child processes.
+#
+# Setting the environment string "prometheus_multiproc_dir" is the only
+# mechanism we have to ensure that the Prometheus client library is in
+# multi-process mode. "promethus_clietn.values.ValueClass" will be an instance
```
replaceafill, line 38:
```suggestion
# multi-process mode. "prometheus_client.values.ValueClass" will be an instance
```

replaceafill (approved):
This looks great @sevein! I appreciate the detailed docstrings and explanations in the code and linked issues and gists. And thank you for removing those Python 2 bits throughout the change.

replaceafill:
Closing in favor of https://github.com/artefactual/archivematica/pull/1845

PR #1780, opened 2021-09-24
Bump version to 1.13.1

sevein:
Connects to https://github.com/archivematica/Issues/issues/1499 and https://github.com/archivematica/Issues/issues/1500.


tw4l (approved):
Thanks!

PR #1781, opened 2021-09-24
Bump version to 1.14.0

sevein:
Connects to https://github.com/archivematica/Issues/issues/1499 and https://github.com/archivematica/Issues/issues/1501.


tw4l (approved):
LGTM! It's a good approach and the build is now working. Thanks!

PR #1782, opened 2021-09-28
parse_external_mets: handle not found (OSError)

sevein:
Change parse_external_mets so it doesn't return an error code when the external
METS file cannot be found, presumably not provided by the user. This avoids the
job to be marked as failed which was misleading.

Connects to https://github.com/archivematica/Issues/issues/886.


tw4l (approved):
Nice, thanks!

sevein:
3f64cc1f in stable/1.13.x

PR #1784, opened 2021-10-05
Remove message attribute from exceptions

replaceafill:
The `message` attribute was [deprecated in Python 2.6 and dropped in Python 3](https://www.python.org/dev/peps/pep-0352/).

Connects to https://github.com/archivematica/Issues/issues/1500.


sevein (approved):

sevein:
4ba16d1780bd5cbee269450f526c31ccd6a92e78 in stable/1.13.x

PR #1787, opened 2021-10-27
Support XML metadata files and modify DMD sections

jraddaoui:
Add support for XML metadata files including configurable validation
and addition to the METS file as DMD sections on ingest and re-ingest.
Modify DMD sections management to improve versioning and indexing.

Refs https://github.com/archivematica/Issues/issues/1531 and https://github.com/archivematica/Issues/issues/1537.

### DIFF
--- a/src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
+++ b/src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python
+#
+# This file is part of Archivematica.
+#
+# Copyright 2010-2021 Artefactual Systems Inc. <http://artefactual.com>
+#
+# Archivematica is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Archivematica is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Archivematica. If not, see <http://www.gnu.org/licenses/>.
+
+"""Management of XML metadata files."""
+
+import csv
+from lxml import etree
+from pathlib import Path
+from urllib.parse import urlparse
+from urllib.request import urlopen
+import requests
+
+from importlib_metadata import version
+import create_mets_v2 as createmets2
+
+from databaseFunctions import insertIntoEvents
+import namespaces as ns
+
+from main import models
+
+
+def process_xml_metadata(mets, sip_dir, sip_uuid, sip_type, xml_validation):
+    if not xml_validation:
+        return mets, []
+    xml_metadata_mapping, xml_metadata_errors = _get_xml_metadata_mapping(
+        sip_dir, reingest="REIN" in sip_type
+    )
+    if not xml_metadata_mapping:
+        return mets, xml_metadata_errors
+    for fsentry in mets.all_files():
+        if fsentry.use != "original" and fsentry.type != "Directory":
+            continue
+        path = fsentry.get_path()
+        if path not in xml_metadata_mapping:
+            continue
+        for xml_type, xml_path in xml_metadata_mapping[path].items():
+            if not xml_path:
+                fsentry.delete_dmdsec("OTHER", xml_type)
+                continue
+            tree = etree.parse(str(xml_path))
+            try:
+                schema_uri = _get_schema_uri(tree, xml_validation)
+            except ValueError as err:
+                xml_metadata_errors.append(err)
+                continue
+            if schema_uri:
+                xml_rel_path = xml_path.relative_to(sip_dir)
+                try:
+                    metadata_file = models.File.objects.get(
+                        sip_id=sip_uuid,
+                        currentlocation="%SIPDirectory%{}".format(xml_rel_path),
+                    )
+                except models.File.DoesNotExist:
+                    xml_metadata_errors.append(
+                        "No uuid for file: {}".format(xml_rel_path)
+                    )
+                    continue
+                valid, errors = _validate_xml(tree, schema_uri)
+                _add_validation_event(
+                    mets, metadata_file.uuid, schema_uri, valid, errors
+                )
+                if not valid:
+                    xml_metadata_errors += errors
+                    continue
+            fsentry.add_dmdsec(
+                tree.getroot(),
+                "OTHER",
+                othermdtype=xml_type,
+                status="update" if "REIN" in sip_type else "original",
+            )
+    return mets, xml_metadata_errors
+
+
+def _get_xml_metadata_mapping(sip_path, reingest=False):
+    """Get a mapping of files/dirs in the SIP and their related XML files.
+
+    On initial ingests, it looks for such mapping in source-metadata.csv
+    files located on each transfer metadata folder. On reingest it only
+    considers the source-metadata.csv file in the main metadata folder.
+
+    Example source-metadata.csv:
+
+    filename,metadata,type
+    objects,objects_metadata.xml,metadata_type
+    objects/dir,dir_metadata.xml,metadata_type
+    objects/dir/file.pdf,file_metadata_a.xml,metadata_type_a
+    objects/dir/file.pdf,file_metadata_b.xml,metadata_type_b
+
+    Example dict returned:
+
+    {
+        "objects": {"metadata_type": Path("/path/to/objects_metadata.xml")},
+        "objects/dir": {"metadata_type": Path("/path/to/dir_metadata.xml")},
+        "objects/dir/file.pdf": {
+            "metadata_type_a": Path("/path/to/file_metadata_a.xml"),
+            "metadata_type_b": Path("/path/to/file_metadata_b.xml"),
+        },
+    }
+
+    :param str sip_path: Absolute path to the SIP.
+    :param bool reingest: Boolean to indicate if it's a reingest.
+    :return dict, list: Dictionary with File/dir path -> dict of type -> metadata
+    file pathlib Path, and list with errors (if a CSV row is missing the filename
+    or type, or if there is more than one entry for the same filename and type).
+    """
+    mapping = {}
+    errors = []
+    source_metadata_paths = []
+    metadata_path = Path(sip_path) / "objects" / "metadata"
+    transfers_metadata_path = metadata_path / "transfers"
+    if reingest:
+        source_metadata_paths.append(metadata_path / "source-metadata.csv")
+    elif transfers_metadata_path.is_dir():
+        for dir_ in transfers_metadata_path.iterdir():
+            source_metadata_paths.append(dir_ / "source-metadata.csv")
+    for source_metadata_path in source_metadata_paths:
+        if not source_metadata_path.is_file():
+            continue
+        with source_metadata_path.open() as f:
+            reader = csv.DictReader(f)
+            for row in reader:
+                if not all(k in row and row[k] for k in ["filename", "type"]):
+                    errors.append(
+                        "A row in {} is missing the filename and/or type".format(
+                            source_metadata_path
+                        )
+                    )
+                    continue
+                if row["type"] == "CUSTOM":
+                    errors.append(
+                        "A row in {} is using CUSTOM, a reserved type".format(
+                            source_metadata_path
+                        )
+                    )
+                    continue
+                if row["filename"] not in mapping:
+                    mapping[row["filename"]] = {}
+                elif row["type"] in mapping[row["filename"]]:
+                    errors.append(
+                        "More than one entry in {} for path {} and type {}".format(
+                            source_metadata_path, row["filename"], row["type"]
+                        )
+                    )
+                    continue
+                if row["metadata"]:
+                    row["metadata"] = source_metadata_path.parent / row["metadata"]
+                mapping[row["filename"]][row["type"]] = row["metadata"]
+    return mapping, errors
+
+
+def _get_schema_uri(tree, xml_validation):
+    key = None
+    checked_keys = []
+    schema_location = tree.xpath(
+        "/*/@xsi:noNamespaceSchemaLocation", namespaces={"xsi": ns.xsiNS}
+    )
+    if schema_location:
+        key = schema_location[0].strip()
+        checked_keys.append(key)
+    if not key or key not in xml_validation:
+        schema_location = tree.xpath(
+            "/*/@xsi:schemaLocation", namespaces={"xsi": ns.xsiNS}
+        )
+        if schema_location:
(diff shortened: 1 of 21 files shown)


src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
 
 
+def createDMDFromXML(job, path, baseDirectoryPath, state):
+    # TODO:
+    # - XML association:
+    #   - Avoid filenames mismatch (metadata XML filename may have changed).
+    #   - Consider equal filenames from different transfer sources?
+    #   - Allow multiple XML metadata per file?
+    #   - Use extra CSV to relate XML file(s)?
+    xmlFilename = os.path.splitext(os.path.relpath(path, "objects"))[0] + ".xml"
```
replaceafill: Could we start using [pathlib](https://docs.python.org/3.6/library/pathlib.html) for path operations and directory traversal? Also, now that Python 2 has been removed entirely from the latest `qa/1.x` I don't think we have to deal with import errors anymore.

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -1141,6 +1240,16 @@ def createFileSec(
                         f.originallocation.replace("%transferDirectory%", "", 1),
                         state,
                     )
+                    XMLDMDID = createDMDFromXML(
```
replaceafill: Should this be renamed to `createDMDIDFromXML` to match the name pattern in the previous function above?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -1141,6 +1240,16 @@ def createFileSec(
...
+                        job,
+                        f.originallocation.replace("%transferDirectory%", "", 1),
+                        baseDirectoryPath,
+                        state,
+                    )
+                    if DMDIDS and XMLDMDID:
+                        DMDIDS += " " + XMLDMDID
+                    elif XMLDMDID:
+                        DMDIDS = XMLDMDID
```
replaceafill: Would something like `DMDIDS = ' '.join(i for i in [DMDIDS, XMLDMDID] if i)` work to avoid the `if`?

src/archivematicaCommon/lib/elasticSearchFunctions.py
```diff
@@ -372,6 +372,7 @@ def _get_index_settings():
     return {
         "index": {
             "mapping": {
+                "ignore_malformed": True,
```
replaceafill: Will this require to recreate the index(es) somehow?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -1141,6 +1240,16 @@ def createFileSec(
                         f.originallocation.replace("%transferDirectory%", "", 1),
                         state,
                     )
+                    XMLDMDID = create_dmd_from_xml(
```
replaceafill:
```suggestion
                    XMLDMDID = create_dmdid_from_xml(
```

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
 
 
+def create_dmd_from_xml(job, path, base_directory_path, state):
+    # TODO:
+    # - XML association:
+    #   - Avoid filenames mismatch (metadata XML filename may have changed).
+    #   - Consider equal filenames from different transfer sources?
+    #   - Allow multiple XML metadata per file?
+    #   - Use extra CSV to relate XML file(s)?
+    xml_filename = os.path.splitext(os.path.relpath(path, "objects"))[0] + ".xml"
```
replaceafill: Could we use [pathlib](https://docs.python.org/3.6/library/pathlib.html) for path handling and directory traversal here? Now that Python 2 has been removed in the latest `qa/1.x` branch I don't think we even need to worry about [the old `ImportErrors`](https://github.com/artefactual/archivematica/blob/bf635157bb829124104b7d0d9278a579c61c261d/src/archivematicaCommon/lib/archivematicaFunctions.py#L39-L42).

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
+    transfer_metadata = os.path.join(
+        base_directory_path, os.path.join("objects", "metadata", "transfers")
+    )
+    if not os.path.isdir(transfer_metadata):
+        return
+    for dir in os.listdir(transfer_metadata):
```
replaceafill: Should we rename [`dir`](https://docs.python.org/3.6/library/functions.html#dir) since it's a built-in function?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
+        xml_path = os.path.join(transfer_metadata, dir, xml_filename)
+        if not os.path.isfile(xml_path):
+            continue
+        try:
+            tree = etree.parse(xml_path)
+            valid, errors = _validate_xml(tree)
```
replaceafill: Would it be worth separating parsing and validation in the `try` block to provide more accurate logging?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
+                    job.pyprint("\t- {}".format(error), file=sys.stderr)
+                return
+        except etree.LxmlError as err:
+            job.pyprint(
+                "Could not parse or validate {}\n\t- {}".format(xml_path, err),
+                file=sys.stderr,
+            )
+            return
+        root = tree.getroot()
+        _, _, tag = root.tag.partition("}")
```
replaceafill: Could we use [`QName(root).localname`](https://lxml.de/api/lxml.etree.QName-class.html) to extract the tag without the namespace?

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
+        state.globalDmdSecCounter += 1
+        DMDID = "dmdSec_" + state.globalDmdSecCounter.__str__()
```
replaceafill:
```suggestion
        DMDID = "dmdSec_{}".format(state.globalDmdSecCounter)
```

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -284,6 +285,104 @@ def createDMDIDsFromCSVMetadata(job, path, state):
...
+        else:
+            return False, ["Unknown schema type: {}".format(schema_type)]
+    except etree.LxmlError as err:
+        return False, ["Could not parse schema file: {}".format(schema_uri), err]
+    if not schema.validate(tree):
+        return False, schema.error_log
+    return True, []
+
+
+def _get_schema_uri(tree):
```
replaceafill: Can we copy the comment from the `common` setting module about this logic?

replaceafill (commented):
@jraddaoui I think this approach is convenient for our current knowledge of the use case and I understand it's still work in progress, but I just left a few comments with potential improvements.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Nice, thanks!!

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Sure, waiting a little bit more to see how much of this will be needed in the end, before adding docstrings and tests for these functions.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Cool, a lot cleaner.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: I had some conflicts about this ... I don't think that pattern represents the actual purpose of those functions. I also thought about renaming the other one and including this call inside it, since the params are the same in the end, but I don't want to mess a lot with existing functions.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Neat, thanks!

src/archivematicaCommon/lib/elasticSearchFunctions.py
jraddaoui: Yes, this is just a hack for now to avoid errors importing XMLs with different structure in the same elements.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Mentioned above.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Addressed above.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: Good point. Since validation has its own exception handling I've moved it out from this try.

jraddaoui:
Thanks for the great feedback @replaceafill, I think I've addressed it all (in commits or comments). Fixing the tests now.

I also added the initial implementation to get the XML metadata to original files mapping from a `source-metadata.csv` file. If you could take a quick look, that would be great.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -0,0 +1,265 @@
...
+import create_mets_v2 as createmets2
+
+from databaseFunctions import insertIntoEvents
+
+from django.conf import settings as mcpclient_settings
+from main import models
+
+
+def process_xml_metadata(mets, sip_dir, sip_uuid, sip_type):
+    if not getattr(mcpclient_settings, "XML_VALIDATION", False):
```
replaceafill: Do you think it's worth moving this condition back to the [caller](https://github.com/artefactual/archivematica/blob/65c54429ba34fe43137c63e20e89b4a98dd48895/src/MCPClient/lib/clientScripts/create_mets_v2.py#L1890-L1893)?

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -0,0 +1,265 @@
...
+                    continue
+                if row["metadata"]:
+                    row["metadata"] = source_metadata_path.parent / row["metadata"]
+                mapping[row["filename"]][row["type"]] = row["metadata"]
+    return mapping, errors
+
+
+def _get_schema_uri(tree):
+    XSI = "http://www.w3.org/2001/XMLSchema-instance"
+    VALIDATION = mcpclient_settings.XML_VALIDATION
```
replaceafill: Do you think is worth trying to avoid the `mcpclient_settings` import and cascade this as a parameter from the caller to the top level function and to here?

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -0,0 +1,265 @@
...
+    key = None
+    checked_keys = []
+    schema_location = tree.xpath(
+        "/*/@xsi:noNamespaceSchemaLocation", namespaces={"xsi": XSI}
```
replaceafill: Could this reuse the [namespaces.NSMAP](https://github.com/artefactual/archivematica/blob/76f6faa1d895e3a83923b45c9f5a5b18cfab1a3e/src/archivematicaCommon/lib/namespaces.py#L26-L35)?

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -0,0 +1,265 @@
...
+def _add_validation_event(mets, file_uuid, schema_uri, valid, errors):
+    event_data = {
+        "eventType": "validation",
+        "eventDetail": 'type="metadata"; validation-source-type="'
+        + schema_uri.split(".")[-1]
+        + '"; validation-source="'
+        + schema_uri
+        + '"; program="lxml"; version="'
+        + etree.__version__
+        + '"',
```
replaceafill: This part is a bit hard to read. What do you think of splitting it like this:

```
    event_detail = {
        "type": "metadata",
        "validation-source-type": schema_uri.split(".")[-1],
        "validation-source": schema_uri,
        "program": "lxml",
        "version": etree.__version__,
    }
...
        "eventDetail": "; ".join(
            [f'{k}="{v}"' for k, v in event_detail.items()]
        ),
...
```

src/MCPClient/lib/clientScripts/create_mets_v2.py
```diff
@@ -36,12 +37,13 @@
 from uuid import uuid4
 
 import django
+import metsrw
 import scandir
 import six
 
 django.setup()
 # dashboard
-from django.utils import timezone
```
replaceafill, line 44: Why is `timezone.now()` not relevant anymore?

src/MCPClient/tests/test_xml_metadata.py
```diff
@@ -0,0 +1,570 @@
+#!/usr/bin/env python
```
replaceafill: Should this module be renamed to `tests_archivematicaCreateMETSMetadataXML` to hint what's covering?

replaceafill (approved):
Great work @jraddaoui! I left a couple of questions and suggestions, but this looks good to me.

src/MCPClient/lib/clientScripts/create_mets_v2.py
jraddaoui: It was used to add the CREATEDATE attribute to the METS header. Now, with the inclusion of the metsrw on the final steps, if that attribute is set when the METS is parsed, metsrw adds the LASTMODDATE attribute thinking it's a re-ingest.

I removed that at some point and then introduced `datetime.datetime` for the new dates, to keep consistency with the metsrw.

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
jraddaoui: Nice, thanks!

src/MCPClient/lib/clientScripts/archivematicaCreateMETSMetadataXML.py
jraddaoui: Since we still need to pass that setting, I think it's okay to have it here.

PR #1788, opened 2021-11-05
Fix Archival Storage date range search

tw4l:
Connected to https://github.com/archivematica/issues/issues/1302

This PR fixes an issue with the Archival Storage tab search where searches on date range were not returning any results because Elasticsearch was throwing an exception when given a range query without a field specified.

In order to fix the issue, `YYYY-MM-DD` input strings are converted into Unix timestamps and then compared against the `aipfiles` index's `indexedAt` field. This is only an approximation of the AIP creation date but close enough for searches at the level of days, and allows us to fix the defect without needing to introduce any additional fields to the `aips` index.

In order to make date range searches inclusive, a day is added to end dates before converting them to Unix timestamps, and `lt` (less than) is used for end dates in the Elasticsearch DSL query rather than `lte` (less than or equal to). In practice this means a Date range search input of "2021-10-01:2021-10:11" will be converted to the following DSL query:

```
{"range": {"indexedAt": {"gte": <Unix timestamp for 2021-10-01T00:00:00>, "lt": <Unix timestamp for 2021-10-12T00:00:00>}}}
```

### DIFF
--- a/src/dashboard/src/components/advanced_search.py
+++ b/src/dashboard/src/components/advanced_search.py
@@ -17,9 +17,8 @@
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import absolute_import
 
+from datetime import datetime, timedelta
 import logging
-
-import dateutil.parser
 
 logger = logging.getLogger("archivematica.dashboard.advanced_search")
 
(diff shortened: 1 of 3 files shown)


sevein (approved):
Thanks!

PR #1791, opened 2021-11-23
Fix the copy_metadata_files view for partial reingests

replaceafill:
When the `copy_metadata_files` view is called on a partial (objects or metadata-only) reingest which is waiting for approval (i.e. before the `Restructure from bag AIP to SIP directory format` job is executed) the SIP structure still looks like a bag causing the view to create an extra `metadata` directory in the wrong location. This affects the associated API endpoint and the `Add metadata files` view of the dashboard UI.

This fixes it by adding a `isPartialReingest` unit variable that is set when the MCPServer creates the SIP package and checking it when the `copy_metadata_files` view is called. The unit variable is later removed once the reingest structucture has been normalized and the reingested METS has been parsed to repopulate the database.

Connected to https://github.com/archivematica/Issues/issues/959

### DIFF
--- a/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
+++ b/src/MCPClient/lib/clientScripts/parse_mets_to_db.py
@@ -551,6 +551,9 @@ def main(job, sip_uuid, sip_path):
 
     parse_rights(job, sip_uuid, root)
 
+    if sip.count():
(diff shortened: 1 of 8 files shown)


sevein (approved):
Nicely done!

PR #1794, opened 2022-01-25
Simplify dict normalization for ES indexing

jraddaoui:
Elasticsearch fields accept a single value or a list of values so there
is no need to make a distinction in that case, only between strings and
dictionaries. The previous normalization named dict fields always as
"tag_dict_list" even if there was only one value and it named string
fields as "tag" (for a single value) or "tag_str_list" (for multiple
values). The new normalization only differentiates between dicts
("tag_dict") and strings ("tag").

Refs https://github.com/archivematica/Issues/issues/1524.


replaceafill (approved):
This looks good to me @jraddaoui. And thanks for recording our chat in https://github.com/archivematica/Issues/issues/1524#issuecomment-1021385570 :bow:

PR #1798, opened 2022-03-04
Remove UI-based metadata only reingest tag

replaceafill:
For context see https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/180

Connected to https://github.com/archivematica/Issues/issues/1531


jraddaoui (approved):
Perfect, thanks!

PR #1800, opened 2022-03-31
WIP: Support Elasticsearch 7.x and OpenSearch 1.x

jraddaoui:
Refs https://github.com/archivematica/Issues/issues/1548.

### DIFF
--- a/hack/docker-compose.opensearch.yml
+++ b/hack/docker-compose.opensearch.yml
@@ -1,0 +1,38 @@
+---
+version: "2.1"
+
+volumes:
+  opensearch_data:
+
+services:
+  opensearch:
+    image: opensearchproject/opensearch:1.3.0
+    environment:
+      - discovery.type=single-node
+      - cluster.name=am-cluster
+      - cluster.routing.allocation.disk.threshold_enabled=false
+      - node.name=am-node
+      - bootstrap.memory_lock=true
+      - "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m"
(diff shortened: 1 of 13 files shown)


FingerlessGlov3s:
Hi José,

Is there any update on running archivematica with ElasticSearch 7 or even ElasticSearch 8?  As I need to do a new installation on EL9 but noticed the version documented in the installation guide is ElasticSearch 6 which went EOL over 2 years ago. So from a security point of view, that's not ideal to install EOL software.

misilot:
Hello, I just wanted to check if there was an update on this? 

I am currently setting this with some security scans

> Description
> 
> The version of Elasticsearch installed on the remote host is prior to 7.17.21 or 8.13.3. It is, therefore, affected by a vulnerability as referenced in the ESA-2024-25 advisory.
> 
>   - An allocation of resources without limits or throttling in Elasticsearch can lead to an OutOfMemoryError exception resulting in a crash via a specially crafted query using an SQL function. (CVE-2024-43709)
> 
> Note that Nessus has not tested for this issue but has instead relied only on the application's self-reported version number.
> 
> Solution
> 
> Upgrade to Elasticsearch version 7.17.21 / 8.13.3 or later.

replaceafill:
@misilot We will upgrade to Elasticsearch 8.x in Archivematica 1.18.0. See https://github.com/archivematica/Issues/issues/1752

replaceafill:
This has been replaced with https://github.com/artefactual/archivematica/pull/2098

PR #1801, opened 2022-05-16
Do not resolve external xml entities

sbreker:
Use a custom parser with lxml that disables external entity resolution
and disables network access.

'no_network' is true by default.

### DIFF
--- a/src/MCPClient/lib/clientScripts/characterize_file.py
+++ b/src/MCPClient/lib/clientScripts/characterize_file.py
@@ -97,7 +97,10 @@ def main(job, file_path, file_uuid, sip_uuid):
             and rule.command.output_format.pronom_id == "fmt/101"
         ):
             try:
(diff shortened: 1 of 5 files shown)


sevein:
I believe this is the default since lxml 5.0.0 was released? See [CHANGES.txt](https://github.com/lxml/lxml/blob/master/CHANGES.txt):

> LP#1742885: lxml no longer expands external entities (XXE) by default to prevent
  the security risk of loading arbitrary files and URLs.  If this feature is needed,
  it can be enabled in a backwards compatible way by using a parser with the option
  ``resolve_entities=True``.  The new default is ``resolve_entities='internal'``.
  
It looks like we can close this PR.

replaceafill:
> It looks like we can close this PR.

Thanks for pointing that out @sevein! I agree with closing this.

PR #1805, opened 2022-09-19
Make ES limits configurable

mamedin:
The total AIPs limit on the "Archival Storage" and "Appraisal" tabs were limited to 10000 items. It was hardcoded so this pull request makes this limit configurable.

Connects to https://github.com/archivematica/Issues/issues/1571

### DIFF
--- a/src/dashboard/install/README.md
+++ b/src/dashboard/install/README.md
@@ -143,6 +143,12 @@ variables or in the gunicorn configuration file.
     - **Type:** `float`
     - **Default:** `10`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_ELASTICSEARCH_MAX_QUERY_SIZE`**:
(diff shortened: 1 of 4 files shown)


mamedin:
The limits on `src/archivematicaCommon/lib/elasticSearchFunctions.py` have not been changed, because iirc it is only to create the indices (from scratch or reindexing all with deletion).

I think this is enough and we can explain in documentation how to run the elasticsearch query to update the limits in indices.

src/dashboard/install/README.md
```diff
@@ -143,6 +143,12 @@ variables or in the gunicorn configuration file.
     - **Type:** `float`
     - **Default:** `10`
 
+- **`ARCHIVEMATICA_DASHBOARD_DASHBOARD_ELASTICSEARCH_MAX_QUERY_SIZE`**:
+    - **Description:** limits the number of results per query that Elasticseach client can retrieve. Note that when using a value greater than 10000, the `total_fields.limit` and `max_result_window` must be increased in indices (For instance, when using `100000` run on ES server linux CLI: `curl -XPUT 'http://localhost:9200/aips,aipfiles,transfers,transferfiles/_settings' -H "Content-Type: application/json"  -d '{"index.mapping.total_fields.limit": 100000, "index.max_result_window": 100000}'`)
+    - **Config file example:** `Dashboard.elasticsearch_max_query_size`
+    - **Type:** `float`
```
replaceafill: Should this be `integer` given that it's `int` in the `CONFIG_MAP`?

replaceafill (approved):
LGTM :+1:

PR #1806, opened 2022-09-23
models: fix Access.get_title() exception handling

hakamine:
Fix JobQuerySet.get_directory_name() exception handling, which is invoked by Access.get_title()

Replace first() with [0] in order to trigger IndexError when required, and correctly handle the exception. Using first() may return a None potentially causing an unhandled exception dowstream (cf. https://docs.djangoproject.com/en/1.11/ref/models/querysets/#first)

Fixes https://github.com/archivematica/Issues/issues/1572

### DIFF
--- a/src/dashboard/src/main/models.py
+++ b/src/dashboard/src/main/models.py
@@ -896,7 +896,7 @@ class JobQuerySet(models.QuerySet):
         TODO(sevein): why is the name not a property of the unit?
         """
         try:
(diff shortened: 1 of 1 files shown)


sevein (approved):

hakamine:
b7c6021e4ee5cc9ad2fc1e26cc6991c43db1b915 in stable/1.13.x branch

PR #1810, opened 2022-12-21
Truncates filenames if they exceed os limit

helrond:
Gets the maximum filename length and truncates the name of the renamed file if it exceeds the max allowable by the underlying OS. Is one path to addressing Archivematica/Issues#1586

### DIFF
--- a/src/MCPClient/lib/clientScripts/change_names.py
+++ b/src/MCPClient/lib/clientScripts/change_names.py
@@ -63,8 +63,9 @@ def change_path(path):
     dirname = os.path.dirname(path)
 
     n = 1
+    max_filename = os.fpathconf(changed_name, 'PC_NAME_MAX')
     file_title, file_extension = os.path.splitext(changed_name)
-    changed_name = os.path.join(dirname, file_title + file_extension)
+    changed_name = os.path.join(dirname, file_title[:max_filename - len(file_extension)] + file_extension)
 
     while os.path.exists(changed_name):
         changed_name = os.path.join(
--- a/src/MCPClient/tests/test_filename_change.py
+++ b/src/MCPClient/tests/test_filename_change.py
@@ -308,6 +308,8 @@ class TestFilenameChange(TempDirMixin, TestCase):
         ("a\x80b", "ab"),
         ("Smörgåsbord.txt", "Smorgasbord.txt"),
         ("🚀", "_"),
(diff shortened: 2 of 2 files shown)


src/MCPClient/lib/clientScripts/change_names.py
```diff
@@ -63,8 +63,9 @@ def change_path(path):
     dirname = os.path.dirname(path)
 
     n = 1
+    max_filename = os.fpathconf(changed_name, 'PC_NAME_MAX')
```
replaceafill, line 66: I think `os.fpathconf` requires a file descriptor in the first parameter. Maybe you wanted `os.pathconf` which supports both file paths and descriptors?

src/MCPClient/lib/clientScripts/change_names.py
```diff
@@ -63,8 +63,9 @@ def change_path(path):
...
     file_title, file_extension = os.path.splitext(changed_name)
-    changed_name = os.path.join(dirname, file_title + file_extension)
+    changed_name = os.path.join(dirname, file_title[:max_filename - len(file_extension)] + file_extension)
 
     while os.path.exists(changed_name):
         changed_name = os.path.join(
```
replaceafill, line 71: Should this check also consider the `PC_NAME_MAX` limit when it calculates a new name because of existing collisions?

replaceafill (changes requested):
Thank you @helrond. Could you please target the `qa/1.x` branch in the PR?

I'm wondering if it makes sense to calculate the `PC_NAME_MAX` limit once based on the SIP directory instead of doing it on each file. Let me know what you think.

src/MCPClient/lib/clientScripts/change_names.py
helrond: I'm not sure I understand what you're getting at here?

helrond:
@replaceafill thank you - I will close this PR and resubmit against `qa/1.x`. I can also address the issues with `fpathconf` and checking the file size max once per SIP rather than per file. I wasn't totally getting what you were saying about considering the PC_NAME_MAX limit and collisions - would you be able to clarify for me?

src/MCPClient/lib/clientScripts/change_names.py
replaceafill: Sorry for not being clear. Suppose the filename hit the maximum length and was truncated, but there's already an existing file with that name. The block inside the `while` overrides the `changed_name` variable with:

```python
        changed_name = os.path.join(
            dirname, file_title + REPLACEMENT_CHAR + str(n) + file_extension
        )
```

and at this point `file_title` won't get truncated anymore, which may lead to the original problem?

helrond: Ah okay, I'm with you now!

PR #1826, opened 2023-06-08
Development environment updates

replaceafill:
This makes the following changes to the Docker environment:

* Installs Python 3.6 using [`pyenv`](https://github.com/pyenv/pyenv).
* Migrates the old [Compose V1 configuration](https://docs.docker.com/compose/migrate/) and upgrades dependencies.
* Adds arguments for setting the `archivematica` user and group IDs so the content created in the external volumes can be owned by the Docker host user.
* Removes unnecessary prefixes from the `tox` environments.
* Updates the `hack/README.md` file to reflect these updates.

Depends on https://github.com/artefactual/archivematica-storage-service/pull/649
Connected to https://github.com/archivematica/Issues/issues/1612
Connected to https://github.com/archivematica/Issues/issues/1557

### DIFF
--- a/hack/docker-compose.acceptance-tests.yml
+++ b/hack/docker-compose.acceptance-tests.yml
@@ -1,5 +1,4 @@
 ---
-version: "2.1"
 
 services:
   archivematica-acceptance-tests:
--- a/hack/docker-compose.yml
+++ b/hack/docker-compose.yml
@@ -108,6 +108,9 @@ services:
       dockerfile: "hack/Dockerfile"
       args:
         TARGET: "archivematica-mcp-server"
+        USER_ID: ${USER_ID:-1000}
+        GROUP_ID: ${GROUP_ID:-1000}
+        PYTHON_VERSION: ${PYTHON_VERSION:-3.6}
     environment:
       DJANGO_SECRET_KEY: "12345"
       DJANGO_SETTINGS_MODULE: "settings.common"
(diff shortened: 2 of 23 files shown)


hack/docker-compose.yml
```diff
@@ -16,11 +16,13 @@ volumes:
...
-    external:
-      name: "am-pipeline-data"
+    name: "am-pipeline-data"
+    external: true
+
   archivematica_storage_service_location_data:
-    external:
-      name: "ss-location-data"
+    name: "ss-location-data"
+    external: true
```
sevein, line 24: Is this getting rid of that annoying warning? Awesome!

hack/docker-compose.acceptance-tests.yml
```diff
@@ -1,5 +1,4 @@
 ---
-version: "2.1"
```
sevein, line 2: What's the benefit of dropping this attribute?

sevein (approved):
Lots of nice changes, including pyenv! :partying_face:

hack/docker-compose.acceptance-tests.yml
replaceafill: No benefit I guess, it's just that I noticed it's deprecated now:
* https://docs.docker.com/compose/compose-file/04-version-and-name/
* https://github.com/compose-spec/compose-spec/issues/13

sevein: Makes sense.

PR #1831, opened 2023-07-21
Development environment fixes

replaceafill:
This fixes a few problems identified after the `1.14` release:

* Removes a workaround in the AMAUATs CI workflow caused by Compose 2.19.0 that prevented services to be restarted
* Sets the home directory of the `archivematica` user and the `pyenv` root directory outside of the `/home` directory to avoid exposing files to the external volumes
* Sets permissions on test databases only instead of creating them on each test run which integrates better with `pytest-django`

Connected to https://github.com/archivematica/Issues/issues/1613

### DIFF
--- a/.github/workflows/acceptance-test.yml
+++ b/.github/workflows/acceptance-test.yml
@@ -33,13 +33,6 @@ jobs:
         uses: "actions/checkout@v3"
         with:
           submodules: true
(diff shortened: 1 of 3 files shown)


sevein (approved):
I ran into the docker compose restart issue yesterday, thanks for the fix! :+1:

PR #1841, opened 2023-08-31
Remove django warnings in development environment

Dhwaniartefact:
Remove Django 3.2 warnings created during "make bootstrap" in development environment. 

**warning:**  
`fpr.fpcommand: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.`

Reference: https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-DEFAULT_AUTO_FIELD

### DIFF
--- a/src/dashboard/src/components/mcp/apps.py
+++ b/src/dashboard/src/components/mcp/apps.py
@@ -1,6 +1,0 @@
-from django.apps import AppConfig
-
-
-class McpAppConfig(AppConfig):
-    default_auto_field = "django.db.models.AutoField"
-    name = "components.mcp"
--- a/src/dashboard/src/components/mcp/apps.py
+++ b/src/dashboard/src/components/mcp/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class McpAppConfig(AppConfig):
+    default_auto_field = "django.db.models.AutoField"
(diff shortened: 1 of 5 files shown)


src/dashboard/src/fpr/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class FprAppConfig(AppConfig):
```
replaceafill:
```suggestion
class FPRAppConfig(AppConfig):
```

src/dashboard/src/components/mcp/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
```
replaceafill: `components.mcp` doesn't have any models or signals, and I don't think we'll ever add any. I think we can remove this one.

replaceafill (approved):
Looks fine @Dhwaniartefact. I don't think we need the `McpAppConfig`.

src/dashboard/src/components/mcp/apps.py
Dhwaniartefact: Thank you for your suggestion.

replaceafill (approved):

PR #1853, opened 2023-09-28
Add small workflow schema improvements

sevein:
I'm submitting a couple of fixes to the workflow schema:
1. Exclude null values
2. Fix validation of "chain_choices" and "replacements" objects


replaceafill (approved):
Thanks!

PR #1862, opened 2023-10-16
Dev docs: complete the installation steps by accessing Archivematica

eviau-artefactual:
As discussed, we can help the reader find the next step after installation.


replaceafill (approved):
Thank you @eviau-artefactual!

Were you also planning to remove the stray `$` prompt symbols in the `Source code auto-reloading` section?

PR #1868, opened 2023-11-02
Remove vcrpy

replaceafill:
`vcrpy` 4.3.1 [started pinning its `urllib3` dependency](https://github.com/kevin1024/vcrpy/blob/0e068369084f393f290b7b0a87ddd06f073058ab/setup.py#L50-L56) for Python versions lower than 3.10. Besides from preventing us from upgrading `urllib3` ourselves, this also makes `pip-compile` to generate production and development requirements files with different versions.

This PR removes the `vcrpy` dependency and its yaml cassette files and replace them with mocks. Some mock set ups and assertions are really verbose because they were generated from the cassette interactions, but I plan to trim them down to contain only the fields needed by each test.


replaceafill:
I've included this change in https://github.com/artefactual/archivematica/pull/1867 to fix a problem with `urllib3` created by `vcrpy`.

PR #1873, opened 2023-11-17
Fix rejected directory settings

replaceafill:
The rejected directory settings are the only ones that use an interpolated variable (i.e. `%sharedPath%`). Their `%%` part was escaped by the old SafeConfigParser and converted into a single `%` making them work in the MCPServer.

This broke when the config parser base was replaced with the RawConfigParser which doesn't escape the double `%%` anymore making the MCPServer to interpolate the value as `%/var/archivematica/sharedDirectory/rejected%`.

We could rewrite the default value of the settings to be `rejectedDirectory = %sharedPath%rejected` and that would work fine, but the rest of the shared directory related settings don't use any interpolated variables. They point to the absolute path of the shared directory.

So, this replaces the `%%sharedPath%%` part of the rejected directory settings with the absolute path to the shared directory to make them match with the rest.

Connected to https://github.com/archivematica/Issues/issues/1636


sevein (approved):
LGTM. Perhaps the change should be mentioned in the release notes.

PR #1874, opened 2023-11-20
Upgrade Frontend dependencies

Dhwaniartefact:
Frontend dependencies are upgraded by removing yarn packages and installing npm version 9. 

`overrides` feature of npm is used to upgrade specific version of `devDependencies` and `dependencies`. 

Reference : https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides


hack/Dockerfile
```diff
@@ -51,25 +51,18 @@ RUN set -ex \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy multiverse" \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy-security universe" \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy-updates multiverse" \
+    && curl -sS https://deb.nodesource.com/setup_14.x | bash - \
```
replaceafill: Should this and line 138 below use tabs instead of spaces?

hack/Dockerfile
```diff
@@ -141,13 +134,11 @@ RUN set -ex \
 		/src/src/dashboard/src/media \
 	' \
 	&& mkdir -p $internalDirs \
-	&& chown -R archivematica $internalDirs
+	&& chown -R archivematica $internalDirs \
+    && npm install -g npm@9.8.1 /src/src/dashboard/frontend
```
replaceafill: I presume this has the same effect of the `yarn ... install ...` line below? If that's the case the output of this command will be owned by `root` and not by the `archivematica` user.

src/dashboard/frontend/package.json
```diff
@@ -51,8 +51,16 @@
...
-  }
+  },
+   "overrides": {
+    "fsevents": "1.2.13",
+    "js-yaml": "3.13.1",
+    "json5": "1.0.2",
+    "webpack": {
+      "loader-utils": "1.4.2"
+    }
+    }
```
replaceafill: Indentation seems a bit off?

replaceafill: I wonder if we should add this file to the pre-commit hook that formats other files in the repo? https://github.com/artefactual/archivematica/blob/155f229fa7a15cc7f01a57eb5a9fc0335a20f5c8/.pre-commit-config.yaml#L7-L11

hack/Dockerfile
Dhwaniartefact: oops. That's something due to editor. Let me fix it.

hack/Dockerfile
```diff
@@ -141,13 +134,11 @@ RUN set -ex \
...
 
 USER archivematica
 
-RUN set -ex \
-	&& yarn --cwd=/src/src/dashboard/frontend install --frozen-lockfile
```
replaceafill, line 149: If this is removed, should we also delete the `yarn.lock` file?

src/dashboard/frontend/package.json
replaceafill: I tested this and the hook works as expected but I'm hesitant about its default behavior of sorting mapping keys. It also has a couple of [arguments](https://github.com/pre-commit/pre-commit-hooks#pretty-format-json) that could be useful: `--no-sort-keys` to get the file formatted with no key reordering and `--top-keys` which allows you to define keys you want to keep in place. We could use the latter for the metadata keys of the package such as `name`, `description` and `version`. In any case we can discuss it after this PR gets merged :slightly_smiling_face:

hack/Dockerfile
Dhwaniartefact: Thanks and done.

hack/Dockerfile
Dhwaniartefact: Yes it is removed in next commit. Thanks for reminding me.

hack/Dockerfile
Dhwaniartefact: Thanks for suggestions. I have installed npm globally by root user separately.

src/dashboard/frontend/package.json
```diff
@@ -51,8 +51,16 @@
...
     "moment": "^2.21.0",
     "restangular": "^1.5.1"
+  },
+  "overrides": {
+    "fsevents": "1.2.13",
+    "js-yaml": "3.13.1",
+    "json5": "1.0.2",
+    "webpack": {
+      "loader-utils": "1.4.2"
+    }
```
sevein, line 64: I didn't know overrides can do this, cool.

hack/Dockerfile
```diff
@@ -51,25 +51,18 @@ RUN set -ex \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy multiverse" \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy-security universe" \
 	&& add-apt-repository --no-update --yes "deb http://archive.ubuntu.com/ubuntu/ jammy-updates multiverse" \
+	&& curl -sS https://deb.nodesource.com/setup_14.x | bash - \
```
sevein, line 54: Can we use 18.x or 20.x? Related: https://endoflife.date/nodejs.

hack/Dockerfile
```diff
@@ -143,10 +136,15 @@ RUN set -ex \
 	&& mkdir -p $internalDirs \
 	&& chown -R archivematica $internalDirs
 
+RUN set ex  \
+    && npm install -g npm@9.8.1
```
sevein, line 140: Why 9.8.1 and not the latest 9.x, e.g. 9.9.2?

sevein (approved):
I haven't tested but this looks good to me. I defer to Douglas for the final decision. It would be great if we can use the latest LTS available.

hack/Dockerfile
Dhwaniartefact: Thank you and noted all your suggestions.  I made changes accordingly in https://github.com/artefactual/archivematica/pull/1875

Dhwaniartefact:
Closing this PR since npm and node are upgraded in https://github.com/artefactual/archivematica/pull/1875

PR #1876, opened 2023-11-29
Update contributing guide

replaceafill:
This updates the contributing guide to reflect our current development process better and address some common problems:

* GitHub automatically uses the `stable/*` branch for new pull requests. This instructs contributors to use the `qa/*` branch instead which is the focus of our development.
* We usually request new contributors to sign the contributor agreement before engaging in the code review
* We have introduced `pre-commit` for formatting and linting our code which replaces the need to use different formatting and linting tools like `pylint` separately. `pre-commit` is available for developers in the Docker development environment and enabled as a continuous integration job.
* Archivematica dropped Python 2.7 compatibility a couple of versions ago.
* We prefer not to add more tests based on Django's `TestCase` anymore since their fixture system is less flexible than `pytest`.
* Maintaining `pyvcr` fixtures is problematic as well and recent versions have introduced a [compatibility problem](https://github.com/artefactual/archivematica/pull/1868) for us.

The code review guidelines document has also been updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,8 +7,8 @@ us assess your changes faster and makes it easier for us to merge your
 submission!
 
 There are many ways to contribute: writing tutorials or blog posts about your
-experience, improving the [documentation], submitting bug reports, answering 
-questions on the [mailing list], or writing code which can be incorporated into 
+experience, improving the [documentation], submitting bug reports, answering
+questions on the [mailing list], or writing code which can be incorporated into
 Archivematica itself.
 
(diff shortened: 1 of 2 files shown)


eviau-artefactual:
Thanks for keeping me in the loop - this looks good to me.

eviau-artefactual (commented):
The one thing I would add is links to any relevant PR, Issue or something else - e.g. I see that we are not linking to VCR.py anymore, but maybe in the future we would like to know why this has been removed.

replaceafill:
Thank you @eviau-artefactual! That's a great point. Apologies for not being explicit. I've explained a few of the changes in the pull request description.

PR #1880, opened 2023-12-14
Use TMPDIR environment variable for temporary file path in createAndRunScript

camlyall:
Avoid using hardcoded temp directory.

This fixes issue https://github.com/archivematica/Issues/issues/1425

Users facing the issue 7z when mounting /tmp directory with noexec can set the environment variable `TMPDIR=/path/to/dir` in /etc/default/archivematica-mcp-client


mamedin:
Reviewed in AM1.14.

* Transfer with compression
* Reingest
* Fixity run

Looks fine!

PR #1882, opened 2023-12-15
Use TMPDIR environment variable for temporary file path in createAndRunScript

camlyall:
Fixes use of hardcoded temp directory.
This fixes issue [archivematica/Issues#1425](https://github.com/archivematica/Issues/issues/1425)

Reviewed in AM1.14
- Transfer with compression
- Reingest
- Fixity run

Users facing the issue 7z when mounting /tmp directory with noexec can set the environment variable `TMPDIR=/path/to/dir` in /etc/default/archivematica-mcp-client

### DIFF
--- a/src/archivematicaCommon/lib/executeOrRunSubProcess.py
+++ b/src/archivematicaCommon/lib/executeOrRunSubProcess.py
@@ -147,11 +147,11 @@ def createAndRunScript(
         env_updates = {}
     # Output the text to a temporary file
     tmpdir_path = os.environ.get('TMPDIR', '/tmp')
-    scriptPath =  tmpdir_path + '/' + uuid.uuid4().__str__()()
-    FILE = os.open(scriptPath, os.O_WRONLY | os.O_CREAT, 0o770)
+    script_path =  tmpdir_path + '/' + uuid.uuid4().__str__()()
+    FILE = os.open(script_path, os.O_WRONLY | os.O_CREAT, 0o770)
     os.write(FILE, text.encode("utf8"))
     os.close(FILE)
-    cmd = [scriptPath]
+    cmd = [script_path]
     cmd.extend(arguments)
 
     # Run it
@@ -164,7 +164,7 @@ def createAndRunScript(
     )
 
     # Remove the temp file
-    os.remove(scriptPath)
+    os.remove(script_path)
 
     return ret
 
--- a/src/archivematicaCommon/lib/executeOrRunSubProcess.py
+++ b/src/archivematicaCommon/lib/executeOrRunSubProcess.py
@@ -147,7 +147,7 @@ def createAndRunScript(
         env_updates = {}
     # Output the text to a temporary file
     tmpdir_path = os.environ.get('TMPDIR', '/tmp')
-    script_path =  tmpdir_path + '/' + uuid.uuid4().__str__()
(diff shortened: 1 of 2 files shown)


src/archivematicaCommon/lib/executeOrRunSubProcess.py
```diff
@@ -145,12 +145,13 @@ def createAndRunScript(
         arguments = []
     if env_updates is None:
         env_updates = {}
-    # Output the text to a /tmp/ file
-    scriptPath = "/tmp/" + uuid.uuid4().__str__()
-    FILE = os.open(scriptPath, os.O_WRONLY | os.O_CREAT, 0o770)
+    # Output the text to a temporary file
+    tmpdir_path = os.environ.get("TMPDIR", "/tmp")
```
replaceafill: What do you think of using `tempfile` here which handles the `TMPDIR`, `TEMP` and `TMP` environment variables transparently. See how [this function](https://github.com/artefactual-labs/a3m/blob/8b7a7c59f804edd6484a7c6a4619818efb1b0d66/a3m/executeOrRunSubProcess.py#L147-L176) is implemented in a3m using a context manager which deletes the file automatically at the end.

replaceafill (commented):
Thanks @camlyall, this looks good. I wonder if we could improve that function a bit and since it lacks test coverage at the moment add a test for it. Maybe we could use `pytest` [`monkeypatch.setenv`](https://docs.pytest.org/en/latest/how-to/monkeypatch.html#how-to-monkeypatch-mock-modules-and-environments) and [`tmp_path`](https://docs.pytest.org/en/latest/how-to/tmp_path.html) to test the `TMPDIR` case.

replaceafill:
@camlyall Out of curiosity, are you planning to address this in the Storage Service which I think has a [similar problem when compressing packages](https://github.com/artefactual/archivematica-storage-service/blob/6f67c1396176da34c84de98ea00a51c6af54d818/storage_service/locations/models/package.py#L1717)?

src/archivematicaCommon/lib/executeOrRunSubProcess.py
camlyall: thanks for the suggestion @replaceafill. I have updated the code to use that library

src/archivematicaCommon/lib/executeOrRunSubProcess.py
```diff
@@ -145,26 +145,25 @@ def createAndRunScript(
...
-        stdIn="",
-        printing=printing,
-        env_updates=env_updates,
-        capture_output=capture_output,
-    )
-
-    # Remove the temp file
-    os.remove(scriptPath)
-
+    # Output the text to a temporary file
```
replaceafill: Could we move this comment below? Maybe above the `tmpfile.write(text)` line?

replaceafill (approved):
Thank you @camlyall! Great work.

src/archivematicaCommon/lib/executeOrRunSubProcess.py
camlyall: Alternatively I could update the comment to `Write script to temporary file and execute it`

replaceafill: That works. Thanks.

PR #1895, opened 2024-02-07
Remove osdeps directories

Dhwaniartefact:
Removal of osdeps directory since it became unwanted after Archivematica 1.15

Connected to : https://github.com/archivematica/Issues/issues/1659


hack/Dockerfile
```diff
@@ -12,24 +12,55 @@ ENV PYTHONUNBUFFERED 1
...
+		logapp \
+		md5deep \
+		mediaconch \
+		mediainfo \
+		nfs-common \
+		nginx \
+		openjdk-8-jre-headless \
+		p7zip-full \
+		pbzip2 \
+		postfix \
```
replaceafill: Why are `ngnix` and `postfix` included? They were filtered out when `osdeps.py` was used:

```
/src/hack/osdeps.py Ubuntu-22 1 | grep -v -E "nginx|postfix|python3.9-dev"
```

hack/Dockerfile
```diff
@@ -56,8 +82,7 @@ RUN set -ex \
 	&& dpkg -i /tmp/repo-mediaarea_1.0-21_all.deb \
 	&& apt-get update \
 	&& apt-get install -y --no-install-recommends \
-		libyaml-dev clamav nodejs \
-	&& /src/hack/osdeps.py Ubuntu-22 1 | grep -v -E "nginx|postfix|python3.9-dev" | xargs apt-get install -y --no-install-recommends \
+		bulk-extractor siegfried fits ffmpeg jhove nailgun clamav nodejs \
```
replaceafill: Why are some packages placed here and some at the top?

hack/Dockerfile
```diff
@@ -12,24 +12,55 @@ ENV PYTHONUNBUFFERED 1
...
 		libsqlite3-dev \
 		libssl-dev \
 		libxml2-dev \
 		libxmlsec1-dev \
+		libxslt1-dev \
 		locales \
+		logapp \
+		md5deep \
+		mediaconch \
+		mediainfo \
```
replaceafill: Shouldn't these be placed **after** the `repo-mediaarea` package has been installed?

replaceafill (changes requested):
Good progress @Dhwaniartefact. If the packages you're adding represent Archivematica specific dependencies I think having them split in two `apt-get install` instructions is confusing.

I wonder if we should move all those new packages to the second `apt-get install` instruction where the `osdeps.py` script was called. Alternatively, we could have a single/combined instruction at the top with everything in it, but I'm not sure how that would affect other instructions.

Dhwaniartefact:
> Good progress @Dhwaniartefact. If the packages you're adding represent Archivematica specific dependencies I think having them split in two `apt-get install` instructions is confusing.
> 
> I wonder if we should move all those new packages to the second `apt-get install` instruction where the `osdeps.py` script was called. Alternatively, we could have a single/combined instruction at the top with everything in it, but I'm not sure how that would affect other instructions.

By placing packages like sigfried, jhove etc on top `apt-get install` is creating building error. But let me place Archivematica specific dependencies at secont `apt-get install` instruction.

hack/Dockerfile
Dhwaniartefact: oh ya!

hack/Dockerfile
```diff
@@ -56,8 +51,48 @@ RUN set -ex \
...
+		pst-utils \
+		rsync \
+		siegfried \
+		sleuthkit \
+		tesseract-ocr \
+		tree \
+		unar \
+		unrar-free \
+		uuid \
+		clamav nodejs \
```
replaceafill: Could we split each package separately and sort them? `clamav` and `nodejs` are two different packages.

There are also a few that are repeated from the top instruction like `gettext`, `libssl-dev`, etc. Should those be removed from this second instruction? Maybe combining both instructions makes sense.

replaceafill (approved):

hack/Dockerfile
Dhwaniartefact: Thanks for the suggestions. Done

PR #1897, opened 2024-02-16
Make uploadedDIPs a real watched directory

replaceafill:
Even though the `watchedDirectories/uploadedDIPs` directory already exists in the system it lacks the necessary `workflow.json` definition to start any processing from it, i.e. content put there doesn't make the MCPServer to "react". This is how that part of the workflow looks right now in `amflow`:

imagen

This PR adds that mentioned definition for the directory and modifies the `Move to the uploadedDIPs directory` job to move the DIP instead of copying and to complete the `uploadDIP` chain. Processing will continue to the same next job (`Store DIP?`) from the new introduced chain.

And this is how it will look now:

imagen

The following jobs have been adjusted for this change too:

* The `Store DIP` job doesn't contain logic for deleting watched directories anymore. Instead the right DIP directory is passed as job argument from the workflow definition. And a new job has been added to remove the DIP copy at the end of the chain.
* The job to handle unstored dips has been replaced with a job that only moves the DIP directory to the `rejected` directory. This rejected copy is prefixed with `DIP-` to avoid name clashes with rejected AIPs and the rejection related labels of the workflow jobs have been updated to be similar to the AIP workflow.

This is how that area of the workflow looks now:

imagen

https://github.com/archivematica/Issues/issues/1665


sevein (approved):
Nice finding, Douglas! :heart_eyes:

Dhwaniartefact (approved):
Finally done. Awesome !

PR #1902, opened 2024-02-23
Restructure unit tests

replaceafill:
A long term goal of the project is to adopt a Python package [src layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) and stop customizing and injecting `PYTHONPATH` environment variables.

This moves all the unit tests to a top level `tests` directory as an initial step in that direction.


sevein (approved):
Nice! :clap:

PR #1906, opened 2024-02-28
Fix script to capture mutable globals

replaceafill:
This updates the script that ensures no mutable globals are used in client scripts. It now uses [dis.get_instructions](https://docs.python.org/3/library/dis.html#dis.get_instructions) to iterate through the global operations instead of checking bytecode offsets.


sevein (approved):
:+1:

PR #1907, opened 2024-02-28
Upgrade frontend dependencies

replaceafill:
This upgrades the configuration for `webpack` following these migration guides:
* https://webpack.js.org/migrate/3/
* https://webpack.js.org/migrate/4/
* https://webpack.js.org/migrate/5/

It also removes the vendored apps `angulare-tree-control` and `angular-ui-bootstrap` which are now installed from the `npm` registry instead.


sevein (approved):
Nice job! :+1:

PR #1908, opened 2024-03-03
Convert script to capture mutable globals into test

replaceafill:
This converts the script into an MCPClient test so its separate `tox` environment and `make` rule can be removed.


sevein (approved):
Nice idea!

PR #1917, opened 2024-03-14
Add CI action to push images to Docker Hub

DanielCosme:
This PR creates a Github Action that builds and push docker images to Docker Hub. For now the trigger to this workflow will be manual and will be tagged only with `latest`. We expect to do a second pass on this workflow so that we have multiple tags that relate to different versions of AM. We will do this once we have a better understanding on how a dockerised production grade deployment of AM looks like.

Also, the Dockerfile has been reworked a little bit to make the resulting images smaller in the registry. We did this by only using dependencies that are strictly needed for all components of AM. Before all 3 components shared dependencies that where not really needed.

I expect to do an additional pass on this file in the future, once I have a better understanding on how the dependencies work. As for the `archivematica-tests` target I left it as it was originally.


.github/workflows/push-images.yml
```diff
@@ -0,0 +1,46 @@
+name: "Push images to Docker Hub"
+on: workflow_dispatch
+jobs:
+  build:
+    name: "Builds images and push them to Docker Hub"
```
replaceafill:
```suggestion
    name: "Build and push images"
```

.github/workflows/push-images.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: "ubuntu-22.04"
+    steps:
+      - name: "Check out repository"
+        uses: "actions/checkout@v4"
+      - name: "Set up buildx"
+        uses: "docker/setup-buildx-action@v3"
+        with:
+          version: latest
+          driver-opts: image=moby/buildkit:v0.13.0
```
replaceafill, line 14: I noticed the Storage Service [related change](https://github.com/artefactual/archivematica-storage-service/blob/dd7b6b288173a1d8117993479fd13c6cbc8ea24a/.github/workflows/push-images.yml#L10-L11) doesn't include this. Why is it necessary for the Archivematica images?

.github/workflows/push-images.yml
```diff
@@ -0,0 +1,46 @@
...
+          tags: artefactual/archivematica-mcp-client:latest
+      - name: "Build and Push MCP-server"
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          load: true
+          file: ./hack/Dockerfile
+          target: "archivematica-mcp-server"
+          tags: artefactual/archivematica-mcp-server:latest
```
replaceafill: Could we add a newline at the end of the file?

hack/Dockerfile
```diff
@@ -150,6 +150,35 @@ ARG USER_ID=1000
 ARG GROUP_ID=1000
 ARG PYENV_DIR=/pyenv
 
+RUN set -ex \
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		coreutils \
+		clamav \
+		libldap2-dev \
+		libsasl2-dev \
```
replaceafill: Could these be removed since they're already installed in the `base-builder`?

hack/Dockerfile
```diff
@@ -150,6 +150,35 @@ ARG USER_ID=1000
 ARG GROUP_ID=1000
 ARG PYENV_DIR=/pyenv
 
+RUN set -ex \
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		coreutils \
+		clamav \
```
replaceafill: I might be wrong, but I think the `clamav` package and the:

```
# Download ClamAV virus signatures
RUN freshclam --quiet
```

instruction below are only necessary in the MCPClient stage.

hack/Dockerfile
```diff
@@ -150,6 +150,35 @@ ARG USER_ID=1000
 ARG GROUP_ID=1000
 ARG PYENV_DIR=/pyenv
 
+RUN set -ex \
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		coreutils \
```
replaceafill: I noticed the [ansible-archivematica-src vars](https://github.com/artefactual-labs/ansible-archivematica-src/blob/62356e1923fd7c1293391244f1befe42d737ccf8/vars/Ubuntu-22.yml#L43) set this as a dashboard dependency. Should it be moved there?

hack/Dockerfile
```diff
@@ -283,7 +296,75 @@ ENTRYPOINT ["pyenv", "exec", "python3", "-m", "gunicorn", "--config=/src/src/das
 
 # -----------------------------------------------------------------------------
 
-FROM base AS archivematica-tests
+FROM base-builder as archivematica-tests
```
replaceafill, line 326: I think the `archivematica-tests` stage can be much more simple. It's needed for running `tox` and building its virtual environments, so it only needs Python, the common development libraries `base-builder` already has and from a quick local test based on the `qa/1.x` branch the following dependencies:

* `gcc` for building C dependencies for the virtual environment.
* `media-types`, `p7zip-full`, `rsync` and `unar` for being able to run the Storage Service tests from the `hack/Makefile`.

hack/Dockerfile
```diff
@@ -283,7 +296,75 @@ ENTRYPOINT ["pyenv", "exec", "python3", "-m", "gunicorn", "--config=/src/src/das
...
+	&& mkdir -p /var/archivematica/sharedDirectory \
+	&& chown -R archivematica:archivematica /var/archivematica
+
+# Download ClamAV virus signatures
+RUN freshclam --quiet
+
+USER archivematica
+
+COPY --chown=${USER_ID}:${GROUP_ID} --from=pyenv-builder --link ${PYENV_DIR} ${PYENV_DIR}
+COPY --chown=${USER_ID}:${GROUP_ID} --link . /src
```
replaceafill: Not sure if it makes sense, but I think if you end up moving all the OS dependencies out of `base` to their corresponding stages, this block might be inherited from the `base` stage instead.

replaceafill (commented):
@DanielCosme great progress! The new workflow for publishing images looks good to me. I haven't tested the `Dockerfile` in the PR yet, but I left a few questions and suggestions.

.github/workflows/push-images.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: "Build and Push Dashboard"
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          load: true
+          file: ./hack/Dockerfile
+          target: "archivematica-dashboard"
+          tags: artefactual/archivematica-dashboard:latest
+      - name: "Build and Push MCP-client"
```
sevein:
```suggestion
      - name: "Build and push MCPClient"
```

.github/workflows/push-images.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          load: true
+          file: ./hack/Dockerfile
+          target: "archivematica-mcp-client"
+          tags: artefactual/archivematica-mcp-client:latest
+      - name: "Build and Push MCP-server"
```
sevein:
```suggestion
      - name: "Build and push MCPServer"
```

hack/Dockerfile
```diff
@@ -283,7 +296,75 @@ ENTRYPOINT ["pyenv", "exec", "python3", "-m", "gunicorn", "--config=/src/src/das
 
 # -----------------------------------------------------------------------------
 
-FROM base AS archivematica-tests
+FROM base-builder as archivematica-tests
+
+ARG USER_ID=1000
+ARG GROUP_ID=1000
+ARG PYENV_DIR=/pyenv
```
sevein: According to the docs, we don't need to redefine the default again when pulling the argument, e.g.:

```suggestion
ARG USER_ID
ARG GROUP_ID
ARG PYENV_DIR
```

.github/workflows/push-images.yml
DanielCosme: Because the environment in which this is executed uses by default buildkit `v0.12` and we need `v0.13`+ in order for the multitarget feature to work. So we have to directly specify the version.

hack/Dockerfile
DanielCosme: I'll try it.

hack/Dockerfile
DanielCosme: I remember. I decided to leave `coreutils` here just because I don't understand the components and dependencies enough. coreutils contains (sometimes) essential packages for the OS. And since AM is so reliant on OS dependencies I left it in the `base-builder` step. Should I remove it?

hack/Dockerfile
DanielCosme: So what I think you are saying is that this block is not really needed because it already happens in `base`. So by having this block here is just redundant, doing the same work again, right?

replaceafill: That's correct.

hack/Dockerfile
replaceafill: I think you can try moving `coreutils` to the dashboard stage and let the Acceptance Test workflow to tell you if anything breaks.

.github/workflows/push-images.yml
DanielCosme: Done.

hack/Dockerfile
DanielCosme: Done

hack/Dockerfile
DanielCosme: Done. Nothing broke in the tests.

hack/Dockerfile
DanielCosme: `coreutils` removed. nothing broke :)

hack/Dockerfile
DanielCosme: Yeah, it works, thank you. Done.

hack/Dockerfile
DanielCosme: Done.

.github/workflows/push-images.yml
DanielCosme: Done.

.github/workflows/push-images.yml
DanielCosme: Done.

hack/Dockerfile
DanielCosme: Thanks!

.github/workflows/push-images.yml
DanielCosme: Done.

DanielCosme:
@replaceafill I think I addressed all feedback. All tests are passing, if you have any further feedback, let me know.

replaceafill:
Closing this after we decided to approach it in two separate pull requests:
* Add a GitHub workflow to publish images to Docker Hub. See https://github.com/artefactual/archivematica/pull/2070
* Make the Docker image sizes as small as possible

PR #1924, opened 2024-04-15
Generate SBOM in CI

sbreker:
Add ability to generate and upload an SBOM to Dependency Track in CI.
    
This CI workflow will autocreate a new Dependency Track project for any
qa or stable branch this workflow is picked to. The projectName
will be set to "archivematica" and the project version will be the branch
name.

Because the parentName is set to 'Archivematica', this SBOM will be grouped under
a project group of the same name which must be present in advance.
    
Dev dependencies will be included in SBOM.

### DIFF
--- a/.github/workflows/generate-sbom.yml
+++ b/.github/workflows/generate-sbom.yml
@@ -1,0 +1,59 @@
+name: Generate and upload SBOM
+
+on:
+  push:
+    branches:
+    - qa/**
+    - stable/**
+
+jobs:
+  generate-sbom:
+    runs-on: ubuntu-latest
+    container:
+      image: aquasec/trivy:latest
+      options: --entrypoint ""
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        
+      - name: Set up cache
+        uses: actions/cache@v4
+        with:
+          path: .trivycache/
+          key: ${{ runner.os }}-trivy-${{ hashFiles('requirements*.txt', 'src/dashboard/frontend/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-trivy-
+
+      - name: Generate SBOM
+        run: trivy fs --cache-dir .trivycache --format cyclonedx --file-patterns "pip:requirements.*\.txt" --include-dev-deps --output sbom.xml .
+        env:
+          TRIVY_NO_PROGRESS: "true"
+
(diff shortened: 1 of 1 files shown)


sbreker:
Sample SBOM analysis for the `dev/produce-sbom` branch can be viewed here:

https://dependency-track.pet.artefactual.dev/projects/7e078039-e97e-4641-9292-c4844639b1ee

image

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,59 @@
...
+      options: --entrypoint ""
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        
+      - name: Set up cache
+        uses: actions/cache@v4
+        with:
+          path: .trivycache/
+          key: ${{ runner.os }}-trivy-${{ hashFiles('**/lockfiles') }}
```
replaceafill: Is this `**/lockfiles` path supposed to exist in the repository?

sbreker: Ah good catch! I've fixed that to refer to the lockfiles - if they change the cache will be invalidated.

.github/workflows/generate-sbom.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Set up cache
+        uses: actions/cache@v4
+        with:
+          path: .trivycache/
+          key: ${{ runner.os }}-trivy-${{ hashFiles('requirements*.txt', 'src/dashboard/frontend/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-trivy-
+
+      - name: Generate SBOM
+        run: trivy fs --cache-dir .trivycache --format cyclonedx --file-patterns "pip:requirements.*\.txt" --include-dev-deps --output sbom.xml .
```
replaceafill, line 28: Does `trivy` look for the `package-lock.json` file automatically without specifying it in the `--file-patterns` flag?

replaceafill (approved):
Thank you @sbreker.

.github/workflows/generate-sbom.yml
sbreker: Yes it will still check package-lock.json - the --file-patterns option only affects the pip file matching pattern (you can see the prefix 'pip' in the option payload). It will still scan for all other manifest file types.

PR #1928, opened 2024-04-21
Remove dead code in MCPServer

sevein:
There's no rush to merge this, it's just the removal of some dead code.

PR #1929, opened 2024-04-22
Remove unused file filter

sevein:
Similar to https://github.com/artefactual/archivematica/pull/1928, no rush on this. `filter_file_start` is unused so there is no point in keeping any related logic unless you have short-term plans to implement it.


replaceafill (approved):
No short term plans to implement it. We can reintroduce it if needed. Thanks!

PR #1930, opened 2024-04-23
Use json.loads in OutputClientScriptJob

sevein:
What [get_aip_storage_locations](https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPClient/lib/clientScripts/get_aip_storage_locations.py) writes to stdout is a JSON blob so we should use the `json` decoder to deserialize it. The payload looks like the following, always hitting the Storage Service API to build the full list of locations:

```json
{
  "default": {"description": "Default Location", "uri": "/api/v2/location/default/AS/"},
  "3191bf34-c803-4bdf-a37a-cf1aaf1364d9": {"description": "...", "uri": "..."}
}
```

This is done twice in the standard Archivematica workflow, i.e. for AIP (AS) and DIP (DS) store locations.

### DIFF
--- a/src/MCPServer/lib/server/jobs/client.py
+++ b/src/MCPServer/lib/server/jobs/client.py
@@ -3,6 +3,7 @@ Jobs remotely executed by on MCP client.
 """
 import abc
 import ast
+import json
 import logging
 
 from dbconns import auto_close_old_connections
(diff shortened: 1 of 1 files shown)


replaceafill (approved):
Nice, thank you!

PR #1932, opened 2024-04-23
Extend test coverage of dashboard app components.file

Dhwaniartefact:
Extend test coverage of components `file/views.py` in dashboard app

### DIFF
--- a/tests/dashboard/components/file/test_views.py
+++ b/tests/dashboard/components/file/test_views.py
@@ -1,0 +1,403 @@
+import json
+import uuid
+from unittest import mock
+
+import elasticSearchFunctions
+import pytest
+from components import helpers
+from django.urls import reverse
+from main import models
+
+
+@pytest.fixture
+def dashboard_uuid(db):
+    helpers.set_setting("dashboard_uuid", str(uuid.uuid4()))
+
+
+@pytest.fixture
+def transfer(db):
+    return models.Transfer.objects.create()
+
+
+@pytest.fixture
+def sip(db):
+    return models.SIP.objects.create()
+
+
+@pytest.fixture
+def file(db):
+    return models.File.objects.create()
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+def test_file_details(_get_client, admin_client, dashboard_uuid, file, sip, mocker):
+    mocker.patch(
+        "elasticSearchFunctions.get_transfer_file_info",
+        return_value={
+            "filename": "LICENSE",
+            "fileuuid": str(file.uuid),
+            "sipuuid": str(sip.uuid),
+            "status": "backlog",
+            "size": 0.032919883728027344,
+            "tags": [],
+            "bulk_extractor_reports": [],
+            "format": [
+                {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+            ],
+            "pending_deletion": False,
+        },
+    )
+    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
+    assert response.status_code == 200
+
+    content = json.loads(response.content.decode("utf8"))
+    assert content["id"] == str(file.uuid)
+    assert content["title"] == "LICENSE"
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+@pytest.mark.parametrize(
+    "mock_file_info, status_code",
+    [
+        ("get_transfer_file_info returned no exact results", 404),
+        ("unknown error", 500),
+    ],
+    ids=["no exact results", "unknown error"],
+)
+def test_file_details_handles_exceptions(
+    _get_client, admin_client, dashboard_uuid, file, mocker, mock_file_info, status_code
+):
+    mocker.patch(
+        "elasticSearchFunctions.get_transfer_file_info",
+        side_effect=elasticSearchFunctions.ElasticsearchError(mock_file_info),
+    )
+
+    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
+    assert response.status_code == status_code
+
+    assert json.loads(response.content.decode("utf8")) == {
+        "success": False,
+        "message": mock_file_info,
+    }
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch(
+    "elasticSearchFunctions.get_transfer_file_info",
+    return_value={
+        "filename": "piiTestDataCreditCardNumbers.txt",
+        "relative_path": "test-bulk-extract-a239e3e1-0391-46da-94d7-25a3e8509b45/data/objects/piiTestDataCreditCardNumbers.txt",
+        "status": "backlog",
+        "size": 0.00026416778564453125,
+        "tags": [],
+        "file_extension": "txt",
+        "bulk_extractor_reports": ["ccn"],
+        "format": [
+            {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+        ],
+    },
+)
+def test_bulk_extractor_if_missing_reports_exist(
+    _get_client, _get_transfer_file_info, admin_client, dashboard_uuid, file, sip
+):
+    response = admin_client.get(reverse("file:bulk_extractor", args=[file.uuid]))
+    assert response.status_code == 400
+
+    assert response.json() == {
+        "success": False,
+        "message": "Requested file is missing the following requested reports: pii",
+    }
+
+
+@mock.patch("components.file.views.len", side_effect=[0])
+def test_bulk_extractor_validates_reports_parameter_with_len_mock(
+    _len, admin_client, dashboard_uuid, file
+):
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == 400
+
+    assert json.loads(response.content.decode()) == {
+        "success": False,
+        "message": "No reports were requested.",
+    }
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+@pytest.mark.parametrize(
+    "mock_file_info, status_code",
+    [
+        ("get_transfer_file_info returned no exact results", 404),
+        ("unknown error", 500),
+    ],
+    ids=["no exact results", "unknown error"],
+)
+def test_bulk_extractor_handles_exceptions(
+    _get_client, admin_client, dashboard_uuid, file, mocker, mock_file_info, status_code
+):
+    mocker.patch(
+        "elasticSearchFunctions.get_transfer_file_info",
+        side_effect=elasticSearchFunctions.ElasticsearchError(mock_file_info),
+    )
+
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == status_code
+
+    assert json.loads(response.content.decode("utf8")) == {
+        "success": False,
+        "message": mock_file_info,
+    }
+
+
+@pytest.mark.django_db
+@mock.patch("elasticSearchFunctions.get_client")
+def test_bulk_extractor_status_code_200(
+    _get_client,
+    admin_client,
+    dashboard_uuid,
+    mocker,
+    transfer,
+    sip,
+):
+    file = models.File.objects.create(uuid=uuid.uuid4(), transfer=transfer, sip=sip)
+    mocker.patch(
+        "elasticSearchFunctions.get_transfer_file_info",
+        return_value={
+            "filename": "piiTestDataCreditCardNumbers.txt",
+            "fileuuid": str(file.uuid),
+            "sipuuid": str(sip.uuid),
+            "status": "backlog",
+            "created": 1713465340.0109284,
+            "size": 0.00026416778564453125,
+            "tags": [],
+            "file_extension": "txt",
+            "bulk_extractor_reports": ["ccn", "pii"],
+            "format": [
+                {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+            ],
+            "pending_deletion": False,
+        },
+    )
+    mocker.patch(
+        "requests.get",
+        return_value=mocker.Mock(
+            status_code=200,
+            text="""64	378282246310005	rican Express - 378282246310005 02. American E
+                                                          104	371449635398431	rican Express - 371449635398431 03. American E
+                                                          154	378734493671000	ess Corporate - 378734493671000 04. Australian
+                                                          197	5610591081018250	lian BankCard - 5610591081018250 05. Discover -
+                                                          230	6011000990139424	 05. Discover - 6011000990139424 06. JCB - 3566
+                                                          258	3566002020360505	9424 06. JCB - 3566002020360505 """,
+        ),
+    )
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == 200
+
+
+@pytest.mark.django_db
+@mock.patch("elasticSearchFunctions.get_client")
+def test_bulk_extractor_status_code_other_than_200(
+    _get_client, admin_client, dashboard_uuid, mocker, caplog, transfer, sip
+):
+    file = models.File.objects.create(uuid=uuid.uuid4(), transfer=transfer, sip=sip)
+    mocker.patch(
+        "elasticSearchFunctions.get_transfer_file_info",
+        return_value={
+            "filename": "piiTestDataCreditCardNumbers.txt",
+            "fileuuid": file.uuid,
+            "sipuuid": sip.uuid,
+            "status": "backlog",
+            "created": 1713465340.0109284,
+            "size": 0.00026416778564453125,
+            "tags": [],
+            "file_extension": "txt",
+            "bulk_extractor_reports": ["ccn", "pii"],
+            "format": [
+                {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+            ],
+            "pending_deletion": False,
+        },
+    )
+    mocker.patch(
+        "requests.get",
+        return_value=mocker.Mock(status_code=500, text="""Internal Server Error"""),
+    )
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == 200
+
+    assert (
+        f"Unable to retrieve ccn report for file with UUID {file.uuid}; response: ('Internal Server Error',)"
+        in ([rec.message for rec in caplog.records[1::2]])
+    )
+    assert f"Unable to retrieve ccn report for file with UUID {file.uuid}" in (
+        [(rec.message).split(";")[0] for rec in caplog.records[1::2]]
+    )
+    assert " response: ('Internal Server Error',)" in (
+        [(rec.message).split(";")[1] for rec in caplog.records[1::2]]
+    )
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+def test_transfer_file_tags(_get_client, admin_client, dashboard_uuid, file, mocker):
+    mocker.patch(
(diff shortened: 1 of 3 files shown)


tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+            ],
+            "pending_deletion": False,
+        },
+    )
+    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
+    assert response.status_code == 200
+
+    content = json.loads(response.content.decode("utf8"))
+    assert content["id"] == str(file.uuid)
+    assert content["title"] == "LICENSE"
```
replaceafill: Would it be possible to assert against the entire `content` mapping instead of just two keys, i.e. `assert content == {...}`? I find that helps you catch mistakes more often.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+@pytest.mark.parametrize(
+    "mock_file_info, status_code",
```
replaceafill: I noticed this is the content of an error message. What's the meaning of `mock_file_info`?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    [
+        ("get_transfer_file_info returned no exact results", 404),
+        ("unknown error", 500),
+    ],
+    ids=["no exact results", "unknown error"],
+)
+def test_file_details_handles_exceptions(
+    admin_client, dashboard_uuid, file, mocker, mock_file_info, status_code
```
replaceafill: Let's keep `parametrize` or `patch` decorators as the initial parameters of the test functions. Fixtures can come after.

```suggestion
    mock_file_info, status_code, admin_client, dashboard_uuid, file, mocker
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+            "size": 0.032919883728027344,
+            "tags": [],
+            "bulk_extractor_reports": [],
+            "format": [
+                {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+            ],
+            "pending_deletion": False,
+        },
+    )
+    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
```
replaceafill, line 52:
```suggestion

    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    assert response.status_code == status_code
+
+    assert json.loads(response.content.decode("utf8")) == {
+        "success": False,
+        "message": mock_file_info,
+    }
+
+
+def test_bulk_extractor_if_missing_reports_exist(
```
replaceafill: This name is not very clear. What is an example of a missing report that exists?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    assert response.status_code == 400
+
+    assert response.json() == {
+        "success": False,
+        "message": "Requested file is missing the following requested reports: pii",
+    }
+
+
+@mock.patch("components.file.views.len", side_effect=[0])
+def test_bulk_extractor_validates_reports_parameter_with_len_mock(
```
replaceafill: There's no need to mention the mechanics of the test in its name (suppose we need to change how the test works, we'd need to update the name). Names should reflect the purpose of the test not how it does it.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    assert response.status_code == 400
+
+    assert json.loads(response.content.decode()) == {
+        "success": False,
+        "message": "No reports were requested.",
+    }
+
+
+@pytest.mark.parametrize(
+    "mock_file_info, status_code",
```
replaceafill: Same comment about `mock_file_info` as above.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    assert response.status_code == status_code
+
+    assert json.loads(response.content.decode("utf8")) == {
+        "success": False,
+        "message": mock_file_info,
+    }
+
+
+@pytest.mark.django_db
+def test_bulk_extractor_status_code_200(
```
replaceafill: Do we need the `status_code_200` in this name?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    admin_client,
+    dashboard_uuid,
+    mocker,
+    transfer,
+    sip,
+):
+    file = models.File.objects.create(uuid=uuid.uuid4(), transfer=transfer, sip=sip)
```
replaceafill: I have not tested, but you could try reusing the `file` fixture here instead of creating a new file. Maybe something like this would work:

```suggestion
    file.transfer = transfer
    file.sip = sip
    file.save()
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+                                                          154	378734493671000	ess Corporate - 378734493671000 04. Australian
+                                                          197	5610591081018250	lian BankCard - 5610591081018250 05. Discover -
+                                                          230	6011000990139424	 05. Discover - 6011000990139424 06. JCB - 3566
+                                                          258	3566002020360505	9424 06. JCB - 3566002020360505 """,
+        ),
+    )
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == 200
```
replaceafill, line 215: Can something be asserted about the contents of the `response` here? What is the view supposed to return?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    mocker.patch(
+        "requests.get",
+        return_value=mocker.Mock(
+            status_code=200,
+            text="""64	378282246310005	rican Express - 378282246310005 02. American E
+                                                          104	371449635398431	rican Express - 371449635398431 03. American E
+                                                          154	378734493671000	ess Corporate - 378734493671000 04. Australian
+                                                          197	5610591081018250	lian BankCard - 5610591081018250 05. Discover -
+                                                          230	6011000990139424	 05. Discover - 6011000990139424 06. JCB - 3566
+                                                          258	3566002020360505	9424 06. JCB - 3566002020360505 """,
```
replaceafill: If you're not asserting anything about the `response` are all these lines necessary? Could a couple be enough?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+        ),
+    )
+    response = admin_client.get(
+        reverse("file:bulk_extractor", kwargs={"fileuuid": file.uuid})
+    )
+    assert response.status_code == 200
+
+
+@pytest.mark.django_db
+def test_bulk_extractor_status_code_other_than_200(
```
replaceafill: Rename this to reflect what it's doing/looking for/ensuring/etc.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+    assert (
+        f"Unable to retrieve ccn report for file with UUID {file.uuid}; response: ('Internal Server Error',)"
+        in ([rec.message for rec in caplog.records[1::2]])
+    )
+    assert f"Unable to retrieve ccn report for file with UUID {file.uuid}" in (
+        [(rec.message).split(";")[0] for rec in caplog.records[1::2]]
+    )
+    assert " response: ('Internal Server Error',)" in (
+        [(rec.message).split(";")[1] for rec in caplog.records[1::2]]
+    )
```
replaceafill: The [caplog](https://docs.pytest.org/en/latest/how-to/logging.html#caplog-fixture) fixture allows you to filter levels and loggers. Use that instead of manually iterating the records.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+def test_transfer_file_tags(admin_client, dashboard_uuid, file, mocker):
```
replaceafill: I think most of the tests of the `TransferFileTags` class should be renamed. They're exposing unnecessary details of what they're testing. The docstring of that class provides a very good example of what the class is doing:

https://github.com/artefactual/archivematica/blob/37ba46a6f90d64282f9f8f94acc635949c4c52a1/src/dashboard/src/components/file/views.py#L19-L21

So, instead of naming things like `get_method`, `put_method`, etc. why don't you use similar wording `fetch`, `set` and `remove` (without any `method` suffix).

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+        "elasticSearchFunctions.set_file_tags",
+        return_value=["test"],
+    )
+    response = admin_client.put(
+        reverse("file:transfer_file_tags", args=[file.uuid]),
+        data=json.dumps(tag_to_update),
+    )
+    assert response.status_code == 200
+
+    assert response.json() == {"success": True}
```
replaceafill, line 356: I've noticed some asserts use `json.loads(...) == {}` and some use `response.json() == {}`. Could we make them consistent?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,405 @@
...
+def sip(db):
+    return models.SIP.objects.create()
+
+
+@pytest.fixture
+def file(db):
+    return models.File.objects.create()
+
+
+def test_file_details(admin_client, dashboard_uuid, file, mocker):
```
replaceafill: Instead of using the `mocker` fixture which needs an extra dependency, let's start moving towards Python's standard [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) module.

replaceafill (changes requested):
This is a good start @Dhwaniartefact, but I think the tests should be named more consistently and not expose details on how they're implemented. I like this [article](https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/) that recommends a few guidelines for naming tests.

tests/dashboard/components/file/test_views.py
Dhwaniartefact: sure noted

tests/dashboard/components/file/test_views.py
Dhwaniartefact: Let me assert here.

tests/dashboard/components/file/test_views.py
Dhwaniartefact: Thanks for suggestion.

Dhwaniartefact:
> This is a good start @Dhwaniartefact, but I think the tests should be named more consistently and not expose details on how they're implemented. I like this [article](https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/) that recommends a few guidelines for naming tests.

I have done my best to rename the test functions. Let me know if anything needs to be changed.

tests/dashboard/components/file/test_views.py
replaceafill: Could we replace `mocker` entirely with `unittest.mock`?

Dhwaniartefact: sure

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    "error_message, status_code",
+    [
+        ("get_transfer_file_info returned no exact results", 404),
+        ("unknown error", 500),
+    ],
+    ids=["no exact results", "unknown error"],
+)
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch("elasticSearchFunctions.get_transfer_file_info")
+def test_file_details_handle_exceptions(
```
replaceafill:
```suggestion
def test_file_details_handles_exceptions(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    error_message,
+    status_code,
+    admin_client,
+    dashboard_uuid,
+    file,
+):
+    get_transfer_file_info.side_effect = elasticSearchFunctions.ElasticsearchError(
+        error_message
+    )
+    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
```
replaceafill, line 91:
```suggestion

    response = admin_client.get(reverse("file:file_details", args=[file.uuid]))
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    error_message,
+    status_code,
+    admin_client,
+    dashboard_uuid,
+    file,
+):
+    get_transfer_file_info.side_effect = elasticSearchFunctions.ElasticsearchError(
+        error_message
+    )
+    response = admin_client.get(
```
replaceafill, line 164:
```suggestion

    response = admin_client.get(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    "error_message, status_code",
+    [
+        ("get_transfer_file_info returned no exact results", 404),
+        ("unknown error", 500),
+    ],
+    ids=["no exact results", "unknown error"],
+)
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch("elasticSearchFunctions.get_transfer_file_info")
+def test_bulk_extractor_handle_exceptions(
```
replaceafill:
```suggestion
def test_bulk_extractor_handles_exceptions(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+        "sipuuid": str(sip.uuid),
+        "status": "backlog",
+        "tags": [],
+        "bulk_extractor_reports": ["ccn", "pii"],
+        "format": [
+            {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+        ],
+        "pending_deletion": False,
+    }
+    with mock.patch(
```
replaceafill: Why does need to be a context manager?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+                "context": "9424 06. JCB - 3566002020360505 ",
+            },
+        ],
+    }
+
+
+@pytest.mark.django_db
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch("elasticSearchFunctions.get_transfer_file_info")
+def test_bulk_extractor_handles_exception_if_no_missing_reports(
```
replaceafill: This name is still not clear. You handle an exception when there are no reports missing?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+        "fileuuid": file.uuid,
+        "sipuuid": sip.uuid,
+        "status": "backlog",
+        "tags": [],
+        "bulk_extractor_reports": ["ccn", "pii"],
+        "format": [
+            {"puid": "x-fmt/111", "format": "Generic TXT", "group": "Text (Plain)"}
+        ],
+    }
+    with mock.patch(
```
replaceafill: Why does need to be a context manager?

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    "exception, status_code",
+    [
+        (elasticSearchFunctions.ElasticsearchError("No tags"), 400),
+        (elasticSearchFunctions.EmptySearchResultError("No tags"), 404),
+    ],
+    ids=["Elasticsearch error", "Empty search result error"],
+)
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch("elasticSearchFunctions.get_file_tags")
+def test_transfer_file_tags_handle_exceptions(
```
replaceafill:
```suggestion
def test_transfer_file_tags_handles_exceptions(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    get_file_tags,
+    get_client,
+    exception,
+    status_code,
+    admin_client,
+    dashboard_uuid,
+    file,
+):
+    get_file_tags.side_effect = exception
+    response = admin_client.get(reverse("file:transfer_file_tags", args=[file.uuid]))
```
replaceafill, line 331:
```suggestion

    response = admin_client.get(reverse("file:transfer_file_tags", args=[file.uuid]))
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+        "message": "No tags",
+    }
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch(
+    "elasticSearchFunctions.set_file_tags",
+    return_value=["test"],
+)
+def test_transfer_file_tags_if_update_tags(
```
replaceafill:
```suggestion
def test_transfer_file_tags_updates_tags(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    response = admin_client.put(
+        reverse("file:transfer_file_tags", args=[file.uuid]),
+        data=json.dumps(tag_to_update),
+    )
+    assert response.status_code == 200
+
+    assert response.json() == {"success": True}
+
+
+def test_transfer_file_tags_fails_if_no_JSON_document_requested(
```
replaceafill:
```suggestion
def test_transfer_file_tags_fails_if_no_document_provided(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    response = admin_client.put(reverse("file:transfer_file_tags", args=[file.uuid]))
+    assert response.status_code == 400
+
+    assert response.json() == {
+        "success": False,
+        "message": "No JSON document could be decoded from the request.",
+    }
+
+
+def test_transfer_file_tags_fails_if_updated_tag_is_not_list(
```
replaceafill:
```suggestion
def test_transfer_file_tags_fails_if_provided_tags_are_not_in_a_list(
```

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+        "message": "No tags",
+    }
+
+
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch(
+    "elasticSearchFunctions.set_file_tags",
+    return_value=["test"],
+)
+def test_transfer_file_tags_if_tags_are_removed(
```
replaceafill: This name is not clear.

tests/dashboard/components/file/test_views.py
```diff
@@ -0,0 +1,459 @@
...
+    "exception, status_code",
+    [
+        (elasticSearchFunctions.ElasticsearchError("No tags"), 400),
+        (elasticSearchFunctions.EmptySearchResultError("No tags"), 404),
+    ],
+    ids=["Elasticsearch error", "Empty search result error"],
+)
+@mock.patch("elasticSearchFunctions.get_client")
+@mock.patch("elasticSearchFunctions.set_file_tags")
+def test_transfer_file_tags_handle_exceptions_if_tags_are_removed(
```
replaceafill:
```suggestion
def test_transfer_file_tags_handle_exceptions_when_removing_tags(
```

replaceafill (changes requested):
I suggested a few changes:
* Maybe a pattern for naming the test functions could be`test_[some_function]_[executes an action/reacts to something]_[when/on some condition]`
* Try to group blocks in the test function as: [given/setup environment] [when/call function] [then/assert] leaving blank lines between them

I'm also curious why a couple of mocks are used as context managers with the `with` statement but most of them are decorators.

replaceafill (approved):

PR #1947, opened 2024-06-02
Fix parsing in load labels from file_labels.csv job

replaceafill:
While working on https://github.com/archivematica/Issues/issues/1556 I noticed the ` Load labels from metadata/file_labels.csv` job had a similar problem, so this fixes it and adds a test for the client script.


Dhwaniartefact:
The `open` function in Python 3 has an argument `newline`. Setting it to None enables universal newline mode. I guess that can be another approach as per this [docs](https://docs.python.org/release/3.2/library/functions.html#open).

Dhwaniartefact (approved):

PR #1948, opened 2024-06-03
Fix backlog files removal in purge data command

replaceafill:
The `elasticSearchFunctions._document_ids_from_field_query` helper [assumes/expects the search value passed to it to be a string](https://github.com/artefactual/archivematica/blob/49833a02f9dce5eaf91130ab6d5869ad3b085952/src/archivematicaCommon/lib/elasticSearchFunctions.py#L1251). However the `purge_transient_processing_data` [passes the Transfer primary key which is a UUID](https://github.com/artefactual/archivematica/blob/49833a02f9dce5eaf91130ab6d5869ad3b085952/src/dashboard/src/main/management/commands/purge_transient_processing_data.py#L184) raising an error. This updates the helper to convert the search value explicitly.

Connected to https://github.com/archivematica/Issues/issues/1714

### DIFF
--- a/src/archivematicaCommon/lib/elasticSearchFunctions.py
+++ b/src/archivematicaCommon/lib/elasticSearchFunctions.py
@@ -1248,7 +1248,7 @@ def _document_ids_from_field_query(client, index, field, value):
     document_ids = []
 
     # Escape /'s with \\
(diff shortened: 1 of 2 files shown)


Dhwaniartefact (approved):
good one!

PR #1949, opened 2024-06-04
Handle NULL values when creating transfer METS

replaceafill:
This converts the nullable fields of the models involved into empty strings during the creation of the transfer METS file.

Connected to https://github.com/archivematica/Issues/issues/1581


Dhwaniartefact (approved):
@replaceafill : LGTM !

PR #1952, opened 2024-06-20
Prevent unnecessary grid repaint

sevein:
This is a simplistic client-based approach to avoid repainting the jobs grid if the server response hasn't changed. It reduces the frequency of grid repaints significantly, minimizing interference with user interactions, especially with decision dropdowns.


jraddaoui (approved):
Nice!

replaceafill (approved):

PR #1953, opened 2024-07-04
Upgrade MediaArea repo version in Dockerfile

replaceafill:
Related to https://github.com/artefactual-labs/ansible-archivematica-src/pull/398


Dhwaniartefact (approved):
Surprise to know that Docker is complaining for `AS` with FROM statement. nice!

PR #1954, opened 2024-07-06
Add test for transcribe_file

Dhwaniartefact:
Extend test coverage of "transcribe_file.py"

### DIFF
--- a/tests/MCPClient/test_transcribe_file.py
+++ b/tests/MCPClient/test_transcribe_file.py
@@ -160,16 +160,16 @@ def test_main(file, task, fprule, file_format_version):
     assert (
         models.File.objects.filter(
             filegrpuse="original",
-            originallocation=b"%SIPDirectory%objects/file.jpg",
-            currentlocation=b"%SIPDirectory%objects/file.jpg",
+            originallocation=file.originallocation,
+            currentlocation=file.currentlocation,
         ).count()
         == 1
     )
     assert (
         models.File.objects.filter(
             filegrpuse="text/ocr",
-            originallocation=b"%SIPDirectory%objects/file.jpg",
-            currentlocation=b"%SIPDirectory%objects/file.jpg",
+            originallocation=file.originallocation,
+            currentlocation=file.currentlocation,
         ).count()
         == 1
     )
--- a/tests/MCPClient/test_transcribe_file.py
+++ b/tests/MCPClient/test_transcribe_file.py
@@ -8,6 +8,16 @@ from django.utils import timezone
 from fpr import models as fprmodels
 from main import models
 
+EXECUTE_OR_RUN_STDOUT = "Hello"
+
+
+@pytest.fixture(scope="module", autouse=True)
+def execute_or_run():
+    with mock.patch(
+        "transcribe_file.executeOrRun", return_value=(0, EXECUTE_OR_RUN_STDOUT, "")
+    ):
+        yield
+
 
 @pytest.fixture
 def sip_directory(tmp_path):
@@ -19,7 +29,7 @@ def sip_directory(tmp_path):
 
 @pytest.fixture
 def sip(sip_directory):
-    return models.SIP.objects.create(currentpath=str(sip_directory) + "/")
+    return models.SIP.objects.create(currentpath=f"{sip_directory}/")
 
 
 @pytest.fixture
@@ -27,18 +37,15 @@ def file(sip, sip_directory):
     location = b"%SIPDirectory%objects/file.jpg"
     file_dir = sip_directory / "objects"
     file_dir.mkdir(parents=True)
+    file_name = pathlib.Path(location.decode()).name
+    (file_dir / file_name).touch()
 
-    file = models.File.objects.create(
+    return models.File.objects.create(
         sip=sip,
         filegrpuse="original",
         currentlocation=location,
         originallocation=location,
     )
-
-    file_name = pathlib.Path(file.currentlocation.decode()).name
-    (file_dir / file_name).touch()
-
-    return file
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


replaceafill:
@Dhwaniartefact is this ready for review? It's still marked as `Draft`?

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+def sip_directory(tmp_path):
+    result = tmp_path / "sip"
+    result.mkdir()
+
+    return result
+
+
+@pytest.fixture
+def sip(sip_directory):
+    return models.SIP.objects.create(currentpath=str(sip_directory) + "/")
```
replaceafill:
```suggestion
    return models.SIP.objects.create(currentpath=f"{sip_directory}/")
```

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+
+    file = models.File.objects.create(
+        sip=sip,
+        filegrpuse="original",
+        currentlocation=location,
+        originallocation=location,
+    )
+
+    file_name = pathlib.Path(file.currentlocation.decode()).name
+    (file_dir / file_name).touch()
```
replaceafill: Could we reorganize this a bit? I think it'd be good to group together and separate the file system manipulation code (creation of `objects` directory and dummy file) from the Django model. They only relate through the `location` variable which is already declared at the top of the fixture.

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+
+
+@pytest.fixture
+def preservation_file_fprule(format_version, command):
+    return fprmodels.FPRule.objects.create(
+        format=format_version,
+        command=command,
+        purpose="transcription",
+    )
```
replaceafill: This fixture doesn't depend on the `preservation_file` fixture which makes me wonder if naming it with something related to `derivatives` would be better. It's also only used for the testing the `fetch_rules_for_derivatives` function, also related to `derivatives`. Hope this makes sense. Otherwise, feel free to ignore.

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+@pytest.mark.django_db
+def test_main(file, task, fprule, file_format_version, capsys):
+    job = mock.Mock(spec=Job)
+
+    result = transcribe_file.main(job, task_uuid=task.taskuuid, file_uuid=file.uuid)
+
+    assert result == 0
+
+    captured = capsys.readouterr()
+    assert captured.out.strip() == "Hello"
```
replaceafill: I appreciate that you've fully populated the `command` fixture so that it can be executed but I think it's best to mock `transcribe_file.executeOrRun` and prevent it from running. I see it similar to what we do with temporary file paths instead of assuming paths exist in the machine that is running the test. We should not assume we can execute commands unless we set up the environment accordingly.

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+    assert models.Event.objects.filter(event_type="transcription").count() == 1
+    assert models.File.objects.count() == 2
+    assert models.File.objects.filter(filegrpuse="original").count() == 1
+    assert models.File.objects.filter(filegrpuse="text/ocr").count() == 1
+    assert (
+        models.Derivation.objects.filter(event__event_type="transcription").count() == 1
+    )
```
replaceafill: Could these `filter` calls include more criteria attributes? Fox example the transcription event has other meaningful attributes like detail, outcome or note that you can assert. The derivation could also have the source and derived files as criteria.

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+):
+    job = mock.Mock(spec=Job)
+
+    result = transcribe_file.main(
+        job, task_uuid=task.taskuuid, file_uuid=preservation_file.uuid
+    )
+
+    assert result == 0
+
+    assert models.Event.objects.count() == 0
```
replaceafill: Does this assert that an event was not created or that all events were deleted?

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+    assert job.print_error.mock_calls == [
+        mock.call(f"{preservation_file.uuid} is not an original; not transcribing")
+    ]
```
replaceafill, line 201: :clap:

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+    result = transcribe_file.main(job, task_uuid=task.taskuuid, file_uuid=file.uuid)
+
+    assert result == 0
+
+    assert models.Event.objects.count() == 0
+    assert job.print_error.mock_calls == [
+        mock.call(
+            f"No rules found for file {file.uuid} and its derivatives; not transcribing"
+        )
+    ]
```
replaceafill: Same question and praise as the test above :grin:

tests/MCPClient/test_transcribe_file.py
```diff
@@ -0,0 +1,210 @@
...
+
+@pytest.mark.django_db
+def test_fetch_rules_for_derivatives(
+    derivation, preservation_file_fprule, preservation_file_format_version
+):
+    derived_file, rules_of_derived_file = transcribe_file.fetch_rules_for_derivatives(
+        file_=derivation.source_file
+    )
+
+    assert rules_of_derived_file.values()[0]["purpose"] == "transcription"
```
replaceafill:
```suggestion
    assert set(rules_of_derived_file.values_list('purpose')) == {('transcription',)}
```

replaceafill (changes requested):
Great progress @Dhwaniartefact.

replaceafill (approved):

PR #1956, opened 2024-07-11
Use only active rules for transcription

Dhwaniartefact:
Active `FPRule` for Tessaract is now correctly filtered by this change.

connected to https://github.com/archivematica/Issues/issues/1630


replaceafill (approved):
Great work fixing this @Dhwaniartefact!

PR #1957, opened 2024-07-16
Enhance OIDC authentication flow

sbreker:
Adds two new enhancements for OIDC authentication flow:

1. Add a new setting to control whether local AM authentication is
available when OIDC authentication is in use. If local AM authentication
is disabled, then users will only be able to authenticate via the OIDC
provider. If the new setting is not configured, local AM authentication
is available.

2. Add ability to define more than one OIDC provider in AM. Specific
providers can be chosen using HTTP query params passed to the server
when authenticating.


src/dashboard/src/templates/layout.html
```diff
@@ -19,7 +19,9 @@
     <script src="{% static 'vendor/jquery.idle-timer.js' %}" type="text/javascript"></script>
     <script src="{% static 'vendor/underscore.js' %}" type="text/javascript"></script>
     <script src="{% static 'vendor/backbone.js' %}" type="text/javascript"></script>
-    <script src="{% static 'js/status.js' %}" type="text/javascript"></script>
+    {% if user.is_authenticated %}
+      <script src="{% static 'js/status.js' %}" type="text/javascript"></script>
+    {% endif %}
```
replaceafill, line 24: Thank you for addressing this :tada:!

For the record, Steve found that the `login` view kept redirecting back to itself after a few seconds which was caused by an attempt to poll data from the MCPServer which is only available to authenticated users.

src/dashboard/src/components/accounts/backends.py
```diff
@@ -42,6 +43,49 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
...
+            "OIDC_OP_TOKEN_ENDPOINT",
+            "OIDC_OP_USER_ENDPOINT",
+            "OIDC_OP_JWKS_ENDPOINT",
+            "OIDC_OP_LOGOUT_ENDPOINT",
+        ]:
+            # Retrieve the request object stored in the instance.
+            request = getattr(self, "request", None)
+
+            if request:
+                provider_name = request.session.get("providername")
```
replaceafill, line 60: Is it necessary to query the session through all these iterations?

src/dashboard/src/components/accounts/backends.py
```diff
@@ -42,6 +43,49 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
...
+        # not in the list, call the superclass's get_settings method.
+        return OIDCAuthenticationBackend.get_settings(attr, *args)
+
+    def authenticate(self, request, **kwargs):
+        self.request = request
+        self.OIDC_RP_CLIENT_ID = self.get_settings("OIDC_RP_CLIENT_ID")
+        self.OIDC_RP_CLIENT_SECRET = self.get_settings("OIDC_RP_CLIENT_SECRET")
+        self.OIDC_OP_TOKEN_ENDPOINT = self.get_settings("OIDC_OP_TOKEN_ENDPOINT")
+        self.OIDC_OP_USER_ENDPOINT = self.get_settings("OIDC_OP_USER_ENDPOINT")
+        self.OIDC_OP_JWKS_ENDPOINT = self.get_settings("OIDC_OP_JWKS_ENDPOINT")
```
replaceafill, line 85: If I understand the `get_settings` method correctly this block is going to try to iterate the 7 attributes these many times?

src/dashboard/src/components/accounts/views.py
```diff
@@ -193,3 +201,178 @@ def delete(request, id):
...
+        # Redirect to the OIDC authentication URL.
+        return redirect(f"{login_url}")
+
+
+class CustomOIDCAuthenticationCallbackView(OIDCAuthenticationCallbackView):
+    """
+    OIDC client authentication callback HTTP endpoint
+    """
+
+    def get_settings(self, attr, *args):
```
replaceafill: Is the logic in these `get_settings` methods similar enough to be extracted to a common base class?

src/dashboard/src/components/accounts/views.py
```diff
@@ -193,3 +201,178 @@ def delete(request, id):
...
+            # If the user authenticated via OIDC, perform the OIDC logout.
+            return super().post(request)
+        else:
+            # If the user did not authenticate via OIDC, perform a local logout and redirect to login.
+            return logout_then_login(request)
+
+
+def get_oidc_logout_url(request):
+    """
+    Constructs the OIDC logout URL for Keycloak.
```
replaceafill: Is this logic Keycloak specific? If not, maybe it's not worth mentioning it in the commment?

replaceafill (changes requested):
Great progress @sbreker! I've left a few comments and please bear with me that this is early feedback since I'm still getting familiar with the approach :sweat_smile: 

I think my biggest concern at this point is about testing the new code. Automated unit tests would be preferably but even having an option to manually test the new feature would be nice.

I tried testing the feature manually by setting the PR branch and noticed that the settings added to the Compose services are turning the feature on  by default in the development environment which now shows a `Log in with OpenID Connect` link in the authentication view. Clicking on it however points me to a non existing `keycloak` service.

I wonder if we should consider making the feature opt-in. Maybe by adding a separate `docker-compose.keycloak.yml` file that [extends](https://docs.docker.com/compose/multiple-compose-files/extends/) the existing services and adds a fully integrated service that runs a `keycloak` container, similarly to the `oidc-demo` project you once showed me.

Should we also try to document this feature more extensively? The PR description mentions new settings and abilities but doesn't list their names or default values. Would it be more convenient to either use:

* The [OIDC variables](https://github.com/artefactual/archivematica/tree/d6b3869ca742c6be955e4bd34f640d59949941b5/src/dashboard/install#oidc-variables) section of the Dashboard README
* The [OIDC setup](https://www.archivematica.org/en/docs/archivematica-1.16/admin-manual/security/security/#oidc-setup) section of the docs
* Maybe both?

I think some design decisions like uppercasing the secondary provider name should be reflected too.

During my brief testing I also noticed some weird interaction while switching between the Dashboard and the Storage Service UIs. I haven't been able to reproduce it consistently but sometimes after logging out from one service the other service starts giving me CSRF errors. I'll comment again if I'm able to find the right sequence of steps to trigger it.

Finally I'm wondering how the new setting to allow "local authentication" relates to Django's own [authentication heuristics](https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#specifying-authentication-backends) where a user should be able to set up multiple authentication backends in a list. I'll try to do some more research on this and comment back if I see any potential problem.

sbreker:
Hi @replaceafill! Thanks for this feedback - I will start addressing these items.

One thing I forgot to mention is the commit that updates the docker compose file will be dropped - the defaults for AM's authentication are not meant to be updated.

I have started crafting some configuration instructions [in this gist](https://gist.github.com/sbreker/2022a1af14adeb665bbe699b62bdce8d).

sbreker:
I think I may need you to elaborate on what you are looking for wrt to documentation here: "I think some design decisions like uppercasing the secondary provider name should be reflected too."

The secondary provider name config is case insensitive. Do you think the code should be changed to force it to be specified as uppercase?

src/dashboard/src/templates/layout.html
sbreker: I will not squash this commit - I will keep it as a separate individual commit since it is unrelated to the OIDC changes.

src/dashboard/src/components/accounts/backends.py
sbreker: Since this session value is the source of record for the current provider, my strategy was to rely on this variable completely rather than make local copies in the oidc derived classes. I thought this was simpler and more clear and avoids the problem of checking if the provider has changed, and comparing it to a stored provider name. This code will only be executed if one of these seven env vars are queried - any others will use the base class get_settings().

Is there a performance problem I am not accounting for with checking a session var?

src/dashboard/src/components/accounts/backends.py
replaceafill: Sorry Steve. I misread the `if` conditions at the beginning of the `get_settings` methods as `for` loops :facepalm:

src/dashboard/src/components/accounts/backends.py
replaceafill: We plan to address this refactoring after adding some basic tests for the new components.

src/dashboard/src/components/accounts/views.py
replaceafill: We plan to address this refactoring after adding some basic tests for the new components.

replaceafill: The callback class has been removed as it seemed unnecessary for the feature.

src/dashboard/src/components/accounts/backends.py
replaceafill: I misread the code here. These were not iterations, but `if` conditions :facepalm:. Sorry for the noise.

sbreker (commented):
Looks great @replaceafill! :+1:

replaceafill (approved):

Dhwaniartefact (approved):

PR #1964, opened 2024-08-02
Fix mypy hook in pre-commit

replaceafill:
This updates the `pre-commit` configuration to list our typing dependencies explicitly in the `mypy` hook.

We were previously using `types-all` to indirectly import a lot of [`typeshed` stubs](https://github.com/python/typeshed/tree/main/stubs), but `types-pkg-resources` has been deprecated and [yanked from PyPI](https://pypi.org/project/types-pkg-resources/).


Dhwaniartefact (approved):
Thanks for fixing this issue. LGTM!

PR #1965, opened 2024-08-02
Add test for the extract_contents client script

replaceafill:
This also updates a couple of shared fixtures in the `conftest` module of the `MCPClient` to make them more generic.


Dhwaniartefact (approved):
LGTM!

PR #1968, opened 2024-08-09
Add release notes

eviau-artefactual:
Related: https://github.com/artefactual/archivematica-docs/pull/470

We currently think release notes should be closer to the code.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,0 +1,4545 @@
+# Release notes {#release_notes}
+
+Release notes are prepared by Artefactual for each release of
+Archivematica. Most releases also have a corresponding release of the
+Storage System, and may also include corresponding releases for various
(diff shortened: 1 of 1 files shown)


eviau-artefactual:
TODO

- fix duplicate heading so the curly brackets content does not show.

RELEASENOTES.md
```diff
@@ -0,0 +1,4545 @@
+# Release notes {#release_notes}
+
+Release notes are prepared by Artefactual for each release of
+Archivematica. Most releases also have a corresponding release of the
+Storage System, and may also include corresponding releases for various
+dependencies and libraries (i.e. FIDO, Automation Tools) - be sure to
```
sromkey: Instead of FIDO as an example (which we do not maintain) I would change to AIPScan, for e.g.

replaceafill (approved):
LGTM.

PR #1972, opened 2024-08-12
Extend test coverage of normalize script

replaceafill:
This also adds type hints to the `normalize` script and its tests.


Dhwaniartefact (approved):
LGTM!

PR #1975, opened 2024-08-13
Extend test coverage of policy check

Dhwaniartefact:
This PR adds type hints to the `policy_check` script and its tests.

### DIFF
--- a/src/MCPClient/lib/clientScripts/policy_check.py
+++ b/src/MCPClient/lib/clientScripts/policy_check.py
@@ -78,7 +78,7 @@ class PolicyChecker:
         self,
         job: Job,
         file_path: str,
-        file_uuid: str,
+        file_uuid: Optional[str],
         sip_uuid: str,
         shared_path: str,
         file_type: str,
@@ -172,9 +172,7 @@ class PolicyChecker:
             return True
         # Access derivatives have Derivation rows with NULL event types (cf.
         # normalize.py client script).
-        event_type: Optional[str] = "normalization"
-        if for_access:
-            event_type = None
+        event_type = None if for_access else "normalization"
         try:
             Derivation.objects.get(
                 derived_file__uuid=self.file_uuid, event__event_type=event_type
@@ -210,7 +208,7 @@ class PolicyChecker:
             return True
         return False
 
-    def _get_manually_normalized_access_derivative_file_uuid(self) -> File:
+    def _get_manually_normalized_access_derivative_file_uuid(self) -> Optional[File]:
         """If the file-to-be-policy-checked is a manually normalized access
         derivative it will have no file UUID in the database. We therefore have
         to retrieve the UUID of the original file that was format-identified,
--- a/src/MCPClient/lib/clientScripts/policy_check.py
+++ b/src/MCPClient/lib/clientScripts/policy_check.py
@@ -164,7 +164,7 @@ class PolicyChecker:
         else:
             return True
 
-    def _file_is_derivative(self, for_access=False) -> bool:
+    def _file_is_derivative(self, for_access: bool = False) -> bool:
         """Return ``True`` if the target file is a derivative; ``False``
         otherwise.
         """
@@ -466,20 +466,6 @@ class PolicyChecker:
         return self._sip_policy_checks_dir
 
 
-def _get_shared_path(argv):
-    try:
-        return argv[4]
-    except IndexError:
-        return None
-
-
-def _get_file_type(argv):
-    try:
-        return argv[5]
-    except IndexError:
-        return "original"
-
-
 def call(jobs: List[Job]) -> None:
     with transaction.atomic():
         for job in jobs:
@@ -487,8 +473,8 @@ def call(jobs: List[Job]) -> None:
                 file_path = job.args[1]
                 file_uuid = job.args[2]
                 sip_uuid = job.args[3]
-                shared_path = _get_shared_path(job.args)
-                file_type = _get_file_type(job.args)
+                shared_path = job.args[4]
+                file_type = job.args[5]
 
                 try:
                     job.set_status(
--- a/src/MCPClient/lib/clientScripts/policy_check.py
+++ b/src/MCPClient/lib/clientScripts/policy_check.py
@@ -18,7 +18,13 @@ import django
 from custom_handlers import get_script_logger
 
 django.setup()
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
+
 import databaseFunctions
+from client.job import Job
 from dicts import replace_string_values
 from django.conf import settings as mcpclient_settings
 from django.core.exceptions import ValidationError
@@ -39,7 +45,14 @@ NOT_APPLICABLE_CODE = 0
 FAIL_CODE = 1
 
 
-def main(job, file_path, file_uuid, sip_uuid, shared_path, file_type):
+def main(
+    job: Job,
+    file_path: str,
+    file_uuid: str,
+    sip_uuid: str,
+    shared_path: str,
+    file_type: str,
+) -> int:
     """Entry point for policy checker."""
     setup_dicts(mcpclient_settings)
 
@@ -61,7 +74,15 @@ class PolicyChecker:
     ``check`` method.
     """
 
-    def __init__(self, job, file_path, file_uuid, sip_uuid, shared_path, file_type):
+    def __init__(
+        self,
+        job: Job,
+        file_path: str,
+        file_uuid: str,
+        sip_uuid: str,
+        shared_path: str,
+        file_type: str,
+    ) -> None:
         """Initiate a new policy check."""
         self.job = job
         self.file_path = file_path
@@ -73,11 +94,11 @@ class PolicyChecker:
         self.is_manually_normalized_access_derivative = (
             self._get_is_manually_normalized_access_derivative()
         )
-        self._sip_logs_dir = None
-        self._sip_subm_doc_dir = None
-        self._sip_policy_checks_dir = None
+        self._sip_logs_dir: Optional[str] = None
+        self._sip_subm_doc_dir: Optional[str] = None
+        self._sip_policy_checks_dir: Optional[str] = None
 
-    def check(self):
+    def check(self) -> int:
         """Check the file identified by ``self.file_uuid`` against any
         policy-check FPR commands that are applicable. If any fail, return a
         non-zero exit code; otherwise return ``0``.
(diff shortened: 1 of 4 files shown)


src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -143,15 +164,15 @@ def _we_check_this_type_of_file(self):
...
         otherwise.
         """
         if self.is_manually_normalized_access_derivative:
             return True
         # Access derivatives have Derivation rows with NULL event types (cf.
         # normalize.py client script).
-        event_type = "normalization"
+        event_type: Optional[str] = "normalization"
         if for_access:
             event_type = None
```
replaceafill: I think expressing this bit differently would allow the type checker to infer the possible values without the annotation:

```suggestion
        event_type = None if for_access else "normalization"
```

src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -189,7 +210,7 @@ def _file_is_for_access(self):
             return True
         return False
 
-    def _get_manually_normalized_access_derivative_file_uuid(self):
+    def _get_manually_normalized_access_derivative_file_uuid(self) -> File:
```
replaceafill: The last line of the method might return `None`, should the return value be `Optional[File]`?

src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -313,14 +334,14 @@ def _get_command_to_execute(self, rule):
         else:
             return (rule.command.command, [self.file_path, self.policies_dir])
 
-    def _save_to_logs_dir(self, output):
+    def _save_to_logs_dir(self, output: Dict[str, str]) -> None:
```
replaceafill, line 335: I think in the long term we need to parse `output = json.loads(stdout)` passing our own decoder that guarantees the types in the `output` mapping. See for example:

https://github.com/artefactual/archivematica/blob/01293dc7696ece0066353b6c83ce348ff4c3f3e5/src/MCPServer/lib/server/workflow.py#L228-L238

src/MCPClient/lib/clientScripts/policy_check.py
```diff
@@ -445,29 +466,15 @@ def sip_policy_checks_dir(self):
...
     with transaction.atomic():
         for job in jobs:
             with job.JobContext(logger=logger):
                 file_path = job.args[1]
                 file_uuid = job.args[2]
                 sip_uuid = job.args[3]
-                shared_path = _get_shared_path(job.args)
-                file_type = _get_file_type(job.args)
+                shared_path = job.args[4]
+                file_type = job.args[5]
```
replaceafill, line 475: :clap:

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+    )
+
+
+STDOUT = json.dumps(
+    {"eventOutcomeInformation": "pass", "eventOutcomeDetailNote": "a note"}
+)
+
+
+@pytest.fixture
+def EXECUTE_OR_RUN() -> Any:
```
replaceafill: Why does this fixture name have to be uppercase?

Also what is advantage of having this mock created as a fixture and not as decorator like in other cases?

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+        {
+            "eventOutcomeInformation": "pass",
+            "eventOutcomeDetailNote": f'format="{format.description}"; version="{format_version.version}"; result="Well-Formed and valid"',
+        }
+    )
+    execute_or_run.return_value = (0, expected_stdout, "")
+
+    policy_check.call([job])
+
+    job.set_status.assert_called_once_with(0)
```
replaceafill: `policy_check` defines constants for its return values:

https://github.com/artefactual/archivematica/blob/01293dc7696ece0066353b6c83ce348ff4c3f3e5/src/MCPClient/lib/clientScripts/policy_check.py#L37-L39

All these `set_status` assertions should use them.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+            STDOUT,
+            "",
+        ),
+    ):
+        yield
+
+
+@pytest.mark.django_db
+@mock.patch("policy_check.executeOrRun")
+def test_policy_check_succeeds_if_rules_exist(
```
replaceafill: Why are some tests named `test_policy_check_...` and some `test_policy_checker_...`?

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+            str(shared_directory_path),
+            sip_file.filegrpuse,
+        ],
+        JobContext=mock.MagicMock(),
+        spec=Job,
+    )
+    expected_stdout = json.dumps(
+        {
+            "eventOutcomeInformation": "pass",
+            "eventOutcomeDetailNote": f'format="{format.description}"; version="{format_version.version}"; result="Well-Formed and valid"',
```
replaceafill, line 108: See how your `event_detail` parameter below is using `tool.description` retrieved from the rule. I think you could also retrieve these from the `sip_file_format_version` fixture.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+@pytest.mark.django_db
+@mock.patch(
+    "policy_check.executeOrRun",
+    return_value=(
+        1,
+        STDOUT,
+        "",
+    ),
```
replaceafill: I think several of these `return_value` tuples can be represented in a single line if you remove the trailing comma.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+@mock.patch("policy_check.executeOrRun")
+def test_policy_check_succeeds_if_rules_exist(
+    execute_or_run: mock.Mock,
+    sip_file: models.File,
+    sip: models.SIP,
+    fprule_policy_check: fprmodels.FPRule,
+    sip_file_format_version: models.FileFormatVersion,
+    format: fprmodels.Format,
+    format_version: fprmodels.FormatVersion,
+    shared_directory_path: str,
```
replaceafill: This fixture returns a `pathlib.Path` object:

https://github.com/artefactual/archivematica/blob/01293dc7696ece0066353b6c83ce348ff4c3f3e5/tests/MCPClient/conftest.py#L216

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+) -> None:
+    job = mock.Mock(
+        args=[
+            "policy_check",
+            "",
+            "",
```
replaceafill: Could we use a valid `UUID` string here? Maybe `str(uuid.uuid4())`.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+@pytest.mark.django_db
+@mock.patch(
+    "policy_check.executeOrRun",
+    return_value=(
+        1,
+        STDOUT,
+        "",
+    ),
+)
+def test_policy_checker_fails_if_execute_rule_command_fails(
```
replaceafill:
```suggestion
def test_policy_checker_fails_if_rule_command_fails(
```

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+
+    policy_check.call([job])
+
+    job.set_status.assert_called_once_with(1)
+    assert job.pyprint.mock_calls == [
+        mock.call("Running", fprule_policy_check.command.description)
+    ]
+    job.print_error.assert_called_once_with(
+        f"Command {fprule_policy_check.command.description} failed with exit status 1; stderr:",
+        "",
```
replaceafill: This could reuse the `stderr` value from the mock if you set its `return_value` inside the test.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,592 @@
...
+    shared_directory_path: str,
+) -> None:
+    job = mock.Mock(
+        args=[
+            "policy_check",
+            access_file.currentlocation.decode(),
+            str(access_file.uuid),
+            str(sip.uuid),
+            str(shared_directory_path),
+            "access",
```
replaceafill: The preservation case of this test is using the `filegrpuse` attribute instead of a hard coded string.

replaceafill (changes requested):

src/MCPClient/lib/clientScripts/policy_check.py
Dhwaniartefact: oh yes, I miss the exception part. Thanks!

tests/MCPClient/test_policy_check.py
Dhwaniartefact: ya good point

tests/MCPClient/test_policy_check.py
Dhwaniartefact: I used uppercase due to name conflicts. But this fixture has been removed since it is used as decorator by other tests with different `stdout`.

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,613 @@
...
+    ),
+)
+def test_policy_checker_warns_if_rules_do_not_exist(
+    execute_or_run: mock.Mock,
+    sip_file: models.File,
+    sip: models.SIP,
+    sip_file_format_version: models.FileFormatVersion,
+    format: fprmodels.Format,
+    format_version: fprmodels.FormatVersion,
+    shared_directory_path: str,
```
replaceafill:
```suggestion
    shared_directory_path: pathlib.Path,
```

tests/MCPClient/test_policy_check.py
```diff
@@ -0,0 +1,613 @@
...
+    shared_directory_path: pathlib.Path,
+) -> None:
+    job = mock.Mock(
+        args=[
+            "policy_check",
+            access_file.currentlocation.decode(),
+            str(access_file.uuid),
+            str(sip.uuid),
+            str(shared_directory_path),
+            "access",
```
replaceafill:
```suggestion
            access_file.filegrpuse,
```

replaceafill (approved):
Thanks @Dhwaniartefact. You missed a couple of observations, but I made suggestions for those.

PR #1976, opened 2024-08-14
Extend test coverage of the validate_file script

replaceafill:
This also adds type hints to the `validate_file` script and its tests.


Dhwaniartefact (approved):
Great work!

PR #1977, opened 2024-08-15
Add pyproject.toml

replaceafill:
This adds a [pyproject.toml](https://peps.python.org/pep-0621/) file to store configuration metadata for the following tools:

* pytest
* coverage
* Ruff
* mypy
* tox

In a future iteration we plan to extend it with project and dependency metadata.


Dhwaniartefact (approved):
Great !

PR #1978, opened 2024-08-16
Fix error in duplicate PREMIS event ID handling

mcantelon:
Attempting to create a UUID object using an existing UUID object, rather than a string, results in an error. Changed duplicate PREMIS event identifier handling to avoid ending up doing this.

### DIFF
--- a/src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
+++ b/src/MCPClient/lib/clientScripts/load_premis_events_from_xml.py
@@ -663,7 +663,7 @@ def ensure_event_id_is_uuid(event_id, printfn=print):
         if Event.objects.filter(event_id=event_id).exists():
             raise ValueError(f"There is already an event with this event_id {event_id}")
     except ValueError:
-        new_event_id = uuid.uuid4()
+        new_event_id = str(uuid.uuid4())
         log_event_id_change(event_id, new_event_id, printfn)
         event_id = new_event_id
     return event_id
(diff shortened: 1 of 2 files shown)


replaceafill:
Thanks for catching and fixing this @mcantelon!

What do you think of updating the test module like this:

```diff
 def test_ensure_event_id_is_uuid(mocker, params):
     if params["message_logged"]:
-        mocker.patch("uuid.uuid4", return_value="f4eea76b-1921-4152-b1b4-a93dbbfeaaef")
+        mocker.patch(
+            "uuid.uuid4", return_value=uuid.UUID("f4eea76b-1921-4152-b1b4-a93dbbfeaaef")
+        )
     printfn = mocker.Mock()
     result = load_premis_events_from_xml.ensure_event_id_is_uuid(
         params["event_id"], printfn
     )
+    assert isinstance(result, str)
     if not params["message_logged"]:
         assert result == params["event_id"]
         printfn.assert_not_called()
```

This updates the test of that specific function `test_ensure_event_id_is_uuid` and fixes the wrong return value of the mock which should be a `uuid.UUID` type, not a string. Then it adds an assertion on `result` similar to yours but using `isinstance` which [is recommended for this use case](https://docs.python.org/3.9/library/functions.html#type).

mcantelon:
@replaceafill Nice! I've updated the PR!

replaceafill (approved):
Looks good @mcantelon, thank you!

PR #1979, opened 2024-08-19
Update policy check tests

Dhwaniartefact:


tests/MCPClient/test_policy_check.py
```diff
@@ -303,6 +301,12 @@ def test_policy_checker_fails_if_event_outcome_information_in_output_is_not_pass
 
     job.set_status.assert_called_once_with(policy_check.FAIL_CODE)
 
+    assert job.print_error.mock_calls == [
+        mock.call(
+            f"Command {fprule_policy_check.command.description} returned a non-pass outcome for the policy check;\n\noutcome: {json.loads(expected_stdout)['eventOutcomeInformation']}\n\ndetails: {json.loads(expected_stdout)['eventOutcomeDetailNote']}."
```
replaceafill: If you extract the value of `eventOutcomeDetailNote` into a separate variable before setting up the `return_value` of the mock, similarly to what you've done below in `test_policy_checker_saves_policy_check_result_into_logs_directory`, you could drop this `json.loads(...)` call entirely and use the extracted variable instead.

replaceafill (commented):
This looks good @Dhwaniartefact. Could you please also add some assertions to `test_policy_checker_checks_manually_normalized_access_derivative_file` that guarantee it's actually doing what you expect? It ends with `job.set_status.assert_called_once_with(policy_check.SUCCESS_CODE)` and by this point we know that might not be accurate because of `NOT_APPLICABLE_CODE`.

replaceafill (approved):

PR #1980, opened 2024-08-20
Add type hints to the characterize_file script

replaceafill:
This also removes an unnecessary argument from the workflow definition and adds a typed argument parser to the script.


Dhwaniartefact (approved):
LGTM!

PR #1983, opened 2024-08-22
Show format version and PRONOM ID in FPRule views

replaceafill:
This updates the `FPRule` tables and detail views of the `Preservation planning` tab to show the format version and PRONOM ID of each rule when available.


Dhwaniartefact (approved):
LGTM!

PR #1984, opened 2024-08-23
Use menu for processing config in Re-ingest form

replaceafill:
This updates the `Re-ingest` form of the AIP view in `Archival storage` to use a dropdown menu instead of a text field for specifying the processing configuration to use in the re-ingest request. It looks like this:

Captura desde 2024-08-23 09-38-35

The `default` processing configuration is the initial value of the menu and the XML files are listed similarly to how the [processing-configuration](https://www.archivematica.org/en/docs/archivematica-latest/dev-manual/api/api-reference-archivematica/#processing-configuration) API endpoint lists them.


Dhwaniartefact (approved):
Great!

PR #1985, opened 2024-08-27
Prefetch related models in FPR views

replaceafill:
Many views in the FPR have the [N+1 Queries problem](https://adamj.eu/tech/2020/09/01/django-and-the-n-plus-one-queries-problem/) where multiple/separate queries are created to select single related rows in the database. For example, to create an FPR rule every single `Format` is retrieved separately, and after its related `FormatGroup`.

This PR adds [prefetch_related](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#prefetch-related) calls to specific `QuerySet`s in views and forms to retrieve the related models beforehand.


Dhwaniartefact (approved):
LGTM!

PR #1990, opened 2024-09-11
Escape output and error in launchSubProcess

replaceafill:
This replaces non `utf-8` characters with [replacement characters](https://docs.python.org/3.9/library/codecs.html#codecs.replace_errors) when decoding the output and error bytes from `subprocess` calls in the `launchSubProcess` function.

Connected to https://github.com/archivematica/Issues/issues/1702


Dhwaniartefact (approved):
LGTM!

PR #1991, opened 2024-09-17
Replace clamd with clamav-client

sevein:
This commit replaces [clamd 1.0.2] with [clamav-client 0.7.0], a modern fork of
clamd that follows current Python practices and tooling. clamav-client is tested
with supported versions of Python and avoids deprecated dependencies such as
`pkg_resources` (which was removed in Python 3.12).

clamav-client provides its own `Scanner` interface to handle integration with
both `clamd` and `clamscan`, so Archivematica doesn't need to. Both backends are
tested with integration tests.

[clamd 1.0.2]: https://pypi.org/project/clamd/1.0.2/
[clamav-client 0.7.0]: https://pypi.org/project/clamav-client/0.7.0/


src/MCPClient/lib/clientScripts/antivirus.py
```diff
@@ -0,0 +1,219 @@
...
+    parser.add_argument("path", metavar="PATH", help="File or directory location")
+    parser.add_argument("date", metavar="DATE")
+    parser.add_argument(
+        "task_uuid", metavar="taskUUID", help="Currently unused, feel free to ignore."
+    )
+    return parser
+
+
+# Map user-provided backend names from configuration to the corresponding
+# internal values used in the clamav_client package. Default to "clamscanner" if
```
replaceafill:
```suggestion
# internal values used in the clamav_client package. Default to "clamdscanner" if
```

tests/MCPClient/test_antivirus.py
```diff
@@ -72,10 +73,17 @@ def __init__(self, should_except=False, passed=False):
...
         if self.should_except:
             raise Exception("Something really bad happened!")
-        return self.passed, None, None
+        result = ScanResult(filename=path, state="OK", details="details", err=None)
+        mock.patch.object(
+            result.__class__,
+            "passed",
+            new_callable=mock.PropertyMock,
+            return_value=self.passed,
+        ).start()
```
replaceafill: Is this patch stopped implicitly?

src/MCPClient/lib/clientScripts/antivirus.py
```diff
@@ -0,0 +1,219 @@
...
+def get_parser():
+    """Return a ``Namespace`` with the parsed arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("file_uuid", metavar="fileUUID")
+    parser.add_argument("path", metavar="PATH", help="File or directory location")
+    parser.add_argument("date", metavar="DATE")
+    parser.add_argument(
+        "task_uuid", metavar="taskUUID", help="Currently unused, feel free to ignore."
+    )
+    return parser
```
replaceafill: Should we remove this helper function if we're not using it?

replaceafill (approved):
Thanks @sevein. This looks great!

tests/MCPClient/test_antivirus.py
sevein: I don't really know what I'm doing, but I've tried returning a a `ScanResultMock` instead and it seems to work.

src/MCPClient/lib/clientScripts/antivirus.py
sevein: Yup! Thanks.

sevein:
@replaceafill, I've finally managed to get around to this! I've introduced a couple of commits because it was hard to see the refactor after also renaming the client module. I hope it's easier to follow now. I've also made the changes you requested.

replaceafill (approved):
Looks good. Thank you @sevein.

PR #1993, opened 2024-09-19
Enable OIDC tests

replaceafill:
This removes the condition that makes `pytest` to [skip](https://docs.pytest.org/en/latest/how-to/skipping.html) the existing OIDC authentication backend tests and mocks the settings necessary for the tests to pass.


sbreker (approved):
Looks good @replaceafill :+1: 

It will be good to run these all the time instead of having them only active when OIDC is turned on.

PR #1996, opened 2024-09-27
Fix broken links in Markdown files

replaceafill:
This also introduces [markdown-link-check](https://github.com/tcort/markdown-link-check/) as a `pre-commit` hook to check broken links in the Markdown files of the repository. The hook has been set as [manual](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages) to not interfere in the usual development workflow.

For example, to check links in a specific file:

```shell
pre-commit run --hook-stage manual markdown-link-check --files src/dashboard/frontend/README.md
```

Or to check all the Markdown files in the repository:

```shell
pre-commit run --hook-stage manual markdown-link-check --all-files
```

We should not include this in the pull request workflow because link checking is flaky. Sometimes working links fail from GitHub hosts because websites block them. But we could potentially create a separate workflow that developers could trigger manually.

### DIFF
--- a/.markdown-link-check.json
+++ b/.markdown-link-check.json
@@ -1,0 +1,7 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^http://(127.0.0.1|localhost)"
+    }
+  ]
+}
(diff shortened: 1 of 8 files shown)


Dhwaniartefact (approved):
Thanks for adding this. LGTM !

PR #2001, opened 2024-10-18
Remove comments from import blocks

replaceafill:
These comments were added before we implemented automatic import sorting.


Dhwaniartefact (approved):
Great work!

PR #2004, opened 2024-10-23
Fix API key generation when user edits are disabled

replaceafill:
When the [ALLOW_USER_EDITS](https://github.com/artefactual/archivematica/blob/b1e75911632966319d13751c90fe08131d2384a3/src/dashboard/src/settings/base.py#L581) setting of the Dashboard is set to `False` the user profile view shows most fields as read-only but allows the user to regenerate their API key.

However, submission handling of the form is currently broken and the API key is always regenerated when the `Save` button is clicked ignoring the state of the `Regenerate API key` checkbox.

This also extends test coverage of the views of the `accounts` and `administration` packages.


Dhwaniartefact (approved):
lgtm!

PR #2007, opened 2024-10-31
Defer FPRule counter updates during normalization

sevein:
This pull request introduces a context manager (`DeferredFPRuleCounter`) to defer the tracked FPRule execution until the `call` function completes the execution of the entire batch, reducing the risk of deadlocks when multiple MCPClients attempt to update the same FPRule concurrently.

This change helps prevent deadlocks, but keep in mind that normalize.py might be reworked later to avoid using long database transactions altogether.

Tentative fix for https://github.com/archivematica/Issues/issues/752 and https://github.com/archivematica/Issues/issues/1525.
Relates to https://github.com/archivematica/Issues/issues/1161.

Updated analysis:
https://gist.github.com/sevein/2ff98ef96e0e195ab57f806796a71000

### DIFF
--- a/src/MCPClient/lib/clientScripts/normalize.py
+++ b/src/MCPClient/lib/clientScripts/normalize.py
@@ -573,8 +573,8 @@ def main(job: Job, opts: NormalizeArgs, counter: DeferredFPRuleCounter) -> int:
     if (
         exitstatus != 0
         and opts.purpose in ("access", "thumbnail")
-        and cl.commandObject.output_location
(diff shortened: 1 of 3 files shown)


sevein:
An interesting aspect of this pull request is that I've never been able to reproduce the original issue, so I'm uncertain if this change is necessary or whether it solves the problem. I've tested multiple batch sizes, number of worker replicas and transfers with hundreds of files unsuccessfully, but the issue has occasionally been experienced by both customers and the Artefactual team. Maybe we can keep this work around until needed.

PR #2012, opened 2025-01-08
Extend test coverage of the create_mets_v2 script

replaceafill:
This adds type hints to the `test_create_mets_v2.py` script and extends the test coverage of the `create_mets_v2` script.

The goal is to guarantee that an `amdSec` element is created from various types of metadata sources:
* bag metadata
* a `transfer_metadata.xml` metadata file
* `*.xml` metadata files contained in `sourceMD` directories.

Connected to https://github.com/archivematica/Issues/issues/1129


Dhwaniartefact (approved):
Great!

PR #2013, opened 2025-01-09
Do not create amdSec when bag metadata is empty

replaceafill:
This prevents the creation and association of a `amdSec` element to the `objects` directory in the `structMap` when the transfer bag metadata is missing or unreadable.

Connected to https://github.com/archivematica/Issues/issues/1129


Dhwaniartefact (approved):
Lgtm!

PR #2024, opened 2025-02-20
Add tests for FPR event detail commands

replaceafill:
This adds tests that assert the output of the existing event detail commands in the FPR.

The tests assume the MCPClient OS tools are installed in the test runner and call `archivematicaCommon`'s `executeOrRun` to execute the event detail commands.

A couple of notes about the implementation:

* The `unrar-nonfree` parameter is skipped at the moment because we don't install that package in the development environment or in the Ansible role. We should probably update the event detail command to use the `unrar` command instead.
* I have added a data migration to update event detail commands. For now it only updates the `Get event detail text for 7z extraction` command to be compatible with newer versions of `7z`, but as I stated in the previous point we might need to update the `unrar-nonfree` command as well (to be decided).
* The regular expression patterns used to match the `version="..."` part of the output are not strict and they only look at the beginning of lines.

Connected to https://github.com/archivematica/Issues/issues/1729


Dhwaniartefact (approved):
Lgtm!

PR #2025, opened 2025-02-25
Remove FPR South migrations

replaceafill:
The South migrations framework was [deprecated in Django 1.7](https://south.aeracode.org/). These FPR migration files are not needed anymore.


Dhwaniartefact (approved):
Great work of removing deprecated dependency.

PR #2031, opened 2025-03-14
Add support for PKCE

scollazo:
refs: https://github.com/mozilla/mozilla-django-oidc/pull/447

### DIFF
--- a/src/archivematica/dashboard/settings/components/oidc_auth.py
+++ b/src/archivematica/dashboard/settings/components/oidc_auth.py
@@ -51,6 +51,17 @@ if OIDC_OP_LOGOUT_ENDPOINT:
 
 OIDC_RP_SIGN_ALGO = os.environ.get("OIDC_RP_SIGN_ALGO", "HS256")
 
+OIDC_USE_PKCE = os.environ.get("OIDC_USE_PKCE", "false").lower() in (
+    "true",
+    "yes",
+    "on",
+    "1",
+)
+
+OIDC_PKCE_CODE_CHALLENGE_METHOD = os.environ.get(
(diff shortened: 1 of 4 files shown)


sbreker (approved):
Thanks @scollazo! :+1:  I've added the same tests on this PR as with the Storage Service PR.

src/archivematica/dashboard/settings/components/oidc_auth.py
```diff
@@ -51,6 +51,11 @@
 
 OIDC_RP_SIGN_ALGO = os.environ.get("OIDC_RP_SIGN_ALGO", "HS256")
 
+OIDC_USE_PKCE = os.environ.get("OIDC_USE_PKCE", "false")
```
replaceafill: Should this setting be converted into a boolean value? As discussed in https://github.com/artefactual/archivematica-storage-service/pull/764 it is initially extracted as a string and will be considered truth-y by `mozilla-django-oidc`.

This issue can be demonstrated by setting `OIDC_USE_PKCE` to `"false"` in the `tests/integration/docker-compose.yml` file and then running the integration OIDC tests. Since the realm configuration in the test mandates the use of PKCE, the existing OIDC tests are expected to fail when PKCE is disabled. However, the tests still pass, indicating the PKCE was used anyway.

replaceafill (changes requested):

src/archivematica/dashboard/settings/components/oidc_auth.py
sbreker: I have added a commit that should address this.

replaceafill (approved):
The linting CI job is failing but this looks good.

PR #2036, opened 2025-03-21
Avoid use of ast.literal_eval in validate_file

sevein:
There aren't many remaining instances of `ast.literal_eval` usage. This pull request addresses `validate_file`, where switching to `json.loads` appears to work without requiring changes to the FPR scripts, since those already seem to use `json.dumps` for encoding.

There is a chance this introduces a backward-incompatible change for users who were relying on Python-style string representations when building custom validation scripts, but I believe the change is justified. If you think this could be a problem, please let me know - I can work on an alternative. For example, gpg-4o suggested this approach: https://gist.github.com/sevein/f58f3dbd8f9db2500eb8d91dc3a9f98d, that we could use as last resort when `json.loads` is unable to do the work.

Excluding #2035, there are two other uses of `ast.literal_eval` in MCPServer (in the client and packages modules) that can be addressed in future pull requests.

Connected to https://github.com/archivematica/Issues/issues/1735

### DIFF
--- a/src/MCPClient/clientScripts/validate_file.py
+++ b/src/MCPClient/clientScripts/validate_file.py
@@ -18,7 +18,7 @@ Arguments:
 import ast
 import os
 import sys
-from collections.abc import Mapping
+from dataclasses import dataclass
 from pprint import pformat
 from typing import Any
 from typing import Optional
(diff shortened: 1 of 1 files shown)


src/MCPClient/clientScripts/validate_file.py
```diff
@@ -66,6 +66,48 @@ def main(
 logger = get_script_logger("archivematica.mcp.client.validateFile")
 
 
+@dataclass
+class ValidationResult:
```
replaceafill, line 70: Thank you for abstracting this behavior!

replaceafill (approved):

PR #2037, opened 2025-03-21
Use json.loads to decode unit vars

sevein:
This pull request removes the remaining uses of ast.literal_eval, excluding an old database migration that doesn't seem worth addressing.

Relates to #2035 and #2036.

Connected to https://github.com/archivematica/Issues/issues/1735

### DIFF
--- a/src/MCPServer/server/packages.py
+++ b/src/MCPServer/server/packages.py
@@ -1,8 +1,8 @@
 """Package management."""
 
 import abc
-import ast
 import collections
+import json
 import logging
 import os
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


src/MCPServer/server/packages.py
```diff
@@ -926,10 +926,9 @@ def load_from_db(cls, uuid):
         # Distinct helps here, at least
         unit_vars_queryset = unit_vars_queryset.values_list("variablevalue").distinct()
         for unit_var_value in unit_vars_queryset:
-            # TODO: nope nope nope, fix eval usage
```
replaceafill, line 929: :rofl:

replaceafill (approved):
Thank you @sevein!

PR #2041, opened 2025-03-28
Do not show user API keys

replaceafill:
This updates the `Administration > General` view of the Dashboard to not show the API key of the user that connects with the Storage Service. The current plain text field has been replaced with a password field to do so.

Also the Dashboard user views and their redirects have been updated to show the API key only once on re-generation. This uses Django's [messages framework](https://docs.djangoproject.com/en/4.2/ref/contrib/messages/) to render an alert with a button to copy the API key to the browser's clipboard.

Captura desde 2025-03-28 16-41-49

Captura desde 2025-03-28 16-41-57

Connected to https://github.com/archivematica/Issues/issues/1738

### DIFF
--- a/src/archivematica/dashboard/media/js/misc.js
+++ b/src/archivematica/dashboard/media/js/misc.js
@@ -138,4 +138,31 @@ $(document).ready(
                 $(this).closest('.content').hide();
                 $(this).closest('.preview-help-text').children('.preview').show();
               });
+
+      // Enable tooltip for the button which allows users to copy the API key to the clipboard.
+      $('#copy-api-key-button').tooltip();
+
+      // Copy the API key to the clipboard on button click, then update the tooltip and icon accordingly.
+      $('#copy-api-key-button').click(function () {
+        var button = $(this);
+        navigator.clipboard.writeText($('#api-key').val()).then(function () {
+          var button_icon = $('#copy-button-icon');
+          var icon_original_class = button_icon.data('icon-original-class');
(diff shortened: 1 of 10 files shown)


src/archivematica/dashboard/media/js/misc.js
```diff
@@ -138,4 +138,31 @@ $(document).ready(
                 $(this).closest('.content').hide();
                 $(this).closest('.preview-help-text').children('.preview').show();
               });
+
+      // Enable tooltip for the button which allows users to copy the API key to the clipboard.
+      $('#copy-api-key-button').tooltip();
+
+      // Copy the API key to the clipboard on button click, then update the tooltip and icon accordingly.
+      $('#copy-api-key-button').click(function () {
```
sevein: You could use chaining here as long as tooltip returns the jQuery object - slightly faster because jQuery doesn't have to look up the same element twice, e.g.:

```js
$('#copy-api-key-button')
  .tooltip()
  .click(function () {
    // ...
  });
```

sevein (approved):
Nicely done!

src/archivematica/dashboard/media/js/misc.js
replaceafill: Thank you @sevein! I implemented chaining in https://github.com/artefactual/archivematica/pull/2041/commits/bab337348d3ff0f2e1fae9ec01d13bdb036e0af3 :muscle:

PR #2047, opened 2025-04-23
Add ability to set admin from OIDC claims

sbreker:
Adds new OIDC setting that will allow admins to configure whether user
roles will be set from the OIDC claims, or assigned the system default
role. The setting name is 'OIDC_OP_SET_ROLES_FROM_CLAIMS' and default
is False.

Adds new OIDC setting that will configure the token path for retrieving
the role info. The setting name is 'OIDC_OP_ROLE_CLAIM_PATH' and
defaults to "realm_access.roles" which is the standard Keycloak location
for user role info.

Added new OIDC setting that will configure what information is extracted
from the OIDC token. The setting name is 'OIDC_ACCESS_ATTRIBUTE_MAP'
and the setting contents should be JSON-decodable.

These settings should be configured individually for each provider by
appending the provider name for secondary providers as necessary.

### DIFF
--- a/src/archivematica/dashboard/components/accounts/backends.py
+++ b/src/archivematica/dashboard/components/accounts/backends.py
@@ -1,7 +1,11 @@
 import json
+from typing import Any
+from typing import Optional
 
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpRequest
 from django_auth_ldap.backend import LDAPBackend
 from django_cas_ng.backends import CASBackend
 from josepy.jws import JWS
@@ -44,7 +48,40 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
     Provide OpenID Connect authentication
     """
 
-    def get_settings(self, attr, *args):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # Store additional settings as instance attributes.
+        self.OIDC_OP_SET_ROLES_FROM_CLAIMS = getattr(
+            settings, "OIDC_OP_SET_ROLES_FROM_CLAIMS", False
+        )
+
+        self.OIDC_OP_ROLE_CLAIM_PATH = getattr(
+            settings, "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
+        )
+
+        self.OIDC_ACCESS_ATTRIBUTE_MAP = getattr(
+            settings, "OIDC_ACCESS_ATTRIBUTE_MAP", settings.DEFAULT_OIDC_CLAIMS
+        )
+
+        # Valid role claim name which may be extracted from OIDC token.
+        self.OIDC_ROLE_CLAIM_ADMIN = getattr(settings, "OIDC_ROLE_CLAIM_ADMIN", "admin")
+        self.OIDC_ROLE_CLAIM_DEFAULT = getattr(
+            settings, "OIDC_ROLE_CLAIM_DEFAULT", "default"
+        )
+        # Valid user roles.
+        self.USER_ROLE_ADMIN = "admin"
+        self.USER_ROLE_DEFAULT = "default"
+
+        # The roles are ordered from highest to lowest permission in this list.
+        # This feature is used in OIDC authentication to determine the highest
+        # permission role of a user based on the claims received when multiple
+        # roles are received.
+        self.USER_ROLE_TO_ROLE_CLAIM_MAP = {
+            self.USER_ROLE_ADMIN: self.OIDC_ROLE_CLAIM_ADMIN,
+            self.USER_ROLE_DEFAULT: self.OIDC_ROLE_CLAIM_DEFAULT,
+        }
+
+    def get_settings(self, attr: str, *args: Any) -> Any:
         if attr in [
             "OIDC_RP_CLIENT_ID",
             "OIDC_RP_CLIENT_SECRET",
@@ -53,6 +90,9 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
             "OIDC_OP_USER_ENDPOINT",
             "OIDC_OP_JWKS_ENDPOINT",
             "OIDC_OP_LOGOUT_ENDPOINT",
+            "OIDC_OP_SET_ROLES_FROM_CLAIMS",
+            "OIDC_OP_ROLE_CLAIM_PATH",
+            "OIDC_ACCESS_ATTRIBUTE_MAP",
         ]:
             # Retrieve the request object stored in the instance.
             request = getattr(self, "request", None)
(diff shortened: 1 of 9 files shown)


src/archivematica/dashboard/components/accounts/backends.py
```diff
@@ -44,7 +48,35 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
...
+            settings, "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
+        )
+
+        self.OIDC_ACCESS_ATTRIBUTE_MAP = getattr(
+            settings, "OIDC_ACCESS_ATTRIBUTE_MAP", settings.DEFAULT_OIDC_CLAIMS
+        )
+
+        # Valid user roles which may be extracted from OIDC token.
+        self.USER_ROLE_ADMIN = getattr(settings, "USER_ROLE_ADMIN", "admin")
+        self.USER_ROLE_DEFAULT = getattr(settings, "USER_ROLE_DEFAULT", "default")
```
replaceafill: Is there a way for the user to override these values from the `settings` modules?

src/archivematica/dashboard/components/accounts/backends.py
```diff
@@ -44,7 +48,35 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
...
+
+        # The roles are ordered from highest to lowest permission in this list.
+        # This feature is used in OIDC authentication to determine the highest
+        # permission role of a user based on the claims received when multiple
+        # roles are received.
+        self.USER_ROLES = [
+            (self.USER_ROLE_ADMIN, "Administrator"),
+            (self.USER_ROLE_DEFAULT, "Default"),
+        ]
```
replaceafill: Why does this need a two elements tuple if the loop below, which is the only part that accesses it, ignores the second element?

src/archivematica/dashboard/components/accounts/backends.py
```diff
@@ -110,10 +152,65 @@ def decode_token(token):
 
         return info
 
-    def create_user(self, user_info):
+    def create_user(self, user_info: dict[str, Any]) -> User:
```
replaceafill:
```suggestion
    def create_user(self, user_info: dict[str, Any]) -> Optional[User]:
```

src/archivematica/dashboard/components/accounts/backends.py
```diff
@@ -110,10 +152,65 @@ def decode_token(token):
...
+        If the setting is False roles are being managed by an admin so do not update the role.
+        """
+        if self.OIDC_OP_SET_ROLES_FROM_CLAIMS:
+            role = self.get_user_role(user_info)
+            if role is None:
+                return None
+            self.set_user_role(user, role)
+        return user
+
+    def set_user_role(self, user, role: str):
```
replaceafill:
```suggestion
    def set_user_role(self, user: User, role: str) -> None:
```

replaceafill (changes requested):
@sbreker I’m a bit unclear about the role handling implementation here. In the context of the Dashboard, it seems less intuitive than in the Storage Service, especially since we haven't introduced the concept of roles yet.

For example the `USER_ROLE_DEFAULT` setting doesn't seem to have any effect in the implementation.

Do you think it’s necessary to mirror the Storage Service implementation this closely, given that our main goal here is just to set the `user.is_superuser` attribute? I wonder if a simpler approach might be more appropriate for now.

src/archivematica/dashboard/components/accounts/backends.py
sbreker: There should be in case the role name coming in the token cannot be changed. I will fix this.

src/archivematica/dashboard/components/accounts/backends.py
sbreker: I wanted to keep this code the same as in the Storage Service as much as possible. I will update this so that it is a list.

sbreker:
> @sbreker I’m a bit unclear about the role handling implementation here. In the context of the Dashboard, it seems less intuitive than in the Storage Service, especially since we haven't introduced the concept of roles yet.
> 
> For example the `USER_ROLE_DEFAULT` setting doesn't seem to have any effect in the implementation.
> 
> Do you think it’s necessary to mirror the Storage Service implementation this closely, given that our main goal here is just to set the `user.is_superuser` attribute? I wonder if a simpler approach might be more appropriate for now.

USER_ROLE_DEFAULT is needed to validate the incoming role in the token. On the IDP side, the roles that an AM user belongs to will be sent in the token identifying an AM admin or an AM regular authenticated user. It is possible in a big institution that a person is not assigned a role that allows them to authenticate in AM - because of this we need to examine the token contents and make sure that even though their credentials are valid, if they do not have a valid AM role then they are not allowed access. So while we do not have roles in AM, we need these two values to validate what is sent from the IDP.

src/archivematica/dashboard/components/accounts/backends.py
replaceafill: I understand there's value in maintaining both implementation in sync. So don't worry about this.

sbreker:
> > @sbreker I’m a bit unclear about the role handling implementation here. In the context of the Dashboard, it seems less intuitive than in the Storage Service, especially since we haven't introduced the concept of roles yet.
> > For example the `USER_ROLE_DEFAULT` setting doesn't seem to have any effect in the implementation.
> > Do you think it’s necessary to mirror the Storage Service implementation this closely, given that our main goal here is just to set the `user.is_superuser` attribute? I wonder if a simpler approach might be more appropriate for now.
> 
> USER_ROLE_DEFAULT is needed to validate the incoming role in the token. On the IDP side, the roles that an AM user belongs to will be sent in the token identifying an AM admin or an AM regular authenticated user. It is possible in a big institution that a person is not assigned a role that allows them to authenticate in AM - because of this we need to examine the token contents and make sure that even though their credentials are valid, if they do not have a valid AM role then they are not allowed access. So while we do not have roles in AM, we need these two values to validate what is sent from the IDP.

@replaceafill I have added a commit that makes this a little more straightforward allowing for external role identifiers to be matched to internal roles.

src/archivematica/dashboard/components/accounts/backends.py
sbreker: I have added a commit that clarifies this. The internal value should not be modifiable - I have added new settings that map what is coming from the OIDC backend to internal roles.

src/archivematica/dashboard/components/accounts/backends.py
sbreker: Fixed - see latest commit.

src/archivematica/dashboard/components/accounts/backends.py
sbreker: Fixed - see latest commit.

src/archivematica/dashboard/install/README.md
```diff
@@ -741,6 +741,32 @@ set.
...
+  - **Type:** `boolean`
+  - **Default:** `False`
+
+- **`OIDC_OP_ROLE_CLAIM_PATH`**:
+  - **Description:** Set OIDC token path for extracting role info
+  - **Type:** `string`
+  - **Default:** `'realm_access.roles'`
+
+- **`OIDC_ROLE_CLAIM_ADMIN`**:
+  - **Description:** The OIDC role claim value which maps to a the Admin role.
```
replaceafill:
```suggestion
  - **Description:** The OIDC role claim value which maps to the Admin role.
```

replaceafill (approved):
Looks good. Thanks @sbreker.

src/archivematica/dashboard/install/README.md
sbreker: Fixed in this and the storage service README files.

PR #2048, opened 2025-04-24
Fixed the Problem (Step 2): Unable to Identifying Certain Data File T…

Meghanxuxx:
Related issue: https://github.com/archivematica/Issues/issues/1745

This PR is still continuing to solve the Issue [#1745](https://github.com/archivematica/Issues/issues/1745), the solution is split into two PRs because both the stroge service files and the archivematica files need to be modified, and here's the first part: https://github.com/artefactual/archivematica-storage-service/pull/772

If the step1 [PR](https://github.com/artefactual/archivematica-storage-service/pull/772) change is taken, the only issue left now is that files extracted using [extract_contents.py](https://github.com/artefactual/archivematica/blob/ad01b681f5bcd568d235121b7daaf2601a8f824f/src/archivematica/MCPClient/clientScripts/extract_contents.py#L214) will end up with different file paths than they did with function [extract_and_remove_bundle](https://github.com/artefactual/archivematica-storage-service/blob/ea58d030ae54642f60ee206b20129d97bda60380/src/archivematica/storage_service/locations/models/dataverse.py#L286). The new paths include dates like:

https://github.com/artefactual/archivematica/blob/ad01b681f5bcd568d235121b7daaf2601a8f824f/src/archivematica/MCPClient/clientScripts/extract_contents.py#L34
So we need to fix the [get_db_objects](https://github.com/artefactual/archivematica/blob/ad01b681f5bcd568d235121b7daaf2601a8f824f/src/archivematica/MCPClient/clientScripts/parse_dataverse_mets.py#L34) function in [parse_dataverse_mets.py](https://github.com/artefactual/archivematica/blob/ad01b681f5bcd568d235121b7daaf2601a8f824f/src/archivematica/MCPClient/clientScripts/parse_dataverse_mets.py#L1) to make sure all files can be correctly matched using the pattern **"[original_directory].zip-[timestamp]/[original_file_path]"**

This PR modifies the get_db_objects function, which now searches for files in the database using three different methods:
1. First tries using the complete METS path to find the file (existing method)
2. Then it uses timestamp pattern matching. If multiple matches are found, it selects the top-level path (newly added)
3. Tries matching by filename (existing method)

### DIFF
--- a/src/archivematica/MCPClient/clientScripts/parse_dataverse_mets.py
+++ b/src/archivematica/MCPClient/clientScripts/parse_dataverse_mets.py
@@ -3,23 +3,25 @@
 validate against the objects expected to be part of the SIP generated during
 transfer.
 """
-
 import json
 import os
 import uuid
 
 import django
 
+# databaseFunctions requires Django to be set up
+
 django.setup()
-import metsrw
-from django.core.exceptions import ValidationError
 from django.utils import timezone
 
-from archivematica.archivematicaCommon import databaseFunctions
-from archivematica.archivematicaCommon.archivematicaFunctions import get_file_checksum
-from archivematica.archivematicaCommon.custom_handlers import get_script_logger
-from archivematica.dashboard.main.models import Agent
-from archivematica.dashboard.main.models import File
+# archivematicaCommon
+from archivematicaFunctions import get_file_checksum
+from custom_handlers import get_script_logger
+import databaseFunctions
+from main.models import Agent, File
+import metsrw
+import re
+from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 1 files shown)


replaceafill:
@Meghanxuxx I think this has similar problems to https://github.com/artefactual/archivematica-storage-service/pull/772#issuecomment-2829061482. The `Test mpc-client` jobs are also failing because of the import changes in your branch.

sarah-mason:
Hi @Meghanxuxx - just letting your know we're going to test 1.18 in a few weeks and will stop accepting contributions for it. If you have any questions about the failing tests, do let us know. Thank you!

PR #2050, opened 2025-05-01
Allow OIDC role config by provider

sbreker:
Allow the OIDC role claim settings to be configured individually for each configured provider.


replaceafill (approved):
Looks good. Thanks @sbreker.

PR #2051, opened 2025-05-05
Add OIDC session refresh middleware

sbreker:
mozilla_django_oidc session refresh middleware will be available when OIDC authentication is in use. Session refresh will default to OFF.


replaceafill (approved):
Looks good @sbreker. Thanks!

PR #2066, opened 2025-05-29
Remove non-functional FPR features

replaceafill:
This removes the [DGI script](https://github.com/anarchivist/fiwalk-dgi/tree/master?tab=readme-ov-file#fiwalk-dgi-scripts) previously used by the `fiwalk` characterization command to run `FIDO` on files contained in disk images via a Python script. The script was never updated for Python 3 and appears to have been broken since at least Archivematica 1.9.

It also adds an irreversible data migration to remove the `Transcoding maildir to mbox` normalization command, along with its related event detail command and associated rules. These commands refer to scripts and paths that were removed from the codebase years ago and are no longer in use.

Connected to https://github.com/archivematica/Issues/issues/1749


sevein (approved):
Noice!

PR #2072, opened 2025-06-11
Reduce docker images sizes

DanielCosme:
WIP implementation on reducing the `mcp-client`, `mcp-server` and `dashboard` docker image size.

With the changes in this PR the size if the resulting images will be as follows:

```sh
REPOSITORY                             TAG       IMAGE ID       CREATED          SIZE
archivematica-dashboard                latest    f2f893aa80dc   11 minutes ago   774MB
archivematica-mcp-client               latest    deffe2421ff3   12 minutes ago   1.72GB
archivematica-mcp-server               latest    bcccf1cc4165   14 minutes ago   621MB
```

Changes are not definitive, as some testing is required.


sevein:
@replaceafill, I suggest closing this PR and instead moving toward distributing a single image that contains all dependencies. While separating images may eventually be a goal, I think that at this stage it would be premature optimization. Maintaining multiple images means we have to build, test, scan, and publish each one separately, which adds unnecessary overhead. In most deployment environments, including Kubernetes, users already rely on a shared container registry, so pulling the same image for different services is not a limitation. We can still run separate components using different entrypoints, while keeping the build and release process simpler and more consistent.

PR #2076, opened 2025-06-21
Add macOS support to dev environment

sevein:
> [!WARNING]  
> This PR has been replaced. I'll be submitting smaller PRs to address individual issues.

Already merged:

- [x] Use `clamav/clamav-debian:1.4.3-45` (published with ARM builds)
- [x] Use ES 8.x ARM builds (for now, search funcionality is removed)
- [x] Use percona and nginx images with ARM builds (https://github.com/artefactual/archivematica/pull/2102)
- [x] Update user management strategy to free macOS default uid/gid (to be discussed)
- [ ] Publish arm64 builds of tools in packages.archivematica.org

### DIFF
--- a/hack/Dockerfile
+++ b/hack/Dockerfile
@@ -18,13 +18,6 @@ ARG PYENV_DIR=/pyenv
 ENV DEBIAN_FRONTEND=noninteractive
 ENV PYTHONUNBUFFERED=1
 
-# Ubuntu 24.04 and later Docker images include a default user with UID (1000)
-# and GID (1000). Remove this user to prevent conflicts with the USER_ID and
-# GROUP_ID build arguments.
-RUN set -ex \
-	&& id -u ubuntu >/dev/null 2>&1 \
-	&& userdel --remove ubuntu || true
-
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 	--mount=type=cache,target=/var/lib/apt,sharing=locked \
 	set -ex \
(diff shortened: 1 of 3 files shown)


sevein:
Note to myself - some other issues I've seen:

#### compilemessages

This gets stuck forever (strace reads: futex wait):

    make manage-dashboard ARG="compilemessages -v 3"

But this works:

    make manage-dashboard ARG="compilemessages -l en"    

Tried without shared volumes - same issue. Also affects the build.

#### qemu version

1. I'm seeing random segfaults while pyenv is building Python 3.9. This solves the problem:

   ```
   docker run --privileged --rm tonistiigi/binfmt:qemu-v9.2.2 --uninstall amd64
   docker run --privileged --rm tonistiigi/binfmt:qemu-v9.2.2 --install amd64
   ```

2. But then the build fails again during the build of [vue frontend](https://github.com/artefactual/archivematica/pull/2082) (namely, esbuild - written in go - crashes), and siegfried will also crash during file format identification (related issue: https://github.com/golang/go/issues/69255). At that point you need to go back to qemu v7:

   ```
   docker run --privileged --rm tonistiigi/binfmt:qemu-v7.0.0 --uninstall amd64
   docker run --privileged --rm tonistiigi/binfmt:qemu-v7.0.0 --install amd64
   ```

This situation makes you question whether it's better to focus on building native arm64 images or to encourage the use of regular VMs instead. VMs are simpler to set up, but they don't offer the same developer experience on macOS - though, realistically, how many of us are actually using macOS for this anyway?

sevein:
I'm in the process of submitting smaller pull requests addressing individual issues.

PR #2077, opened 2025-06-23
Move db access logic out of the elasticSearchFunction module

replaceafill:
This updates the `index_aip_and_files` and `index_transfer_and_files` functions in the `archivematica.archivematicaCommon.elasticSearchFunction` module to receive the pipeline UUID, transfer details, and SIP identifiers as parameters, eliminating the need for direct database access to retrieve this information.

It also introduces a separate class to encapsulate the logic for parsing the AIP METS to extract AIP metadata such as creation date, AIC identifier or `isPartOf` attribute, which is used when indexing the AIP and its files.

Connected to https://github.com/archivematica/Issues/issues/1752


sevein (approved):
Nice cleanup! 👍

PR #2081, opened 2025-07-04
Improve Docker build performance with caching

replaceafill:
This optimizes Docker build performance by:

- Using BuildKit cache mounts for apt and pip in the `hack/Dockerfile` and removing manual cache cleanup steps.
- Adding a separate job in the `test.yml` GitHub workflow to pre-build and cache the `pyenv-builder` stage layers and using cache keys based on requirements file changes and build contexts.


sevein (approved):
LGTM

PR #2086, opened 2025-07-25
Update OIDC var description in README

sbreker:
Provide additional detail for setting: `OIDC_ACCESS_ATTRIBUTE_MAP`.

### DIFF
--- a/src/archivematica/dashboard/install/README.md
+++ b/src/archivematica/dashboard/install/README.md
@@ -766,8 +766,7 @@ set.
 - **`OIDC_ACCESS_ATTRIBUTE_MAP`**
   - **Description:** Set OIDC token details to extract. This string should be
     JSON-decodable.  If `OIDC_OP_SET_ROLES_FROM_CLAIMS` is set to `True` then
-    `"realm_access": "realm_access"` must be added to the default config for
-    this setting.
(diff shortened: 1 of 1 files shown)


src/archivematica/dashboard/install/README.md
```diff
@@ -765,7 +765,9 @@ set.
 
 - **`OIDC_ACCESS_ATTRIBUTE_MAP`**
   - **Description:** Set OIDC token details to extract. This string should be
-    JSON-decodable.
+    JSON-decodable.  If `OIDC_OP_SET_ROLES_FROM_CLAIMS` is set to `True` then
+    `"realm_access": "realm_access"` must be added to the default config for
+    this setting.
```
replaceafill: Nitpick: I find the "default" part of the new sentence a bit hard to understand. When users customize and provide their own setting value it's not the **default** anymore, right? What about something along the lines of:

```suggestion
    JSON-decodable.  If `OIDC_OP_SET_ROLES_FROM_CLAIMS` is set to `True` then
    the entry `"realm_access": "realm_access"` must be included in this setting.
```

replaceafill (approved):
Thanks @sbreker. I recommended an alternative sentence, but yours is good.

sbreker:
@replaceafill I have accepted your suggestion and added as a commit

PR #2087, opened 2025-07-31
Remove wiki references from README

sarah-mason:
Updating references and wording in the README.md to remove any references to the Archivematica wiki and instead point to current resources such as the Archivematica documentation or contributing guidelines.


sarah-mason:
Ready for review @replaceafill

replaceafill (approved):
Thank you @sarah-mason! I just made a small edit to the paragraph about the `RELEASENOTES` file so we don't have to update it on every release.

PR #2090, opened 2025-08-06
Add current password validation to user edit form

replaceafill:
This updates the user edit form to require the current password from the user setting a new password. It also improves the form field ordering and labels to enhance the user experience. Tests have been added to cover all scenarios handled by the view and the form.

Connected to https://github.com/archivematica/Issues/issues/1757


sevein (approved):
LGTM

PR #2092, opened 2025-08-08
Fix cleanup job after SIP arrangement

replaceafill:
The `Clean up after storing AIP` job currently fails to contact the Storage Service after arranging a SIP, resulting in the backlog transfer not being deleted and the post-store callback not being triggered for the stored AIP.

imagen

Captura desde 2025-08-08 12-13-23


sevein (approved):
👍

PR #2093, opened 2025-08-13
Refactor search functionality with service layer

replaceafill:
This pull request consolidates Archivematica's search logic into a single service layer, replacing scattered implementations while preserving full backward compatibility.

The new search service centralizes all Elasticsearch operations behind a consistent interface, replacing the previous approach where search code was dispersed across multiple modules.

MCPClient scripts and dashboard components, such as views and management commands, no longer make direct Elasticsearch calls. Instead, they now use the service for AIP indexing, backlog handling, post-storage hooks, and index maintenance or rebuilding.

The legacy search modules have been removed, making future Elasticsearch upgrades easier to maintain.

Connected to https://github.com/archivematica/Issues/issues/1752


sevein (approved):
Nice job! 👍

PR #2097, opened 2025-08-20
Remove warnings about autocreated index fields

scollazo:
Removes models.W042 warning when running manage.py commands

Connected to https://github.com/archivematica/Issues/issues/1759

### DIFF
--- a/src/archivematica/dashboard/settings/base.py
+++ b/src/archivematica/dashboard/settings/base.py
@@ -753,3 +753,6 @@ if PROMETHEUS_ENABLED:
 
 # Apply email settings
 globals().update(email_settings.get_settings(config))
+
+# Remove models.W042 warning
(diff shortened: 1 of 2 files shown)


src/archivematica/dashboard/settings/base.py
```diff
@@ -709,3 +709,6 @@ def _get_settings_from_file(path):
 
 # Apply email settings
 globals().update(email_settings.get_settings(config))
+
+# Remove models.W042 warning
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
replaceafill: I'd prefer to have a way to customize the [SILENCED_SYSTEM_CHECKS](https://docs.djangoproject.com/en/4.2/ref/settings/#silenced-system-checks) setting from configuration. That way we could turn the warning off in production but keep it for development in case we add new models.

PR #2102, opened 2025-09-01
Update Docker images in dev services

sevein:
This commit updates several Docker images while ensuring backward compatibility. The Percona and Nginx images used support ARM64 builds.

Pinning images in Docker Compose requires updating them regularly, but Dependabot can handle that. At Artefactual, we've generally adopted this practice because it provides reproducibility without the risk of falling behind on updates, and the same approach could make sense here.

Relates to https://github.com/artefactual/archivematica/pull/2076.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  - package-ecosystem: "docker-compose"
+    directory: "/hack"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 1
+    groups:
+      docker-compose-dev:
+        patterns:
+          - "*"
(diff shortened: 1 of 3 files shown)


replaceafill (approved):
This is great, thanks @sevein!

Could this approach be extended to update the versions in the `docker-compose.yml` files in the integration tests?
* [Archivematica](https://github.com/artefactual/archivematica/blob/52385843c5d2bd02973c7d7e11246f1230e3f87c/tests/integration/docker-compose.yml)
* [Storage Service](https://github.com/artefactual/archivematica-storage-service/blob/4c0f5152f9ac69dcbc61c25e9c6b4a80c629f359/tests/integration/docker-compose.yml)

PR #2103, opened 2025-09-02
Prevent UID/GID clashes more robustly in Docker builds

sevein:
The previous method for preventing UID/GID clashes in the Docker build process was limited to removing a specific "ubuntu" user with default IDs. This approach was insufficient in environments where different default UIDs/GIDs are used, or when user IDs on the host system (such as common macOS UIDs like 501 or GIDs like 20 for "staff") clash with IDs that might be present in the container. This change dynamically checks for and remove any existing user/group that occupies the requester `USER_ID` or `GROUP_ID`, ensuring that the "archivematica" user can be created with the specified IDs without conflict.


replaceafill (approved):
LGTM :+1:

PR #2108, opened 2025-09-08
Disable Gunicorn's autoreload by default

sevein:
For a long time we've been asking Gunicorn to "poll" the filesystem for changes,
which is very inefficient. When your dev machine is idle, there's still a
significant amount of load just to keep the Gunicorn apps up to date with local
filesystem changes.

While this may be desirable in some development workflows, in most cases you
don't actually need it. Disabling it by default keeps your laptop from getting
warm and saves a lot of energy. If you don't believe me - try it!

Gunicorn would be far more efficient if it could use the inotify API to detect
changes. This feature exists but that doesn't work with our gevent config. This
problem was reported back in 2017 (https://github.com/benoitc/gunicorn/issues/1494).
Eight years later, I'm suggesting we make the poll-based auto-reload opt-in by
default.

Use the Compose's `.env` file to enable `AM_GUNICORN_RELOAD` and
`SS_GUNICORN_RELOAD` just when you need it. Remember to recreate the
`archivematica-dashboard` and `archivematica-storage-services` when
changing the configuration.


replaceafill (approved):
Thanks @sevein!

PR #2110, opened 2025-09-11
Monitor GitHub Actions with Dependabot

sevein:


replaceafill (approved):
Just as I started wondering about how to maintain the hashes I reached the `Enforce commit SHA pinning` job :grin: Thanks @sevein!

PR #2119, opened 2025-10-03
Use dynamic OS codename and architecture in Dockerfile

replaceafill:
This ensures compatibility across different `arm64` and `amd64` Ubuntu builds.


sevein (approved):
LGTM

PR #2145, opened 2025-12-03
Upgrade Django to 5.2

replaceafill:
This also adds support for Python 3.13 and 3.14.

Connected to https://github.com/archivematica/Issues/issues/1773


sevein (approved):
👍

PR #2160, opened 2026-01-06
Fix Log out link

replaceafill:
[Django 5.0](https://docs.djangoproject.com/en/5.2/releases/5.0/#features-removed-in-5-0) requires the logout requests to be POST not GET:

> Support for logging out via `GET` requests in the `django.contrib.auth.views.LogoutView` and `django.contrib.auth.views.logout_then_login()` is removed.

This replaces the `Log out` link in the navigation bar with a form and submit button.

Connected to https://github.com/archivematica/Issues/issues/1773


sevein (approved):
👍

PR #2162, opened 2026-01-09
Fix processing config API endpoint paths

replaceafill:
This corrects typos in the frontend's processing configuration list API endpoint URLs. These were masked because the urls module route regex was overly permissive, so the broken paths still matched.

It also adds a job in the `test.yml` workflow for running the tests of the Vue application.


sevein (approved):
👍

PR #2164, opened 2026-01-13
Fix metadata-only reingest to reprocess updated metadata.csv

thales0591:
Related issue: https://github.com/archivematica/Issues/issues/1771

### Overview

- Problem: `add_new_files` only considers files “new” if they are absent from METS; on second reingest, `metadata.csv` already exists in METS so it is skipped even when content changes.
- Root cause: metadata.csv is persisted at `metadata.csv` after first reingest, so subsequent reingests reuse the same METS path and are ignored by the “new file” check.
- Fix: compare the on-disk checksum of metadata.csv with the METS PREMIS fixity; when the content differs, reprocess `metadata.csv` even if the file already exists in METS.
- Implementation: add checksum/fixity helpers and trigger `update_metadata_csv` on content change without forcing a full “new file” path.
- Tests: add a regression test that simulates a METS with existing `metadata.csv` and verifies it is reprocessed when content changes.

Obs: I've just sent in the signed contributor's agreement

### DIFF
--- a/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSReingest.py
+++ b/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSReingest.py
@@ -447,10 +447,8 @@ def add_new_files(job, mets, sip_uuid, sip_dir):
 
     If a new file is a metadata.csv, parse it to create dmdSecs.
     """
-    # Find new files
-    # How tell new file from old with same name? Check hash?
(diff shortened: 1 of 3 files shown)


replaceafill:
@thales0591 Thank you for this contribution! We received your contributors agreement.

I've triggered the CI workflows on the pull request, and I added the `triage-1.19` label to the linked issue for the development team to discuss if it'll be included in the next release. I'll try to code review it as soon as possible.

PR #2179, opened 2026-01-27
Drop jobs select2 modal

sevein:
Select2 is only used once in the Dashboard, to show more than 10 options during processing (such as AIP or DIP storage locations). A native dropdown is good enough for this use case, so maintaining the Select2 integration is not necessary.


src/archivematica/dashboard/media/js/ingest.js
```diff
@@ -224,13 +224,6 @@ $(function()
             var $select = $('<select />').append('<option>' + gettext('Actions') + '</option>')
               , numberOfChoices = Object.keys(choices).length;
 
-            // use pop-up action selector for long choice lists
-            if (numberOfChoices >= 10)
```
replaceafill, line 228: Could the `numberOfChoices` assignment above be removed? It doesn't seem to be used anywhere else anymore. Same in the `transfer.js` file.

replaceafill (approved):
Thank you @sevein!

PR #2178, opened 2026-01-26
Migrate filesystem browsers to Vue

sevein:
This pull request rewrites the treeview browsers in "Add metadata files" (lib/md-editor) and "Review AIP" (lib/aip-browser) using Vue. It introduces a new TreeView.vue component based on Reka UI with shared functionality. The transfer browser (lib/browser) has also been updated to leverage this shared component.

I could continue working on incremental improvements, but the changes are in a good state for review now. The `lib/aip-browser` implementation is relatively small and straightforward, while `lib/md-editor` is larger and more complex. I have started breaking it down into smaller components and composables, though there is likely more refactoring that could be done down the road. Similarly, the already merged `lib/browser` could benefit from further improvements.

PR #2180, opened 2026-01-27
Create Vue app to replace {online,status}.js

sevein:
This pull request adds a small Vue app to show MCP connection status and
pending jobs in the Transfer and Ingest tabs, demonstrating how Vue
components can be injected into the existing Dashboard to gradually
replace Backbone.js code. It uses Vue Teleport to attach the Vue
components to the existing DOM from the Django template.


.gitattributes
```diff
@@ -0,0 +1,2 @@
+src/archivematica/dashboard/locale/**/*.po linguist-generated=true
```
replaceafill, line 1: Thanks for adding these!

src/archivematica/dashboard/media/js/jobs.js
```diff
@@ -881,14 +879,6 @@ BaseAppView = Backbone.View.extend({
         dataType: 'text',
         type: 'GET',
         url: this.statusUrl + '?' + new Date().getTime(),
-        beforeSend: function()
-          {
-            window.statusWidget.startPoll();
-          },
-        error: function()
-          {
-            window.statusWidget.text(gettext('Error trying to connect to database. Trying again...'), true);
```
replaceafill, line 817: I noticed the new widget now only says `Disconnected` for both of the previous error cases in this file:
* `Error trying to connect to database. Trying again...`
* `Error trying to connect to MCP server. Trying again...`

Do we want to keep that separation for debugging?

src/archivematica/dashboard/vue/lib/topbar/components/ConnectionStatus.vue
```diff
@@ -0,0 +1,72 @@
...
+  if (state.connected) {
+    return t('topbar.connected')
+  }
+  return t('topbar.disconnected')
+})
+
+const connectionIconClass = computed(() => {
+  if (state.loading) return 'status-icon--loading'
+  if (state.connected === null) return 'status-icon--initializing'
+  return state.connected ? 'status-icon--connected' : 'status-icon--disconnected'
```
replaceafill: What's the meaning of the double dashes in these class names?

src/archivematica/dashboard/vue/lib/topbar/components/NavCounts.vue
```diff
@@ -0,0 +1,63 @@
+<script setup lang="ts">
+import { computed, onMounted, ref } from 'vue'
+import { useStatus } from '@/topbar/composables/useStatus'
+
+const { state } = useStatus()
+
+const transferSelector = 'ul.nav > li > a[href="/transfer/"]'
+const ingestSelector = 'ul.nav > li > a[href="/ingest/"]'
```
replaceafill: I know these selectors match the previous approach, but they seem fragile. Would it be better to add stable hooks like IDs or `data-*` attributes on the nav links and target those instead?

replaceafill (commented):
This looks good to me @sevein. Thanks! I just left a question about the change in the error messages that were displayed in the previous version. Let me know what you think.

And I like the new shared HTTP client you added. I wonder if some of the existing functionality in the new transfer browser could be abstracted the same way. For example the `fetchWithCsrf` in `useTransferAPI.ts`. I don't have a clear idea about it but would it make sense to have an API client defined as a shared library that other components can use? Of course, we don't have to do it in this pull request.

src/archivematica/dashboard/vue/lib/topbar/components/NavCounts.vue
sevein: Good idea!

src/archivematica/dashboard/media/js/jobs.js
sevein: 👍

sevein:
> And I like the new shared HTTP client you added. I wonder if some of the existing functionality in the new transfer browser could be abstracted the same way. For example the fetchWithCsrf in useTransferAPI.ts. I don't have a clear idea about it but would it make sense to have an API client defined as a shared library that other components can use? Of course, we don't have to do it in this pull request.

Yes, this is coming! In https://github.com/artefactual/archivematica/pull/2178 all Vue code should be using that http package and we'll need to deal with CSRF, etc. Thank you!

src/archivematica/dashboard/vue/vite.config.ts
```diff
@@ -92,6 +92,18 @@ export default defineConfig(({ command }) => {
             formats: ['es'],
           },
           cssCodeSplit: true,
+          rollupOptions: {
+            output: {
+              chunkFileNames: (chunkInfo) => {
+                const facade = chunkInfo.facadeModuleId
+                if (facade && facade.includes('/lib/shared/i18n/locales/')) {
+                  const locale = basename(facade).replace('.json', '')
+                  return `locale-${locale}-[hash].js`
```
replaceafill, line 101: Ah nice! I was wondering about those strange names during `npm run build` :sweat_smile:

replaceafill (approved):

PR #2181, opened 2026-01-27
Remove the 1x1 placeholder image

sevein:
Missing thumbnails now return 404 and the UI replaces failed thumbnail loads with a localized “N/A” label, removing the need for the 1x1 placeholder asset.

image


src/archivematica/dashboard/media/css/archival_storage.css
```diff
@@ -23,6 +23,7 @@
   width: 150px;
 }
 
+
```
replaceafill: Is this a leftover?

replaceafill (approved):
:+1:

PR #2187, opened 2026-02-04
Delete unused code

sevein:
We should try to understand how people were using this pagination feature, but it seems pretty useless since you had to manually add a URL parameter.


replaceafill (approved):
LGTM.

sevein:
For the record, we've identified the background behind the client pagination capability. Back in 2013, there was a report of high CPU usage in the ingest tab (browser), and client pagination was introduced at the time as a solution, presumably to avoid rendering too many SIPs and their associated jobs simultaneously. While this may still be partially true, especially since the code has not changed significantly since then, this is something we should be able to revisit in the future, along with other inefficiencies, when we have the opportunity to rewrite it.

PR #2188, opened 2026-02-05
Delete unused filename fragment from template

sevein:
This commit updates the legacy ingest job-row template fragment by removing the conditional (file: <filename>) rendering, since the current ingest status payload does not include a job filename field. The change is a cleanup of unreachable template logic and is not expected to alter runtime behavior.

Related: https://github.com/artefactual/archivematica-history/commit/a7b8878318e9e7d4891dd0deb79c56feb6b81f05

### DIFF
--- a/src/archivematica/dashboard/templates/ingest/grid.html
+++ b/src/archivematica/dashboard/templates/ingest/grid.html
@@ -91,10 +91,7 @@
   <script type="text/template" id="job-template">
     <div class="job-detail-microservice">
       <span class="job-type-label">{% trans "Job" %}</span>
-      <span title="<%= uuid %>">
-        <%= type %>
-        <% if (obj.filename) { %> (file: <i><%= filename %></i>)<% } %>
-      </span>
+      <span title="<%= uuid %>"><%= type %></span>
       <% tmp = window.microservices_help[type]; %>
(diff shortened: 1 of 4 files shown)


src/archivematica/dashboard/templates/ingest/grid.html
```diff
@@ -91,10 +91,7 @@
   <script type="text/template" id="job-template">
     <div class="job-detail-microservice">
       <span class="job-type-label">{% trans "Job" %}</span>
-      <span title="<%= uuid %>">
-        <%= type %>
-        <% if (obj.filename) { %> (file: <i><%= filename %></i>)<% } %>
-      </span>
+      <span title="<%= uuid %>"><%= type %></span>
```
replaceafill, line 94: I see. This is similar to the change to `transfer/grid.html` in https://github.com/artefactual/archivematica/commit/d069446332069bb33e506026f38cbc51aeba8f12

replaceafill (approved):
👍

src/archivematica/dashboard/templates/ingest/grid.html
sevein: Ha wow!

PR #2190, opened 2026-02-05
Implement processing monitor using Vue

sevein:
This pull request introduces a new Vue-based process monitor for the
Transfer and Ingest tabs, intended as a drop-in replacement for the legacy
monitor with matching behavior and visual parity.

image


src/archivematica/dashboard/components/ingest/views.py
```diff
@@ -62,18 +62,20 @@ def ingest_grid(request):
         messages.warning(
             request,
             _(
-                "Error retrieving originals directory locations: is the storage server running? Please contact an administrator."
+                "Error retrieving originals directory locations: '"
```
replaceafill: There's a new `'` at the end here.

src/archivematica/dashboard/vue/lib/shared/http/unit.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (typeof value === 'boolean') {
+    return { removed: value }
+  }
+  if (Array.isArray(value) && value.every(item => typeof item === 'string')) {
+    return { removed: value }
+  }
+  throw new Error('Unexpected "removed" field type in delete response.')
+}
+
+const tryParseConflictResponse = (err: unknown): UnitDeleteResponse | null => {
```
replaceafill: Is the `try` needed in this name or would `parseConflictResponse` work like it does above in `parseDeleteResponse`?

replaceafill (changes requested):
This looks great @sevein. I only noticed a problem with the `Hide non-tasked jobs` functionality you introduced recently. The jobs always display the cogs.

Image
Image

The separation between transfer and ingest seems repetitive, maybe there could be a centralized `unit` abstraction/component but I understand there are difference like the transfer browser, etc.

And I'd suggest making the `?ng=1` approach the default and removing the legacy components.

src/archivematica/dashboard/vue/lib/shared/http/unit.ts
sevein: The try prefix follows what it seems to be a naming convention in which tryX indicates a best-effort operation that does not throw, unlike X, which is typically stricter and may throw on invalid input. In this case, we are already inside a catch block and want to safely probe whether the error matches a specific 409 shape. If it does not we simply return null so we don't see us having to deal with another layyer of exception handling.

I'm going to rename as `safeParseConflictResponse` since safeX seems to be even more common than safeX, e.g. see [safeParse](https://github.com/colinhacks/zod/blob/7d98c909329713cb2f478620f8a67aaf3ef40ce2/packages/zod/src/v4/mini/schemas.ts#L31) in Zod.

sevein:
@replaceafill, feedback addressed in [1531390](https://github.com/artefactual/archivematica/pull/2190/commits/1531390287d0d28a1543968a32423c1364abef43):

* The cog is now hidden when expected.
* `lib/monitor` is now shared in both Transfer and Ingest pages.
* Old code removed, `?ng=1` behavior is now gone.

I've also added a few commits that make the polling bheavior feel much snappier. Let me know what you think:
1. Polling pauses when the document is not visible and resumes with an immediate refresh when it becomes visible again (see [visibilitychange](https://caniuse.com/?search=visibilitychange)),
2. The next poll is accelerated when a user executes a job action, and
3. The next poll is accelerated when a new transfer is submitted.

src/archivematica/dashboard/vue/lib/shared/assets/icons.ts
```diff
@@ -0,0 +1,21 @@
+import accept from '../../../../media/images/accept.png'
+import arrowRefresh from '../../../../media/images/arrow_refresh.png'
+import bell from '../../../../media/images/bell.png'
+import cancel from '../../../../media/images/cancel.png'
+import cog from '../../../../media/images/cog.png'
+import deleteIcon from '../../../../media/images/delete.png'
+import tableEdit from '../../../../media/images/table_edit.png'
+import zoom from '../../../../media/images/zoom.png'
```
replaceafill, line 8: Are these relative paths breaking the build? See https://github.com/artefactual/archivematica/actions/runs/22186388711

replaceafill (approved):
👏

PR #2192, opened 2026-02-05
Replace raster logos with SVG

sevein:
image


replaceafill (changes requested):
Thanks @sevein. This looks good, but I noticed the new svg files do not have the trademark symbol on the right like the old raster logos do. I looked in the Artefactual's logos directory and couldn't find any svg logos with it either.

replaceafill:
@sevein What do you think of these?

logo-light-v2
logo-dark-v2

replaceafill (approved):

PR #2193, opened 2026-02-08
Migrate Dublin Core contextual help into forms

sevein:
This makes contextual_help.js redundant and removes contextual help setup from templates.


replaceafill (approved):
LGTM.

PR #2201, opened 2026-02-14
Hide non-tasked jobs

sevein:
This pull request updates MCPServer to include an additional field in the job listing that indicates when a job does not produce tasks. The dashboard uses this field to hide the Tasks button for those jobs, so users are not redirected to an empty tasks page.

image

### DIFF
--- a/src/archivematica/MCPServer/server/jobs/base.py
+++ b/src/archivematica/MCPServer/server/jobs/base.py
@@ -32,6 +32,9 @@ class Job(metaclass=abc.ABCMeta):
     STATUS_EXECUTING_COMMANDS = models.Job.STATUS_EXECUTING_COMMANDS
     STATUS_FAILED = models.Job.STATUS_FAILED
 
+    # Whether this job type writes rows in the Tasks table.
+    produces_tasks = False
+
     def __init__(self, job_chain, link, package):
         self.uuid = uuid.uuid4()
         self.job_chain = job_chain
--- a/src/archivematica/MCPServer/server/jobs/client.py
+++ b/src/archivematica/MCPServer/server/jobs/client.py
@@ -35,6 +35,8 @@ class ClientScriptJob(Job, metaclass=abc.ABCMeta):
     For task details, see the `Task` and `GearmanTaskBackend` classes.
     """
 
+    produces_tasks = True
+
     # The number of files we'll pack into each MCP Client job.  Chosen somewhat
     # arbitrarily, but benchmarking with larger values (like 512) didn't make much
     # difference to throughput.
--- a/src/archivematica/MCPServer/server/rpc_server.py
+++ b/src/archivematica/MCPServer/server/rpc_server.py
@@ -28,6 +28,7 @@ from archivematica.archivematicaCommon.gearman_encoder import JSONDataEncoder
 from archivematica.dashboard.main.models import SIP
 from archivematica.dashboard.main.models import Job
 from archivematica.dashboard.main.models import Transfer
+from archivematica.MCPServer.server.jobs.chain import get_job_class_for_link
 from archivematica.MCPServer.server.packages import create_package
 from archivematica.MCPServer.server.packages import get_approve_transfer_chain_id
 from archivematica.MCPServer.server.processing_config import get_processing_fields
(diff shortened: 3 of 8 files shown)


src/archivematica/MCPServer/server/workflow.py
```diff
@@ -47,6 +47,13 @@ def _invert_job_statuses():
 # where we're using labels instead of IDs.
 _STATUSES = _invert_job_statuses()
 
+# Workflow managers that run client scripts and can generate Tasks rows.
+TASK_PRODUCING_MANAGERS = {
```
replaceafill: A couple of thoughts:

I don't think workflow module should be the place where the decision is made, and I don't think the decision should be based on the manager type.

From your pull request description:

> when a job does not produce tasks.

I think the job should be starting point of the decision, and looking at the logic of the current job subclasses the only one that creates tasks is the `ClientScriptJob`.

We could add a new boolean attribute `Job.produces_tasks = False` and have the `ClientScriptJob` subclass to set it to `True`. In the `RPCServer._units_statuses_handler` method you could use the link manager to get the job class attribute. Something along the lines of:

```diff
diff --git a/src/archivematica/MCPServer/server/jobs/base.py b/src/archivematica/MCPServer/server/jobs/base.py
index 5847d4f0a..7a3150c1c 100644
--- a/src/archivematica/MCPServer/server/jobs/base.py
+++ b/src/archivematica/MCPServer/server/jobs/base.py
@@ -31,6 +31,8 @@ class Job(metaclass=abc.ABCMeta):
     STATUS_COMPLETED_SUCCESSFULLY = models.Job.STATUS_COMPLETED_SUCCESSFULLY
     STATUS_EXECUTING_COMMANDS = models.Job.STATUS_EXECUTING_COMMANDS
     STATUS_FAILED = models.Job.STATUS_FAILED
+    # Whether this job type writes rows in the Tasks table.
+    produces_tasks = False
 
     def __init__(self, job_chain, link, package):
         self.uuid = uuid.uuid4()
diff --git a/src/archivematica/MCPServer/server/jobs/client.py b/src/archivematica/MCPServer/server/jobs/client.py
index 334a4a34e..1947a7344 100644
--- a/src/archivematica/MCPServer/server/jobs/client.py
+++ b/src/archivematica/MCPServer/server/jobs/client.py
@@ -34,6 +34,7 @@ class ClientScriptJob(Job, metaclass=abc.ABCMeta):
     Tasks are batched into groups of TASK_BATCH_SIZE, sent to mcp client via gearman.
     For task details, see the `Task` and `GearmanTaskBackend` classes.
     """
+    produces_tasks = True
 
     # The number of files we'll pack into each MCP Client job.  Chosen somewhat
     # arbitrarily, but benchmarking with larger values (like 512) didn't make much
diff --git a/src/archivematica/MCPServer/server/rpc_server.py b/src/archivematica/MCPServer/server/rpc_server.py
index f71ad87f1..11c36c51d 100644
--- a/src/archivematica/MCPServer/server/rpc_server.py
+++ b/src/archivematica/MCPServer/server/rpc_server.py
@@ -28,6 +28,7 @@ from archivematica.archivematicaCommon.gearman_encoder import JSONDataEncoder
 from archivematica.dashboard.main.models import SIP
 from archivematica.dashboard.main.models import Job
 from archivematica.dashboard.main.models import Transfer
+from archivematica.MCPServer.server.jobs.chain import get_job_class_for_link
 from archivematica.MCPServer.server.packages import create_package
 from archivematica.MCPServer.server.packages import get_approve_transfer_chain_id
 from archivematica.MCPServer.server.processing_config import get_processing_fields
@@ -404,6 +405,12 @@ class RPCServer(GearmanWorker):
                 )
                 new_job["microservicegroup"] = link.get_label("group", lang)
                 new_job["type"] = link.get_label("description", lang)
+                try:
+                    new_job["produces_tasks"] = get_job_class_for_link(
+                        link
+                    ).produces_tasks
+                except ValueError:
+                    new_job["produces_tasks"] = False
                 try:
                     new_job["choices"] = _pull_choices(
                         str(job_.jobuuid), lang, jobs_awaiting_for_approval
``` 

Let me know if that makes sense to you.

sevein: I agree, thanks! It keeps the responsibility closer to runtime behavior.

replaceafill (approved):

PR #2205, opened 2026-02-17
Remove dead rights-holder model/field

sevein:
This change removes the legacy rightsholder linkage (fkAgent) and the unused RightsStatementLinkingAgentIdentifier table because they were no longer part of active rights capture, display, import, or export workflows.

### DIFF
--- a/src/archivematica/dashboard/components/rights/views.py
+++ b/src/archivematica/dashboard/components/rights/views.py
@@ -15,10 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with Archivematica.  If not, see <http://www.gnu.org/licenses/>.
 import logging
-import re
 
 from django.forms.models import inlineformset_factory
-from django.http import HttpResponse
 from django.shortcuts import redirect
 from django.shortcuts import render
 from django.urls import reverse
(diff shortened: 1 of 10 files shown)


src/archivematica/dashboard/components/rights/views.py
```diff
@@ -90,25 +88,6 @@ def ingest_rights_grants_edit(request, uuid, id):
     return rights_grants_edit(request, uuid, id, "ingest")
 
 
-def rights_parse_agent_id(input):
-    return 0
```
sevein, line 94: !!!

src/archivematica/dashboard/components/rights/views.py
```diff
@@ -168,10 +136,7 @@ def rights_edit(request, uuid, id=None, section="ingest"):
         )
     else:
         if request.method == "POST":
-            postData = request.POST.copy()
-            agentId = rights_parse_agent_id(postData.get("rightsholder"))
```
sevein, line 172: No-op.

src/archivematica/dashboard/media/js/repeating-ajax-data.js
```diff
@@ -131,102 +131,17 @@ Type will default to textarea.
 var RepeatingDataView = Backbone.View.extend({
```
sevein, line 131: Rights pages only use this view to load and update existing child rows.

Creation/deletion controls were dead paths.

src/archivematica/dashboard/media/js/rights_edit.js
```diff
@@ -314,35 +291,4 @@ $(document).ready(function() {
...
-
-  // establish rightsholder autocomplete
-  if ($('#id_rightsholder').length > 0) {
-    // lookup rightsholder
-    $.get('lookup/rightsholder/' + $('#id_rightsholder').val(), function(data) {
-      $('#id_rightsholder').val(data);
-    });
-
-    // attach autocomplete
-    $("#id_rightsholder").autocomplete({
```
sevein, line 326: Looks like we once had a "rights holder" linkage workflow (autocomplete UI, lookup endpoints, and DB relationship), but it was never fully active in practice and has effectively been dead for years.

replaceafill (approved):
Thanks @sevein!

PR #2210, opened 2026-02-20
Fix TypeError in _capture_transfer_failure

liam-lloyd:
Currently, _capture_transfer_failure makes a call to `isinstance` with three arguments. `isinstance` takes only two arguments, so this results in a TypeError. This commit updates the `isinstance` call to instead use a tuple to pass multiple classes. It also adds unit tests for _capture_transfer_failure.


liam-lloyd:
This address https://github.com/archivematica/Issues/issues/1784

replaceafill (approved):
Thank you @liam-lloyd!

PR #2211, opened 2026-02-20
Fix AS matcher JSON serialization in dashboard views

replaceafill:
Follow-up to 45865b1245cc15982cc3cf0fab57d0ee838a2176, which covered UUID stringification in pair_matcher DIP object paths and MCPClient AS upload pairs.

This change covers the remaining Dashboard ArchivesSpace match/review view paths that consume persisted DIP-object/resource pairings:

- Match DIP objects to resource component levels.
- Match DIP objects to resource levels.
- Review current DIP-object-to-resource matches.


sevein (approved):
LGTM

PR #2214, opened 2026-02-23
Migrate FPR tables from DataTables to Vue/TanStack

sevein:
image


src/archivematica/dashboard/fpr/templatetags/revisions_link.py
```diff
@@ -31,11 +32,12 @@ def render(self, context):
         revision_type = self._resolve_if_template_variable(self.revision_type, context)
         object_uuid = self._resolve_if_template_variable(self.object_uuid, context)
 
-        return '<a class="revisions_link" href="{}">Revision history</a>'.format(
+        return '<div class="well well-sm" style="margin-bottom: 12px;"><a class="btn btn-default btn-xs" href="{}">{}</a></div>'.format(
```
replaceafill: Should this use [format_html](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.html.format_html) to build the link so interpolated values (including translated text) are escaped?

```python
  url = reverse(
      "fpr:revision_list",
      kwargs={"entity_name": revision_type, "uuid": object_uuid},
  )
  return format_html(
      '<div class="well well-sm" style="margin-bottom: 12px;"><a class="btn btn-default btn-xs" href="{}">{}</a></div>',
      url,
      _("Revision history"),
  )
```

replaceafill (commented):
@sevein Is the search now substring based? I used to be able to search rules (e.g validation) by using the format version (e.g. 'jpeg 1.02') but now when I start typing `jpeg 1` no results are displayed anymore.

src/archivematica/dashboard/fpr/payloads.py
```diff
@@ -0,0 +1,588 @@
...
+
+    is_superuser = request.user.is_superuser
+    rows: list[PayloadRow] = []
+    for command in idcommands:
+        row: PayloadRow = {
+            "id": str(command.uuid),
+            "command": command.description,
+            "type": command.script_type,
+            "tool": str(command.tool),
+            "toolSlug": command.tool.slug if command.tool else None,
```
replaceafill: Should the `str(command.tool)` call in `tool` be conditional in case the `command.tool` is NULL?

replaceafill (approved):
Thanks @sevein. This looks great!

I only noticed two minor differences compared with the previous implementation:

* Empty search used to show this message: `No matching records found`.
* Search results displayed the total number of entries in the footer: `Showing 1 to 10 of 15 entries (filtered from 1,639 total entries)`

src/archivematica/dashboard/fpr/payloads.py
sevein: Good catch!

PR #2216, opened 2026-02-25
Remove transfer backlog and transfer index support

replaceafill:
This change removes Archivematica's Transfer Backlog feature
end-to-end in the Dashboard, including its routes, views, templates,
JS/CSS assets, download endpoints, and all related navigation/context
wiring.

The legacy backlog workflow branch ("Send to backlog") has been
removed from MCPServer workflow definitions, and the associated
MCPClient scripts and module registrations have been deleted so new
processing decisions can no longer route into backlog-only
tasks. post_store_aip_hook is also updated to request transfer package
cleanup in the Storage Service without attempting transfer-index
document deletion.

Configuration and administration messaging now reflects that only AIP
search indexes are supported. The search service constants and
implementation are simplified accordingly by removing transfer index
support (transfers and transferfiles) along with the related APIs,
mappings, and rebuild command paths.

To keep upgrades safe, search_enabled parsing treats legacy transfers
token(s) as deprecated compatibility input: accepted with warnings and
normalized away instead of hard-failing startup.

Additionally, historical records that completed via the former "Move
transfer to backlog" terminal job are still classified as COMPLETE in
status/completed-unit logic, preventing pre-removal data from
regressing into a permanent PROCESSING state or disappearing from
completed-transfer listings.

Connected to https://github.com/archivematica/Issues/issues/1776


src/archivematica/MCPServer/server/workflow_abilities.py
```diff
@@ -5,19 +5,7 @@
...
-        "enabled_attr": "SEARCH_ENABLED",
-        "enabled_condition": "in",
-        "enabled_value": "transfers",
-        "dependencies": (("Create SIP(s)", "Send to backlog"),),
-    },
-)
+ABILITIES = ()
 
 
 def choice_is_available(link, chain):
```
sevein: The entire module could go too!

sevein (approved):
👍

src/archivematica/MCPServer/server/workflow_abilities.py
replaceafill: Thanks for catching that! I addressed it in https://github.com/artefactual/archivematica/pull/2216/commits/6ab549bd80917a75c224ea344c2e1aeb3f52c66b

PR #2218, opened 2026-02-26
Isolate browser dropdown from global dropdown handlers

sevein:


replaceafill (changes requested):
I tried to build the branch but got permission errors. See https://github.com/artefactual/archivematica/actions/runs/22450473671

replaceafill (approved):
👍

PR #2221, opened 2026-02-26
Pin overrides deps in vue package.json

replaceafill:
This fixes the `Test Vue dashboard` job of the `test.yml` GitHub workflow which started failing after a new `minimatch` release was [published today](https://github.com/isaacs/minimatch/releases/tag/v9.0.9).


sevein (approved):
👍

PR #2225, opened 2026-03-05
Extend XML validation feature

jens-st:
https://github.com/archivematica/Issues/issues/1786

- Forward the metadata XML validation dictionary to the resolver when parsing XSD schemas themselves.
- Always check the dictionary first before falling back to resolving the original schemaLocation.
- Implement a timeout for schema URL requests.

This change should provide a way forward for broken external schema resources, potentially prevent being rate-limited or blocked by external sites, and overall stop or heavily reduce outgoing schema-related traffic (while possibly decreasing loading times). Additionally, it will enable the user to have a more controlled local schema file catalog for security concerns.

Edit: AFAIK, the only alternative to using a custom resolver in this way would be to edit and hardcode new (local) paths into each and every original schema file that's used.


jens-st:
Short update: I got time now to look at the test errors. It's related to mocked `etree.parse` and the resolver utilization I changed. I think I can fix the problem and add a few more tests.

jens-st:
Existing tests passed :tada:

As mentioned before I'd like to add more test scenarios for `archivematicaCreateMETSMetadataXML.process_xml_metadata()` involving variations of XML metadata and XML schemas.


Additionally, I'm not sure if this a bug but with an empty `XML_VALIDATION`, the calling of `process_xml_metadata()` is immediately quit.
* https://github.com/artefactual/archivematica/blob/qa/1.x/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py#L38-L39

Wouldn't you want to continue and validate depending on `ARCHIVEMATICA_MCPCLIENT_MCPCLIENT_METADATA_XML_VALIDATION_ENABLED` alone (despite `XML_VALIDATION` being empty or not)?
* https://github.com/artefactual/archivematica/blob/stable/1.18.x/src/archivematica/MCPClient/install/README.md#metadata-xml-validation-variables

jens-st:
@replaceafill Thank you for the feedback during the latest [Community Development Meeting](https://github.com/archivematica/Issues/wiki/Community-Development-Meetings). :+1: 

I'd like to suspend this PR. Using a local xml catalog is generally a better solution and I'd like to test that for our scenarios. I just did not know that the Archivematica environment allows/supports ENV variables for [libxml2](https://gitlab.gnome.org/GNOME/libxml2/-/wikis/Catalog-support#how-to-tune-catalog-usage) / [xmlcatalog](https://manpages.debian.org/testing/libxml2.9-utils/xmlcatalog.1#ENVIRONMENT). Leaving hints in the documentation about this fact (just like with [AWS-specific environment variables](https://github.com/artefactual/archivematica-storage-service/blob/v0.24.0/install/README.md#aws-specific-environment-variables)) would be great and may help others too.
* https://www.archivematica.org/en/docs/archivematica-1.18/search/?q=catalog
* https://github.com/artefactual/archivematica/blob/stable/1.18.x/src/archivematica/MCPClient/install/README.md#metadata-xml-validation-variables

sarah-mason:
@jens-st - I've marked it as a draft so that you can take the time to test using local xml in your scenarios and see how that might impact this PR. If you decide to carry on with updates following your testing, go ahead and when it's complete you can remove the draft status. The documentation feedback is useful, we'll get that into a docs issue, thank you!

jens-st:
Canceling PR, reason: https://github.com/archivematica/Issues/issues/1786#issuecomment-4622338148.
As discussed in the AM Community Development Meeting, I may open a new PR with just more xml/xsd test cases.

PR #2229, opened 2026-03-06
Fix frontend dist shadowing in integration tests

replaceafill:
This prevents bind-mounted /src from hiding built frontend assets
during integration tests by using an image-owned dist path for
staticfiles, reducing flakiness.


sevein (approved):
👍

PR #2306, opened 2026-06-17
Use forkserver for MCPClient workers

### LINKED ISSUE #1801: Problem: MCPClient forks workers from a threaded parent process
**Expected behaviour**

MCPClient should recycle workers without forking directly from its long-lived, threaded parent process. Using `forkserver` keeps worker startup isolated from parent-side logging, metrics, queue, and maintenance threads, reducing the chance that workers inherit stale locks, open resources, or inflated memory state while still preserving process isolation.

**Current behaviour**

Workers are created with multiprocessing.Process, which defaults to fork on Linux/POSIX in older Python versions.

**Steps to reproduce**

Run Archivematica MCPClient with Prometheus enabled, set `max_tasks_per_child = 1` and `workers = 2`, then process a transfer with many small files so workers are constantly recycled. In the affected implementation, each replacement worker is created with the default `multiprocessing.Process` context, so on Linux it is forked directly from the long-lived MCPClient parent after logging, metrics, HTTP, queue, and maintenance threads are already running.

This behaviour was introduced in Archivematica 1.16.0, when MCPClient moved to the `WorkerPool` model that starts and replaces workers with `multiprocessing.Process`. It affects Linux/POSIX deployments where the default multiprocessing start method is `fork`, especially when MCPClient has Prometheus enabled and workers are recycled during large transfers.

---

**For Artefactual use:**

sevein:
MCPClient keeps long-lived logging, metrics, and maintenance threads in the
parent process. Starting worker processes from a single forkserver context
avoids forking directly from that threaded parent while preserving process
isolation.

Create the worker queues, shutdown event, and worker processes from the same
context so all shared multiprocessing primitives are compatible.

Fixes https://github.com/archivematica/Issues/issues/1801.

### DIFF
--- a/src/archivematica/MCPClient/client/pool.py
+++ b/src/archivematica/MCPClient/client/pool.py
@@ -18,6 +18,7 @@ import logging.handlers
 import multiprocessing
 import threading
 import time
+from multiprocessing.process import BaseProcess
 from multiprocessing.synchronize import Event
 from types import ModuleType
 from typing import Optional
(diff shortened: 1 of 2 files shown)


sevein:
Note: because we are not using `forkserver` yet, child workers can inherit the parent MCPClient `SIGTERM` handler. On container restart, some workers handle `SIGTERM` as if they were the parent, call `pool.stop()`, and try to join parent/thread state that should not exist in the child, which can raise `RuntimeError: cannot join current thread`.

replaceafill (approved):
LGTM.

PR #2313, opened 2026-06-22
Add MCPClient transfer-source retrieval workflow

### LINKED ISSUE #1803: Problem: Transfer-source retrieval is not visible in the workflow
**Expected behaviour**

When a transfer is started from a Storage Service transfer source, the retrieval of the transfer contents should be visible as normal transfer processing. Operators should be able to see that Archivematica is retrieving the transfer contents, whether that retrieval is queued, running, failed, or complete. This work should be represented in the same workflow/job model used for the rest of transfer processing.

**Current behaviour**

Transfer-source retrieval currently happens before the transfer enters the normal workflow. MCPServer creates the Transfer row, then copies the selected contents from Storage Service into Archivematica’s shared directory before scheduling the first workflow job. During this phase there may be no normal workflow job that explains what the transfer is doing. If retrieval blocks on Storage Service, rsync, or shared storage I/O, the transfer can appear stuck or misleadingly idle, and failures may be hard to distinguish from later workflow problems.

**Steps to reproduce**

Start multiple transfers from a Storage Service transfer source, especially using large transfers or a storage backend where retrieval can take a long time. Watch the Dashboard or MCPServer logs while the transfer contents are being copied into the shared directory. Before the first normal workflow job is scheduled, retrieval activity is not represented as a normal workflow step. If the retrieval blocks or fails, operators may not get clear workflow-level visibility into the retrieval phase.

This affects current Archivematica versions where transfer-source retrieval is performed by MCPServer before the normal transfer workflow starts.

---

**For Artefactual use:**

sevein:
This pull request moves transfer-source retrieval for auto-approved API transfers into the normal Archivematica workflow. Instead of copying transfer-source files inside MCPServer before the first workflow job exists, MCPServer now only plans the retrieval, records the workflow continuation, and schedules a new retrieval chain. MCPClient then performs the Storage Service copy and moves the transfer into the processing area as a normal workflow task.

This change reduces the chance that slow or stuck transfer-source retrieval blocks MCPServer bootstrap workers, and it makes retrieval failures visible through the usual workflow/task model. Along the way, the PR improves failure handling for early startup errors, executor handoff failures, Storage Service copy failures, missing transfer paths during cleanup, and Gearman batch failures before MCPClient returns task results. It also updates dashboard/API status handling so transfers waiting for their first workflow job are shown as waiting, and failures before normal processing starts are reported as failed instead of unknown.

Fixes https://github.com/archivematica/Issues/issues/1803.
Relates to https://github.com/archivematica/Issues/issues/1790.
Relates to https://github.com/archivematica/Issues/issues/1791.
Relates to https://github.com/archivematica/Issues/issues/1807.


src/archivematica/MCPServer/server/rpc_server.py
```diff
@@ -379,7 +403,21 @@ def _units_statuses_handler(self, worker, job, payload):
...
+            # created their first Job. Add those real units with a synthetic
+            # timestamp so the monitor can show them as waiting.
+            transfer_jobs = Job.objects.filter(
+                sipuuid=OuterRef("uuid"), unittype=model_attrs[1]
+            )
+            queued_transfers = Transfer.objects.filter(
+                status=PACKAGE_STATUS_PROCESSING, hidden=False
+            ).filter(~Exists(transfer_jobs))
+            sipuuids_and_timestamps.extend(
+                (str(transfer_uuid), 0)
```
replaceafill: Not sure if I'm missing something, but is this synthetic `0` clashing with the [sort by timestamp behavior](https://github.com/artefactual/archivematica/blob/f632a89637939f3d9433a359dc748405c9b9b4e0/src/archivematica/dashboard/frontend/lib/monitor/composables/useProcessingMonitor.ts#L65-L67) of the frontend? When my transfer has no job yet it gets displayed at the bottom of the `Transfer` tab.

Image

replaceafill (changes requested):
This is a great improvement @sevein! I really appreciate how the commits are structured and the detailed explanations.

I left one question about the Transfer Monitor sorting behavior when a processing transfer does not have a job yet. Since those transfers get a synthetic timestamp of `0`, they appear at the bottom of the monitor even though they are active. Let me know if I'm missing something. Otherwise, this looks good to me.

src/archivematica/MCPServer/server/rpc_server.py
sevein: Douglas, I've rebased and pushed a tentative fix for this issue. See commit `Feedback: Use queue marker for sorting`. It seems to be pretty efficient but it's not super clean, a better alternative is likely something like a `processing_started_at` field where we can record the timestamp. Let me know if you'd like to postpone that work or you'd rather tackle it now.

replaceafill: @sevein Nice! I would not have thought of using `UnitVariable.createdtime` as a workaround for the lack of transfer datetime metadata :wink: 

One question: is `LINK_AFTER_TRANSFER_SOURCE_RETRIEVAL` the right marker for this? It is written inside `_start_package_transfer_with_auto_approval()`, after the executor has accepted and started the background work. That still leaves an earlier window where `create_package()` has marked the transfer as `PROCESSING`, but the background function has not created that `UnitVariable` yet.

What do you think of using a `UnitVariable` written synchronously in `create_package()` instead, like [`processingConfiguration`](https://github.com/artefactual/archivematica/blob/5186f02c42683a0b9b883a62451a3d7611e7edbd/src/archivematica/dashboard/main/models.py#L1557), which is set right [after the transfer is created](https://github.com/artefactual/archivematica/blob/5186f02c42683a0b9b883a62451a3d7611e7edbd/src/archivematica/MCPServer/server/packages.py#L267) and before  the transfer is marked `PROCESSING` / handed off to the executor?

If we use `processingConfiguration`, I think this should use `Min("createdtime")` instead of `Max("createdtime")`, since we would be treating that `UnitVariable` as the package-start marker and want the earliest creation timestamp.

replaceafill (approved):
Thanks @sevein!

PR #2317, opened 2026-06-25
Avoid closing active file query results

### LINKED ISSUE #1793: Repeated calls to `auto_close_old_connections()` in a single request can cause a `MySQLdb.OperationalError`
**Expected behaviour**

Within a single request, a connection to MySQL shouldn't be closed while still in use.

**Current behaviour**

We recently hit an issue where MCPServer unexpectedly threw an exception during processing, reporting `MySQLdb.OperationalError: (2006, '')` (which is "mysql server has gone away").  Usually this happens when MySQL gets restarted or the underlying connection is closed for some other reason, but in this case there was no system-level reason for this to have happened.

After some digging, I eventually worked out that the culprit was `auto_close_old_connections()` and a pretty large dose of bad luck.  The root cause is `auto_close_old_connections()` being called more than once during the same request, when the request happens to start just *before* the underlying database connection is due to be expired by Django's `CONN_MAX_AGE` setting (which defaults to 1 hour).

In my case, I think the sequence of events was:

  * A request starts, which calls `auto_close_old_connections()` up front.  No problem there, because the connection is a mere 59 minutes and 59 seconds old.
  * In `lib/server/jobs/client.py`, we find ourselves in `submit_tasks` in a loop that is iterating over the `self.package.files` result set.
  * Within the body of that loop, we call `self.task_backend.submit_task`, which ultimately calls the `submit` method of `server/tasks/backends/gearman_backend.py`.  In turn, that method calls `Task.bulk_log`, which is decorated with `@auto_close_old_connections()`
  * Now our underlying database connection is over an hour old, and `auto_close_old_connections()` closes it.
  * But eventually we return back to our loop that is still iterating over `self.package.files`.  It finishes successfully and attempts to close the underlying result set, but finds that its database connection has gone away.

The stack trace we see when that happens:

```
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 342, in _invoke_callbacks
    callback(self)
  File "/usr/lib/archivematica/MCPServer/server/queues.py", line 220, in _job_completed_callback
    next_job = future.result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/lib/archivematica/MCPServer/server/jobs/client.py", line 122, in run
    self.submit_tasks()
  File "/usr/lib/archivematica/MCPServer/server/jobs/client.py", line 225, in submit_tasks
    for file_replacements in self.package.files(
  File "/usr/lib/archivematica/MCPServer/server/packages.py", line 663, in files
    for file_obj in queryset.iterator():
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/django/db/models/query.py", line 516, in _iterator
    yield from iterable
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/django/db/models/query.py", line 121, in __iter__
    for row in compiler.results_iter(results):
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/django/db/models/sql/compiler.py", line 1496, in apply_converters
    for row in map(list, rows):
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/django/db/models/sql/compiler.py", line 2096, in cursor_iter
    cursor.close()
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/MySQLdb/cursors.py", line 103, in close
    self._discard()
  File "/usr/share/archivematica/virtualenvs/archivematica/lib/python3.10/site-packages/MySQLdb/cursors.py", line 95, in _discard
    while con.next_result() == 0:  # -1 means no more data.
MySQLdb.OperationalError: (2006, '')
```

There are a couple of other issues that might not be exactly the same bug, but have the same kind of shape:
https://github.com/archivematica/Issues/issues/1758 https://github.com/archivematica/Issues/issues/1750

**Steps to reproduce**

It's difficult to reproduce in the wild because you do have to get very unlucky :)  In my development setup, I was able to replicate the issue by:

  * Changing my CONN_MAX_AGE setting to `1` to get it turning over connections after 1 second
  * Adding a `sleep` statement to the body of `lib/server/jobs/client.py`'s `submit_tasks` to slow down the loop enough to make hitting the bug likely.
  * Submitting a transfer as normal

I'm running `1:1.17.0-1~22.04` under Ubuntu Jammy (Ubuntu 22.04.5 LTS)

---

**For Artefactual use:**

sevein:
Package.files() returned database files directly from QuerySet.iterator(). ClientScriptJob.submit_tasks() consumes that generator while creating tasks, and task creation can run auto_close_old_connections() before Package.files() resumes. On MySQL, that can invalidate the active database result.

Load database files in UUID-ordered pages before yielding file mappings. Each page is fully read before task creation runs, so no active database result stays open across generator yields.

This trades one streaming query for bounded page queries. The extra database round trips are about one SELECT per 1000 database files, while memory stays bounded and pagination uses File.uuid, the primary key.

Fixes https://github.com/archivematica/Issues/issues/1793.

### DIFF
--- a/src/archivematica/MCPServer/server/packages.py
+++ b/src/archivematica/MCPServer/server/packages.py
@@ -531,6 +531,9 @@ def get_file_replacement_mapping(file_obj, unit_directory):
 class Package(metaclass=abc.ABCMeta):
     """A `Package` can be a Transfer, a SIP, or a DIP."""
 
+    # Limit file rows held in memory while avoiding long-lived DB cursors.
+    FILE_QUERYSET_BATCH_SIZE = 1000
+
     def __init__(self, current_path, uuid):
         self._current_path = current_path.replace(
             r"%sharedPath%", _get_setting("SHARED_DIRECTORY")
@@ -641,6 +644,25 @@ class Package(metaclass=abc.ABCMeta):
 
         return mapping
 
+    def _database_file_replacement_mappings(self, queryset):
+        """Yield file mappings without holding a database cursor open."""
+        queryset = queryset.order_by("uuid")
+        last_uuid = None
+
+        while True:
+            with auto_close_old_connections():
+                page_queryset = queryset
+                if last_uuid is not None:
+                    page_queryset = page_queryset.filter(uuid__gt=last_uuid)
+                file_objs = list(page_queryset[: self.FILE_QUERYSET_BATCH_SIZE])
+
+            if not file_objs:
+                break
+
+            for file_obj in file_objs:
+                last_uuid = file_obj.uuid
+                yield get_file_replacement_mapping(file_obj, self.current_path)
+
     def files(self, filter_filename_end=None, filter_subdir=None):
         """Generator that yields all files associated with the package or that
         should be associated with a package.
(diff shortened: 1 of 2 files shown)


src/archivematica/MCPServer/server/packages.py
```diff
@@ -641,6 +644,25 @@ def get_replacement_mapping(self, filter_subdir_path=None):
 
         return mapping
 
+    def _database_file_replacement_mappings(self, queryset):
+        """Yield file mappings without holding a database cursor open."""
+        queryset = queryset.order_by("uuid")
```
replaceafill, line 649: Would it be worth adding new indexes to support this `order_by`:

* transferUUID + fileUUID
* sipUUID + fileUUID

replaceafill (approved):
Thanks @sevein!

src/archivematica/MCPServer/server/packages.py
sevein: Good catch! I didn't think about that, but I've tested the new query and it looks like the extra composite indexes would not help which is surprising. Apparently InnoDB can already use the existing sipUUID/transferUUID indexes to speed up the ordering because secondary indexes include the primary key!

TIL that you can combine `EXPLAIN ANALYZE` with `FORCE INDEX` to test the same query against different indexes, e.g. the following query showed that the proposed composite index was slower on my sample dataset:

```mysql
EXPLAIN ANALYZE
SELECT *
FROM Files FORCE INDEX (Files_transfer_fileUUID_probe_idx)
WHERE transferUUID = @transfer_uuid AND fileUUID > @last_transfer_file_uuid
ORDER BY fileUUID
LIMIT 1000;
```

PR #2318, opened 2026-06-25
Fix DSpace multiple license file grouping

replaceafill:
DSpace METS file groups can contain more than one file with `USE="LICENSE"`, but the DSpace file identification script only read the first `FLocat` in each matching file group. As a result, later license files kept the default original file group and were written to the wrong AIP METS fileGrp.

The script now iterates over every FLocat in `TEXT` and `LICENSE` groups, so all matching files are assigned the correct filegrpuse before AIP METS generation.

Connected to https://github.com/archivematica/Issues/issues/1808


sevein (approved):
👍

PR #2327, opened 2026-07-01
Include pending deletion requests in storage totals

replaceafill:
The Archival Storage summary totals were excluding AIPs with pending deletion requests by fetching every matching package from the Storage Service and adding one Elasticsearch must_not term per UUID. Large installations could exceed Elasticsearch's max clause count and fail to load the tab.

Use match_all queries for the total size and indexed file count calculations so pending deletion requests remain included in the storage-management totals and the summary no longer depends on building a large exclusion query.

Connected to https://github.com/archivematica/Issues/issues/1792


sevein (approved):
👍

PR #2339, opened 2026-07-20
Close stale MCPServer worker connections

sevein:
703d8573e1f5b9cab5a7d6ba1f1fd6d5a704e464 added a direct `transfer.save()` inside an unguarded executor worker. After inactivity, the worker attempts to execute it using a connection that MySQL has already closed.


replaceafill (approved):
👍

PR #2341, opened 2026-07-21
Improve resilience of XML metadata validation

jens-st:
https://github.com/archivematica/Issues/issues/1810

- Extended XML metadata validation tests with additional scenarios
- Added timeout to prevent GET requests to hang indefinitely (https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts)
- Gracefully handled failed metadata parsing and added info about the failed file
- Gracefully handled failed schema access by catching schema URL errors as well

Notes:
1. Some of the scenarios may already be covered in other test functions, but my focus was on XML validation.
2. I wasn't able to reproduce a scenario where a custom XML catalog is used by `lxml` to resolve schemas with nested imports.
  * Setting `XML_CATALOG_FILES` dynamically at runtime (with `monkeypatch.setenv`) doesn't seem to affect `lxml` at all.
  * Also, appending a custom resolver to the default resolver for in-memory schemas at runtime (with `etree.get_default_parser().resolvers.add`) has no effect on `lxml`. Ignoring the default resolver might be a security mechanism in `lxml` or related to caching.
  * The only way I found to test schemas with nested imports was to use `schemaLocation` hints with purely local paths in every used schema.

### DIFF
--- a/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
+++ b/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
@@ -189,7 +189,9 @@ class Resolver(etree.Resolver):
         url_scheme = urlparse(url).scheme
         if url_scheme in ("http", "https"):
             try:
-                response = requests.get(url)
+                # timeout to prevent indefinite hangs, see
+                # https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
+                response = requests.get(url, timeout=10)
             except requests.RequestException:
                 return super().resolve(url, id, context)
             else:
--- a/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
+++ b/src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
@@ -21,6 +21,7 @@
 import csv
 from importlib.metadata import version
 from pathlib import Path
+from urllib.error import URLError
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
@@ -52,7 +53,14 @@ def process_xml_metadata(mets, sip_dir, sip_uuid, sip_type, xml_validation):
             if not xml_path:
                 fsentry.delete_dmdsec("OTHER", xml_type)
                 continue
-            tree = etree.parse(str(xml_path))
+            try:
+                tree = etree.parse(str(xml_path))
+            except Exception as err:
+                xml_metadata_errors += [
+                    f"Could not parse metadata file: {xml_path}",
+                    err,
+                ]
+                continue
             try:
                 schema_uri = _get_schema_uri(tree, xml_validation)
             except ValueError as err:
@@ -230,7 +238,7 @@ def _validate_xml(tree, schema_uri):
                 schema = etree.RelaxNG(schema_contents)
             else:
                 return False, [f"Unknown XML validation schema type: {schema_type}"]
-    except etree.LxmlError as err:
+    except (etree.LxmlError, URLError) as err:
         return False, [f"Could not parse schema file: {schema_uri}", err]
     if not schema.validate(tree):
         return False, schema.error_log
--- a/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
+++ b/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
@@ -23,9 +23,10 @@ from archivematica.MCPClient.clientScripts.archivematicaCreateMETSMetadataXML im
 METADATA_DIR = Path("objects") / "metadata"
 TRANSFER_METADATA_DIR = METADATA_DIR / "transfers" / "transfer_a"
 TRANSFER_SOURCE_METADATA_CSV = TRANSFER_METADATA_DIR / "source-metadata.csv"
-VALID_XML = '<?xml version="1.0" encoding="UTF-8"?><foo><bar/></foo>'
-INVALID_XML = '<?xml version="1.0" encoding="UTF-8"?><foo/>'
-SCHEMAS = {
+DUMMY_EXTERNAL_SCHEMA_URI = "http://foo.com/my.xsd"
+DUMMY_SCHEMA_NAMESPACE = "http://foo.com/1.0"
+
+DUMMY_SCHEMAS = {
     "xsd": """<?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
   <xs:element name="foo">
@@ -48,18 +49,21 @@ SCHEMAS = {
   </oneOrMore>
 </element>
 """,
-}
-IMPORTED_SCHEMA = """<?xml version="1.0" encoding="UTF-8"?>
+    "xsd_imported": f"""<?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
-           xmlns="http://foo.com/1.0" targetNamespace="http://foo.com/1.0">
-</xs:schema>"""
+           xmlns="{DUMMY_SCHEMA_NAMESPACE}" targetNamespace="{DUMMY_SCHEMA_NAMESPACE}">
+</xs:schema>
+""",
+}
 
+VALID_XML = '<?xml version="1.0" encoding="UTF-8"?><foo><bar/></foo>'
+INVALID_XML = '<?xml version="1.0" encoding="UTF-8"?><foo/>'
 
 @pytest.fixture
 def make_schema_file(tmp_path):
     def _make_schema_file(schema_type):
         schema_path = tmp_path / (schema_type + "." + schema_type)
-        schema_path.write_text(SCHEMAS[schema_type])
+        schema_path.write_text(DUMMY_SCHEMAS[schema_type])
         return schema_path
 
     return _make_schema_file
@@ -150,7 +154,7 @@ def requests_get():
     # Return a mock response good enough to be parsed as an XML schema.
     with mock.patch(
         "requests.get",
-        return_value=mock.Mock(text=IMPORTED_SCHEMA),
+        return_value=mock.Mock(text=DUMMY_SCHEMAS["xsd_imported"]),
     ) as result:
         yield result
 
@@ -192,9 +196,9 @@ def etree_parse():
 @pytest.fixture
 def schema_with_remote_import(tmp_path):
     # Create a schema that imports a remote schema.
-    schema = """<?xml version="1.0" encoding="UTF-8"?>
+    schema = f"""<?xml version="1.0" encoding="UTF-8"?>
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
-  <xs:import namespace="http://foo.com/1.0" schemaLocation="http://foo.com/my.xsd" />
+  <xs:import namespace="{DUMMY_SCHEMA_NAMESPACE}" schemaLocation="{DUMMY_EXTERNAL_SCHEMA_URI}" />
   <xs:element name="foo">
     <xs:complexType>
       <xs:sequence>
@@ -228,7 +232,7 @@ def schema_with_local_import(tmp_path):
     schema_path.write_text(schema)
 
     local_schema = tmp_path / "my.xsd"
-    local_schema.write_text(IMPORTED_SCHEMA)
+    local_schema.write_text(DUMMY_SCHEMAS["xsd_imported"])
 
     return schema_path
 
@@ -618,7 +622,7 @@ def test_resolver(
         xml_validation,
     )
     assert not errors
-    requests_get.assert_called_once_with("http://foo.com/my.xsd")
+    requests_get.assert_called_once_with(DUMMY_EXTERNAL_SCHEMA_URI)
 
 
 @pytest.mark.django_db
--- a/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
+++ b/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
@@ -5,7 +5,9 @@ Tests for XML metadata management on the METS creation process:
 archivematicaCreateMETSMetadataXML.process_xml_metadata()
 """
 
+import re
 from importlib.metadata import version
+from os.path import abspath
 from pathlib import Path
 from unittest import mock
 from uuid import uuid4
@@ -23,8 +25,18 @@ from archivematica.MCPClient.clientScripts.archivematicaCreateMETSMetadataXML im
 METADATA_DIR = Path("objects") / "metadata"
 TRANSFER_METADATA_DIR = METADATA_DIR / "transfers" / "transfer_a"
 TRANSFER_SOURCE_METADATA_CSV = TRANSFER_METADATA_DIR / "source-metadata.csv"
+PLACEHOLDER_LOCAL_DIR = "INJECT_LOCAL_DIR_HERE"
+
 DUMMY_EXTERNAL_SCHEMA_URI = "http://foo.com/my.xsd"
+DUMMY_EXTERNAL_SCHEMA_URI_0 = "http://foo.com/layer0.xsd"
+DUMMY_EXTERNAL_SCHEMA_URI_1 = "http://foo.com/layer1.xsd"
+DUMMY_EXTERNAL_SCHEMA_URI_2 = "http://foo.com/layer2.xsd"
+DUMMY_EXTERNAL_SCHEMA_URI_3 = "http://foo.com/layer3.xsd"
 DUMMY_SCHEMA_NAMESPACE = "http://foo.com/1.0"
+DUMMY_SCHEMA_NAMESPACE_0 = "http://foo.com/layer0"
+DUMMY_SCHEMA_NAMESPACE_1 = "http://foo.com/layer1"
+DUMMY_SCHEMA_NAMESPACE_2 = "http://foo.com/layer2"
+DUMMY_SCHEMA_NAMESPACE_3 = "http://foo.com/layer3"
 
 DUMMY_SCHEMAS = {
     "xsd": """<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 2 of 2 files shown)


replaceafill:
Thank you for this contribution @jens-st! Could you please sign and send the [Contributor's Agreement](https://github.com/artefactual/archivematica/blob/qa/1.x/CONTRIBUTING.md#contributors-agreement) before we review this pull request?

jens-st:
> Could you please sign and send the [Contributor's Agreement](https://github.com/artefactual/archivematica/blob/qa/1.x/CONTRIBUTING.md#contributors-agreement) before we review this pull request?

In the works! :)

There is an additional side effect when catching parsing errors on XML metadata files (e.g. broken, maybe truncated, not well-formed XML metadata) directly. The job "Generate METS.xml document" is more resilient and can still pull through and the AIP can get archived if `XML_VALIDATION_FAIL_ON_ERROR` is `False` (default). The particular XML file in question is just skipped from import into the AIP METS to not damage it. The given error message (https://github.com/artefactual/archivematica/pull/2341/changes#diff-cb94ebe0cbc1604cc921d0ceac8f19093f92239d638cc2a9bcba6f03e44ed31aR60) should probably reflect that fact too.

jens-st:
@replaceafill The CLA is signed, any feedback regarding the pull request would be appreciated :)

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
     )
     assert not errors
     requests_get.assert_not_called()
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "xml, schemas, xml_validation_factory, should_pass, expected_error",
```
replaceafill, line 835: Several scenarios repeat existing tests (or each other). They only check `errors`, while most of the tests they repeat also check the `dmdSec` or the PREMIS validation event. I'll comment on each one with the test it repeats.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+    [
+        # scenario 01 - XML metadata without namespace
+        #               XML_VALIDATION provides schema based on root element node
```
replaceafill, line 838: This overlaps with [`test_validation[valid.xml-xsd-True]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L265): same document (`XML_WITHOUT_NAMESPACE` is `VALID_XML`), same schema (`DUMMY_SCHEMAS["xsd"]`) and same key (`foo`). It only checks `not errors`; `test_validation` also checks that the `dmdSec` is added with root tag `foo`, and the PREMIS validation event.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            XML_WITH_NAMESPACE_AND_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE}": str(schema_paths["custom.xsd"])
+            },
+            True,
+            None,
+        ),
+        # scenario 04: XML metadata with namespace & schemaLocation hint, but metadata is invalid
+        #              XML_VALIDATION forces validation skip, overrules schemaLocation hint
```
replaceafill: This overlaps with `test_skipped_validation`: the key maps to `None`, so validation is skipped. The only difference is how the key is found: through the namespace after the `xsi:schemaLocation` URL misses, instead of through the root element name. That fall-through is already covered by `test_schema_uri_retrieval[http://foo.com/foo_namespace.xsd-False]`. `scenario 04` doesn't check that the `dmdSec` is still added or that no PREMIS event is recorded. Its comment also says the metadata is invalid, but the document is valid.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            XML_WITH_NAMESPACE_AND_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns_v2"]},
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE}": str(schema_paths["custom.xsd"])
+            },
+            False,
+            "bar[^2].*element is not expected",
+        ),
+        # scenario 09: XML metadata with namespace
+        #              XML_VALIDATION provides schema based on namespace, but schema path is not absolute
```
replaceafill: This scenario is the first block of `test_validation_schema_errors`: a relative schema path gives "XML schema local path ... must be absolute". Only the key differs (namespace instead of `foo`). `scenario 09` also writes a `custom.xsd` it never uses.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            XML_WITH_NAMESPACE_NO_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE}": abspath("does_not_exist.xsd")
+            },
+            False,
+            "Could not parse schema file|No such file or directory",
+        ),
+        # scenario 11: XML metadata without namespace
+        #              XML_VALIDATION does not provide a schema matching the root node
```
replaceafill: This overlaps with [`test_schema_uri_retrieval[bar-True]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L472): no key matches, so "XML validation schema not found for keys". The existing test's document runs all four lookups (`noNamespaceSchemaLocation`, `schemaLocation`, namespace, root name) and checks that no PREMIS event is recorded. `scenario 11`'s document only yields two keys (`''` and `foo`).

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            BROKEN_XML_WITH_NAMESPACE_AND_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE}": str(schema_paths["custom.xsd"])
+            },
+            False,
+            "Could not parse metadata file",
+        ),
+        # scenario 07: XML metadata with namespace, but no schemaLocation hint, metadata is invalid
+        #              XML_VALIDATION provides schema based on namespace, metadata invalid against user provided schema
```
replaceafill: This and `scenario 08` below are namespaced versions of [`test_validation[invalid.xml-xsd-False]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L268), and the same case as each other: a document that doesn't match the mapped schema. They check only the error text; `test_validation` also checks that no `dmdSec` is added and that the event outcome is `fail`.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+        #               XML_VALIDATION provides schema based on root element node
+        (
+            XML_WITHOUT_NAMESPACE,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd"]},
+            lambda schema_paths: {"foo": str(schema_paths["custom.xsd"])},
+            True,
+            None,
+        ),
+        # scenario 02: XML metadata with namespace, but no schemaLocation hint
+        #              XML_VALIDATION provides schema based on namespace
```
replaceafill, line 847: This scenario only differs from `scenario 03` by an `xsi:schemaLocation` hint. Its URL isn't a key in `XML_VALIDATION`, and `libxml2` doesn't load hints when validating against a compiled schema, so one of them is enough.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
replaceafill, line 869: This overlaps with `test_skipped_validation`: [the key maps to `None`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L337), so validation is skipped. The only difference is how the key is found: through the namespace after the `xsi:schemaLocation` URL misses, instead of through the root element name. That fall-through is already covered by [`test_schema_uri_retrieval[http://foo.com/foo_namespace.xsd-False]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L470). `scenario 04` doesn't check that the `dmdSec` is still added or that no PREMIS event is recorded. Its comment also says the metadata is invalid, but the document is valid.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
replaceafill, line 898: This and `scenario 08` below are namespaced versions of [`test_validation[invalid.xml-xsd-False]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L268), and the same case as each other: a document that doesn't match the mapped schema. They check only the error text; `test_validation` also checks that no `dmdSec` is added and that the event outcome is `fail`.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
replaceafill, line 920: This scenario is the [first block of `test_validation_schema_errors`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L386): a relative schema path gives "XML schema local path ... must be absolute". Only the key differs (namespace instead of `foo`). `scenario 09` also writes a `custom.xsd` it never uses.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
replaceafill, line 940: This overlaps with [`test_schema_uri_retrieval[bar-True]`](https://github.com/artefactual/archivematica/blob/ac9f2faa8bfeeac744cf31b107d308f5aeb7b4e3/tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py#L472): no key matches, so "XML validation schema not found for keys". The existing test's document runs all four lookups (`noNamespaceSchemaLocation`, `schemaLocation`, namespace, root name) and checks that no PREMIS event is recorded. `scenario 11`'s document only yields two keys (`''` and `foo`).

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -302,7 +452,7 @@ def test_validation(
     objects_fsentry = mock_mets.get_file(label="objects")
     metadata_fsentry = mock_mets.get_file(file_uuid=str(metadata_file.uuid))
     if should_pass:
-        # Use ANY to avoid comparision with etree.Element, but confirm element tag.
+        # Use ANY to avoid comparison with etree.Element, but confirm element tag.
```
replaceafill, line 455: Thank you for catching these typos!

src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -52,7 +53,14 @@ def process_xml_metadata(mets, sip_dir, sip_uuid, sip_type, xml_validation):
             if not xml_path:
                 fsentry.delete_dmdsec("OTHER", xml_type)
                 continue
-            tree = etree.parse(str(xml_path))
+            try:
+                tree = etree.parse(str(xml_path))
+            except Exception as err:
```
replaceafill, line 58: `etree.parse()` raises `OSError` for a missing file and `etree.LxmlError` (as `XMLSyntaxError`) for an empty or malformed one, so catching just those covers the cases this is meant for while letting unexpected errors surface.

```suggestion
            except (OSError, etree.LxmlError) as err:
```

src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -189,7 +197,9 @@ def resolve(self, url, id, context):
         url_scheme = urlparse(url).scheme
         if url_scheme in ("http", "https"):
             try:
-                response = requests.get(url)
+                # timeout to prevent indefinite hangs, see
+                # https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
+                response = requests.get(url, timeout=10)
```
replaceafill, line 202: Could we convert this literal `10` into a module constant (e.g. `SCHEMA_FETCH_TIMEOUT = 10`)? The `test_resolver` test you've modified could reuse it.

Separately, a pre-existing issue that doesn't need to be handled in this PR: `requests.get()` doesn't raise for HTTP error responses, so when an imported schema URL returns a 404, the resolver passes the HTML error page to `lxml` as if it were the schema. The error reported then is "Opening and ending tag mismatch: meta line 4 and head", which makes it look like the schema in `XML_VALIDATION` is broken. Calling `response.raise_for_status()` inside the `try` would send `4xx/5xx` responses through the existing `RequestException` fallback instead, and the reported error would become the unresolved import (`"The QName value '{...}bar' does not resolve to a(n) element declaration"`):

```diff
                 response = requests.get(url, timeout=10)
+                response.raise_for_status()
             except requests.RequestException:
```

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+        #              XML_VALIDATION provides schema based on namespace, but schema path is not absolute
+        (
+            XML_WITH_NAMESPACE_NO_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
+            lambda schema_paths: {f"{DUMMY_SCHEMA_NAMESPACE}": "relative_path.xsd"},
+            False,
+            "schema local path.*must be absolute",
+        ),
+        # scenario 10: XML metadata with namespace
+        #              XML_VALIDATION provides schema based on namespace, but schema path is not accessable, parsing schema should fail gracefully
```
replaceafill, line 929:
```suggestion
        #              XML_VALIDATION provides schema based on namespace, but schema path is not accessible, parsing schema should fail gracefully
```

src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
replaceafill: The timeout you added properly handles one kind of hang: fetching imported schemas through the resolver. There's a pre-existing one right here, though: the schema listed in `XML_VALIDATION` is opened with `urlopen()` without a timeout (here and again at line 233), and `urlopen()` waits indefinitely by default. So if the schema host accepts the connection but never answers, the job still hangs forever. Since it predates this PR, it doesn't need to block it; it can be dealt with in a future PR if needed.

For reference, the fix would be to pass the timeout (the constant from my comment on line 202) to both calls, plus one change in the `except` at line 241: a read timeout from `urlopen()` raises `TimeoutError`, which isn't a `URLError`, so it should catch `OSError` instead. `URLError` is a subclass of `OSError`, so the cases you already handle stay covered, and the `URLError` import can go:

```diff
-        with urlopen(schema_uri) as f:
+        with urlopen(schema_uri, timeout=SCHEMA_FETCH_TIMEOUT) as f:
 ...
-                    with urlopen(schema_uri) as f2:
+                    with urlopen(schema_uri, timeout=SCHEMA_FETCH_TIMEOUT) as f2:
 ...
-    except (etree.LxmlError, URLError) as err:
+    except (etree.LxmlError, OSError) as err:
```

With those changes, a host that never replies gives "Could not parse schema file: ..." and `TimeoutError: timed out` after 10 seconds instead of blocking the job. A test for it can point `XML_VALIDATION` at a local socket that accepts connections but never replies, and monkeypatch `SCHEMA_FETCH_TIMEOUT` down to a fraction of a second.

src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -228,7 +238,7 @@ def _validate_xml(tree, schema_uri):
                 schema = etree.RelaxNG(schema_contents)
             else:
                 return False, [f"Unknown XML validation schema type: {schema_type}"]
-    except etree.LxmlError as err:
+    except (etree.LxmlError, URLError) as err:
```
replaceafill, line 241: The timeout you added to the resolver (line 202) properly handles one kind of hang: fetching imported schemas. There's a pre-existing one in this function, though: the schema listed in `XML_VALIDATION` is opened with `urlopen()` without a timeout (lines 221 and 233), and `urlopen()` waits indefinitely by default. So if the schema host accepts the connection but never answers, the job still hangs forever. Since it predates this PR, it doesn't need to block it; it can be dealt with in a future PR if needed.

For reference, the fix would be to pass the timeout (the constant from my comment on line 202) to both calls, plus one change in this `except`: a read timeout from `urlopen()` raises `TimeoutError`, which isn't a `URLError`, so it should catch `OSError` instead. `URLError` is a subclass of `OSError`, so the cases you already handle stay covered, and the `URLError` import can go:

```diff
-        with urlopen(schema_uri) as f:
+        with urlopen(schema_uri, timeout=SCHEMA_FETCH_TIMEOUT) as f:
 ...
-                    with urlopen(schema_uri) as f2:
+                    with urlopen(schema_uri, timeout=SCHEMA_FETCH_TIMEOUT) as f2:
 ...
-    except (etree.LxmlError, URLError) as err:
+    except (etree.LxmlError, OSError) as err:
```

With those changes, a host that never replies gives "Could not parse schema file: ..." and `TimeoutError: timed out` after 10 seconds instead of blocking the job. A test for it can point `XML_VALIDATION` at a local socket that accepts connections but never replies, and monkeypatch `SCHEMA_FETCH_TIMEOUT` down to a fraction of a second.

src/archivematica/MCPClient/clientScripts/archivematicaCreateMETSMetadataXML.py
```diff
@@ -52,7 +53,14 @@ def process_xml_metadata(mets, sip_dir, sip_uuid, sip_type, xml_validation):
             if not xml_path:
                 fsentry.delete_dmdsec("OTHER", xml_type)
                 continue
-            tree = etree.parse(str(xml_path))
+            try:
+                tree = etree.parse(str(xml_path))
+            except Exception as err:
+                xml_metadata_errors += [
+                    f"Could not parse metadata file: {xml_path}",
```
replaceafill, line 60: As you noted in your earlier comment, this changes what happens to a metadata file that can't be parsed. Today `etree.parse()` raises, "Generate METS.xml document" fails with a traceback and no AIP is stored, whatever `XML_VALIDATION_FAIL_ON_ERROR` says. With this PR and the default `XML_VALIDATION_FAIL_ON_ERROR = False`, the job succeeds and the AIP is stored without that file's metadata. Unlike a file that fails validation, which gets a PREMIS `validation` event with outcome `fail`, nothing in the METS records the skip; the only trace is the job's error output. I see a few options:

1. Keep the skip and, as you suggested, make this message say what happened, e.g. `"Could not parse metadata file objects/metadata/transfers/…/broken.xml; its metadata was not added to the METS"`, using the SIP-relative path like the `"No uuid for file"` message below instead of the full shared-directory path.
2. Keep the skip and also record it in the AIP with a PREMIS event on the metadata file. No schema has been looked up at this point, so the event details would differ from the ones `_add_validation_event()` writes.
3. Treat a file that can't be parsed as an error regardless of `XML_VALIDATION_FAIL_ON_ERROR`, as today, but with this message instead of a traceback.

With 1 or 2, ingests that fail today would succeed without that file's metadata, so the change would need a release note and a mention in the "Importing and validating XML metadata with source-metadata.csv" documentation. What do you think?

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            lambda schema_paths: {f"{DUMMY_SCHEMA_NAMESPACE}": None},
+            True,
+            None,
+        ),
+        # scenario 05: XML metadata is missing, empty file
+        #              parsing metadata should fail gracefully
+        (
+            "",
+            {},
+            lambda schema_paths: True,  # just enable xml validation
```
replaceafill, line 882: `XML_VALIDATION` is always a dict in practice (the MCPClient settings refuse anything else), and `True` only works here because parsing fails before the mapping is used; if it didn't, `_get_schema_uri()` would raise a `TypeError`. A real mapping keeps the scenario valid either way, e.g. `lambda schema_paths: {"foo": None}` as in `test_no_source_metadata_csv`.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            False,
+            "schema local path.*must be absolute",
+        ),
+        # scenario 10: XML metadata with namespace
+        #              XML_VALIDATION provides schema based on namespace, but schema path is not accessable, parsing schema should fail gracefully
+        (
+            XML_WITH_NAMESPACE_NO_SCHEMALOCATION,
+            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE}": abspath("does_not_exist.xsd")
```
replaceafill, line 934: `abspath()` depends on where the tests are run from. Pointing at a missing file next to `custom.xsd` keeps it inside the test's temporary directory:

```suggestion
                f"{DUMMY_SCHEMA_NAMESPACE}": str(
                    schema_paths["custom.xsd"].with_name("does_not_exist.xsd")
                )
```

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -5,7 +5,9 @@
 archivematicaCreateMETSMetadataXML.process_xml_metadata()
 """
 
+import re
 from importlib.metadata import version
+from os.path import abspath
```
replaceafill, line 10: No longer needed with the change to scenario 10:

```suggestion
```

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -678,3 +828,239 @@ def test_resolver_with_local_import(
...
+            },
+            lambda schema_paths: {
+                f"{DUMMY_SCHEMA_NAMESPACE_0}": str(schema_paths["layer0.xsd"])
+            },  # XML_VALIDATION only handles layer0
+            False,
+            "invalid_node.*element is not expected",
+        ),
+    ],
+)
+def test_validation_in_additional_xml_scenarios(
```
replaceafill, line 1000: We are not very consistent about setting explicit test IDs, but the ones pytest generates for these scenarios are hard to read.

Searching for `test_validation_in_additional_xml_scenarios` in [the CI log for this PR](https://github.com/artefactual/archivematica/actions/runs/29845663811/job/88701051676) shows entries like these, 99 to 624 characters long:

Image

That's because pytest builds each ID from the parameter values: strings are used as they are, so each XML document ends up in the ID with its newlines shown as `\n`; dicts become the argument name plus an index (`schemas1`); and lambdas show up as `<lambda>`. Giving each case its own ID avoids this. With `pytest.param`, the ID stays next to its case:

```python
        pytest.param(
            XML_WITH_NAMESPACE_NO_SCHEMALOCATION,
            {"custom.xsd": DUMMY_SCHEMAS["xsd_with_ns"]},
            lambda schema_paths: {
                f"{DUMMY_SCHEMA_NAMESPACE}": str(schema_paths["custom.xsd"])
            },
            True,
            None,
            id="namespace_key",
        ),
```

The test then shows up as `test_validation_in_additional_xml_scenarios[namespace_key]`. For the other scenarios I'd keep, something like `empty_metadata_file`, `malformed_metadata_file`, `missing_schema_file`, `nested_imports`, `nested_imports_circular` and `nested_imports_invalid`.

tests/MCPClient/test_archivematicaCreateMETSMetadataXML.py
```diff
@@ -48,18 +61,155 @@
...
+    </xs:complexType>
+  </xs:element>
+</xs:schema>
+""",
+    "xsd_with_nested_imports_layer0": f"""<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema targetNamespace="{DUMMY_SCHEMA_NAMESPACE_0}"
+           xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:layer1="{DUMMY_SCHEMA_NAMESPACE_1}"
+           elementFormDefault="qualified">
+  <xs:import namespace="{DUMMY_SCHEMA_NAMESPACE_1}" schemaLocation="file://{PLACEHOLDER_LOCAL_DIR}/layer1.xsd"/>
```
replaceafill, line 96: About your note on nested imports: the `file://` locations with `INJECT_LOCAL_DIR_HERE` aren't needed. A relative location like `schemaLocation="layer1.xsd"` is resolved against the location of the schema that imports it, and all four layers are written to the same temporary directory, so the chain resolves on its own (`schema_with_local_import` already relies on this). With relative locations here and in lines 111, 126 and 147, `PLACEHOLDER_LOCAL_DIR` and both `re.sub` calls in the test body can go (`re` is still needed to match the expected errors), and scenarios 12 to 14 still pass, including the circular and invalid ones.

Separately, the documents for these scenarios keep `xsi:schemaLocation="{DUMMY_SCHEMA_NAMESPACE} {DUMMY_EXTERNAL_SCHEMA_URI}"` (lines 183 and 197), which points at the other scenarios' namespace. It isn't used to load anything (Archivematica only uses it as a lookup key, and `libxml2` doesn't load it when validating against a compiled schema), so it can go too.

replaceafill (changes requested):
Thanks for signing the CLA, @jens-st, and for this PR! Overall it looks good. Most of my comments are about trimming and tightening the tests, and the details are inline.

About your notes in the PR description:

> Some of the scenarios may already be covered in other test functions, but my focus was on XML validation.

As I pointed out in the inline comments, several scenarios repeat existing tests: dropping 01, 04, 07, 08, 09, 11 and one of 02 and 03 leaves the line and branch coverage of `archivematicaCreateMETSMetadataXML.py` unchanged.

The ones worth keeping are 02 (or 03), 05, 06, 10 and 12 to 14, and 05, 06 and 10 could check more than `errors`, since they cover what this PR changes: that no `dmdSec` is added for a file that can't be parsed (plus whatever we settle on in the line 60 thread), and that a `validation` event with outcome `fail` is recorded when a schema can't be opened.

> I wasn't able to reproduce a scenario where a custom XML catalog is used by `lxml` to resolve schemas with nested imports.
> Setting `XML_CATALOG_FILES` dynamically at runtime (with `monkeypatch.setenv`) doesn't seem to affect `lxml` at all.

Catalogs do resolve nested imports, and what got in the way is the test process. `monkeypatch.setenv` reaches `libxml2`, but `libxml2` only reads `XML_CATALOG_FILES` once per process, the first time it has to look up a resource that isn't a local file. `tests/MCPClient/conftest.py` already sets that variable for every test (the autouse fixture `set_xml_catalog_files`), and in this module `test_resolver_with_requests_error` triggers the first lookup, so a test like the one you describe passes on its own but fails whenever that test runs first. I wouldn't add a catalog test to this PR, since the `metadata-xml` acceptance tests already run with sampledata's catalog; if you want one, you could add a test catalog to the list the autouse fixture sets, or run the validation in a subprocess with `XML_CATALOG_FILES` in its environment.

> Also, appending a custom resolver to the default resolver for in-memory schemas at runtime (with `etree.get_default_parser().resolvers.add`) has no effect on `lxml`. Ignoring the default resolver might be a security mechanism in `lxml` or related to caching.

Resolvers added with `etree.get_default_parser().resolvers.add()` are actually used while the schema is compiled; I resolved a three-level chain through one. They can look ignored for two reasons: `resolve_string(text, context)` without `base_url` hands nested relative locations to the resolver as bare names (`layer2.xsd`), and the schema listed in `XML_VALIDATION` is opened with `urlopen()`, which neither resolvers nor catalogs take part in. Since the default parser is global state, such a test would also have to remove its resolver afterwards.

> The only way I found to test schemas with nested imports was to use `schemaLocation` hints with purely local paths in every used schema.

My comment on line 96 shows how relative locations remove the need for absolute paths. For imports from `http(s)` locations, `requests.get` can be mocked per URL so each level goes through the module's `Resolver`, which also covers the new timeout:

```python
remote = {
    f"http://foo.com/layer{n}.xsd": schema_paths[f"layer{n}.xsd"].read_text()
    for n in (1, 2, 3)
}
with mock.patch(
    "requests.get", side_effect=lambda url, timeout: mock.Mock(text=remote[url])
) as get:
    mock_mets, errors = process_xml_metadata(...)
assert not errors
assert get.call_args_list == [mock.call(url, timeout=SCHEMA_FETCH_TIMEOUT) for url in remote]
```

Separately, the two pre-existing problems I mentioned inline, the `urlopen()` timeout and 404 responses in the resolver, don't need to be handled here; we can address them in a follow-up PR.

If you'd like a hand finishing any of this, just let me know. I'm happy to help, and since the PR allows edits from maintainers, I can also push changes to your branch if that's easier for you.

Thanks again for working on this!

PR #2346, opened 2026-07-23
Adopt uv project workflows

sevein:
Replace pip-tools and pyenv workflows with a universal uv lock and a pinned Python 3.14.6 default. Use uv for local, CI, and container commands.

Retain tox as the Compose test orchestrator and use tox-uv to synchronize its environments from the lock.

Connected to https://github.com/archivematica/Issues/issues/1788


hack/docker-compose.acceptance-tests.yml
```diff
@@ -9,7 +9,7 @@ services:
         TARGET: "archivematica-acceptance-tests"
         USER_ID: ${USER_ID:-1000}
         GROUP_ID: ${GROUP_ID:-1000}
-        PYTHON_VERSION: ${PYTHON_VERSION:-3.10}
+        PYTHON_VERSION: ${PYTHON_VERSION:-3.14.6}
```
replaceafill: Could we use `${PYTHON_VERSION:-}` here, as the [Storage Service does](https://github.com/artefactual/archivematica-storage-service/blob/2359e2a68070f351b17114140292eb1341275d22/tests/integration/docker-compose.yml#L15), and for the equivalent build arguments in the other Compose files? This lets each build context use its own `.python-version` by default while preserving explicit overrides from CI or the environment.

replaceafill (approved):
Thanks @sevein!

PR #2351, opened 2026-07-31
Batch antivirus database operations

sevein:
This pull request reduces the database work performed by antivirus jobs by using the batch of files already supplied to each MCPClient worker. It:

- Loads file sizes and existing virus-check events once per batch.
- Creates PREMIS events and their agent associations in bulk.
- Resolves shared agent associations once instead of separately for every event.
- Reuses the configured antivirus scanner for the lifetime of the batch.

## Scanner reuse and observability

The scanner is created lazily and reused only for the lifetime of one worker batch. Its configuration and cached version or virus-definition metadata are therefore not refreshed between files in that batch; a subsequent batch creates a new scanner when needed.

This pull request does not add mid-batch refresh logic or new instrumentation for measuring batch sizes, database query counts, or scanner reuse. Those improvements can be addressed separately.

## Testing

The changes include focused regression tests and a database-backed test of the complete batch workflow, covering:

- Loading file sizes and existing virus-check events in batch.
- Reading file sizes from the filesystem for files not represented in the database.
- Propagating batch lookup failures before scanning or event insertion begins.
- Skipping previously scanned files.
- Scanning new files.
- Reusing the configured scanner within a batch.
- Creating PREMIS events with the expected agents.

### DIFF
--- a/src/archivematica/MCPClient/clientScripts/antivirus.py
+++ b/src/archivematica/MCPClient/clientScripts/antivirus.py
@@ -18,19 +18,25 @@
 import multiprocessing
 import os
 import uuid
+from collections.abc import Callable
+from collections.abc import Collection
+from collections.abc import Mapping
+from dataclasses import dataclass
+from functools import cache
 
 import django
 
 django.setup()
+from clamav_client.scanner import Scanner
 from clamav_client.scanner import get_scanner
 from django.conf import settings as mcpclient_settings
-from django.core.exceptions import ValidationError
 from django.db import transaction
 
 from archivematica.archivematicaCommon.custom_handlers import get_script_logger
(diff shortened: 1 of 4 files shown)


replaceafill (commented):
> ## Compatibility
> ...
> * If the initial batch lookup fails, processing falls back to the existing per-file query and event-insertion paths.

@sevein I can see the new implementation using the existing per-file query but I've noticed the new `insert_events` helper is always used?

sevein:
> @sevein I can see the new implementation using the existing per-file query but I've noticed the new insert_events helper is always used?

You're right. I've ended up removing the per-file database fallback completely, because if the batch query fails, I think it is better to fail the jobs instead of trying the queries again one file at a time. `insert_events` remains the only event insertion path.

replaceafill (approved):

PR #2352, opened 2026-07-31
Make transfer submission idempotent

sevein:
Transfer submission previously had no request identity. If a client lost the
response after Archivematica accepted a request, retrying it created and
scheduled another Transfer, potentially producing a duplicate AIP.

Accept an optional Idempotency-Key header scoped to the authenticated user.
Return the original transfer UUID for an identical retry and reject changed
request parameters with HTTP 422.

Persist SHA-256 key and request digests in a dedicated IdempotencyRecord.
Keep request digests independent of processing configurations and metadata sets
so retries remain stable if those resources disappear. Keeping this separate
from Transfer lets the reservation survive normal transfer purging and provides
the unique constraint needed to resolve concurrent submissions safely. Expire
records after a configurable 90 days.

> [!WARNING]
> There is a small crash window after an idempotency record is saved as `pending` and before it is marked as `completed`. If MCPServer stops during this window, retries with the same key may return HTTP 409 until the record expires. A follow-up could add a short pending lease with recovery, or use a durable database outbox based on [transactional enqueueing](https://riverqueue.com/docs/transactional-enqueueing).

### Usage example

Enduro's start-transfer Temporal activity calls this endpoint with an
activity policy that retries timed-out or failed requests. Archivematica may
accept and schedule the transfer while the response is lost, so Temporal
cannot tell whether it is safe to retry. Without a request identity, a retry
starts another transfer.

This support lets Enduro derive a stable key from its workflow or activity
identity and reuse it for every attempt. Archivematica then returns the
original transfer UUID instead of scheduling the same preservation work
twice. Enduro still needs a corresponding client change to send the key.


src/archivematica/dashboard/components/api/views.py
replaceafill:
```python
except Exception as err:
    if isinstance(err, RPCServerError) and err.status_code == 422:
        return helpers.json_response(
            {"error": True, "message": str(err)},
            err.status_code,
        )

    msg = "Package cannot be created"
    LOGGER.exception("%s: %s", msg, err)
    return helpers.json_response({"error": True, "message": msg}, 500)
```

src/archivematica/MCPServer/server/packages.py
```diff
@@ -297,13 +388,20 @@ def create_package(
...
-            raise
-    else:
-        result = executor.submit(_start_package_transfer, *params)
+        if idempotency_key is not None:
+            logger.exception(
+                "Transfer %s could not be handed off after its idempotency "
+                "reservation was committed",
+                transfer_uuid,
+            )
+            return transfer_uuid
```
replaceafill: Not sure if I'm missing something, but wouldn't returning the UUID here cause the API endpoint to treat the failed handoff as successful and respond with 202 Accepted, even though no processing work was scheduled?

To check, I inserted an artificial raise before `executor.submit`. A retry with the same idempotency key then replayed this UUID without attempting the handoff again, leaving the transfer permanently failed. It looks like `test_create_package_keeps_keyed_result_when_executor_rejects_handoff` currently codifies this behavior.

Would it make sense to propagate the exception and revert or otherwise make the reservation retryable so a later request can attempt the handoff again?

sevein: Thanks, good catch! I was recording the identifier too early. I have changed this so the idempotency record remains pending until executor.submit succeeds. If the handoff fails, the error is propagated, the transfer is marked as failed, and the idempotency reservation is released. A retry with the same key will then try the handoff again. Only after a successful handoff will later requests return the same UUID. I also updated the test to cover this case.

replaceafill (approved):

PR #2353, opened 2026-08-01
Contain unexpected MCPServer job failures

sevein:
Handle exceptions from PackageQueue job futures so one failed job cannot retain an active package slot and stall unrelated processing.

On failure, log the full exception; mark the job, unfinished tasks, and package as failed; release the package slot; start the next package; replace the worker's Gearman backend; and increment a Prometheus counter. Run each cleanup action independently so one error does not block the rest.

User-visible impact:

- Package: Processing indefinitely -> Failed with a completion time.
- Job: may remain executing -> Failed.
- Unfinished tasks: no result -> exit code 1 with an explanatory error.
- Queue: leaked slot and eventual stall -> release the slot and continue.
- Recovery: MCPServer restart -> no restart, but no automatic retry.
- Normal success and configured exit-code paths remain unchanged.

Gearman may already have accepted remote work before the connection is reset, so a task can still finish or leave side effects after the package fails. Report that uncertainty and avoid automatic retries of potentially non-idempotent work.

This contains failures such as the reported python-gearman unknown-handle KeyError. It does not fix the unconfirmed Gearman root cause.

### DIFF
--- a/src/archivematica/MCPServer/server/jobs/base.py
+++ b/src/archivematica/MCPServer/server/jobs/base.py
@@ -90,6 +90,12 @@ class Job(metaclass=abc.ABCMeta):
         )
 
     @auto_close_old_connections()
+    def mark_failed(self):
+        return models.Job.objects.filter(jobuuid=self.uuid).update(
+            currentstep=self.STATUS_FAILED
+        )
+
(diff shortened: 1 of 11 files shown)


replaceafill (approved):
👏

PR #2367, opened 2026-08-14
Lazy-load processing monitor jobs

sevein:
The Transfer and Ingest tabs currently load all jobs for every processing unit every ten seconds. This does not scale well when a Transfer or SIP has thousands of jobs, making the Dashboard slow even when the user does not need to see those jobs.

This PR uses new Django views that let the Dashboard load a lightweight list of processing units first. Jobs are loaded only when the user expands a unit, and only that unit is refreshed until it is collapsed or another unit is opened.

## Demo

demo-dashboard-perf


src/archivematica/MCPServer/server/rpc_server.py
```diff
@@ -428,6 +455,279 @@ def _get_processing_config_fields_handler(self, worker, job, payload):
...
+        aggregate_rows = (
+            jobs.exclude(currentstep=Job.STATUS_AWAITING_DECISION)
+            .values("microservicechainlink", "currentstep")
+            .annotate(
+                count=Count("jobuuid"),
+                job_uuid=Subquery(representative_jobs.values("jobuuid")[:1]),
+                first_time=Min("createdtime"),
+                latest_time=Max("createdtime"),
+            )
+        )
```
replaceafill, line 674: The representative-job `Subquery` becomes a grouping expression here, so Django generates `GROUP BY 1, 2, 4` instead of grouping only by `(microservicechainlink, currentstep)`. In my development environment, `EXPLAIN ANALYZE` shows the dependent, non-cacheable subquery executing once for each of the 25,800 input Job rows of a synthetic SIP.

This makes the query take 10 to 13 seconds for that SIP, exceeding the monitor's five-second polling interval. Exact timings will vary, but the execution plan exposes the same per-row behavior. As a diagnostic baseline, the aggregate takes approximately 0.11 to 0.13 seconds when representative selection is removed entirely.

Could this be structured so the aggregate groups only by the intended Job fields and representative selection does not run once per input row? For example, retaining the subquery but adding it in a separate annotation generates `GROUP BY 1, 2` and reduced the query to approximately 0.48 seconds in the same environment:

```python
.annotate(
    count=Count("jobuuid"),
    first_time=Min("createdtime"),
    latest_time=Max("createdtime"),
)
.annotate(
    job_uuid=Subquery(representative_jobs.values("jobuuid")[:1]),
)
```

src/archivematica/dashboard/frontend/lib/monitor/components/ProcessMonitor.vue
```diff
@@ -92,6 +118,64 @@ const groupedJobsByUnitUuid = computed<Record<string, JobGroup[]>>(() => {
...
+
+const loadUnitJobGroups = async (unitUuid: string): Promise<void> => {
+  if (loadingUnitGroupUuids.value[unitUuid]) {
+    // A newer summary poll supersedes the in-flight detail response. Queue a
+    // replacement request so stale decision controls cannot be restored.
+    unitGroupRequestGenerations.value[unitUuid]
+      = (unitGroupRequestGenerations.value[unitUuid] ?? 0) + 1
+    pendingUnitGroupReloadUuids.value[unitUuid] = true
+    return
+  }
```
replaceafill, line 162: Periodic summary polling can starve the lazy request for an expanded unit. Every successful summary poll, including an unchanged one, triggers `loadUnitJobGroups()`; when a detail request is already pending, the pending-request path advances the generation, making the in-flight response stale. Its completed response is then discarded and another request is started. If detail responses keep overlapping the default five-second polls, the expansion can remain in `Loading…` for an unbounded time while the server repeatedly performs the same work.

I reproduced this locally: a 20,400-job transfer remained in `Loading…` after five completed detail responses, and the 25,800-job SIP after four. The same discarded-response behavior occurred on a 51-job unit with an artificial endpoint delay, confirming that the slow database query triggers the race but is not its cause.

Image

Could we ensure that background summary polling cannot continually invalidate an expanded unit's detail response, while retaining the protection against genuinely stale decision controls? There is a useful test gap here: the current test proves that one stale response is rejected, but it does not keep a detail request pending across repeated successful polls. A regression test for that case would expose the starvation behavior.

src/archivematica/MCPServer/server/rpc_server.py
```diff
@@ -428,6 +455,279 @@ def _get_processing_config_fields_handler(self, worker, job, payload):
...
+            row = {
+                "key": f"{link_id}:{aggregate['currentstep']}",
+                "link_id": str(link_id),
+                "currentstep": aggregate["currentstep"],
+                "timestamp": _datetime_to_unix_timestamp(aggregate["latest_time"]),
+                "first_timestamp": _datetime_to_unix_timestamp(aggregate["first_time"]),
+                "microservicegroup": link.get_label("group", lang),
+                "type": link.get_label("description", lang),
+                "count": count,
+                "uuid": str(aggregate["job_uuid"]),
```
replaceafill, line 697: Grouping repeated Jobs as `× N` is reasonable, but the grouped row retains only the newest Job UUID. As a result, its single Tasks button opens only the newest run; output from the earlier runs is still stored but is no longer reachable from the monitor. This differs from `qa/1.x`, where repeated Jobs were separate rows with separate Tasks actions.

This occurs in a normal workflow: choosing **Redo** at **Approve normalization** produced `Normalize for preservation × 2`. Both runs had seven Tasks, but only the second run was linked from the grouped row; the first Tasks page remained accessible only because its URL had been saved beforehand.

Could the aggregated representation retain some on-demand way to inspect every constituent Job? A regression test with two same-link/status Jobs carrying different Tasks would also help: the existing aggregation test checks the count and representative UUID, but not whether the other runs remain accessible.

src/archivematica/dashboard/components/api/views.py
```diff
@@ -163,7 +163,9 @@ def authenticate_request(request):
     if authorized is not True:
         error = "API key not valid."
 
-    elif not allowed_by_allowlist(client_ip):
+    elif not isinstance(
+        request._authentication_backend, SessionAuthentication
+    ) and not (allowed_by_allowlist(client_ip)):
```
replaceafill: This narrows the existing IP allowlist: previously it applied after either API-key or session authentication; now every API request authenticated by a Dashboard session bypasses it. Because `authenticate_request()` is shared by all `_api_endpoint` handlers, deployments relying on this setting lose the ability to restrict session-authenticated access to any of those endpoints by source IP, not only the endpoints used by the transfer browser.

I understand the immediate benefit: the transfer browser can use its same-origin Dashboard session for `/api/processing-configuration/` and `/api/v2beta/package/` without exposing an API key or requiring the browser/proxy address to be allowlisted. However, this is a broader authentication-policy and backward-compatibility change, independent of the monitor lazy-loading work.

Could this be handled as a separate, explicit policy change? The in-app administration text is updated, but the external documentation still says that API keys are always required and describes the allowlist as controlling access to the API generally. If this behavior is intentional, the documentation should define the two authentication paths and their different allowlist behavior, and the loss of IP restriction for Dashboard sessions should be called out in the release or upgrade notes.

src/archivematica/dashboard/frontend/lib/monitor/components/ProcessMonitorUnit.vue
```diff
@@ -105,64 +135,81 @@ const getIngestStartTime = (unit: ProcessingUnit): string => {
...
-        >
-          <SilkDeleteIcon
-            class="monitor-action-icon"
-            aria-hidden="true"
-            size="16"
-            alt=""
-          />
-          <span>{{ t('monitor.remove') }}</span>
-        </a>
+        <template v-if="isExpandable">
```
replaceafill: Waiting transfers no longer show either Metadata or Remove because `waiting_for_processing` rows are marked non-expandable. `qa/1.x` continued to show both actions for these rows.

Hiding Remove does not materially change behavior because active transfers were already protected in both the frontend and backend. Hiding Metadata does: the transfer’s detail and metadata pages remain available and usable while it waits, but the monitor no longer provides a way to reach them.

Could waiting transfers retain their existing detail/metadata navigation even though their Job details cannot yet be expanded?

replaceafill (changes requested):
This is looking great @sevein! I tested the transfer and ingest monitors in my development environment, created synthetic Jobs to reproduce the aggregation and polling behavior at different scales, and compared the behavior with `qa/1.x`.

I found that the representative-job subquery makes detail loading expensive and that summary polling can continually invalidate those detail responses. I also found some behavior and compatibility regressions around access to earlier Jobs' Tasks, Metadata navigation for waiting transfers, and the scope of the API allowlist change.

src/archivematica/dashboard/components/api/views.py
sevein: Sorry, leftover from the first iteration when I was making use of new API endpoints. Now gone.

src/archivematica/MCPServer/server/rpc_server.py
sevein: Fixed in e3740178b using the separate annotation you suggested.

src/archivematica/dashboard/frontend/lib/monitor/components/ProcessMonitor.vue
sevein: Fixed in fbd440e0d and 1fb8ecfaf.

src/archivematica/MCPServer/server/rpc_server.py
sevein: Fixed in 1ac704833.

src/archivematica/dashboard/frontend/lib/monitor/components/ProcessMonitorUnit.vue
sevein: Fixed in 4bf1c6779.

replaceafill (approved):

PR #2369, opened 2026-08-18
Add CAS authentication overlay and integration tests

replaceafill:
Add a docker-compose.cas.yml overlay that runs an Apereo CAS server with predefined demo and admin users for testing CAS authentication from a browser. The overlay turns on the existing CAS support in the Dashboard and the Storage Service and maps membership in the administrators CAS group to the superuser flag.

The CAS image is built from the Apereo CAS WAR overlay template pinned by commit (CAS 7.3.8) with the JSON service registry and JSON user store modules, since the published apereo/cas images are demo builds that do not include them. The runtime configuration, the registered services and the users are mounted from hack/etc/cas and can be edited without rebuilding the image.

Add a Playwright integration test suite covering the CAS behavior surface: login redirection, local user creation, existing user authentication, superuser promotion and demotion from CAS attributes, email autoconfiguration and CAS session termination on logout. CAS and OIDC cannot be enabled in the same Django process, so the suite runs through a dedicated Compose service selected with the INTEGRATION_SERVICE variable in tests/integration/run.sh, and the new manual CAS Integration Test workflow runs it in CI.

### DIFF
--- a/.github/workflows/cas-integration-test.yml
+++ b/.github/workflows/cas-integration-test.yml
@@ -1,0 +1,113 @@
+---
+name: "CAS Integration Test"
+on:
+  workflow_dispatch:
+env:
+  UBUNTU_VERSION: "24.04"
+  PYTHON_VERSION: "3.10"
+jobs:
+  cache-python-builder:
+    name: "Cache layers of the python-builder stage"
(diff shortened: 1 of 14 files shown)


sevein (approved):
👍

PR #2371, opened 2026-08-20
Upgrade Gearman client and harden recovery

sevein:
Archivematica now upgrades python-gearman to [f216d89], which adds bounded connection attempts, hostname re-resolution, and TCP keepalive. The previous client could leave idle connections unaware of some Gearman replacements.

Gearman replacements also left MCP workers and task producers holding stale connections. Workers now reconnect in process with bounded jittered backoff, and MCPServer executor threads invalidate their cached producers after an outage. Dashboard clients consistently close their connections.

Ambiguous task submissions now fail visibly instead of being replayed. Later work establishes fresh connections and resumes after Gearman returns.

[f216d89]: https://github.com/artefactual-labs/python-gearman/commit/f216d89faead4ff5f7f1d51bf767881b2a30bb47

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -70,7 +70,7 @@ repos:
     - "django-stubs>=6,<7"
     - django-tastypie
     - elasticsearch>=8.0.0,<9.0.0
-    - gearman3 @ git+https://github.com/artefactual-labs/python-gearman.git@b68efc868c7a494dd6a2d2e820fb098a6da9f797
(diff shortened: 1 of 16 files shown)


sevein:
Closing this for now. This was an exploration after the Gearman job server started becoming unresponsive in a customer deployment, but we do not yet know whether the problem will happen again or whether these changes are the right solution.

These changes could help Gearman reconnect, but they also have some risks. If the connection is lost while a task is running, MCPServer may not know whether it finished, and retrying it could run it twice. We need more testing to be sure this is safe, and it is not a priority right now. We can come back to it if the problem happens again or we learn more about the cause.

PR #2376, opened 2026-08-24
Import objects/ CSV rights at transfer scope

sevein:
Rights CSV imports previously treated every file value as an original file path. This prevented one statement from representing rights for a whole transfer, even though transfer and AIP METS generation already support transfer-scoped statements.

Recognize objects/ as the transfer target and store those statements against the transfer UUID. Continue resolving every other value as a file path. Mixed CSVs therefore preserve file-specific rights, and METS generation writes both scopes for affected original files.

Clarify this behavior in the transfer rights editor. Regression tests cover mixed-scope imports, scope-aware duplicate handling, and transfer plus file rights in transfer and AIP METS.

Related:
* https://github.com/archivematica/Issues/issues/1340
* https://github.com/wellcomecollection/archivematica-infrastructure/issues/114
* https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/307
* https://github.com/artefactual/archivematica-sampledata/pull/88
* https://github.com/artefactual/archivematica-docs/pull/515

### DIFF
--- a/src/archivematica/MCPClient/clientScripts/rights_from_csv.py
+++ b/src/archivematica/MCPClient/clientScripts/rights_from_csv.py
@@ -33,7 +33,10 @@ class RightsRowException(Exception):
 
 
 class RightCsvReader:
-    metadata_applies_to_type = None
+    TRANSFER_TARGET = "objects/"
+
+    file_metadata_applies_to_type = None
+    transfer_metadata_applies_to_type = None
     current_row = None
     rows_processed = 0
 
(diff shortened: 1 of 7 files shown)


src/archivematica/MCPClient/clientScripts/rights_from_csv.py
```diff
@@ -33,7 +33,10 @@ def __init__(self, message, reader):
 
 
 class RightCsvReader:
-    metadata_applies_to_type = None
+    TRANSFER_TARGET = "objects/"
```
replaceafill, line 36: Should this accept `objects` (no trailing slash) as well?

src/archivematica/MCPClient/clientScripts/rights_from_csv.py
```diff
@@ -162,17 +168,23 @@ def generate_rights_statement(self):
...
-        )
+
+        if filepath == self.TRANSFER_TARGET:
+            metadata_applies_to_type = self.transfer_metadata_applies_to_type
+            metadata_applies_to_identifier = self.transfer_uuid
+        else:
+            transfer_file = models.File.objects.get(
+                originallocation=("%transferDirectory%" + filepath).encode(),
+                transfer_id=self.transfer_uuid,
+            )
```
replaceafill, line 180: Feel free to ignore this one, but could we catch `File.DoesNotExist` here and report a clear error when `filepath` doesn't match a file, for example because of a typo?

replaceafill (approved):
Thanks @sevein. While testing this today, I noticed a pre-existing mismatch in the [Importing rights metadata with rights.csv documentation](https://www.archivematica.org/en/docs/archivematica-1.18/user-manual/transfer/import-metadata/#importing-rights-metadata-with-rights-csv):

> Rights metadata can be applied to individual objects within a transfer, to directories within the transfer, or both.

The existing importer only supports individual files. This PR adds `objects/` as a transfer-wide target, while other directory paths remain unsupported. We should update the documentation to reflect that distinction.

PR #2379, opened 2026-08-27
Measure coverage across the whole archivematica package

replaceafill:
The coverage configuration listed the archivematicaCommon, dashboard, MCPClient and MCPServer packages in an include list, so any module added outside them was silently left out of the reports; the search package was never measured.

Replace the list with a source directory pointing at src/archivematica. Coverage now measures every package under the namespace and also reports modules the test suite never imports, which an include list cannot do. A package name would not work here: archivematica is a namespace package shared with the Storage Service, and coverage only walks packages that have an __init__.py for unexecuted files.

The path is resolved against the current directory and tox runs pytest from the tests directories, so the tox environments export COVERAGE_SOURCE_DIR with the absolute path; the default keeps pytest working from the repository root.

The XML report now lists filenames relative to the source directory instead of absolute paths, and the overall percentage drops because never-imported modules are counted.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,10 @@ DJANGO_SETTINGS_MODULE="settings.test"
 norecursedirs = ".svn _build tmp* node_modules bower_components share .tox .venv"
 
 [tool.coverage.run]
+# The source path is resolved against the current directory. tox runs pytest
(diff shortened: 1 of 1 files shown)


sevein (approved):
👍

PR #2380, opened 2026-08-27
Fix search document purge in purge_transient_processing_data

replaceafill:
The command passed the package UUID object to SearchService.delete_aip and delete_aip_files, whose _delete_by_field treats any non-str value as a set of values to iterate, so purging search documents failed with "'UUID' object is not iterable". The error handler then crashed too, because it wrote the return value of traceback.print_exc(), which is None, to stdout. Pass the string form of the UUID, as every other caller does, and use traceback.format_exc() like the transfer loop.

The existing test did not catch this because it asserted the mocked service was called with the UUID object; it now expects the string. A new test makes the mocked service raise and checks that the error is reported and the remaining packages are still purged, since nothing exercised that handler before.

Add the command to the strict mypy list to prevent regressions. Supporting annotations and minor refactors let mypy check the affected code.

Connected to https://github.com/archivematica/Issues/issues/1814

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,7 @@ ignore_errors = true
 module = [
     "archivematica.archivematicaCommon.transfer_source_retrieval",
     "archivematica.archivematicaCommon.executeOrRunSubProcess",
+    "archivematica.dashboard.main.management.commands.purge_transient_processing_data",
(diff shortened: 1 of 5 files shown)


sevein (approved):
LGTM

PR #2381, opened 2026-08-28
Add LDAP authentication overlay and integration tests

replaceafill:
Replaces the mockldap-based unit test, which never ran because it required `LDAP_AUTHENTICATION` to be enabled, with a Playwright suite that exercises the Dashboard's LDAP backend against a real OpenLDAP server. The same server is available as a `docker-compose.ldap.yml` overlay for local testing.

Along the way this fixes a real bug: `CustomLDAPBackend` stripped `AUTH_LDAP_USERNAME_SUFFIX` with `str.rstrip`, so `"paula_ldap"` became the local user `"pau"`.

See the individual commits for details. Run the suite locally with `INTEGRATION_SERVICE=archivematica-dashboard-ldap tests/integration/run.sh` or by dispatching the `LDAP Integration Test` workflow.


hack/docker-compose.ldap.yml
```diff
@@ -0,0 +1,82 @@
...
+      AUTH_LDAP_MANAGER_GROUP: "managers"
+      AUTH_LDAP_REVIEWER_GROUP: "reviewers"
+    depends_on:
+      openldap:
+        condition: service_healthy
+
+  openldap:
+    # The image only publishes the latest tag, so the build is pinned by its
+    # digest, which Dependabot keeps up to date.
+    image: "vegardit/openldap:latest@sha256:8161e8fa74c518697776fff27b403cc4c273aede07b6d87f473a45201050d0de"
```
sevein:
```yaml
# Upstream publishes versioned tags, but may rebuild them to pick up base-image
# security updates. Keep the version tag for readability and pin the exact
# build by digest; Dependabot keeps the tag and digest up to date.
image: "vegardit/openldap:2.6.10@sha256:8161e8fa74c518697776fff27b403cc4c273aede07b6d87f473a45201050d0de"
```

Btw, it looks like https://hub.docker.com/layers/vegardit/openldap/2.6.10/ already lists a newer build with index digest `sha256:4378fb00077a0dd4a969f35694d7a3660c171ccb2bd62e9260647490882f17c5`.

hack/docker-compose.ldap.yml
```diff
@@ -0,0 +1,82 @@
...
+        - -D
+        - uid=ldapadmin,dc=archivematica,dc=org
+        - -w
+        - test
+        - -b
+        - dc=archivematica,dc=org
+        - -s
+        - base
+        - (objectClass=*)
+        - dn
```
sevein: Are you sure you don't want to do this instead? 😆 

```yaml
test: ["CMD", "ldapsearch", "-x", "-H", "ldap://127.0.0.1:389", "-D", "uid=ldapadmin,dc=archivematica,dc=org", "-w", "test", "-b", "dc=archivematica,dc=org", "-s", "base", "(objectClass=*)", "dn"]
```

sevein (approved):
LGTM.

The LDAP workflow is almost identical to the OIDC and CAS workflows. Could we reduce duplication in a follow-up by sharing the common steps between the authentication integration workflows, e.g. a reusable workflow or a matrix?

hack/docker-compose.ldap.yml
replaceafill: Thanks, now that I know about the `configs` element I can start abusing it :grin: https://github.com/artefactual/archivematica/pull/2381/commits/318191c45eeb7281fb29663cd5e983402da86bc1

PR #2387, opened 2026-09-02
Ignore deleted files when parsing METS during reingest

replaceafill:
Reingest retains deleted entries in the METS as provenance tombstones after their files have been removed from the AIP. They therefore have no FLocat and should not be loaded into the database as files to process again.

Connected to https://github.com/archivematica/Issues/issues/1044


sevein (approved):
LGTM

PR #2390, opened 2026-09-04
Consolidate the authentication integration test workflows

replaceafill:
Follow-up to the review on https://github.com/artefactual/archivematica/pull/2381, which asked to share the common steps between the authentication integration workflows.

The OIDC, CAS and LDAP workflows were copies of one another, so this replaces them with a single workflow that runs the three providers as a matrix. A dispatch input keeps the ability to run one provider on its own, which until now worked only because each had its own workflow.

Along the way, the suites now run when the provider configuration or the Dashboard's authentication code changes, and on the release branches, where before only a change to the compose file started them. And a failing run keeps the logs of the services it was testing against, which previously went away with the stack before anything could collect them.


sevein (approved):
LGTM

PR #2394, opened 2026-09-09
Add Shibboleth authentication overlay and integration tests

replaceafill:
Add a `docker-compose.shibboleth.yml` overlay that puts a Shibboleth service provider in front of the Dashboard and the Storage Service, with a SAML realm in a Keycloak service like the OIDC overlay's, so the Shibboleth backends of both applications can be tried locally for the first time. The realm comes with predefined users whose group memberships map to the Dashboard's superuser flag and the Storage Service's roles.

Replace the Dashboard's Shibboleth unit tests, skipped since 2017 because the setting cannot be turned on per test, with a Playwright suite that runs with Shibboleth enabled: most tests send the attribute headers a service provider would set, and a few drive a real SAML login through the overlay's service provider and Keycloak.

Fix two defects in the Dashboard that the suite exposed. Logging in failed when the identity provider released no attribute that maps to a user field, and logging out was broken twice over: the "Log out" button posts to a library view that only accepted GET, and the return page was handed to the service provider under a parameter it ignores. The Storage Service has the same logout problem, its "Log out" button gets a 405 from the same library view; that will be fixed separately in its repository.

Switch the OIDC overlay to `keycloak.localhost` along the way, so browsers and containers share one Keycloak URL, as the CAS overlay already does with `cas.localhost`.


sevein (approved):
LGTM

